Issue #952, opened 2014-12-09
socket.sendto() Invalid parameter in TunnelCommunity


LipuFei:
Got the backtrace from http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/3006

```
Traceback (most recent call last):
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/community.py", line 1109, in exit_data
    self.exit_sockets[circuit_id].sendto(data, destination)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/community.py", line 198, in sendto
    self.transport.write(data, destination)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/udp.py", line 271, in write
    return self.socket.sendto(datagram, addr)
error: [Errno 22] Invalid argument
```

synctext:
Got this again when running V6.4.0 for almost 3 weeks on my Ubuntu box. Fixed in V6.4.1 ?

```
ERROR   1418979352.59       community:617   Could not remove relay 3290279600 no activity
ERROR   1418979353.76       community:853   Don't know how to decrypt incoming message for circuit_id 1377147404
ERROR   1418979355.85       community:1112  Dropping data packets while EXITing
Traceback (most recent call last):
  File "/usr/share/tribler/Tribler/community/tunnel/community.py", line 1110, in exit_data
    self.exit_sockets[circuit_id].sendto(data, destination)
  File "/usr/share/tribler/Tribler/community/tunnel/community.py", line 198, in sendto
    self.transport.write(data, destination)
  File "/usr/lib/python2.7/dist-packages/twisted/internet/udp.py", line 271, in write
    return self.socket.sendto(datagram, addr)
error: [Errno 22] Invalid argument
ERROR   1418979362.59       community:617   Could not remove relay 670744770 no activity
ERROR   1418979367.59       community:617   Could not remove relay 4118921637 no activity
ERROR   1418979372.58       community:617   Could not remove relay 3430112127 no activity
```

whirm:
@synctext It should. Could you see if it works for you too?

synctext:
Still happening on current Next branch
commit: 385dfb1ddbace2bf964d5bfe5fe7fc7a005a2cc7
```
ERROR   1420644596.38       community:618   Could not remove relay 2453034024 no activity
[...]

PR #986, opened 2014-12-15
Check for short datagram payloads

whirm:
To try to find out if that's the cause for #952


whirm:
Reopening as it is not fixed

PR #1017, opened 2014-12-18
Changed Download directory for windows users.

GerAnon4113:
It probably feels unnatural to save files to desktop for some users so the downloads directory should be used instead. (Many people only want their most important stuff on their desktop)


synctext:
This is a nice fix that many users asked for.
I'm kind of in favor of adding this to a 6.4.x fix version, instead of waiting for V6.5

whirm:
@synctext OK, Ill cherry pick @GerAnon4113 commits from this PR and do the migration step.

whirm:
I'm closing this one in favor of the fixed/squashed/refined one I just pushed.
Thanks again @GerAnon4113!

PR #1105, opened 2014-12-31
Add default anonymous level in settings

### LINKED ISSUE #1001: The settings panel needs a default privacy level slider
as reported on forum:
I had a .torrent file on my desktop then in Tribler, I clicked the "+" icon on the top right and I browsed for the local .torrent file. As soon as I pressed "Open" in the file browser window the torrent is added to Tribler's download list and starts downloading without any prompt about anonymity. The newly added torrent has a tag that says "Anonymous: No". Is this what should be happening? I feel as if I should be seeing something like the what's shown in the picture in the first post in this topic:

http://forum.tribler.org/viewtopic.php?f=2&t=6722

I have three log files and I'm not sure which one would have useful information in it so I've attached all of them.

EDIT: I probably should have played around with the settings a little more before posting because I just found the problem. I had the option "Let me choose a location for every download" disabled and so the window I was talking about was completely bypassed when I added the torrent. It would be good if anonymous download preferences could be set in the options window so that this anonymity slider window is not necessary if any users might not like always having to pick a download location. Or, perhaps the anonymity slider should just always be shown when adding a torrent despite the user's preference of choosing or not choosing a download location for each download.

LipuFei:
Fixes #1001

### DIFF
--- a/Tribler/Core/defaults.py
+++ b/Tribler/Core/defaults.py
@@ -121,8 +121,9 @@ sessdefaults['video']['preferredmode'] = PLAYBACKMODE_INTERNAL
 #  Version 7: swift params
 #  Version 8: deleted many of the old params that were not used anymore (due to the switch to libtorrent)
 #  Version 9: remove swift
+#  Version 10: add default anonymous level
 
-DLDEFAULTS_VERSION = 9
+DLDEFAULTS_VERSION = 10
 dldefaults = OrderedDict()
 
 # General download settings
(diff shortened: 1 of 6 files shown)


whirm:
retest this please

whirm:
retest this please

LipuFei:
retest this please

LipuFei:
retest this please

synctext:
retest this please

LipuFei:
retest this please

LipuFei:
retest this please

whirm:
retest this please

LipuFei:
retest this please

LipuFei:
retest this please

PR #1109, opened 2015-01-01
Remove unnecessary grep in SingleInstanceChecker.py

RunningDroid:
pgrep removes itself from the list of results and with grep 2.21 or greater grep will print `grep: warning: GREP_OPTIONS is deprecated; please use an alias or script` if GREP_OPTIONS is set, causing Tribler to believe there is already a process running.


whirm:
Thanks!

PR #1127, opened 2015-01-07
Fix anon dialog

LipuFei:
use `wx.MessageDialog` instead


Tribler/Main/tribler_main.py
```diff
@@ -407,8 +407,21 @@ def InitStage1(self, installdir):
...
+                                            'trouble in various countries.\n'
+                                            'This privacy enhancement will not protect you against spooks or '
+                                            'government agencies.\n'
+                                            'We are a torrent client and aim to protect you against lawyer-based '
+                                            'attacks and censorship.\n'
+                                            'With help from many volunteers we are continuously evolving and improving.'
+                                            '\n\nI you aren\'t sure, press Cancel to disable the \n'
+                                            'experimental anonymity feature',
+                                            'Do you want to use the experimental anonymity feature?',
+                                            wx.ICON_WARNING | wx.OK | wx.CANCEL)
```
NielsZeilemaker: I would replace the wx.OK and wx.CANCEL flags by wx.YES_NO | wx.NO_DEFAULT, and change the text accordingly.

Tribler/Main/tribler_main.py
```diff
@@ -407,8 +407,21 @@ def InitStage1(self, installdir):
...
             enable_tunnel_community = optin_dialog.ShowModal() == wx.ID_OK
```
NielsZeilemaker: Moreover, I think after that change you should match agains wx.ID_YES

synctext: Yes, very good. Originally I asked for ID_DISABLE and ID_ENABLE. This was during our X-Mas rush fix.

NielsZeilemaker: Those don't exist, but wx.YES_NO does. 
See http://www.wxpython.org/docs/api/wx.MessageDialog-class.html

Issue #1115, opened 2015-01-02
After updating from v6.3.5 to v6.4.1, it stuck on "Checking tribler version"


vi:
No messages get logged to console.

vi:
After moving away `~/.Tribler` to `~/.Tribler.bak` it got though. After `cp  ~/.Tribler.bak/dlcheckpoints/* ~/.Tribler/dlcheckpoints/` it also remembered the downloads.

whirm:
@vi Could you confirm that copying the old DB back to .Tribler makes Tribler fail again?

vi:
Backed up current `.Tribler`, restored old v6.3.5 `.Triber`. 

`Checking Tribler version...`

I have just noticed that it's also big: sqlite directory is 543.7MiB. Maybe it's just slow? I see some CPU and IO activity...

whirm:
@vi we've seen that the torrent migration is quite slow on some machines but we haven't investigated further as we are seriously thinking about killing the torrent collection altogether and just share magnets with some metadata associated.

I've sent a pull request that should deal with the cross-filesystem issues you where having.

Do you still have the 6.3.5 data around to test again with 6.4.2?

Thanks!

vi:
The migration can be slow, but in this case it should be mentioned in the message to use (not just `Checking Tribler version`), 

I have `~/.Tribler` backup for 6.3.5, but not the torrent collection was removed.

> just share magnets with some metadata associated.
Will Tribler be self-sufficient, i.e. will the network of only other Tribler clients, without other non-Tribler torrent peers still work?

whirm:
The part I've seeing being slow was the torrent parsing/migration and you see the infohashes passing trough. 
In your case it was before that?
I'll have a look and see what's going on before we get to that point.

RE being self-sufficient: yes, it would collect the torrent info from the DHT as when you click on a standard magnet link.

whirm:
@vi: We are working on a branch  which uses a leveldb based torrent store instead of the collected_torrent_files directory:

https://github.com/whirm/tribler/tree/torrent_store_migrator

That branch includes the upgrade code too, so it will ingest all your torrent files into the new store.

It will also store the torrent stuff in ```.Tribler``` instead of ```TriblerDownloads```, just so you know.

Could you put a copy or your old backup in place and tell us how it behaves for you?

Have in mind that you will need to install python-leveldb to be able to run this branch.

Thanks!

whirm:
[...]

PR #1147, opened 2015-01-15
Use shutil.move instead of os.rename in several places

whirm:
To allow to move files across filesystems. 
Related to: #1115


whirm:
@LipuFei ready

PR #1179, opened 2015-01-30
use avconv or ffmpeg, whatever is available on Linux

whirm:
Do not merge until tested on Windows


whirm:
@LipuFei ready to merge

PR #1182, opened 2015-01-30
Show an error dialog when add magnet download fails

### LINKED ISSUE #1002: Tribler goes to foreground when Magnet:: is clicked and download is started.
Now it looks like nothing is happening, other clients also pop to the foreground.

LipuFei:
Fixes #1002

### DIFF
--- a/Tribler/Main/vwxGUI/MainFrame.py
+++ b/Tribler/Main/vwxGUI/MainFrame.py
@@ -418,8 +418,15 @@ class MainFrame(wx.Frame):
         name, infohash, _ = parse_magnetlink(url)
         if name is None:
             name = ""
-        tdef = TorrentDefNoMetainfo(infohash, name, url=url)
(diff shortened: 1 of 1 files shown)


LipuFei:
retest this please

LipuFei:
retest this please

LipuFei:
retest this please

whirm:
@LipuFei I would be a bit more explicit about why we failed (maybe The magnet link you clicked on is invalid?)

LipuFei:
@whirm So far I have found only one failure which is caused by a None info hash, so for this failure, I can add an if to check if info hash is None and show a more meaningful message (missing infohash in magnet link). But I don't know if there can be any other problems, so the best we can do is to show the exception itself.

Do you want me to add a check for None info hashes?

whirm:
I would show a user friendly in case we can identify the error. In that case I would show something like:
```The magnet link is invalid: Missing infohash```
If we don't know why it failed:
```The magnet link is invalid: "exception str"```

LipuFei:
@whirm Done. Feel free to merge when it passes.

LipuFei:
retest this please

LipuFei:
@whirm ready

PR #1184, opened 2015-01-30
Some fixes from the torrent store branch

whirm:
I've splitted them into a new branch so the can get merged while the torrent store stuff gets tested.


whirm:
@LipuFei ready to merge if it passes

PR #1190, opened 2015-02-02
Set delimiter for linehandler in tunnel main to \n

rjruigrok:
Small fix - for some reason Ubuntu on my mac stopped recognizing the enter key, rendering the main debugger useless. By specifying \n hardcoded, problems is solved.

### DIFF
--- a/Tribler/community/tunnel/main.py
+++ b/Tribler/community/tunnel/main.py
@@ -202,7 +202,7 @@ class Tunnel(object):
 
 class LineHandler(LineReceiver):
     
-    delimiter = '\n'
+    delimiter = os.linesep
     
     def __init__(self, anon_tunnel, profile):
         self.anon_tunnel = anon_tunnel
--- a/Tribler/community/tunnel/main.py
+++ b/Tribler/community/tunnel/main.py
@@ -201,6 +201,9 @@ class Tunnel(object):
 
 
 class LineHandler(LineReceiver):
+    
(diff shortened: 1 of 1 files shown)


whirm:
@rjruigrok what is the main debugger? whatever it is it should be sending CR LF, Not just LF.

rjruigrok:
I mean tunnel/main.py, where you can see the circuits by entering key `c`, then `enter` repeatedly. I tried setting it back to \r\n, obviously the default-value, because the linereader does not work anymore here in that case. :)

whirm:
Ah I see, it's not telnet stuff but a commandline interface.
I guess it never worked for you? You probably get CRLF on windows, LF in Linux and CR on OSX.
I guess we should be using ```delimiter = os.linesep``` instead then.

rjruigrok:
@whirm I changed it to os.linesep, which works perfectly (although the twisted linereader should have that by default :)

Strangest thing about this is that it worked in the past without any problems, and since a few days it stopped working. I guess this problem started after an apt-get update two weeks ago.

whirm:
@rjruigrok I don't think it should as I guess it's designed to implement a telnet type protocol.

whirm:
@rjruigrok also, it would be nice if you get used to squash the overlaping commits together to not pollute the commit history too much.

whirm:
https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i

PR #1208, opened 2015-02-06
Fix a few missed ds->download_status replacements

whirm:
And download_status->download_state

Ready if it passes


NielsZeilemaker:
@whirm looks great

PR #1214, opened 2015-02-09
Remove Torrent Source Table

### LINKED ISSUE #1213: Remove TorrentSource table and source_id column in Torrent table in Tribler Database
The `TorrentSource` table and the `source_id` column in `Torrent` table are no longer being used. The only place that uses them is the `test_sqlitecachedbhandler` unit test. We should remove them.

- [x] Remove `TorrentSource` and related code
- [ ] Modify database upgrade script

LipuFei:
Partially fix #1213 
We need to modify the database upgrade script.


LipuFei:
@whirm ready

PR #1219, opened 2015-02-11
Guitaskqueue taskmanager

whirm:
Drop TimedTasQueue and GUITaskQueue and use TwistedRawServer instead.


whirm:
retest this please

whirm:
retest this please

NielsZeilemaker:
@whirm I was looking at the changes, and similar to the other discussion we had, I'm would remove most of the blocking_ decorators. Those are really dangerous.

Tribler/Core/Utilities/configparser.py
```diff
@@ -15,59 +15,58 @@ class CallbackConfigParser(RawConfigParser):
     def __init__(self, *args, **kwargs):
         RawConfigParser.__init__(self, *args, **kwargs)
         self.callback = None
-        self.lock = RLock()
 
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: Removing the lock and replacing it with a blocking call is worse than before. I would revert these changes.

Tribler/Core/Session.py
```diff
@@ -296,25 +305,42 @@ def set_download_states_callback(self, usercallback, getpeerlist=None):
     #
     # Config parameters that only exist at runtime
     #
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: This should be blocking

Tribler/Core/Session.py
```diff
@@ -174,15 +182,16 @@ def get_instance(*args, **kw):
...
-    get_instance = staticmethod(get_instance)
 
+    @staticmethod
+    @blocking_call_on_reactor_thread
     def has_instance():
         return Session.__single is not None
-    has_instance = staticmethod(has_instance)
 
+    @staticmethod
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: No need to make this blocking, if you want you could have a singleton_lock

Tribler/Core/Session.py
```diff
@@ -174,15 +182,16 @@ def get_instance(*args, **kw):
         if Session.__single is None:
             Session(*args, **kw)
         return Session.__single
-    get_instance = staticmethod(get_instance)
 
+    @staticmethod
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: No need for blocking

Tribler/Core/Session.py
```diff
@@ -166,6 +172,8 @@ def prestart(self):
     #
     # Class methods
     #
+    @staticmethod
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: No need for blocking

Tribler/Core/Session.py
```diff
@@ -58,7 +63,7 @@ def __init__(self, scfg=None, ignore_singleton=False, autoload_discovery=True):
             if Session.__single:
                 raise RuntimeError("Session is singleton")
             Session.__single = self
-
+        assert isInIOThread()
```
NielsZeilemaker: Why does this need to be constructed on the ioThread again?

Tribler/Core/SessionConfig.py
```diff
@@ -161,10 +167,16 @@ def set_permid_keypair_filename(self, keypairfilename):
         """
         self.sessconfig.set(u'general', u'eckeypairfilename', keypairfilename)
 
+    @blocking_call_on_reactor_thread
     def get_permid_keypair_filename(self):
         """ Returns the filename of the Session's keypair.
         @return An absolute path name. """
-        return self.sessconfig.get(u'general', u'eckeypairfilename')
+        file_name = self.sessconfig.get(u'general', u'eckeypairfilename')
+        if not file_name:
```
NielsZeilemaker: What happened here?

Tribler/Core/SessionConfig.py
```diff
@@ -618,12 +630,12 @@ def get_default_config_filename(state_dir):
         return os.path.join(state_dir, STATEDIR_SESSCONFIG)
 
 
-
 class SessionStartupConfig(SessionConfigInterface):
 
     """ Class to configure a Session """
 
     def __init__(self, sessconfig=None):
+        assert isInIOThread()
```
NielsZeilemaker: Why does this need to be on the ioThread?

Tribler/Main/Dialogs/GUITaskQueue.py
```diff
@@ -21,11 +21,11 @@ def __init__(self):
...
         TimedTaskQueue.__init__(self, nameprefix="GUITaskQueue")
 
+    @staticmethod
     def getInstance(*args, **kw):
         if GUITaskQueue.__single is None:
             GUITaskQueue(*args, **kw)
         return GUITaskQueue.__single
-    getInstance = staticmethod(getInstance)
 
     def delInstance(*args, **kw):
```
NielsZeilemaker: Then update the delInstance as well to a decorated version

Tribler/Test/API/test_seeding.py
```diff
@@ -32,6 +38,7 @@ def setUp(self):
         self.seeding_event = threading.Event()
         self.downloading_event = threading.Event()
 
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: Why does this need to be on the reactor thread?

Tribler/Test/test_as_server.py
```diff
@@ -152,7 +155,10 @@ def setUp(self):
 
         self.quitting = False
 
-        self.session = Session(self.config)
+        @blocking_call_on_reactor_thread
```
NielsZeilemaker: I guess this is due to the assert isIOThread stuff, doesn't make a lot of sense

Tribler/Test/test_as_server.py
```diff
@@ -166,8 +172,10 @@ def setUp(self):
 
         self.annotate(self._testMethodName, start=True)
 
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: Same here, why must the preSession be on the ioThread?

Tribler/Core/SessionConfig.py
whirm: I did that a while ago, I don't remember anymore, I guess it was getting ```None```. I will try removing that and seeing if I can reproduce whatever issue I had.

whirm:
@NielsZeilemaker about the whole pushing stuff to the reactor thread: I was seeing if I could move as much logic to the reactor thread as possible easily. It won't be that easy so I'm leaving it for a future PR/

NielsZeilemaker:
@whirm that makes a lot of sense. However, I would go about it the other way. By removing all other threads, this eventually will result in the same endpoint.

NielsZeilemaker:
what about removing the taskqueue and the guitaskqueue and switching to the twistedrawserver.add_task_in_thread

Tribler/Core/Session.py
```diff
@@ -106,21 +110,22 @@ def set_and_create_dir(dirname, setter, default_dir):
             #
             # 1. keypair
             #
-            pairfilename = os.path.join(scfg.get_state_dir(), 'ec.pem')
```
NielsZeilemaker: Ah, I see you pushed this logic into sessionconfig

whirm: Ah, there it is! :D

whirm:
> what about removing the taskqueue and the guitaskqueue and switching to the twistedrawserver.add_task_in_thread

Yes, I was thinking about having a few utility methods to deal with this on TaskManager but maybe we could just keep it on the rawserver for the time being.

BTW, why is the rawserver called rawserver?

whirm:
@NielsZeilemaker I'll this pass and then proceed with the refactoring to use the rawserver.

Tribler/Core/Session.py
NielsZeilemaker: This should not be blocking :smile:

Tribler/Core/Session.py
```diff
@@ -296,25 +301,36 @@ def set_download_states_callback(self, usercallback, getpeerlist=None):
...
+          startup and due to the Dialback mechanism, this method may return
+          False while the Session is actually already reachable. Note that True
+          doesn't mean the Session is reachable from the open Internet, could just
+          be from the local (otherwise firewalled) LAN.
+          @return A boolean. """
+
+        # Arno, LICHT: make it throw exception when used in LITE versie.
+        raise NotYetImplementedException()
+
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: Same here

NielsZeilemaker:
It's getting there, my current worries are now that the TimedTaskQueue used to be a separate thread, and now it's running on the reactor. It might be better to run the tasks on the reactor thread pool instead.

whirm:
@NielsZeilemaker yes, now that is passing, I guess we can directly migrate to the rawserver.add_task_in_thread()

NielsZeilemaker:
Yes, that would be a nice improvement. And maybe lateron rename rawserver to something else.

whirm:
retest this please

whirm:
Ok, I'll squash some commits and rename the rawserver too. What about PooledTaskManager or ThreadPoolTaskManager?

whirm:
Rebase and cleanup done, only the class renaming is left.

whirm:
Github is being silly and sorts the commits by creation date, but https://github.com/Tribler/tribler/commit/6d4fe018b6c09eb8a89a0af50061ce80cefe31ca is the last one really.

Tribler/Main/tribler_main.py
```diff
@@ -216,7 +211,7 @@ def __init__(self, params, installdir, autoload_discovery=True,
 
             # Schedule task for checkpointing Session, to avoid hash checks after
             # crashes.
-            self.guiserver.add_task(self.guiservthread_checkpoint_timer, SESSION_CHECKPOINT_INTERVAL)
+            session.lm.rawserver.call_in_thread(SESSION_CHECKPOINT_INTERVAL, self.guiservthread_checkpoint_timer)
```
NielsZeilemaker: In the gui I would use the start_worker stuff and then specify that it should run in the threadpool.

whirm:
Ready if you guys like the new name :)

LipuFei:
@whirm @NielsZeilemaker Looks good :)

whirm:
@LipuFei rebased and ready when it passes

whirm:
retest this please

LipuFei:
@whirm I will merge this one. That PyDeadObject problem in ListBody happens quite often these days, but I have no idea where it comes from. The backtrace is not informative enough.

whirm:
@LipuFei yes, maybe we should have another shot at separating the GUI into a component that can be started and stopped by itself. I have a branch with some work on that but It'll need heavy refactoring. Let's talk about it on Monday and decide what to do next.

PR #1235, opened 2015-02-18
Fix Mac build script

LipuFei:


LipuFei:
@whirm tested. ready.

PR #1243, opened 2015-02-19
Add trackers into torrent files

### LINKED ISSUE #262: Torrents in the megacache don't get the trackers updated from more recent torrent files.
Adding a torrent with the same infohash does not update MegaCache. If the new torrent has updated trackers they won't get updated in MegaCache either and download will rely on DHT only. This might gravely affect torrents with a small number of seeds.

LipuFei:
Closes #262


LipuFei:
@whirm ready

PR #1248, opened 2015-02-19
Skip livinglab reporter tests, not used.

whirm:
Also they output xml which seems to break jenkins xunit plugin.


whirm:
@LipuFei ready

Issue #1259, opened 2015-02-21
Unable to install dependencies on clean Ubuntu 14.04


JeffGoderie:
I was unable to install libsodium13 and python-cryptography through the command that you provided in the README.md. I'm using a clean Ubuntu 14.04 Install, and I was not the only one encountering this problem. It returned the following errors:

    E: Unable to locate package libsodium13
    E: Unable to locate package python-cryptography

JeffGoderie:
I've opened a pull-request with an update to the README.md that explains a way to solve this issue.

PR #1260, opened 2015-02-21
Added Install Manual for dependencies to README.md

JeffGoderie:
Added description how to install required dependencies in Ubuntu 14.04, which could not be installed by calling the command listed under "Installing Dependencies". Related to Issue #1259


whirm:
Thanks!

PR #1267, opened 2015-02-23
Replace torrent_file_name column

LipuFei:
1. `thumbnail` column is removed from the `Torrent` table
2. `torrent_file_name` column is replaced with `is_collected`.
3. some other fixes


LipuFei:
@whirm ready

PR #1291, opened 2015-02-25
Reinstated README.md package dependencies one-liner

qstokkink:
By using pip in the dependency one-liner, users will no longer have to build libsodium and python-cryptography themselves.


whirm:
We are not going to tell people to install stuff with pip. At least the manually downloaded .debs will get updated when they switch to a new distro release.

That said, if you want to make a PR with instructions on how to create a virtualenv and install all the dependencies in there it would be welcomed. (you could even provide a requeriments.txt to use with virtualenv)

PR #1296, opened 2015-02-28
No more _encode_decode

NielsZeilemaker:


NielsZeilemaker:
retest this please

NielsZeilemaker:
@whirm @LipuFei it seems to be ready but it fails on the pioneer filter thingy.
However, have you guys looked at the debug screen recently? 
http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester_devel/3571/artifact/output//Screenshot-10.png
It has 0 collected torrents, but it actually downloaded two of them as can be seem in the recently collected. Are you sure the collectedtorrents are working? That would break remotesearch....

synctext:
@NielsZeilemaker  Thank you for that solid cleanup, much more readable! 

@LipuFei @whirm Did the TFTP change broke the "0 collected torrents" stats in the Debug screen?

LipuFei:
@synctext No, I fixed it already so the Debug screen works with TFTP.

NielsZeilemaker:
@LipuFei are you sure, because this branch is devel which already included the tftp stuff

LipuFei:
@NielsZeilemaker @synctext I think the collected torrent number problem has something to do with one of my earlier cleanup. I have removed the torrent_file_name column and added an integer column called "is_collected". I have also changed the `CollectedTorrent` view to use `is_collected` instead.

We still haven't updated the database archive for testing yet, so every time a test starts, it has to run the upgrade script, and the script can be assigning wrong values (zeros) to the `is_collected` column.

I will check this problem.

NielsZeilemaker:
It's an empty database, so that shouldn't be the problem.

2015-03-02 10:53 GMT+01:00 Lipu Fei <notifications@github.com>:

LipuFei:
@NielsZeilemaker Found the problem... I forgot to set the `is_collected` flag when storing the torrent into database. Making PR right now.

LipuFei:
retest this please

PR #1303, opened 2015-03-04
Remove some torrent collecting directory related stuff

LipuFei:


LipuFei:
@whirm ready

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -1078,31 +1077,8 @@ def freeSpace(self, torrents2del):
         # sql_insert =  "insert into Torrent (torrent_id, infohash, relevance) values (?,?,?)"
         # self._db.executemany(sql_insert, torrent_id_infohashes)
 
-        torrent_dir = self.session.get_torrent_collecting_dir()
         deleted = 0  # deleted any file?
         insert_files = []
```
synctext: Is this still used?

LipuFei: This `freeSpace()` used to delete torrent files in the torrent collecting directory to free space. Since we have migrated to `leveldb`, those code doesn't work any more. So no, it's not used. Later, we also need to fix this `freeSpace()` or simply remove it if we don't need it any more.

Tribler/Main/vwxGUI/TorrentStateManager.py
```diff
@@ -89,8 +89,9 @@ def create_and_seed_metadata_thumbnail(self, thumbnail_file, torrent, modificati
                 file_name = u"%s.%s" % (thumbnail_hash_str, fmt)
 
                 sub_file_path = os.path.join(hexlify(torrent.infohash), file_name)
-                sub_dir_path = os.path.join(self.session.get_torrent_collecting_dir(), hexlify(torrent.infohash))
-                file_path = os.path.join(self.session.get_torrent_collecting_dir(), sub_file_path)
+                # TODO
```
whirm: todo what?

Tribler/Main/vwxGUI/home.py
```diff
@@ -968,8 +968,9 @@ def SetData(self, delayedResult):
         torrents = delayedResult.get()
 
         for torrent in torrents:
+            # TODO
```
whirm: idem

whirm:
there's a bunch of TODO's you added, could you add a description to them?

LipuFei:
@whirm Those are the thumbnail stuff. They are disabled now, so we need to fix them later if we plan to use the same design as before. I will add more descriptions.

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
NielsZeilemaker: I would create an issue for this, as without this limit some users will start collecting ridiculous amounts of torrents.
I think @synctext had more than 300.000 at some point.

LipuFei: @NielsZeilemaker Made an issue for this  #1305

PR #1309, opened 2015-03-06
Remove unused rameezSort()

LipuFei:


LipuFei:
@whirm ready. The splash window has been causing a lot of dead object errors recently. We need to investigate after the wx threading work is finished.

PR #1310, opened 2015-03-06
Remove TorrentStatus, Category, and MiscDBHandler

### LINKED ISSUE #1285: Remove TorrentStatus and Category tables
For now we are using an `id (integer) <-> string` mapping for torrent status and categories, but we only have a fixed number of torrent status and categories and they haven't been changed for years, so this whole thing seems to be unnecessary. Also, in the SearchCommunity we are sending torrents with actual category name instead of category_id.

So, I think it is better to remove the two tables `TorrentStatus` and `Category`, and we will get the following benefits:
1. `TorrentStatus` and `Category` tables will be gone
2. `MiscDBHandler` will be gone
3. All code will not need to use `MiscDBHandler` to convert status and categories between id and string all the time, and GUI will also be cleaner.

LipuFei:
Closes #1285


LipuFei:
@whirm ready

Issue #1174, opened 2015-01-27
No longer serve by default as an "Exit node" flag


synctext:
By default Tribler acts as an exit node and relay for others.

Numerous users demand that we only enable relaying by default.
The "exit node" functionality should only be enabled when the user explicity configures this in the configuration panel. 
An exit node user also needs to be connectable in order to support the hidden seeding functionality.
Provide an informational popup _or_ check the Tribler connectability status and only then enable this exit flag.

Implemented as a simple addition to the introduction-request message. Each anonymous relay tunnel community member advertises with two flags if: they are willing to serve as an exit node and if they are connectable.

UPDATE: We're thinking of disabling the "exit" node functionality. Only offer downloading from hidden seeds in upcoming versions.

UPDATE2: in the upcoming version we can add with minor effort an "Exit node" flag to the introduction-request (plus also a connectability flag). So we don't need the credit mining fully operational for proper seeding and finding seeds.

whirm:
Depends on #1066

synctext:
After more thinking and forum discussion I'm now in favor of just adding an "Exit node" flag  in the config screen. It's a quick fix and then we can decide later if we go with the "only darknet mode".
Forum post by Roli: http://forum.tribler.org/viewtopic.php?f=2&t=7143&p=11662&sid=83bbed72b5ce09a38cf3432c8f4110d0#p11635

rjruigrok:
I'll try to get this working in the coming week.

synctext:
@rjruigrok Good to hear progress on this ticket.

@whirm @LipuFei : please review the approach taken by Rob
https://github.com/Tribler/tribler/compare/devel...rjruigrok:exitnode-switch

As suggested we should release this soon. When this is done and there are no other pressing open issues (DB upgrade fully done?) we can build an RC1 and start testing a new release!

Issue #3, opened 2013-03-05
Perfect Darknet Roadmap Ticket


synctext:
- [ ] _long-term tit-for-tat for anonymity_

  InternetOpen community to disseminate sharing ratio of participants. You need to help and relay traffic of others. Related experimental work in the Tor community: [GoldStar, PAR, BRAIDS, LIRA, TEARS, and TorCoin](https://blog.torproject.org/blog/tor-incentives-research-roundup-goldstar-par-braids-lira-tears-and-torcoin) (Tribler Tickets: partly #1230, also #5)
- [ ] _Tribler users **never** become an exit node._ 

  We need to protect our users from lawyer-based attacks. The Internet should have privacy by default. End-to-end encryption is the default and no exceptions. Only offer downloading from hidden seeds in upcoming versions. Additionally, no unprotected uploading/seeding, only through hidden service. We need a require connectability flag: #1174. Faulty approach: #885,     @rjruigrok 

- [ ] _Collective intelligence around reliability of anonymous relay partners_

  Share with the rest of the network who is actively helping others and who is freeriding. Introduce a Bitcoin-like accounting system. Instead of pointless hashing we use proxy helping bandwidth as our proof-of-work. Our protocol follows the Tor specification as much as possible, our circuits have a maximum lifetime of 10 minutes and also terminate after 55MByte of traffic (security limiters). Similar to private Bittorrent communities we use the sharing ratio to identify freeriders. Double signed signature record is created after proxy helping. These proof-of-work records are shared in the community (like Bitcoin public ledger). Each peer builds a database with the reliability of their neighbors, visualize in a graph (locally; no central accountant, controlling server).  #5

- [ ] _real-time statistics of you contributions and tunnel ratio

Notify users they are freeriding and disable anonymous downloading. <be harsh, Tribler is only for social people>.

- [ ] _real-time anonymous tunnels reliability statistics_

Each member of the collective measures their perception of network performance and shows it to the user. Advanced programmer-minded feature and debugging. Add to the Tribler debug screen how many unique neighbors have been seen. Measure the proxy bandwidth donations and online duration (reliability) of yourself and neighbors. Show broken tunnel events. Show overall tunnel + hidden seeding reliability and performance for a tunnel tests and 50MB downloads.

colin1497:
[...]

PR #1325, opened 2015-03-12
Exiting data over tunnels is only allowed after opt-in via settings

rjruigrok:
See #1174


rjruigrok:
retest this please

rjruigrok:
retest this please

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -643,6 +657,10 @@ def check_create(self, messages):
             if self.crypto.key.pub().key_to_bin() != message.payload.node_public_key:
                 yield DropMessage(message, "TunnelCommunity: public keys do not match")
                 continue
+            
+            if not self.settings.become_exitnode and message.payload.become_exit:
+                yield DropMessage(message, "Exit-node functionality disabled")
```
NielsZeilemaker: I would reply with a message saying denied or something like that, then choose a different peer if you receive a denied response. Otherwise creating a tunnel might be ridicously difficult.

Moreover, how are you going to link two tunnels? Are introduction/rendezvous points exit points? Because they do exit_data.

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -440,10 +451,13 @@ def create_circuit(self, goal_hops, ctype=CIRCUIT_TYPE_DATA, callback=None, max_
         self.circuits[circuit_id] = circuit
         self.waiting_for.add(circuit_id)
 
+        # FIXME
```
NielsZeilemaker: Fixme

Tribler/Test/test_tunnel_community.py
```diff
@@ -269,8 +269,9 @@ def create_proxy(index):
             def load_community(session):
                 keypair = dispersy.crypto.generate_key(u"curve25519")
                 dispersy_member = dispersy.get_member(private_key=dispersy.crypto.key_to_bin(keypair))
-                settings = TunnelSettings()
+                settings = TunnelSettings(tribler_session=session)
                 settings.do_test = False
+                settings.become_exitnode = True #become_exit_node
```
NielsZeilemaker: Fixme

NielsZeilemaker:
@whirm same thing is happening here, look it asserts http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester-isolated_devel/292/artifact/output//Screenshot-33.png but "passes" passes all the tests

whirm:
@NielsZeilemaker it's weird.

I'm only hooking on the sys.excepthook to capture unhandled exceptions I don't see how that can happen (If the test raises an exception my wrapper does nothing else than raising the exception again. It's when the test passes that I check for unhandled exceptions: https://github.com/Tribler/tribler/blob/devel/Tribler/Test/test_as_server.py#L56

I can't think of a way this would fail.

Any idea?

NielsZeilemaker:
Maybe because callconditional is setting do_assert to False? Hence, it doesn't assert, it only quits.

LipuFei:
@whirm I agree with @NielsZeilemaker, can be the `do_assert` thing.

whirm:
Well, then why is that test not setting it to True? I thought you where saying this has something to do with the unhandled exception catcher stuff I made.

NielsZeilemaker:
Because we didn't want to assert on the wx thread, but maybe that's OK now. You've modified this code

whirm:
@NielsZeilemaker and how was this supposed to work? how where the tests marked as failed if you weren't asserting? To me it looks like they have been always broken. the last time the assert thing was modified was in 2013...

NielsZeilemaker:
You removed the assert list, we did the asserts after closing tribler

NielsZeilemaker:
@whirm its this one https://github.com/Tribler/tribler/commit/ebe98d61da5bb1aeea880264751d672c4ac99516

whirm:
Oh man, I don't know how I went over that one. So we just need to remove the assertion skip everywhere. I'll make a PR for that.

NielsZeilemaker:
I guess, simply remove the option. But have a look if everything still works. Maybe you need to rewrite it such that if do_assert is false it adds the assert to the list and still only asserts after tribler quit

LipuFei:
@whirm I think we still need the option. For example, in `remote_search` test, the `CallConditional` is used to check if we have enough connections or not. In this case, we may still want to continue the test even if we don't have enough connections because the search function may still work.

NielsZeilemaker:
Passing an assertcallback will allow for that, it will simply replace the assert, and continue

NielsZeilemaker:
@rjruigrok sorry for hijacking your pull request btw

rjruigrok:
@NielsZeilemaker Thanks for your feedback! I'll fix these problems asap. 

Linking the two tunnels is indeed using the exit_data to connect to the rendezvouspoint. A workaround is to force the tunnel creation by pretending it's not going to exit any data, and allow the link-e2e message to be exited .

LipuFei:
Got an error message from `python-cryptograph` on Mac OS X, @rjruigrok @NielsZeilemaker Do you know the problem is?
```
  File "Tribler/dispersy/dispersy.pyo", line 1719, in _update
  File "Tribler/community/tunnel/tunnel_community.pyo", line 806, in on_create
  File "Tribler/community/tunnel/crypto/tunnelcrypto.pyo", line 47, in generate_session_keys
  File "cryptography/hazmat/primitives/kdf/hkdf.pyo", line 111, in derive
  File "cryptography/hazmat/primitives/kdf/hkdf.pyo", line 94, in _expand
  File "cryptography/hazmat/primitives/hmac.pyo", line 32, in __init__
  File "cryptography/hazmat/backends/multibackend.pyo", line 105, in create_hmac_ctx
UnsupportedAlgorithm: sha256 is not a supported hash on this backend.
```

LipuFei:
@rjruigrok @NielsZeilemaker Never mind. I found the issue.

rjruigrok:
retest this please

synctext:
Form the ticket: ``An exit node user also needs to be connectable in order to support the hidden seeding functionality.``

When this finally passes, another pull request is needed to take care of detecting connectability?
Some of the code suggests it already includes this logic, but it looks incomplete (``self.connectable_candidates = {}``)

synctext:
retest this please

synctext:
@rjruigrok Looks broken:
```
17:38:23 [-] [wrap_in_vnc.sh run_n...] ERR: test_anon_download (Tribler.Test.test_tunnel_community.TestTunnelCommunity) ... ok
17:38:40 [-] [wrap_in_vnc.sh run_n...] ERR: test_anon_tunnel (Tribler.Test.test_tunnel_community.TestTunnelCommunity) ... [xcb] Unknown request in queue while dequeuing
17:38:40 [-] [wrap_in_vnc.sh run_n...] ERR: [xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
17:38:40 [-] [wrap_in_vnc.sh run_n...] ERR: [xcb] Aborting, sorry about that.
17:38:40 [-] [wrap_in_vnc.sh run_n...] ERR: python: ../../src/xcb_io.c:179: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.
```

rjruigrok:
Strange error. It's not fully clear to me why the test fails - yesterday the creation of anonymous tunnels was working with the same code, error doesn't look directly related to the tunnelcommunity code. Will have a look into this tomorrow. 

Regarding the connectability of peers: I already added this to the payload messages for dispersy-introduction-request / response as a stub so we don't need to change the community identifier again on the next pull request which supports the connectability flag (although I'm not sure about this because other hidden services related payloads may change, e.g. for circumventing the connectability problem on the end-to-end linking to the rendezvous-point which I like to be punctured)

NielsZeilemaker:
@rjruigrok the automatic download should be disabled during the unit-tests, can you make that happen?

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -758,6 +800,30 @@ def _ours_on_created_extended(self, circuit, message):
             from Tribler.Core.simpledefs import NTFY_TUNNEL, NTFY_CREATED, NTFY_EXTENDED
             self.notifier.notify(NTFY_TUNNEL, NTFY_CREATED if len(circuit.hops) == 1 else NTFY_EXTENDED, circuit)
 
+
+    def on_introduction_request(self, messages):
+        connectable = False # Todo. Unused so far
```
NielsZeilemaker: There is actually a connectable flag implemented in Dispersy, each walkcandidate reports its connectability.
See https://github.com/Tribler/dispersy/blob/devel/candidate.py#L130

synctext:
"Showing  65 changed files  with 1,238 additions and 58 deletions"

Why the rename to ``TestOrginal``, this touches a lot of files?

btw you use your own Dispersy, mainline had several fixes past 48h.

rjruigrok:
I'm trying to figure out why the 'anonymous test' is added to the downloads in the test. The only way of testing this is by running it on jenkins, but with the complete test-suite this takes hours. Therefory I temporarily moved the tests and revert them when I found the cause of it.

NielsZeilemaker:
Locally you can simply run one test. I think that's a better approach than
exploiting jenkins in this manner.

2015-03-25 13:45 GMT+01:00 Rob Ruigrok <notifications@github.com>:

rjruigrok:
@NielsZeilemaker You're right. I also find the cause - in the test_tunnel_community.py it creates a TunnelSettings object with do_test = False. But because the tests run via the GUI, the tunnelsettings got overwritten in tribler_main.py. The tunnelsettings in test_tunnel_community have been obsolete all the time, but problems are now caused because I increased the time-limit for the hidden-services test. 

I don't have an idea how to fix this at the moment, so have disabled adding the anonymous download for now. This bug looks related to #1353.

NielsZeilemaker:
@rjruigrok I think your problem is different. We simply don't want the anon download during the unit-tests, hence a is_unit_testing variable similar to
https://github.com/Tribler/tribler/commit/597c4038729260abea09e90ce021865516dd02f8
would suffice.

NielsZeilemaker:
Or add another global in tribler_main, like 
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/tribler_main.py#L129

rjruigrok:
retest this please

synctext:
retest this please

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -184,7 +186,16 @@ def __init__(self, install_dir=None):
         self.max_traffic = 55 * 1024 * 1024
 
         self.max_packets_without_reply = 50
-
+        
+        if tribler_session:
+            state_dir = tribler_session.get_state_dir()
```
NielsZeilemaker: use can actually replace these lines by tribler_session.get_tunnel_community_exitcode_enabled()

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -277,24 +290,47 @@ def start_download_test(self):
...
 
     def initiate_meta_messages(self):
-        return super(TunnelCommunity, self).initiate_meta_messages() + \
+        messages = super(TunnelCommunity, self).initiate_meta_messages()
+        copy_messages = messages[:]
+        for message in copy_messages:
+            if message.name == "dispersy-introduction-request" or message.name == "dispersy-introduction-response":
+                messages.remove(message)
+            
+        messages += [Message(self, u"dispersy-introduction-request",
```
NielsZeilemaker: Have a look at how we're replacing the the payload in the discovery community
https://github.com/Tribler/dispersy/blob/devel/discovery/community.py#L232
I think its slightly nicer

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -758,6 +799,28 @@ def _ours_on_created_extended(self, circuit, message):
...
 
+
+    def on_introduction_request(self, messages):
+        exitnode = self.settings.become_exitnode
+        extra_payload = [exitnode]
+        super(TunnelCommunity, self).on_introduction_request(messages, extra_payload)
+        for message in messages:
+            pubkey = message.candidate.get_member().public_key
+            connectable = message.candidate.connection_type == u"public"
+            self.exit_candidates[pubkey] = message.payload.exitnode and connectable
```
NielsZeilemaker: Are you also cleaning this dict?

synctext:
@rjruigrok  Updated the roadmap. Now rendezvous  and exit nodes are shown as different issues (connectable versus volunteer). https://github.com/Tribler/tribler/issues/3

Tribler/community/tunnel/tunnel_community.py
rjruigrok: You're right! I already thought that this should be much easier, but first did not work (got the error that this method was not existing in class Session), but after manual removing of the compiled python files it worked... :+1:

Tribler/community/tunnel/tunnel_community.py
rjruigrok: Added cleanup for this dict, by asking for the list of verified dispersy candidates (which I assume are already cleaned up from time to time) and prune all non-existing candidates from the dict.

NielsZeilemaker: Then just remove the doctor and simply switch to yield verified peers, those have the connection type flag

rjruigrok: Hi Niels,
What do you exactly mean with removing the doctor, just yielding verified
peers? The exitnode flag for each peer that gets introduced is the most
important thing to record here, and connectability is an extra constraint
on this, should connectability be checked somewhere else?

Op donderdag 26 maart 2015 heeft Niels Zeilemaker <notifications@github.com>
het volgende geschreven:

NielsZeilemaker: Sorry about that, the autocorrect didn't really help. The peers yielded by the veriffied peers stilll have the connection type flag, so you can just use that method.
However, if you keep-alive/ping connecties to other peers as well, then make sure to only clean the dict if you drop a connection from the keep-alive list.

Tribler/Main/vwxGUI/MainFrame.py
```diff
@@ -614,8 +615,12 @@ def do_gui():
     def monitorHiddenSerivcesProgress(self, ds, tdef, dscfg, selectedFiles):
         if ds.get_status() in [DLSTATUS_ALLOCATING_DISKSPACE, DLSTATUS_HASHCHECKING, DLSTATUS_WAITING4HASHCHECK]:
             return (5.0, True)
-
-        if ds.get_current_speed(DOWNLOAD) == 0:
+        
+        state_dir = self.utility.session.get_state_dir()
```
NielsZeilemaker: swtich to using session.get_tunnel...

Tribler/Main/vwxGUI/settingsDialog.py
```diff
@@ -207,6 +209,19 @@ def saveAll(self, event):
...
+        becomeExitNode = self._become_exitnode.IsChecked()
+        if becomeExitNode != scfg.get_tunnel_community_exitnode_enabled():
+            scfg.set_tunnel_community_exitnode_enabled(becomeExitNode)
+            self.saveDefaultSessionConfig(scfg)
+            restart = True
+
+        switchHsOnTimeout = self._switch_hs_timeout.IsChecked()
+        if switchHsOnTimeout != scfg.get_tunnel_community_hs_timeout_switch():
+            scfg.set_tunnel_community_hs_timeout_switch(switchHsOnTimeout)
+            self.saveDefaultSessionConfig(scfg)
```
NielsZeilemaker: There's no need for this, the scfg is saved lateron, and also remove the saveDefaultSessionConfg method.

synctext:
retest this please

Tribler/Main/vwxGUI/settingsDialog.py
```diff
@@ -729,3 +749,29 @@ def __create_s5(self, tree_root, sizer):
...
+        exp_s1_sizer.Add(self._become_exitnode, 0, wx.EXPAND)
+        self._switch_hs_timeout = wx.CheckBox(exp_panel, label="Switch from hidden services to exit nodes")
+        exp_s1_sizer.Add(self._switch_hs_timeout, 0, wx.EXPAND)
+        
+        exp_s1_faq_text = wx.StaticText(
+            exp_panel, label="By allowing Tribler to be an exit node, it's possible to become a proxy for someone elses traffic. \nThis may cause problems in some countries.")
+        exp_vsizer.Add(exp_s1_faq_text, 0, wx.EXPAND | wx.TOP, 10)
+
+        # load values
+        state_dir = self.utility.session.get_state_dir()
```
NielsZeilemaker: Don't load the sessionconfig use session:
```
self.utility.session.get_tunnel....

File: Tribler/community/tunnel/tunnel_community.py [hunk 11]
```diff
@@ -277,25 +294,46 @@ def start_download_test(self):
...
         return [master]
 
     def initiate_meta_messages(self):
-        return super(TunnelCommunity, self).initiate_meta_messages() + \
-            [Message(self, u"cell", NoAuthentication(), PublicResolution(), DirectDistribution(),
+        meta_messages = super(TunnelCommunity, self).initiate_meta_messages()
+        for i, mm in enumerate(meta_messages):
+            if mm.name == "dispersy-introduction-request":
+                meta_messages[i] = Message(self, u"dispersy-introduction-request",
+                                     MemberAuthentication(),
```
NielsZeilemaker: I would switch to mm.authentication etc. like we do in the discoverycommunity

Tribler/community/tunnel/tunnel_community.py
rjruigrok: Not fully clear to me yet :) Do you mean line 879 where I should use the connection_type flag, so you mean it's not necessary to maintain the connectable flag in the dictionairy? 

Does dispersy have some method taking care of removing candidates of the candidates list, so this method can be overrided in the tunnel-community? That would be even nicer because it would not be necessary anymore to iterate over lists of verified peers / dictionaries to clean the dictionairy up.

NielsZeilemaker: The candidates which are returned by self.dispersy_yield_verified_candidates() have a connection_type property. I think those will suffice. 
No need for this new dict. You can simply check the connectiontype in line 876

rjruigrok:
retest this please

synctext:
retest this please

rjruigrok:
retest this please

rjruigrok:
Ready to merge

NielsZeilemaker:
@rjruigrok its looking great, but one last question. Could you add a test that checks that if we don't have any peers which are marked as an exit, that the downoad does not pass. E.g. check if the do not exit flag actually works.

rjruigrok:
Thanks! I'll add these tests. 

I'm trying for hours now to run tests locally, it worked in the past, but they keep crashing without any meaningful error now - so am forced to run these tests with jenkins. My apologies for this 'spamming'. :)

NielsZeilemaker:
@rjruigrok they are looking great. Maybe one more, a peer which changes his mind. I.e. first being an exit-node and then switching to not-being an exit-node.
This also allows you to test the exit_data check.

rjruigrok:
@NielsZeilemaker You're right - that would be a good test. I started implementing it in another branch for easier testing on jenkins - but am stuck on this. It looks like when no data is exited (due to refusing of exit_data method) - the test will hang forever. Do you have an idea how to circumvent this loop, or (force) end the test? 

See http://jenkins.tribler.org/job/Experiment_Tunnel_isolated_gorillas/119/ for the result in jenkins. 
Implementation of the test is in https://github.com/rjruigrok/tribler/blob/exitnode-switch-copy/Tribler/Test/test_tunnel_community.py#L170

NielsZeilemaker:
No clue, I would have to check with a debugger. I would add some error log statements to see if the test methods are being called or not.

rjruigrok:
@NielsZeilemaker I was a bit busy last week, but the extra test is added and now working as expected. :)

Tribler/Main/vwxGUI/MainFrame.py
```diff
@@ -56,6 +56,7 @@
 from Tribler.Main.vwxGUI.SRstatusbar import SRstatusbar
 
 from Tribler.Core.Video.utils import videoextdefaults
+from Tribler.Core.SessionConfig import SessionStartupConfig
```
NielsZeilemaker: Remove this line

Tribler/community/tunnel/conversion.py
```diff
@@ -58,13 +85,13 @@ def _decode_cell(self, placeholder, offset, data):
 
     def _encode_create(self, message):
         payload = message.payload
-        packet = pack("!IHH20s", payload.circuit_id, len(payload.node_public_key),
-                      len(payload.key), payload.node_id) + payload.node_public_key + payload.key
+        packet = pack("!IHH20s?", payload.circuit_id, len(payload.node_public_key),
+                      len(payload.key), payload.node_id, payload.exit_candidates) + payload.node_public_key + payload.key
```
NielsZeilemaker: Hmm, so now the node before the exit-node knows he's n-1. Is that a problem?

Tribler/dispersy
```diff
@@ -1 +1 @@
-Subproject commit fec55be1708ac9fd70bdde9714f71e58f88d7697
+Subproject commit 2c6efd6d2661f96dc0180d0acdc96afac5590382
```
NielsZeilemaker: Revert this, fec55 is newer.

Tribler/community/tunnel/conversion.py
rjruigrok: Node n-1 still does not know what n (length of the circuit) is and what data is exited over the circuit. To achieve the same level of anonimity the circuit length should be at least 3 hops to prevent node n-1 from knowing both its predecessor and succesor (in the case of 2 hops node n-1 knows both the exit-node and initator, that's unsafe). 

I don't think there is a way to prevent this leaking, because the last hop is always responsible of returning a list of candidates. Btw, it is a temporary solution for anonymous downloading. When hidden seeding will become the standard in Tribler, exit-nodes are not relevant anymore.

NielsZeilemaker: The initiator of a circuit is responsible for choosing the exit-node. It will choose a peer which has the exit-flag enabled. Do we really need this flag here? I'm not so sure.

rjruigrok: The problem is the number maximum of 4 possible candidates returned by the last hop. When the next hop becomes the endpoint/exit-node, n-1 should at least return 4 candidates with at least 1 exit-node-enabled candidate. Chances are small, because most users prefer not to be exit-node. That's why the flag is added here.

NielsZeilemaker: But the exit-node must be connectable, hence we could just send an ip-address to which the n-1 node should connect to. Won't solve the "n-1 node knows he's the n-1 node". But it could be a nice change to the create/extend messages (to be able to specify an ip in addition to specifying a pubkey).

rjruigrok: You mean by appending the payloads of create and extend with an ip-field? But there is already support for creating circuits to any 'required_exit'. Are there more benefits for adding such an ip-field to the payload?

NielsZeilemaker: I don't have the details of the protocol in my head, but I really like if peers would choose the exit-node instead of relying on the n-1 peer. Additionally, it would remove the need of this exit_candidates flag

Tribler/dispersy
rjruigrok: Strange, I think it's older? 
See https://github.com/Tribler/dispersy/commits/devel

NielsZeilemaker: Sorry, my bad

synctext:
retest this please

rjruigrok:
I stll have to clean up code in the payloads, but in the current code the
initiator of a circuit will choose the exit itself now.

Op donderdag 9 april 2015 heeft tribler-ci <notifications@github.com> het
volgende geschreven:

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -401,6 +435,17 @@ def do_remove(self):
             elif exit_socket.bytes_up + exit_socket.bytes_down > self.settings.max_traffic:
                 self.remove_exit_socket(circuit_id, 'traffic limit exceeded')
 
+        # Remove exit_candidates that are not returned as dispersy verified candidates
+        current_candidates = {}
```
NielsZeilemaker: you could use a set here
```
current_candidates = set(c.get_member().public_key for c in self.dispersy_yield_verfified_candidates())

**Comment by NielsZeilemaker** [comment]
Overall, its looking great. Maybe add some comments when choosing which candidate to select etc. But its ready to be merged. Great work @rjruigrok

**Comment by rjruigrok** [comment]
Thanks for all your feedback @NielsZeilemaker :+1: 
I'll add some more comments tonight on the crucial places where choices are made etc. 

In my last commit there is a new problem introduced, I suspect the pull request tester / jenkins is not connectable at all. Am I right?

**Comment by NielsZeilemaker** [comment]
I guess it's not, but why did it start to fail now and not before?

File: Tribler/Test/test_tunnel_community.py [hunk 16]
```diff
@@ -66,6 +66,135 @@ def do_create_local_torrent(_):
 
         self.startTest(do_create_local_torrent)
 
+    @timed(180)
```
NielsZeilemaker: I would remove these @timed stuffs, thats why we created the self.Call stuffs.

rjruigrok:
Until today the check on connectability was implemented in the on_create method, where a list of possible (exitnode + connectable) candidates were returned, and connectability was checked per candidate from the list of verified candidates. 

The connectability check is now performed by the peer itself, and causing problems, but I thought self._dispersy would know about its own connectability? Or is there a better way for doing this?

rjruigrok:
I think i've found the issue, the introduction walker in dispersy returns the exitnode-flag as extra payload, but at this moment the node is not yet connectable, thus returning False as the exitnode-flag. Will fix this.

NielsZeilemaker:
Dispersy will guess its own connectability, but maybe its not know at the start. Maybe add a fallback which will attempt to connect to a non-connectable exit-node if no connectable ones are available and hope for the best?

rjruigrok:
I moved the check for connectability for the exit node candidate to the place where the initial peer prepares the extension to the exit-node. This seems to work, because the dispersy introduction has already taken place and it guessed connectability on that point. :)

rjruigrok:
Yay - it works!

synctext:
retest this please

rjruigrok:
In my opinion ready for merge.

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -697,19 +745,33 @@ def _ours_on_created_extended(self, circuit, message):
...
+                    # The exit node is a verified connectable exit node peer chosen by the circuit initiator
+                    extend_hop_public_bin = None
+                    for c in self.dispersy_yield_verified_candidates():
+                        pubkey = c.get_member().public_key
+                        exit_candidate = self.exit_candidates[pubkey]
+                        if exit_candidate.become_exit and self.candidate_is_connectable(c) and pubkey not in ignore_candidates:
+                            wan = c._wan_address
+                            extend_hop_public_bin = pubkey
+                            extend_hop_addr = wan
+                            break
```
NielsZeilemaker: I would move this logic into the create_circuit method. Then if a required_exit is passed, do nothing, but if the required_exit is None then look for an exit-node in the verified candidates.
This will ensure that we never attempt to create a circuit without having an exit-node.

NielsZeilemaker:
retest this please

rjruigrok:
It looks like the connectability check fails so it won't find an exit node. Strange, so have to debug why it does not work...

NielsZeilemaker:
You're creating 3 exit-nodes by default, and are waiting for 4 tunnels. Because the peer is now choosing the exit-nodes beforehand, it will only create 3 tunnels instead of 4, as its checking if an exit-node isn't used before.

NielsZeilemaker:
Hmm, its not checking if the exit-node is used before. But still, it seems highly coincidental.

rjruigrok:
Ah - that might be the cause. What do you think, is it necessary for checking whether an exit node is used before? I was not expecting this 'checking', but from a privacy perspective it might be a good idea not to reuse exit-nodes. 

Have to test, but think it is caused by the check `c.sock_addr not in hops`?

NielsZeilemaker:
That was my guess, but im not so sure anymore. I think its wise to attempt
to construct a new tunnel to an exit-node we're not using at the moment.
But have a fallback if only used exit-nodes are available.

2015-04-15 16:02 GMT+02:00 Rob Ruigrok <notifications@github.com>:

rjruigrok:
@NielsZeilemaker It still crashes, apparently not because the checking if the exit-node is used before (last jenkins run failed with just 2 circuits), but because of crypto errors. I think the public key passed to the required_exit fails, but don't know why because it's the same as in the original method. Do you have an idea where it's going wrong? 

I can't get the tunnel thing to work on my computer since yesterday, in dispersy debug frame the tunnelcommunity has 0 candidates, sometimes it takes more than half an hour to get candidates, which makes it almost impossible to debug the code. I'm getting frustrated more and more :)

rjruigrok:
I Figured it out, the `_ours_on_created_extended` was always extending to the `required_exit` for any hop. That's not what should happen, fixed this and the circuit creation works again. :+1: 

Just one strange thing (not going to look at this because it is not supposed to happen anyway): the crypto fails when the same hop is used multiple times in a circuit... But why?

synctext:
retest this please

rjruigrok:
Test failed because of remote torrent search tests. Tunnels passed all tests now. :+1:

NielsZeilemaker:
That's great news, I'll merge it tomorrow. Then I'll have some time. Maybe I can create a unit test to verify the same member used in one circuit crypto bug and attempt to fix it

NielsZeilemaker:
retest this please

rjruigrok:
The hidden sevices test failed - the error is crypto-related, looks the same as I had this afternoon where the same member was used twice in a circuit. 


> DEBUG   1429130028.67 tunnel_community:857   Got establish-rendezvous (624069349) from ('95.211.198.142', 35504)
DEBUG   1429130028.67   SocketHandler:498   SocketHandler: Got UDP data ('95.211.198.142', 14371) len 333
ERROR   1429130028.67        dispersy:1724  exception during handle_callback for cell
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/dispersy/dispersy.py", line 1719, in _update
    messages[0].handle_callback(messages)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/tunnel_community.py", line 865, in on_cell
    encrypted = self.crypto_in(circuit_id, encrypted)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/tunnel_community.py", line 1179, in crypto_in
    content, hop.session_keys[ORIGINATOR], hop.session_keys[ORIGINATOR_SALT])
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/crypto/tunnelcrypto.py", line 81, in decrypt_str
    return cipher.update(content[24:]) + cipher.finalize()
  File "/usr/lib/python2.7/dist-packages/cryptography/hazmat/primitives/ciphers/base.py", line 112, in finalize
    data = self._ctx.finalize()
  File "/usr/lib/python2.7/dist-packages/cryptography/hazmat/backends/openssl/ciphers.py", line 146, in finalize
    raise InvalidTag
InvalidTag

rjruigrok:
retest this please

rjruigrok:
retest this please

NielsZeilemaker:
Hi @rjruigrok let's attempt to debug this thing. Let's start by fixing some stuff which I can see during the unit-tests.
First, there are multiple thing going wrong in this image:
http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester_devel/3963/artifact/output//Screenshot-36.png
* I can see an ip adress (or the start of one) in the bytes column, that can't be right 
* There are 8 circuits created, I would expect only 4 right? (who's filling that circuits_needed dict?)

NielsZeilemaker:
This is magic, I've found the circuits_needed logic.
https://github.com/Tribler/tribler/blob/a831c2d24f85f81b51d74a74b4b70be78c5bd708/Tribler/Core/Libtorrent/LibtorrentMgr.py#L134

We should eventually move this logic to the community itself, but basically an anonymous download needs 1 circuit, and a normal download requests max_circuits and waits unitl min_circuits to start downloading.

NielsZeilemaker:
Any progress on the bytes send thingy?
I've seen this bug btw.
```
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/hidden_community.py", line 54, in on_timeout
    self._logger.debug("RPRequestCache: no response on establish-rendezvous (circuit %d)", self.rp.circuit_id)
exceptions.AttributeError: 'RendezvousPoint' object has no attribute 'circuit_id'

**Comment by rjruigrok** [comment]
I'm currently working on this. :+1: Still have to figure out why there is so much traffic on one of those circuits - I suspect this is the circuit causing the cryptography errors too.

**Comment by NielsZeilemaker** [comment]
Its also generating quite some cpu load (probably caused by the high traffic), which might also cause the crypto to mess up.

**Comment by rjruigrok** [comment]
I moved the code with the `needed_circuits` logic to the tunnelcommunity. But after this (small) change it was not creating circuits at all anymore, very strange. Figured out that this was because of the connectability check in `create_circuit`. Not really clear to me what caused the connectability checks to fail after days of passed tests, but I've commented out this part for now - will have to figure out if there is a better way to check someone's connectability. Currently I'm checking the `conn_type` for equality to `public`, but is this always correct / reliable?

**Comment by NielsZeilemaker** [comment]
It should be, but i'm not sure if it always is when starting a couple peers in the background. Maybe you could (similarly to choosing an already used exit-node) fallback to a "non-connectable" exit-node.

**Comment by NielsZeilemaker** [comment]
Another small bug
```
16:54:21 [-] [wrap_in_vnc.sh run_n...] ERR:   File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/hidden_community.py", line 55, in on_timeout
16:54:21 [-] [wrap_in_vnc.sh run_n...] ERR:     self.community.remove_circuit(self.circuit.circuit_id, 'establish-rendezvous timeout')
16:54:21 [-] [wrap_in_vnc.sh run_n...] ERR: exceptions.AttributeError: 'RPRequestCache' object has no attribute 'circuit'

rjruigrok:
Yep already saw that one, fixed it in the later commit. :) I also added the fallback thing for the connectability check too, that's indeed a good idea. Still something is going wrong on tunnel creation, even the downloading is not working anymore (but without any errors). 

Btw last jenkins test got strange results, with exiting images in it. http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/3971/

NielsZeilemaker:
@rjruigrok have a look at test_downloadfromfile that seems to have broken, forgetting to remove a session instance, cause all other tests to refuse to run as well.

NielsZeilemaker:
Maybe add a 
```
self.assertFalse(Session.has_instance(), .....)
```
here
https://github.com/Tribler/tribler/blob/devel/Tribler/Test/test_as_server.py#L101
and here
https://github.com/Tribler/tribler/blob/devel/Tribler/Test/test_as_server.py#L305

Then we'll know whom is leaving behind a Session instance, and all gui tests wont start if a Session instance is present.

rjruigrok:
@NielsZeilemaker  You're a genius! 
Looks like the test_downloadfromfile failed because of the tunnels_ready = 1 thing, which crashes, and leaves a session behind which messed up all other tests. Now waiting for the remaining test results.

rjruigrok:
Hmm weird, failure without any errors. It just hangs...

NielsZeilemaker:
retest this please

rjruigrok:
retest this please

synctext:
retest this please

NielsZeilemaker:
retest this please

rjruigrok:
The crypto bug still occasionly occurs, let's wait for the test result... Hopefully it passes again but it's a wild guess :)

When looking to the output in jenkins, receiving/decoding data on the circuits to the rp / ip sometimes fails. The error was the same as the error when the same peer is used more than once in a circuit, but with the current code I don't see a place where this is still possible. What's the best way of debugging this kind of crypto problems?

rjruigrok:
retest this please

synctext:
retest this please

rjruigrok:
I tried to figure out the crypto bug but really have no clue. It seems it always occurs in the rp /ip circuit (not the linking one) with length 2, while peeling off data from the first layer. sometimes the first cell sent for the circuit succeeds and the next cells all fail. And sometimes, like the last test, all tunnel tests run without issues. Anyone having a suggestion or idea what the cause could be? Problem started after moving the exit node choice to the create_circuit method.

NielsZeilemaker:
It's decrypting with the wrong key, that's why we get the error. I'll have
a look what causing it when I have some time.  But for now, I don't think
it related to your changes
Op 19 apr. 2015 8:38 PM schreef "Rob Ruigrok" <notifications@github.com>:

synctext:
Thanks for the insight @NielsZeilemaker. Does that mean this PR can be pushed through? Plus newer ones are opened for connectability, move the 4-8 creation magic into community.py etc..

NielsZeilemaker:
Jup, that would be my suggestion
Op 19 apr. 2015 10:12 PM schreef "Johan Pouwelse" <notifications@github.com

rjruigrok:
@NielsZeilemaker Finally resolved the issue! You were wrong - It __was__ related to my changes. :-)

The problem was caused by the selection of the `first_hop` in `create_circuit`. I discovered the issue after a lot of debugging with the tunnel messages send along. It was selecting the first hop without checking whether it would also become the required_exit, hence the circuit got the same hop twice in the circuit (as first hop and as exit node). It's clear to see in the following log. Note the `using port 17410`, and a few lines further the `send created` is sent to candidate 17410. 

```
Exit-node enabled
ERROR   1429565522.64   LibtorrentMgr:109   could not restore dht state, starting from scratch
Using port 17410
ERROR   1429565563.45 tunnel_community:868   Got create (2270751903) from ('10.0.2.15', 45321)
ERROR   1429565564.30 tunnel_community:677   send created to 1 candidates: ['{10.0.2.15:45321 84.105.131.99:57987}']
ERROR   1429565565.28 tunnel_community:868   Got extend (2270751903) from ('10.0.2.15', 45321)
ERROR   1429565565.28 tunnel_community:1211  Decrypt layer for relaying own circuit 2270751903 with session key 적�Ŏ@:Q�?P$	
ERROR   1429565565.28 tunnel_community:677   send create to 1 candidates: ['{10.0.2.15:17410}']
ERROR   1429565565.28 tunnel_community:868   Got create (2584804787) from ('10.0.2.15', 17410)
ERROR   1429565565.29 tunnel_community:677   send created to 1 candidates: ['{10.0.2.15:17410}']
ERROR   1429565565.31 tunnel_community:868   Got created (2584804787) from ('10.0.2.15', 17410)
ERROR   1429565565.31 tunnel_community:1189  Encrypt layer for relaying own circuit 2270751903 with session key �����?7����6
ERROR   1429565565.31 tunnel_community:677   send extended to 1 candidates: ['{10.0.2.15:45321}']
ERROR   1429565569.17 tunnel_community:1226  Decrypt layer for relaying data to exit node for circuit 2270751903 with session key 적�Ŏ@:Q�?P$	
ERROR   1429565569.17 tunnel_community:677   send data to 1 candidates: ['{10.0.2.15:17410}']
ERROR   1429565569.17 tunnel_community:1221  Encrypt layer for relaying data to origin for circuit 2584804787 with session key 8�K����{���
ERROR   1429565569.17 tunnel_community:677   send data to 1 candidates: ['{10.0.2.15:45321}']
```
Although it should theoretically be possible to use the same hop multiple times in a circuit, the crypto in its current state will crash. You might solve this bug, but it eventually helped us to identify the problem of reusing hops in the same circuit... As long as everybody knows that failures in the crypto can be caused by reuse of hops, this should be fine. What do you think? 

BTW I'm already preparing a pull request which resolves this issue (and which adds some debugging info)

NielsZeilemaker:
For now, avoiding the same member twice seems to be a solution. I'll have a look if I can fix this bug later on.

PR #1336, opened 2015-03-16
remove do_assert arg as now we are registering all raised exceptions.

whirm:


whirm:
retest this please

NielsZeilemaker:
It doesn't look good, I would revert both this and the assert_list commit and try again.
I'm afraid that asserting in the wx-thread will break quitting tribler. Basically the self.quit call is simply a request to quit. It's not blocking. Hence, asserting just after will cause some serious issues while trying to quit.

When we implementing this before, it took us a long time before we could reliably close tribler whenever something happened. It might not be pretty, but at least it worked.

whirm:
I'll temporarily revert this while we work on completely separating the UI from the rest of the code.

NielsZeilemaker:
I would revert the assert_list as well, in order to fix the unit-tests.

whirm:
That's what I meant, this PR hasn't been merged. I've created a PR with the reversal.

PR #1348, opened 2015-03-20
Fix FamilyFilter capital letter problem

LipuFei:


NielsZeilemaker:
Ok, so I've traced the FF problems. The ff works on lowercase categories, see:
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/list.py#L830

But here it could be comparing it against a 0:
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/list.py#L862
I guess that needs to be changed to ''

Here I would add a lower():
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/list.py#L863
because category has a capital, and the enabled_category_list only contains lowercase categories.

Add an assert to the guidbtuple torrent constructor which enforces that category is a string, as currently it seems to be all over the place.

This one seems ok, but it's a bit strange that we have a ff in two places?
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/SearchGridManager.py#L1354

LipuFei:
@NielsZeilemaker It used to use the `Category` table in the database to get category names and ids. This problem showed up after I removed that table. I am also refactoring the Category and FamilyFilter at the moment and I will fix this uppercase lowercase problem as well. Thanks for the help!

After the refactory, the `family_filter` option will be moved from tribler GUI config part to `SessionConfig`. `Category` and `FamilyFilter` will know by themselves if the family filter is enabled or not, so we can remove a lot of those family-filter-related code from GuiUtility and GUI.

LipuFei:
@NielsZeilemaker I will first make a fix for this problem.

NielsZeilemaker:
But be very carefull, because this stuff is as fragile as can be.

LipuFei:
retest this please

LipuFei:
@NielsZeilemaker @whirm Should be OK now.

NielsZeilemaker:
I guess this one: https://github.com/LipuFei/tribler/blob/fix_ff_channel/Tribler/Main/vwxGUI/list.py#L862
Should be u"unknown" instead of 0

NielsZeilemaker:
And have a look to see if you can remove _applyFF from searchgridmanager, I have a feeling were actually filtering channeltorrents twice.

LipuFei:
@NielsZeilemaker Thanks!

PR #1350, opened 2015-03-23
Force use of system's python.

### LINKED ISSUE #1346: Custom python installation breaks tribler
When there is a file /usr/local/bin/python, tribler (6.4.3) won't start. After removing or renaming this file, tribler works fine. It seems tribler doesn't explicitly rely on system python (/usr/bin/python), but to python set by the current environment, which, if custom python is installed, is /usr/local/bin/python.

Please update all the shebangs in tribler's python files so that they explicitely refer to /usr/bin/python.
 .

whirm:
Instead of the first one on the PATH.

Closes #1346

### DIFF
--- a/debian/bin/tribler
+++ b/debian/bin/tribler
@@ -7,4 +7,4 @@ export PYTHONPATH="$PYTHONPATH":$_TRIBLERPATH
 
 echo "Starting Tribler..."
 cd $_TRIBLERPATH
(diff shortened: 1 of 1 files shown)


whirm:
@LipuFei ready

LipuFei:
@whirm Should it be `/usr/bin/python2.7` instead?

whirm:
No if we are still compatible with 2.6.

PR #1338, opened 2015-03-17
Revert "Don't raise the assert exceptions at tearDown"

whirm:
This reverts commit ebe98d61da5bb1aeea880264751d672c4ac99516.

PR #1343, opened 2015-03-18
Revert searchcommunity update

whirm:
Until we decide what to do about the backward incompatibility.

PR #1352, opened 2015-03-24
Fix unit tests

LipuFei:
This one includes everything in PR #1338


LipuFei:
retest this please

NielsZeilemaker:
@LipuFei for my sanity, you created a couple peers in the new search community, in order to help the unit-tests right? Could you verify if those are still running/restart them?

LipuFei:
@NielsZeilemaker No, I didn't. This only happens in the isolated PR tester. It will create 20 peers for each test. Since it has been disabled, there is no peers that we can reach.

NielsZeilemaker:
I have a feeling that there a simply no more peers than the 5 we discover. Let's merge #1343 to see if that fixes them.

LipuFei:
@NielsZeilemaker I will need to add some changes to #1343 before we can merge it. The search community used to send **a list of categories** in the torrent messages. Because we never have more than one category and we have changed the identifier, I changed it to a single value. So I need to revert this as well.

LipuFei:
@NielsZeilemaker I've cherry-picked the commits in #1343 and reverted the category list. Let's see what happens now.

NielsZeilemaker:
Nice

LipuFei:
retest this please

NielsZeilemaker:
So I guess it's fixed. Let's implement a new test which only uses the allchannel community to search. That is probably still broken. 
I guess overwriting the search method of the search community will do.

LipuFei:
@NielsZeilemaker Ok, I will do the refactory in the same way as the SearchCommunity. Then, we can use only the allchannel community to search.

NielsZeilemaker:
Seems like a great plan, but first merge all/most pull requests before opening new ones

LipuFei:
@NielsZeilemaker I finished the refactory. It's not very small (+118, -52). Shall we merge this one first and then I create another PR for the refacotry?

NielsZeilemaker:
Yes, I would merge this one and some of the other pull requests and then add the new test

PR #1353, opened 2015-03-24
Refactor AllChannel and separate channel and torrent search tests

LipuFei:
1. Refactor AllChannelCommunity in the same way as SearchCommunity
2. Separate channel and torrent search tests.
3. Make search tests to use a custom session config to control which communities to load.


LipuFei:
@NielsZeilemaker I've run several times and the channel search (in the GUI) is not working properly. I will try to figure this out tomorrow.

NielsZeilemaker:
I'm seeing
```
raise RuntimeError("Session is singleton")
```
you must be missing a delInstance somewhere.

LipuFei:
@NielsZeilemaker The "Session is singleton" problem is a side effect of the megacache issue #1355. I'm fixing them now.

Tribler/Test/API/test_seeding.py
```diff
@@ -36,6 +36,7 @@ def setUpPreSession(self):
         super(TestSeeding, self).setUpPreSession()
 
         self.config.set_libtorrent(True)
+        self.config.set_megacache(True)
```
NielsZeilemaker: This doesn't make sense

LipuFei: @NielsZeilemaker I made `SearchManager` to be always initialized and it depends on ChannelDBHandler, so that's why I have to add enabled megacache in a lot of tests.

LipuFei:
@NielsZeilemaker `SearchManager` is now only initialized if either torrent or channel search is enabled, so there is no need to enable megacache for all those tests. I will work on the megacache thing in another PR after this one.

Tribler/Core/Session.py
```diff
@@ -625,7 +625,14 @@ def search_remote_torrents(self, keywords):
...
         """
-        return self.lm.torrent_search_manager.search(keywords)
+        return self.lm.search_manager.search_for_torrents(keywords)
+
+    def search_remote_channels(self, keywords):
+        """
+        Searches for remote channels through AllChannelCommunity with the given keywords.
+        :param keywords: The given keywords.
+        """
+        self.lm.search_manager.search_for_channels(keywords)
```
NielsZeilemaker: I would add a check to see if the torrentsearch/channelsearch is enabled and raise an OperationNotEnabledByConfigurationException() if they aren't.

Tribler/Test/test_as_server.py
```diff
@@ -187,7 +188,7 @@ def tearDown(self):
         """ unittest test tear down code """
         if self.session is not None:
             self._shutdown_session(self.session)
-            Session.del_instance()
+        Session.del_instance()
```
NielsZeilemaker: This is intentional, if we created a session, call del_instance. If not (self.session = None) don't.

Tribler/Test/test_as_server.py
```diff
@@ -301,6 +302,7 @@ class TestGuiAsServer(TestAsServer):
 
     def setUp(self):
         AbstractServer.setUp(self, annotate=False)
+        self.setUpPreSession()
```
NielsZeilemaker: This shouldn't be neccesary, we're using the session from Tribler -> and its default settings.

Tribler/Test/test_as_server.py
```diff
@@ -420,6 +424,7 @@ def tearDown(self):
         del self.frame
         del self.lm
         del self.session
+        del self.config
```
NielsZeilemaker: Same here

Tribler/Test/test_remote_search.py
```diff
@@ -1,19 +1,48 @@
...
 from Tribler.Main.vwxGUI.list_item import ChannelListItem
 
 
-class TestRemoteQuery(TestGuiAsServer):
+class BaseRemoteTest(TestGuiAsServer):
 
     def setUpPreSession(self):
-        super(TestRemoteQuery, self).setUpPreSession()
+        super(BaseRemoteTest, self).setUpPreSession()
+        self.config.set_megacache(True)
```
NielsZeilemaker: Dont touch the session config, you're testing tribler with its defaults.

Tribler/Test/test_remote_torrent_handler.py
```diff
@@ -28,6 +28,8 @@ def setUpPreSession(self):
         super(TestRemoteTorrentHandler, self).setUpPreSession()
         self.config.set_dispersy(True)
         self.config.set_torrent_store(True)
+        self.config.set_megacache(True)
+        self.config.set_torrent_collecting(True)
```
NielsZeilemaker: You're testing a single torrent collect, not the automatic torrent collecting. Revert these

NielsZeilemaker:
I get why you're donig these changes to be able to pass a custom session config, but you should do it in a different manner. Maybe add a callback that is called after the session is created, but to "started" yet. 
Passing a custom session config will overwrite all the Tribler defaults, and hence not test tribler anymore. We need to be testing Tribler, not a custom version of it.

Tribler/Test/test_remote_torrent_handler.py
LipuFei: If we do not enable torrent_collecting, RemoteTorrentHandler will not be created, and there is no way to test it.

Tribler/Test/test_as_server.py
LipuFei: @NielsZeilemaker This is only for `test_remote_search` because it needs to create a SessionConfig, set some options and pass it to `tribler_main`. The line doesn't have any effect on other tests because the other GUI tests are simply running full Tribler. In `test_remote_search`, the config is the same as full Tribler except `torrent_search` and `channel_search`, so we are still testing full Tribler, but for `torrent_seach` only the `SearchCommunity` is enabled and for `channel_seach` the `AllChannelCommunity`.

LipuFei:
@NielsZeilemaker Because it's still very difficult to replace the `ABCApp` thing, so if we want to test `channel_search` only with `AllChannelCommunity`, the easiest way is to pass a custom `SessionConfig` with everything enabled except the `SearchCommunity`. That's why I did it this way.

LipuFei:
@NielsZeilemaker I've fixed most issues except the customized SessionConfig thing. I will try to find another way to do it.

Tribler/Test/test_remote_torrent_handler.py
NielsZeilemaker: why did this work before then?

NielsZeilemaker:
If it's only for this unit-test, I would overwrite the property in the lmc. It's nasty, but I like it better than this custom session config stuffs. 
Maybe add an assert which checks that the torrent/channel search is actually returning none after overwriting it in lmc.

Tribler/Test/test_remote_torrent_handler.py
LipuFei: @NielsZeilemaker Good point. I think I saw an exception because of the missing RemoteTorrentHandler, but I forgot where. I will remove those two lines.

LipuFei:
@NielsZeilemaker @rjruigrok Rob and I both need a customised session in the GUI. I think we can still add some flags as parameters just like the `autoload_discovery` thing, but this must not keep growing otherwise we will end up having hundreds of parameters in `tribler_main.py`.

@NielsZeilemaker If you agree, we can use the flag approach at the moment.

LipuFei:
@NielsZeilemaker I made two flags to disable the communities.

LipuFei:
@NielsZeilemaker Nice! It worked.

Tribler/Core/Session.py
LipuFei: @NielsZeilemaker I removed the `OperationNotEnabledByConfigurationException()` here because in `torrent_search` and `channel_search`, only one community is enabled, but the test uses `GuiUtility` to search for results, which always calls both `search_remote_torrents` and `search_remote_channels`, so it will fail if we add this check.

NielsZeilemaker: Or you could catch and ignore the exception in the gui

LipuFei: @NielsZeilemaker ok, I will let GUI check if the function is enabled before using it.

NielsZeilemaker: Even nicer

LipuFei:
retest this please

LipuFei:
retest this please

NielsZeilemaker:
retest this please

LipuFei:
retest this please

synctext:
retest this please

NielsZeilemaker:
retest this please

LipuFei:
Something wrong with github. It has been slow since this Monday.

NielsZeilemaker:
I saw, probably the Chinese again :smile:

2015-03-27 9:22 GMT+01:00 Lipu Fei <notifications@github.com>:

LipuFei:
retest this please

synctext:
Wow, saw the China+Github headlines https://news.ycombinator.com/item?id=9275771

NielsZeilemaker:
retest this please

NielsZeilemaker:
@LipuFei ready to be merged

PR #1378, opened 2015-04-21
Review torrent checking

whirm:


whirm:
@LipuFei ready

Tribler/Core/TorrentChecker/torrent_checker.py
```diff
@@ -406,11 +422,8 @@ def _update_torrent_result(self, response):
         torrent_id = result[u'torrent_id']
         retries = result[u'tracker_check_retries']
 
-        # the result logic
-        is_good_result = seeders > 0 or leechers > 0
-
         # the status logic
-        if is_good_result:
+        if seeders > 0:
```
NielsZeilemaker: No more leechers > 0?

LipuFei: @NielsZeilemaker We decided to check seeders only because the swarm can be dead if there is no leecher (although it might be alive, but the chance is relatively low).

NielsZeilemaker: I would be worried about crappy/intentionally crappy tracker responses. Not marking a torent as good will schedule it for removal. It has quite some consequences.

whirm: Yeah, I was thinking about this when on the train yesterday, we where/are not taking into account that we may find more peers trough other trackers and/or PEX/DHT.

NielsZeilemaker: In the past the dht checker would set a torrent to good if it found some peers. That worked great

whirm: RE lying trackers: We are already keeping the best response across all trackers for a given torrent (torrent_checker.py:407)

The issue comes when we randomly select a torrent to fill a multiscrape request for a specific tracker triggered by a manually selected tracker (from the GUI). We will then be only checking one of the (possibly) multiple trackers it contains and still increasing the retry counter for it. One possible solution would be to mark those randomly selected torrents as such so we can avoid increasing the retry counter. The other would be that just before marking a torrent as dead, we should do a final full check on it and only then mark it at dead so it can be purged from the system.

Also I guess it's a good moment to re-add DHT support to the checker.

NielsZeilemaker: It is still there right? see session.py UdpTrackerSession

whirm: @LipuFei If I understood correctly, you said it was not being used yet?

whirm: Sorry, I misread that, I was talking about DHT checks not UDP trackers/

NielsZeilemaker: My bad, I was thinking you guys meant UDP trackers. The DHT stuff used to simply do a get_peers which if peers were returned set the status to good. Should be pretty trivial to implement.

whirm: I will have a look

whirm:
@NielsZeilemaker @LipuFei ready

LipuFei:
@whirm @NielsZeilemaker Looks good.

synctext:
retest this please

LipuFei:
@whirm By the way, could you also make the automatic selection algorithm to select and check DHT torrents?

You will need to change this: https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Modules/tracker_manager.py#L147 to include `DHT`.

whirm:
sure

whirm:
retest this please

whirm:
@LipuFei ready, I don't know what's up with GH not going green :)

PR #1381, opened 2015-04-23
Stop wrapping stdout/stderr, we use logging now :D

### LINKED ISSUE #1365: Remove Tribler/Debug
I think we should get rid of Tribler/Debug, the console wrapper stuff seems dodgy at best and since we're using logger we shouldn't need any protection wrt stderr anymore.

whirm:
Closes #1365


whirm:
@LipuFei ready

NielsZeilemaker:
Do we need to keep this one:
https://github.com/Tribler/tribler/blob/devel/Tribler/Debug/memory.py
?

LipuFei:
I think this one can go as well. I will remove that.

Issue #787, opened 2014-08-22
Illegal IP address string in TunnelCommunity


egbertbouman:
Got the following error when dest_address[0] is a hostname instead of an IP:

```
ERROR   1408700660.57             log:443   Unhandled Error
Traceback (most recent call last):
  File "C:\Python272\Lib\site-packages\twisted-13.2.0-py2.7-win32.egg\twisted\python\log.py", line 73, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "C:\Python272\Lib\site-packages\twisted-13.2.0-py2.7-win32.egg\twisted\python\context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "C:\Python272\Lib\site-packages\twisted-13.2.0-py2.7-win32.egg\twisted\python\context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "C:\Python272\Lib\site-packages\twisted-13.2.0-py2.7-win32.egg\twisted\internet\selectreactor.py", line 151, in _doReadOrWrite
    why = getattr(selectable, method)()
--- <exception caught here> ---
  File "C:\Python272\Lib\site-packages\twisted-13.2.0-py2.7-win32.egg\twisted\internet\udp.py", line 234, in doRead
    self.protocol.datagramReceived(data, addr)
  File "C:\Users\Egbert\git\tribler\Tribler\community\tunnel\Socks5\server.py", line 61, in datagramReceived
    circuit.tunnel_data(request.destination, request.payload)
  File "C:\Users\Egbert\git\tribler\Tribler\community\tunnel\routing.py", line 86, in tunnel_data
    return self.proxy.tunnel_data_to_end(destination, payload, self)
  File "C:\Users\Egbert\git\tribler\Tribler\community\tunnel\community.py", line 802, in tunnel_data_to_end
    packet = TunnelConversion.encode_data(circuit.circuit_id, ultimate_destination, ('0.0.0.0', 0), data)
  File "C:\Users\Egbert\git\tribler\Tribler\community\tunnel\conversion.py", line 201, in encode_data
    return pack("!I4sH4sH", circuit_id, inet_aton(dest_address[0]), dest_address[1],
socket.error: illegal IP address string passed to inet_aton
```

In libtorrent we have set proxy_hostnames to True, which ensures that libtorrent will not attempt to resolve hostnames locally, but through the proxy. We need to adapt the tunneling protocol to support this.

[...]

PR #1382, opened 2015-04-23
sento method allows a hostname in its destination

rjruigrok:
Fix for #787

### DIFF
--- a/Tribler/community/tunnel/tunnel_community.py
+++ b/Tribler/community/tunnel/tunnel_community.py
@@ -5,7 +5,6 @@ import time
 import random
 import logging
 
-from traceback import print_exc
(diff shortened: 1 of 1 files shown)


Tribler/community/tunnel/tunnel_community.py
```diff
@@ -123,10 +123,21 @@ def enabled(self):
     def sendto(self, data, destination):
         if self.check_num_packets(destination, False):
             if TunnelConversion.is_allowed(data):
+                ip_address = destination[0]
```
whirm: Better to check if it's an actual IP address than to capture all exceptions and then try to resolve the name. (maybe using ```dispersy.util.is_valid_address()```)

PR #1383, opened 2015-04-23
fix the problem that cannot open torrent file with Tribler on Mac OS X

### LINKED ISSUE #1379: Double clicking .torrent file does not add it to queue in OS X
OSX Yosemite.

When Tribler is not active and you click a .torrent file, tribler is started and the option to add the torrent file is presented.

When Tribler is active and you click a .torrent file, tribler does not present the option to add a .torrent file.  you have to click the + button to add.

LipuFei:
Closes #1379


LipuFei:
@whirm ready

PR #1388, opened 2015-04-23
Enable oportunistic encryption

### LINKED ISSUE #887: Support BitTorrent Protocol Encryption
Support BitTorrent protocol Encryption, so that Tribler can communicate with peers who have configured their clients to require encryption.

whirm:
on the non-tunneled session

Closes #887


whirm:
@LipuFei  ready

PR #1392, opened 2015-04-24
TFTP with retry mechanism

### LINKED ISSUE #903: Finish remaining features for TFTP
In the future we will remove Libswift, as it is not sufficiently stable and too complex.

We will test TFTP protocol, from 22 years ago. (https://tools.ietf.org/html/rfc1350)

Priorities and steps:
- [x] integrate TFTP with rawserver.py
- [x] most basic service: single 1MByte hard-coded test file transfer
- [x] ability to serve small files from a given working directory
- [x] add a first and simple timeout
- [x] a downloader gets no response for 5 seconds, we timeout and abort
- [x] do a clean abort, use a temp file or something
- [x] only deliver fully downloaded files
- [x] Fancy timeout and retransmit mechanism
- [x] do a re-transmit of last data packet after 5 seconds, if no ACK is received
- [x] receiver has the ability to ignore duplicated data packets
- [x] simple and effective content correctnes
- [x] the filename contains or is equal to a SHA1 hash check
- [x] after file is downloaded, check the hash.
- [x] if it is incorrect, simply abort
- [x] we do not try to recover from this type of error, too complex

LipuFei:
DO NOT MERGE YET
Closes #903

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -161,7 +161,7 @@ class TriblerLaunchMany(Thread):
                 # register TFTP service
                 from Tribler.Core.TFTP.handler import TftpHandler
                 self.tftp_handler = TftpHandler(self.session, u'', endpoint,
-                                                "fffffffd".decode('hex'), block_size=1024)
+                                                "fffffffc".decode('hex'), block_size=1024)
                 self.tftp_handler.initialize()
 
             if self.session.get_enable_torrent_search() or self.session.get_enable_channel_search():
--- a/Tribler/Core/TFTP/handler.py
+++ b/Tribler/Core/TFTP/handler.py
@@ -214,18 +214,19 @@ class TftpHandler(TaskManager):
             self._handle_new_request(ip, port, packet)
             return
 
-        if (ip, port, packet['session_id']) not in self._session_dict:
+        session_key = (ip, port, packet['session_id'])
+        if session_key not in self._session_dict:
             self._logger.warn(u"got non-existing session from %s:%s, id = %s", ip, port, packet['session_id'])
             return
 
         # handle the response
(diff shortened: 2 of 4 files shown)


LipuFei:
@whirm I think it's ready, but this will break the old TFTP protocol because they are not compatible with each other, so we need a discussion before merging it.

whirm:
Ok

NielsZeilemaker:
Why is it incompatible?

LipuFei:
I added a packet sequence number which doesn't exist in the old TFTP version, and it breaks the compatibility.

NielsZeilemaker:
Why do you need a sequence number?

LipuFei:
To know if an incoming packet is a re-sent or not. If we receive a re-sent packet and there is no sequence number, it is difficult to know whether it is a re-sent packet or the other person sent a wrong packet.

NielsZeilemaker:
Explain "a wrong packet"

LipuFei:
Wrong packet means a packet that is not what we are expecting or is not what we have seen before (a re-sent packet). Suppose we are expecting a DATA packet with data_id 20, and someone sends us a DATA packet with data_id 22, then this id 22 packet is wrong.

NielsZeilemaker:
But would it cause any problems? I would expect a peer which receives a request for part 22 to reply with part 22. Even if I request it multiple times.

LipuFei:
It is fine for DATA and ACK packets because they contain part numbers, but the negotiation packets that are sent in the beginning to initiate a request don't have sequence numbers. If we receive an old retransmitted negotiation packet, it is impossible to tell if it is a retransmission or a wrong packet. And that's why I added this sequence number.

I will think about it again to see if there is a better way to do it.

NielsZeilemaker:
Let's not retransmit any negotiation packets then. Abort if negotiation failed else retransmit request.

LipuFei:
ok

PR #1394, opened 2015-04-24
record ip for peer1 in interaction_log 

corpaul:
instead of member id so we can generate a coherent bartercast graph


corpaul:
ready to merge

PR #1402, opened 2015-05-06
Add .coveragerc to exclude dispersy and pymdht

whirm:


LipuFei:
@whirm strange, the coverage is still around 66% after we exclude `dispersy` and `pymdht`

PR #1429, opened 2015-05-22
MultiChain community

snorberhuis:
This is a pull request to merge the MultiChain community into devel.


synctext:
Starting to look good. Please collapse into 1 commit and fix the tests:

```
	Test Result (20 failures / +20)
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_add_block
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_add_block
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_add_two_blocks
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_add_two_blocks
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_block_id_negative
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_block_id_negative
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_block_id_positive
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_block_id_positive
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_signature_pk_false
Tribler.Test.test_doubleentry_persistence.TestPersistence.test_contains_signature_pk_false
Show all failed tests >>>
```

synctext:
retest this please

snorberhuis:
retest this please

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+    def initiate_conversions(self):
+        return [DefaultConversion(self), DoubleEntryConversion(self)]
+
+    def publish_signature_request_message(self, candidate):
+        """
+        Creates and sends out signature_request message.
+        """
+        self._logger.info("Sending signature request.")
+        message = self.create_signature_request_message(candidate)
+        self.dispersy.store_update_forward([message], True, True, True)
```
NielsZeilemaker: You're attempting to store, and update and message which is destined for candidate.
I would switch to _forward(

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+        :param messages: Potential valid signature_request messages.
+        :return: Series of correct signature_request messages
+        """
+        self._logger.info("Received " + str(len(messages)) + " signature requests.")
+        for message in messages:
+            if self.validate_signature_request(message):
+                self._logger.info("Received valid request.")
+                yield message
+            else:
+                DropMessage(message, "Invalid signature request message")
```
NielsZeilemaker: You need to yield such a message, and I would integrate the validate_signature_request method in de check method

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def _on_signature_request(self, messages):
+        """
+        Handles behaviour of the community when it receives a signature_request message.
+        It will send out a signature response
+        :param messages: Signature_request messages that needs to be handled.
+        """
+        for message in messages:
+            # Check if the messages are not from ourselves.
+            if message.payload.public_key_requester != self._public_key:
```
NielsZeilemaker: This check needs to be performed in the check, and if it fails yield a DropMessage

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+        :return: Series of correct signature_response messages
+        """
+        self._logger.info("Received " + str(len(messages)) + " signature responses.")
+        for message in messages:
+            # For now do no checking.
+            if self.validate_signature_response(message):
+                self._logger.info("Received valid response")
+                yield message
+            else:
+                DropMessage(message, "Invalid signature response message")
```
NielsZeilemaker: Same here integrate the validate method, and there's a yield missing

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def _on_signature_response(self, messages):
+        """
+        Handles behaviour of the community when it receives a signature_response message.
+        It persist the signature response.
+        :param messages: Signature_response messages that needs to be handled.
+        """
+        for message in messages:
+            # Check if the messages are not from ourselves.
+            if message.payload.public_key_responder != self._public_key:
```
NielsZeilemaker: Same here

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def persist_signature_response(self, message):
+        """
+        Persist the signature response message.
+        A hash will be created from the message and this will be used as an unique identifier.
+        :param message:
+        """
+        message_hash = self.hash_signature_response(message)
+        self._logger.info("Persisting sr: %s." % base64.encodestring(message_hash))
+        self.persistence.add_block(message_hash, message.payload)
```
NielsZeilemaker: Move this stuff into the on method.

Tribler/community/doubleentry/community.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def on_introduction_response(self, messages):
+        super(DoubleEntryCommunity, self).on_introduction_response(messages)
+        for message in messages:
+            self._logger.info("in on_introduction_response: Saving %s candidate.", message.candidate)
+            self.candidate_queue.append(message.candidate)
```
NielsZeilemaker: You can get to the candidates of dispersy using self.candidate

Tribler/community/doubleentry/conversion.py
```diff
@@ -0,0 +1,148 @@
...
+                                 self._encode_signature_response, self._decode_signature_response)
+
+    def _encode_signature_request(self, message):
+        """
+        Encode a signature_request message.
+        :param message: The message to be encoded.
+        :return: encoded signature request message.
+        """
+        payload = message.payload
+        return encode((payload.up, payload.down, payload.total_up, payload.total_down,
```
NielsZeilemaker: If you like to do some more work, you could switch to using structs. Have a look at the other communities for an example. They're a bit more efficient + you can force the type of the parameters.

Tribler/community/doubleentry/database.py
```diff
@@ -0,0 +1,266 @@
...
+
+        self.db.execute(
+            u"INSERT INTO double_entry (block_hash, previous_hash_requester, public_key_requester, "
+            u"signature_requester, previous_hash_responder, public_key_responder, signature_responder,"
+            u" sequence_number_requester, sequence_number_responder, up, down, total_up, total_down) "
+            u"values(?,?,?,?,?,?,?,?,?,?,?,?,?)",
+            data)
+
+        self._set_previous_id(block_id)
+        self.db.commit()
```
NielsZeilemaker: Although, you would like to commit. This is not how Dispersy works. It will commit periodically.
Commiting yourself might cause Dispersy and your database to differ in content

Tribler/community/doubleentry/database.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def _set_previous_id(self, block_id):
+        """
+        Update the id of the latest block chain.
+        :param block_id: The id of the block
+        """
+        self._init_database()
```
NielsZeilemaker: This shouldn't be neccesary. Remove all those init_database calls and do a single init from somewhere else.

synctext:
Many thanks for those detailed comments Niels!

Tribler/community/doubleentry/database.py
```diff
@@ -0,0 +1,266 @@
+""" This file contains everything related to persistence for DoubleEntry.
+"""
+import base64
+
+from os import path
+
+from Tribler.dispersy.database import Database
+
+
+def encode_db(db_object):
```
NielsZeilemaker: This method is just plain lazy. If you're inserting binary into the database wrap it into a buffer object, if not don't.
You should be able to know the datatypes beforehand.

NielsZeilemaker:
Btw, this all seems familiar to DoubleMemberAuthentication, which is implemented in Dispersy.
Have a look at https://github.com/Tribler/dispersy/blob/devel/tests/debugcommunity/community.py#L63

NielsZeilemaker:
Here's the documentation to create a signature request
https://github.com/Tribler/dispersy/blob/devel/community.py#L2245

Tribler/community/doubleentry/community.py
snorberhuis: The _forward method is private and gives a pylint error. There are two solutions: 
- Use store_update_forward([message], false, false, true)
- Change the method to public. It is accessed publicly in a lot of other places aswell, so this might be a good change overall. 

What do you prefer?

NielsZeilemaker: Do the first one, but create an issue in the dispersy repo stating that these methods should be public.

Tribler/community/doubleentry/community.py
snorberhuis: This could be moved but would result in duplicate code in _on_signature_response and publish_signature_response_message.

Tribler/community/doubleentry/database.py
snorberhuis: This was a remnant of testing.

Tribler/community/doubleentry/database.py
snorberhuis: Do you object to the use of mapping this function to the list of data to be inserted into the database? I could transform it manually at the point where I have that list.

Or do you object to the laziness of the inner workings of the method?
I can change it so that it checks the type and handles more accordingly for binary.

NielsZeilemaker: You're encoding a binary string into base64 and then storing it as a unicode string to get around the asserts of  dispersy regarding not accepting strings. Just use buffer object in the appropriate places. 

And explicitly convert some columns into buffer and not use a method like this.

synctext:
retest this please

synctext:
@NielsZeilemaker : studied the DoubleMemberAuthentication code a bit.. Do you think the overly complex should be repaired or removed from the core? Clearly the "technical debt" theory applies :-)

DoubleMemberAuthentication is not a simple one-shot just-sign the message thingie. It has negotiation and various exceptions. "Assuming that he does the new message M2, which now includes Bob's signature while Alice's is still \0, is wrapped in a dispersy-signature-response message (E) and sent back to Alice. " https://github.com/Tribler/dispersy/blob/devel/authentication.py#L203

Shall we ask @snorberhuis to get his work operational in community.py and then move his merged code into Dispersy-core, thus replacing the un-used and complex DoubleMemberAuthentication?

btw _very_ scary comment: ```# todo: verify the signature```
https://github.com/Tribler/dispersy/blob/devel/authentication.py#L308 Trust incoming data without a verify or a comment about your own signature?

NielsZeilemaker:
The DoubleMemberAuth stuff is perfectly working, and I actually like it better than the custom stuff @snorberhuis is creating. You just have to implement two callbacks and you're done.

The negotiation is handled in the callbacks, so if you're not interested in negotiating then is should be pretty simple.

Moreover, the scary comment isn't that scary actually, all the verification is done in the conversion layer.

The basic protocol is as follows:
Alice creates a message, wraps it in a signature request message. This triggers a callback at Bob, which can modify of accept the message if he agrees. Alice then get's a callback triggered with a modified or the same message, which she then can accept or not. If she does, she adds her signature and you'll have your double signed message.

Again, I like it better than the implementation of @snorberhuis and this code is actually tested etc. Moreover, it was simplified a bit by me and @whirm during the Dispersy v2.0 refactor.

synctext:
```Alice then get's a callback triggered with a modified or the same message, which she then can accept or not. If she does, she adds her signature and you'll have your double signed message.```
OK, that is almost exactly what we need. Can Alice also copy the signed message from Bob, append it with additional info, and DoubleSign this whole extended message? (still a one-shot primitive)

Then the custom code from @snorberhuis can be cleaned up considerably.

**Comment by NielsZeilemaker** [comment]
Bob can return any message, so if Bob creates a new message containing the values of Alice and him, then yes if Alice accepts it, a double signed message is created.

**Comment by synctext** [comment]
```Alice creates a message, wraps it in a signature request message.```
Is the first message already signed by Alice? Or was that vital efficiency boost only added in Dispersy 2.0 refactor? (Vaguely remember something was not ideal).

**Comment by NielsZeilemaker** [comment]
Honestly, I'm not sure. I think it's optional, but that doesn't make any sense.
We can refactor it such that it's always signed by Alice, and if Bob modifies the message she signs it again.

**Comment by synctext** [comment]
Thnx Niels! @snorberhuis please dive into the code handling the Dispersy DoubleMemberAuth and try to understand if it delivers what we need + operates as expected.

**Comment by synctext** [comment]
https://github.com/Tribler/dispersy/pull/432 vital changes merged.

**Comment by NielsZeilemaker** [comment]
I've created a pull request which contains a refactor of the community using the DoubleMemberAuth. It's not tested, and hence won't work but it should be close.
https://github.com/snorberhuis/tribler/pull/35

It also contains the new dispersy pointer.

**Comment by snorberhuis** [comment]
I used your example in snorberhuis#35 to implement the community using the `create_signature_request`. It was working and messages were passed along (see my branch devel_core in my repository), but then I found a problem with the `create_signature_request`. 

The `create_signature_request` implementation does not follow an append only methodology to the signatures. Bob always needs to add information he only haves. Only he knows his current `previous_hash`.  Alice will therefor always need to sent a third message back to Bob with a new valid signature. 

I am also doubtful how well the current implementation of `create_signature_request` is tested, because it did not even work for me at first, because method calls were done on the wrong objects, see Tribler/dispersy#445.

Because of this, it was decided with @synctext to continue forward with the current way for now.

**Comment by NielsZeilemaker** [comment]
#445 doesn't seem like a big bug, it's a wrong default value for the timeout. it was tested with a passed timeout which was a float.
Btw, what's the problem with alice needing to send a third message to bob? And does bob even need to receive this message?

**Comment by synctext** [comment]
@NielsZeilemaker So you think repairing the ```create_signature_request``` in the Dispersy core is not a difficult undertaking? We should try to avoid useless work so this PR essentially duplicates existing half-broken functionality. 3 messages is not a problem, but needs to be 2 messages eventually.

**Comment by NielsZeilemaker** [comment]
It's fixed already, Tribler/dispersy#445 was a bug in a default value for the timeout.

**Comment by synctext** [comment]
So you're @NielsZeilemaker arguing to get-it-right the first time by using the Dispersy-Core, instead of merging this PR with duplicated functionality?
So ```create_signature_request``` does not contain bugs, it just a 3-message approach.

@snorberhuis just checked the code (https://github.com/Tribler/dispersy/blob/devel/community.py#L1673) it's not clear to me that these signatures are automatically spread in the community as you incicated. We don't want auto-full-sync-spreading. The directdistribution seems also the default. No showstoppers.

**Comment by NielsZeilemaker** [comment]
Its not a three message approach, after Alice signs the message she received from Bob, she will attempt broadcast it to 10 neighbors (usnig the dispersy.py _forward method). 

However, if you're using direct distribution it's not forwarded at all. We could change that behaviour, to always send it at least back to Bob.

**Comment by NielsZeilemaker** [comment]
Basically, that requires a small change here:
https://github.com/Tribler/dispersy/pull/445/files#diff-b16d767d4c9e8d1efb4e9dd33041ab76L2438
adding 
```
if isinstance(new_submsg.meta.destination, (CommunityDestination, CandidateDestination)):
    new_submsg.destination.candidates.add(message.candidate)

NielsZeilemaker:
After that change, the dispersy.py _forward method will forward it back to Bob.

snorberhuis:
# Core version
Tribler/dispersy#445 fixes two bugs. The timeout bug is a really minor bug. But the `self.store_update_forward([new_submsg], True, True, True)`  line fix is a bigger one. But that is not a reason to not choose for `create_signature_request`. I fixed it for myself when implementing `create_signature_request` and wanted to share this back.

I think part of the misunderstanding is that Bob also needs the full payload with valid signatures.
Both Bob and Alice add the signature response as a block in their own chain. So they both need to have a valid block with valid signatures

Because the previous_hash and sequence_number of Bob is always filled in by Bob, Alice's signature always becomes invalid. Alice needs to resign the packet after Bob. Alice can now save the block to his own chain. Bob has to wait to receive the new signature of Alice on the changed payload. Alice will have to resent this block back to Bob. 

We could change the behaviour of `create_signature_request` to only forward it back to Bob, but this has to be deterministic because Bob has to block until this happens before he can create new blocks.
A timeout is possible for Bob.

But this change will still result in three messages. 

# Non core version
The "non-core" version in the pr uses a append only signature scheme. 
Alice signs [payload, previous_hash_A, pk_A] 
Bob signs [[payload, previous_hash_A, pk_A] signature A, previous_hash_B]
Alice does not sign the previous_hash of Bob and Bob only adds his information and signs the total block. When Bob returns the message both Alice and Bob has the same block and valid signatures.
They both can immediately save the block and continue creating new blocks.

Do you think it is a problem that A does not sign information of B?
And, do you think it is a problem B does not have a signature of A on his information?

I don't think this is currently a problem. Because A doesn't have authority over that part of the information. A and B both have authority to say the up and download part are correct, but only they themselves can say the ordering of their chain. 

If B is fraudulent and changes his chain this would have to be detected, but this is outside the scope of my work.

NielsZeilemaker:
What's the added value of having Bob sign the message after adding his previous_hash?

synctext:
Hey @NielsZeilemaker 

>  Why not create separate messages instead?
We want to have a simple single datastructure that you can build upon, crawl and publish. A graph with 1 million records compressed is a long-term goal.

Just talked at length about the code and browsed the Dispersy community.py code.
Seems this is just 100 lines in the core. To make this core signature functionality usable 3 changes are required:
 - [ ] Alter from a 3 msg approach into an efficient 2 msg approach
 - [ ] Instead of a push to 10 peers, use the DirectDistribution
 - [ ] use append-only semantics, no modification of the signature_request. You can withe refuse a signature (Drop or error?) or add data and sign it. 

Correct? Would this be acceptable as a PR on the Dispersy core @NielsZeilemaker, with decent code coverage? If OK, then @snorberhuis can get to work :-)

NielsZeilemaker:
Sure, but its going to be a bit tricky. Btw, by using the DirectDistribution you're already not pushing to 10 peers.

NielsZeilemaker:
@snorberhuis I made some changes to dispersy today to make these changes possible. Basically it allows you to specify a method which determines on which part of the payload is signed/verified. 
It's almost done, hopefully I'll have it ready tomorrow.

NielsZeilemaker:
have a look at https://github.com/Tribler/dispersy/pull/446 I implemented the append feature to the doublememberauth

synctext:
thnx for adding that to the core Niels!

snorberhuis:
I updated the community to use `create_signature_request` with modify behavior.
I used the version of dispersy in Tribler/dispersy#446. Currently this commit breaks the chain, because Bob saves a different version then Alice.

I tried implementing the append behaviour that would fix the chain again. I  looked for some time at the example you have provided. But I could not yet figure out how the split functionality works. @NielsZeilemaker Could you give a pointer how to implement it using my community as it now is clean and using `create_signature_request?

This commit also includes concurrency fixes that would break a chain. No buffering is implemented anymore like previously said to @syntext. Instead incoming requests are dropped if the chain is being modified.

snorberhuis:
Looking with fresh pair of eyes and having though about it I now have an idea how to implement this:
1. Change `_encode_signature` and `_decode_signature` to introduce a split symbol
1. Create a function that can split the result of `_encode_signature` on the split symbol.
@NielsZeilemaker is this correct?

What kind of symbol should I introduce to the encoding that can be split upon?
 This symbol cannot occur in the normal payload, or the split would go wrong.

NielsZeilemaker:
This i implemented using the splitpayload thingy.

NielsZeilemaker:
Have a look at the split_double_payload method in debugcommunity/community.py

snorberhuis:
I used my proposed implementation to implement it. But I used `,,` now to split, because `,` occurs in the payload. But I am not confident `,,` does not occur as well.

NielsZeilemaker:
I would modify the conversion to use struct instead, that way the payload would have a fixed size in bytes. And splitting the payload would be easy/wont suffer from these special characters to split on

synctext:
What is the usage of ```Tribler/Test/output/annotations.txt``` ?

Please use it multi-chain and modify the comments/naming to make it explicit that this is append-only and modification are not allowed.

Furthermore, please add a comment: %this code _always_ signs incoming request. insert checks and rejections HERE

snorberhuis:
I think I now implemented all remarks. The split function using structs is much better! Thanks @NielsZeilemaker. The known issues I posted in the first comment are now fixed. Please review if I missed any remark or if there are new remarks.

Afterwards I will squash the commits to a single commit.

Tribler/Test/output/annotations.txt
```diff
@@ -0,0 +1,5 @@
+annotation start end
```
NielsZeilemaker: Remove this file

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def allow_signature_request(self, message):
+        """
+        We've received a signature request message, we must either:
+            a. append to this message our data (Afterwards we sign the message.).
+            b. None (if we want to drop this message)
+        """
+        self._logger.info("Received signature request.")
+        self._logger.info("Chain Lock: process request: %s" % self.chain_lock.locked())
+        if self.chain_lock.acquire(False):
```
NielsZeilemaker: Unfortunately, this is never going to work. UDP is lossy, hence you cannot expect "normal" locking to work.

NielsZeilemaker: Moreover, attackers can easily lock a peer in a deadlock by simply not replying to a message

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,251 @@
...
+        return previous_hash if previous_hash else GENESIS_ID
+
+    def unload_community(self):
+        self._logger.debug("Unloading the MultiChain Community.")
+        super(MultiChainCommunity, self).unload_community()
+        # Close the persistence layer
+        self.persistence.close()
+
+
+class MultiChainSettings(object):
```
NielsZeilemaker: You probably don't need this object

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,108 @@
...
+
+"""
+Hashlength used in the MultiChain Community
+"""
+HASH_LENGTH = 20
+"""
+Formating of the signature packet
+"""
+request_format = '!I I I I i ' + str(HASH_LENGTH) + 's'
+signature_format = request_format + ' i ' + str(20) + 's'
```
NielsZeilemaker: replace the 20 with a HASH_LENGTH as well

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,108 @@
...
+            raise DropPacket("Unable to decode the payload")
+
+        values = unpack_from(signature_format, data, offset)
+        offset += signature_size
+
+        if len(values) != 8:
+            raise DropPacket("Unable to decode the signature")
+
+        up = values[0]
+        if not isinstance(up, int):
```
NielsZeilemaker: No need to check the types here, struct will make an int due to the signature_format

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def split_function(payload):
+    """
+    This function splits the SIGNATURE MESSAGE in parts.
+    The first to be signed by the requester, and the second the whole message to be signed by the responder
+    :param payload: Encoded message to be split
+    :return: (first_part, whole)
+    """
+    return payload[:request_size], payload
```
NielsZeilemaker: This is wrong, you're now only signing the first request_size bytes of the payload.
You need to sign everything except the part of Bob, so probably
```
return payload[:-24], payload

File: Tribler/community/multichain/database.py [hunk 18]
```diff
@@ -0,0 +1,204 @@
...
+        # Unpack the db_result tuples and decode the results.
+        return [str(x[0]) for x in db_result]
+
+    def contains(self, block_id):
+        """
+        Check if a block is existent in the persistence layer.
+        :param block_id: The id t hat needs to be checked.
+        :return: True if the block exists, else false.
+        """
+        db_query = u"SELECT block_hash from multi_chain where block_hash == ? LIMIT 0,1"
```
NielsZeilemaker: Limit 0,1?

Tribler/community/multichain/payload.py
```diff
@@ -0,0 +1,67 @@
...
+        @property
+        def sequence_number_responder(self):
+            return self._sequence_number_responder
+
+        @property
+        def previous_hash_responder(self):
+            return self._previous_hash_responder
+
+
+def encode_signing_format(data):
```
NielsZeilemaker: Still used?

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        Persist the signature response message.
+        A hash will be created from the message and this will be used as an unique identifier.
+        :param message:
+        """
+        payload = message.payload
+        (signature_requester, pk_requester) = message.authentication.signed_members[0]
+        (signature_responder, pk_responder) = message.authentication.signed_members[1]
+
+        block = DatabaseBlock([payload.previous_hash_requester, pk_requester.public_key, signature_requester,
```
NielsZeilemaker: I would use pk_requester.mid instead of public_key. The mids are the hashed public keys, no need to worry about hash collisions as dispersy uses these mids internally. So if a hash collision occurs, dispersy won't work either.

NielsZeilemaker: Btw, why do you want to store the signature of the packet?

Tribler/community/multichain/community.py
snorberhuis: Would the timeout releasing the lock not prevent this attack and solve the UDP lossy problem?
I have tested this in gumby scenario's and it seems to work. Check my gumby repo for the scenario.

See `allow_signature_response` for the timeout processing.

The chain will be designed to accept a positive inconsistency in the chains between requester and responder.

Tribler/community/multichain/community.py
snorberhuis: We want to have a chain where the blocks are verifiable with their interactors. The signature have to be verifiable and you need a non hashed public key here later on.

NielsZeilemaker: You can fetch the public key from dispersy by getting the member object for it. Call get_member(mid=the_pubkey_hash). Similarly, you can fetch/decode a message from the dispersy database if you need to access it lateron. 
message.packet_id is the row identifier where the packet is stored in the dispersy database, use dispersy.load_message_by_packetid to convert it back to a message if you need to.

Tribler/community/multichain/community.py
NielsZeilemaker: No, you need to have a think about this. Why do you need this lock btw?

snorberhuis: I don't see why blocking would not work. Because it does work in tests. Please give an example of a situation, different from tests, where this would fail. 

The previous hash pointer is given out by A to be used in a new block. 
If A would not lock, then it can create other blocks using the same hash pointer. This creates branches and inconsistencies between A and several B's. Several potential next blocks can come in. A has to do some ordering and decide what the real next block is.

With blocking inconsistencies can only occur if B fails and only at B. B also does not have to do ordering and decide what the real next block is. A will know in a simpler way if B has failed and can make a decision about the continuation of interaction with B.

NielsZeilemaker: There is no data loss in the tests, as all peers run at the same machine.

So basically, what you're attempting to prevent is A creating two chains with different peers at the same time. Right?

Why would you need a lock for that? My guess is that you're going to call publish_signature_request_message periodically right? If so, set the interval higher than the timeout and you're done.

Moreover, a lock won't prevent branching. If for some reason the message singed by Bob is lost, Alice will have a timeout, create another message with the same hash and hence a branch.

My guess is this can only be prevented by letting Alice sign the message last, preventing a branch if the message from Bob is lost. E.g. use the orginal doublesignedmember auth thingy

snorberhuis: There is a test with data loss in the tests. There is a test with a client ignoring the signature request. This is the same situation as a attacker not replying, signature response being lost, or a signature request being lost.

Locking is still needed because of incoming requests for A. Next to that, the code should not be dependent on timing issues to work correctly. 

An inconsistency is allowed in B's chain. 

Alice signing last creates the same problem, but then for Bob having to wait upon Alice's response.

NielsZeilemaker: I am completely lost at this point, I have no clue what this community is attempting to do. @synctext I'll leave it up to you if you think what's implemented here is ok.

Tribler/community/multichain/community.py
snorberhuis: Currently I am working on a crawler and this is an example of a that situation a mid will not work. Because you are not certain that the pk is available in Dispersy. But you want to be able to verify the block.

NielsZeilemaker: The pk must be available in Dispersy as otherwise it wouldn't have been able to verify the signature. If it accepted a message (with a signature), without having the pk available you've found a bug.

synctext:
Please re-use the existing deployed crawler code and operational ```stats``` message:

--crawl https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/main.py#L40
```stats``` https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/tunnel_community.py#L1146

File: Tribler/community/multichain/community.py (hunk 20 above)
Comment (snorberhuis) [inline Tribler/community/multichain/community.py]: You are correct. I made a thinking error that you would send the mid's instead of the pks.

I notice I didn't mention another problem. That is that you will need to be able to query the blocks for several reasons. For example counting the total_up total_down in a chain.

Comment (NielsZeilemaker) [inline Tribler/community/multichain/community.py]: That you could store in this database. And then a pointer to the packet-id to be able to fetch the packet from the dispersy database if required.

Comment (snorberhuis) [inline Tribler/community/multichain/community.py]: I only gave one example of a query and there are many more. In the end you will probably keep most fields in the database. Next to this, I do not know what fields will be required in the future with new incremental steps to this reputation system to improve resilience. If you strip fields now, I think it will be hard to be backward compatible.

You would need to retrieve data from two sources and have to aggregate these. I think the code will become more complicated. Conversion right now is very readable and maintainable. Data is only supplied from one clear source.

What is the benefit of the change other then reduced database size?

Comment (NielsZeilemaker) [inline Tribler/community/multichain/community.py]: It doesn't make any sense to store the public_keys of each block in the database. That's beyond redundant. Just use the mids. I'm not saying you should remove those field from the message format, only that you shouldn't be storing them.

File: Tribler/community/multichain/conversion.py [hunk 21]
```diff
@@ -0,0 +1,78 @@
...
+        :param data: ByteStream containing the message.
+        :return: (offset, SIGNATURE.impl)
+        """
+        if len(data) < offset + signature_size:
+            raise DropPacket("Unable to decode the payload")
+
+        values = unpack_from(signature_format, data, offset)
+        offset += signature_size
+
+        if len(values) != 8:
```
NielsZeilemaker: This cannot happen.

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,334 @@
...
+        self._logger.info("Crawler: Valid %s block response(s) received." % len(messages))
+        for message in messages:
+            block = DatabaseBlock.from_block_response_message(message)
+            # Create the hash of the message
+            if not self.persistence.contains(block.id):
+                self._logger.info("Crawler: Persisting sr: %s" % base64.encodestring(block.id))
+                try:
+                    self.persistence.add_block(block)
+                except IntegrityError:
+                    # A race condition can occur between here and
```
NielsZeilemaker: Can't happen. Dispersy is single threaded. If a block wasn't stored in the database at line 273, it can't have been stored now.

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,160 @@
...
+        :param data: ByteStream containing the message.
+        :return: (offset, BlockRequest.impl)
+        """
+        if len(data) < offset + block_request_size:
+            raise DropPacket("Unable to decode the payload")
+
+        values = unpack_from(block_request_format, data, offset)
+        offset += block_request_size
+
+        return \
```
NielsZeilemaker: No need to break the line here. We're using 120 chars wide max.

Tribler/community/multichain/database.py
```diff
@@ -0,0 +1,314 @@
...
+ down                       INTEGER NOT NULL,
+ total_up_requester         INTEGER NOT NULL,
+ total_down_requester       INTEGER NOT NULL,
+ sequence_number_requester  INTEGER NOT NULL,
+ previous_hash_requester	TEXT NOT NULL,
+ total_up_responder         INTEGER NOT NULL,
+ total_down_responder       INTEGER NOT NULL,
+ sequence_number_responder  INTEGER NOT NULL,
+ previous_hash_responder	TEXT NOT NULL,
+ public_key_requester		TEXT NOT NULL,
```
NielsZeilemaker: I would still like you to consider using mids. Mids are used everywhere in Dispersy, so why differ from this behaviour here. Moreover, they're smaller than public-keys resulting in a slightly smaller database.

snorberhuis: I am still undecided and I am going to try and implement mid and see how it works. I can see the merit in it. I just pushed code right now as I had a big backlog that I wanted up for review.

Tribler/community/multichain/community.py
snorberhuis: I didn't know it was single threaded. But you are right if Dispersy is single threaded, then this is not necessary. I will change this.

snorberhuis:
The scheduler was introduced so:
1. Future schedulers can be easily tested 
2. Decouple block creation and scheduling of block creation.
In the future the scheduler has to validate when to approve requests and when to stop collaborating with other peers.

snorberhuis:
The amounts stored in the scheduler are session amounts. Running totals not credited yet within the chain. These are combined with the amounts stored within the chain and send out for a signature request. The total build up of amounts can be verified in that way.

Amounts in barter cast are probably not usable as they are not verifiable within the chain. They do not have corresponding transactions in the chain and these transactions can not be reproduced. But I will discuss this with @synctext. I have not yet as he was away last week and the scheduler is just a very simple first version to make the integration work in the simplest

whirm:
retest this please

snorberhuis:
Tribler/dispersy#456 needed for the tests to work

snorberhuis:
Should I squash the commits further?

whirm:
Yes, please (If you want to remove the changes that a commit brings to a branch, just kill it instead of adding reversal commit).
Also, rebase your commits on top of devel instead of merging the branch.

Tell me when it's ready and I'll review it further.

Tribler/community/multichain/test
```diff
@@ -0,0 +1,6 @@
+INSERT INTO multi_chain (block_hash, up, down,        total_up_requester, total_down_requester, sequence_number_requester, previous_hash_requester, total_up_responder, total_down_responder, sequence_number_responder, previous_hash_responder, public_key_requester, signature_requester, public_key_responder, signature_responder)
+VALUES("hash",1,2,3,3,1,"hash_a",4,4,1,"hash_a","pk_a","sig_a","pk_a","sig_a")
+
+SELECT MAX(sequence_number) FROM (SELECT total_up_requester as total_up, total_down_requester as total_down, sequence_number_requester AS sequence_number " \
+FROM multi_chain WHERE public_key_requester == ? UNION
+SELECT total_up_requester as total_up, total_down_requester as total_down, sequence_number_responder FROM multi_chain WHERE public_key_responder = ? )
```
whirm: Is this file suposed to be here?

whirm:
You need to add the default setting (to False) for your community enable flag in `Tribler/Core/defaults.py`

snorberhuis:
I added the default setting and removed the incorrect file.

whirm:
@snorberhuis you need to rebase against latest devel

snorberhuis:
I have rebased.

whirm:
Could you add some tests that cover the not yet covered functions and important branches such as in crawl_down(), _convert_to_database_block(), DatabaseBlock.__str__()?

NielsZeilemaker:
So I've seen you do stuff like
```
target = other.my_candidate
target.associate(other.my_pub_member)
# Act
node.call(node.community.publish_signature_request_message, target, 5, 5)
```
That's very bad, as you're basically sharing an object between two different dispersy object. If you need access to the candidate of target you need to use the node object

NielsZeilemaker:
I would rewrite the unites into something like
https://github.com/Tribler/dispersy/blob/devel/tests/test_missingidentity.py
There we use give packet instead of relying on a socket.

NielsZeilemaker:
Won't this 
```
db_query = u"SELECT block_hash, MAX(sequence_number) FROM " \
+                   u"(SELECT block_hash, sequence_number_requester AS sequence_number FROM multi_chain " \
+                   u"WHERE mid_requester == ? UNION SELECT block_hash, sequence_number_responder " \
+                   u"FROM multi_chain WHERE mid_responder = ?)"

```
Still return the hash of the first matching row instead of the row with the highest sequence number? 
I would rewrite the outer select to order by sequence number desc and add a limit 1

NielsZeilemaker:
Same bug in the get_total

NielsZeilemaker:
Shouldn't you be resetting these amount after receiving a signed reply? 
```if request_sent:
+                    """ Reset the outstanding amounts and send a signature request for the outstanding amount"""
+                    self._outstanding_amount_send[peer] = 0
+                    self._outstanding_amount_received[peer] = 0

**Comment by NielsZeilemaker** [comment]
BTW why are you storing a half signed message? 
```
+        if not response:
+            self.logger.info("Timeout received for signature request.")
+            # Unpack the message from the cache object and store a half-signed record.
+            self.persist_signature_response(request.request.payload.message)
+            self.chain_exclusion_flag = False
+            return False

snorberhuis:
I added a commit to make it easily reviewable.

1. `crawl_down` cannot be tested without more support from Dispersy Test further. The method `self.get_candidate_mid` does not properly retrieve a candidate.
1. Some dead code was removed that was not tested.
1.  
```
target = other.my_candidate
target.associate(other.my_pub_member)
# Act
node.call(node.community.publish_signature_request_message, target, 5, 5)
```
No, the candidate is not shared:
```
@property
    def my_candidate(self):
        """
        A Candidate instance for this node.
        """
        return Candidate(self.lan_address, self.tunnel)
```
I also don't understand your proposed solution.

1. Your link shows no code using `give_packet` and it wouldn't matter anyway. As `give_message` uses `give_packet`

1.
```
db_query = u"SELECT block_hash, MAX(sequence_number) FROM " \
+                   u"(SELECT block_hash, sequence_number_requester AS sequence_number FROM multi_chain " \
+                   u"WHERE mid_requester == ? UNION SELECT block_hash, sequence_number_responder " \
+                   u"FROM multi_chain WHERE mid_responder = ?)"
```
No, it does not or you should give a counter example. I have tested the scenario you described.

1. `Shouldn't you be resetting these amount after receiving a signed reply?`
This is a much easier solution. Storing the amounts would not matter, because you would never use them anyways. If the reply fails or not, you cannot request them again. If you would the other peer might disagree with you and say I did sign them!

1. `BTW why are you storing a half signed message?`
Because you need to add these to your chain. If a peer fails this is your next block or you would potentially branch. This has been discussed before, but you were lost during that discussion.

NielsZeilemaker:
No, the candidate is not shared.
@snorberhuis I was talking about the member instance. The pub member actually has a database id in there which is the id of this member in the database of other. Using that object in node (which has a different database) will cause strange and unexpected stuff to happen.

Why don't you want to use give_packet/give_message?

If a peer would disagree, he should prove that using the signed message. If he sends such a signed message, you can then subtract those values and correct them. I think the goal of this community is to ~ estimate the data flowing between two peers. If that's the case, the simply resetting these values does not make a lot of sense to me.

Aren't you branching regardless if you store the message or not. If you store this half message then a peer could eventually reveal the double signed message to create a branch. If the message was really lost in translation, then it doesn't matter if you store the half message or not.

Maybe the only solution to this problem is to switch from a push based protocol to a pull based one.
E.g. instead of sending a half signed message, and letting the other peer add the second signature. You could request a half signed message, which you could the complete by setting the second signature.

And really, I'm only here to help. This community should be the start of more things to come, so if these details aren't implemented correct, we can start all over in 6 months time.

snorberhuis:
Do you mean with using give_packet/give_message to change:
```
node.call(node.community.publish_signature_request_message, target_other_from_node, 5, 
 _, signature_request = other.receive_message(names=[u"dispersy-signature-request"]).next()
other.give_message(signature_request, node)
```

to something like:

```
message = node.call(node.community.create_signature_request_message, candidate, up, down) 
other.give_message(message, node)
```
(`create_signature_request_message` is a MultiChain method)
I don't want to do this, because then the tests would not test the complete flow of the community.

snorberhuis:
`Aren't you branching regardless if you store the message or not. `

The idea is that parts of the block correspond and this shows your willingness to not cheat. And as such does not constitute a real branch. Your part of the block is still part of your chain and only the `total_up`, `total_down`, and `signature` of the other is not present . Because the `signature` is not present you do not up your total amounts in the next block.

snorberhuis:
`If a peer would disagree, he should prove that using the signed message` 
This could be a good improvement in the future. But it does sound very complicated. It might just be much easier to be less altruistic and validate the amounts more often and still estimate the data flow accurately enough.

snorberhuis:
retest this please

whirm:
retest this please

NielsZeilemaker:
@snorberhuis Sending back a message if a peer disagrees is something which is quite common in Dispersy. 
Basically, in the check method of a message you do something like:
```
# the sender apparently does not have the lower dispersy-undo message, lets give it back
self._dispersy._send_packets([message.candidate], [db_msg.packet], self, db_msg.name)

snorberhuis:
The feature creep of this PR has already been huge. The PR has been open since may 22th. I would propose you create a separate issue where we can discuss the proposal about the disagree protocol.

snorberhuis:
retest this please

synctext:
retest this please

PR #1440, opened 2015-06-01
Fix critical bugs related to circuits

rjruigrok:
This PR fixes two bugs related to anonymity in current devel:
* Downloading over circuits won't work when the circuit length was 1. 
* No blocking ok-button in the save-as dialog, this was only required for setting the anonymous=1 flag correctly but is not relevant anymore.


rjruigrok:
retest this please

rjruigrok:
Currently fails on remote_torrent_search only. Ready for merge.

PR #1442, opened 2015-06-02
Use ubuntu + desktop instead of vodo + pioneer

whirm:
As the content is disappearing from the network so the search tests fail
more often.


whirm:
@LipuFei ready

whirm:
@LipuFei I aborted that one

PR #1447, opened 2015-06-02
Rebuild tunnels when rendezvous or introduction circuits break down

rjruigrok:
Hidden services should recover from circuits that break down because of time-limits / throughput limits.

### DIFF
--- a/Tribler/Test/test_tunnel_community.py
+++ b/Tribler/Test/test_tunnel_community.py
@@ -469,6 +469,10 @@ class TestTunnelCommunity(TestGuiAsServer):
                 if not crypto_enabled:
                     settings.crypto = NoCrypto()
                 settings.become_exitnode = become_exit_node
+
+                # To test recovery: circuit breaks down after each 2 megebaytes
+                settings.max_traffic = 2 * 1024 * 1024
+
                 return dispersy.define_auto_load(HiddenTunnelCommunity,
                                                  dispersy_member,
                                                  (session, settings),
--- a/Tribler/Test/test_tunnel_community.py
+++ b/Tribler/Test/test_tunnel_community.py
@@ -470,7 +470,7 @@ class TestTunnelCommunity(TestGuiAsServer):
                     settings.crypto = NoCrypto()
                 settings.become_exitnode = become_exit_node
 
-                # To test recovery: circuit breaks down after each 2 megebaytes
+                # To test circuit recovery: circuit breaks down after each 2 megabytes
                 settings.max_traffic = 2 * 1024 * 1024
 
                 return dispersy.define_auto_load(HiddenTunnelCommunity,
--- a/Tribler/community/tunnel/hidden_community.py
+++ b/Tribler/community/tunnel/hidden_community.py
@@ -13,7 +13,7 @@ from Tribler.Core.DecentralizedTracking.pymdht.core.identifier import Id
 from Tribler.Core.Utilities.encoding import encode, decode
 
 from Tribler.community.tunnel import CIRCUIT_TYPE_IP, CIRCUIT_TYPE_RP, CIRCUIT_TYPE_RENDEZVOUS, \
-                                     EXIT_NODE, EXIT_NODE_SALT, CIRCUIT_ID_PORT
+    EXIT_NODE, EXIT_NODE_SALT, CIRCUIT_ID_PORT
 
 from Tribler.community.tunnel.payload import (EstablishIntroPayload, IntroEstablishedPayload,
                                               EstablishRendezvousPayload, RendezvousEstablishedPayload,
@@ -107,10 +107,14 @@ class HiddenTunnelCommunity(TunnelCommunity):
 
         self.intro_point_for = {}
         self.rendezvous_point_for = {}
+        self.infohash_rp_circuits = defaultdict(list)
+        self.seeding_introduction = {}
+
+        self.infohashes_ip_waiting_for_circuit = defaultdict(list)
 
         self.dht_blacklist = defaultdict(list)
         self.last_dht_lookup = {}
-        
+
         self.hops = {}
 
     def initiate_meta_messages(self):
@@ -147,15 +151,18 @@ class HiddenTunnelCommunity(TunnelCommunity):
                      self.on_rendezvous_established)]
 
     def remove_circuit(self, circuit_id, additional_info='', destroy=False):
+        if circuit_id in self.circuits:
+            info_hash = self.circuits[circuit_id].info_hash
+            self.infohashes_ip_waiting_for_circuit.pop(info_hash, None)
+
         super(HiddenTunnelCommunity, self).remove_circuit(circuit_id, additional_info, destroy)
 
-        # Remove introduction/rendezvous points
         if circuit_id in self.my_intro_points:
             self._logger.debug("removed introduction point")
             self.my_intro_points.pop(circuit_id)
 
         if circuit_id in self.my_download_points:
-            self._logger.error("removed rendezvous point")
+            self._logger.debug("removed rendezvous point")
             self.my_download_points.pop(circuit_id)
 
     def ip_to_circuit_id(self, ip_str):
@@ -178,7 +185,7 @@ class HiddenTunnelCommunity(TunnelCommunity):
                 info_hash = self.get_lookup_info_hash(real_info_hash)
                 hops[info_hash] = download.get_hops()
                 new_states[info_hash] = ds.get_status()
-                
+
         self.hops = hops
 
         for info_hash in set(new_states.keys() + self.download_states.keys()):
@@ -186,9 +193,32 @@ class HiddenTunnelCommunity(TunnelCommunity):
             old_state = self.download_states.get(info_hash, None)
             state_changed = new_state != old_state
 
-            force_dht_lookup = (time.time() - self.last_dht_lookup.get(info_hash, 0)) >= 300
+            time_elapsed = (time.time() - self.last_dht_lookup.get(info_hash, 0))
+            force_dht_lookup = time_elapsed >= self.settings.dht_lookup_interval
+
+            # If the introducing circuit does not exist anymore or timed out: Built a new circuit
+            if info_hash in self.seeding_introduction:
+                circuit_id, required_exit = self.seeding_introduction[info_hash]
+                if circuit_id not in self.my_intro_points and info_hash not in self.infohashes_ip_waiting_for_circuit:
+                    self._logger.debug('Recreate the introducing circuit for %s' % info_hash.encode('hex'))
+                    # check whether required_exit is still a candidate here
+                    if (required_exit[0], required_exit[1]) not in self._candidates:
+                        self._logger.debug('Candidate for exiting introducing circuit does not exist anymore')
+                        self.seeding_introduction.pop(info_hash)
+                        required_exit = None
+                    self.create_introduction_point(info_hash,
+                                                   required_exit=required_exit)
+
+            # If the rendezvous point circuit does not exist anymore: Initiate circuit generation again
+            if info_hash in self.infohash_rp_circuits:
+                for circuit_id in self.infohash_rp_circuits[info_hash]:
+                    if circuit_id not in self.my_download_points:
+                        self.infohash_rp_circuits.pop(info_hash)
+                        self._logger.debug('Force a new lookup for infohash %s' % info_hash.encode('hex'))
+                        force_dht_lookup = True
 
             if (state_changed or force_dht_lookup) and new_state == DLSTATUS_DOWNLOADING:
+                self._logger.debug('Do dht lookup for hidden services download %s' % info_hash.encode('hex'))
                 self.do_lookup(info_hash)
 
             elif state_changed and new_state == DLSTATUS_SEEDING:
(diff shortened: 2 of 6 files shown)


Tribler/community/tunnel/hidden_community.py
```diff
@@ -363,21 +394,29 @@ def check_created_e2e(self, messages):
...
 
-            _, rp_info = decode(self.crypto.decrypt_str(message.payload.rp_sock_addr, session_keys[EXIT_NODE], 
+            _, rp_info = decode(self.crypto.decrypt_str(message.payload.rp_sock_addr,
+                                                        session_keys[EXIT_NODE],
                                                         session_keys[EXIT_NODE_SALT]))
-            
-            # Since it is the seeder that chose the rendezvous_point, we're essentially losing 1 hop of anonymity
-            # at the downloader end. To compensate we add an extra hop.
+
+            # Since it is the downloader that chose the rendezvous_point, we're essentially losing 1 hop of anonymity
```
egbertbouman: I could be wrong, but as I recall the downloader sends a create-e2e to the seeder. The seeder selects and builds a rendezvous point, and sends a created-e2e back to the download, containing the sock_addr of the rendezvous point.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -437,22 +476,30 @@ def check_linked_e2e(self, messages):
...
             cache = self.request_cache.pop(u"link-request", message.payload.identifier)
-            
             download = self.find_download(cache.info_hash)
             if download:
                 download.add_peer((self.circuit_id_to_ip(cache.circuit.circuit_id), CIRCUIT_ID_PORT))
             else:
                 self._logger.error('On linked e2e: could not find download!')
 
-    def create_introduction_point(self, info_hash, amount=1):
+            for download in self.trsession.get_downloads():
```
egbertbouman: This looks like a merge error. You can remove this block of code.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -186,9 +193,32 @@ def monitor_downloads(self, dslist):
...
+            # If the introducing circuit does not exist anymore or timed out: Built a new circuit
+            if info_hash in self.seeding_introduction:
+                circuit_id, required_exit = self.seeding_introduction[info_hash]
+                if circuit_id not in self.my_intro_points and info_hash not in self.infohashes_ip_waiting_for_circuit:
+                    self._logger.debug('Recreate the introducing circuit for %s' % info_hash.encode('hex'))
+                    # check whether required_exit is still a candidate here
+                    if (required_exit[0], required_exit[1]) not in self._candidates:
+                        self._logger.debug('Candidate for exiting introducing circuit does not exist anymore')
+                        self.seeding_introduction.pop(info_hash)
+                        required_exit = None
```
NielsZeilemaker: So if the required exit does not exist anymore, we're going to fallback to a random one. Hmm.
I guess we should put th create_intro_point in an else

NielsZeilemaker: Or always force a new required_exit, by dropping the required_exit argument from the create_introduction_point method. Maybe that's more consistent.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -437,22 +476,30 @@ def check_linked_e2e(self, messages):
...
         def callback(circuit):
-            # We got a circuit, now let's create a introduction point
+            # We got a circuit, now let's create an introduction point
             circuit_id = circuit.circuit_id
-            self.my_intro_points[circuit_id].append(info_hash)
+            self._logger.debug("We got an introduction circuit ending in %s:%s" %
+                               (circuit.required_exit[0], circuit.required_exit[1]))
+            self.my_intro_points[circuit_id].append((info_hash))
+            self.seeding_introduction[info_hash] = (circuit_id, circuit.required_exit)
+            self.infohashes_ip_waiting_for_circuit.pop(info_hash)
```
NielsZeilemaker: So if we want 5 introduction points, we're fine if we only actually construct one

Tribler/community/tunnel/hidden_community.py
rjruigrok: If an introduction circuit breaks down (this happens every ten minutes by design, and in cases where a peer disconnects), it is prefered to try to build a circuit to the original introduction point again. If we select a new required_exit all the time, there is a high chance the DHT will become poluted with old introducers. That's the reason for only selecting a new required_exit: when the original introductionpoint seems to be disconnected.

Tribler/community/tunnel/hidden_community.py
rjruigrok: You're right! I made a mistake with the rebasing.

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: Makes sense, add a comment explaining this. Do we actively maintain connections to the peers in _candidates?

Tribler/community/tunnel/hidden_community.py
rjruigrok: Hmm, was a bit confused / tunnelvisioned by the original Tor design, it's indeed the case that the seeder selects the rendezvous. :+1:

Tribler/community/tunnel/hidden_community.py
```diff
@@ -107,10 +107,14 @@ def __init__(self, *args, **kwargs):
 
         self.intro_point_for = {}
         self.rendezvous_point_for = {}
+        self.infohash_rp_circuits = defaultdict(list)
+        self.seeding_introduction = {}
```
NielsZeilemaker: This seeding_introduction should be a defaultdict(list), to be able to have more than one circuit for a single infohash

Tribler/community/tunnel/hidden_community.py
```diff
@@ -437,22 +476,30 @@ def check_linked_e2e(self, messages):
...
 
         def callback(circuit):
-            # We got a circuit, now let's create a introduction point
+            # We got a circuit, now let's create an introduction point
             circuit_id = circuit.circuit_id
-            self.my_intro_points[circuit_id].append(info_hash)
+            self._logger.debug("We got an introduction circuit ending in %s:%s" %
+                               (circuit.required_exit[0], circuit.required_exit[1]))
+            self.my_intro_points[circuit_id].append((info_hash))
+            self.seeding_introduction[info_hash] = (circuit_id, circuit.required_exit)
```
NielsZeilemaker: Append instead of replace, we could have more than one introduction point for a single infohash.

Tribler/community/tunnel/hidden_community.py
rjruigrok: I replaced the `_candidates` with self.candidates(), which is a bit nicer. But I'm not maintaining connections to these peers. I assume dispersy is doing this (as I'm calling a dispersy function)?

Tribler/community/tunnel/hidden_community.py
```diff
@@ -493,18 +548,23 @@ def on_intro_established(self, messages):
             self.request_cache.pop(u"establish-intro", message.payload.identifier)
             self._logger.info("Got intro-established from %s", message.candidate)
 
-    def create_rendezvous_point(self, hops, finished_callback):
+    def create_rendezvous_point(self, hops, finished_callback, info_hash):
         def callback(circuit):
             # We got a circuit, now let's create a rendezvous point
             circuit_id = circuit.circuit_id
+            self.infohash_rp_circuits[info_hash].append(circuit_id)
```
NielsZeilemaker: See if we can get some consitstency in naming. I like infohash_rp_circuits, but then rename seeding_points to infohash_sp_circuits

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: Nope, I would assume that those peers are gone after 1 minute. 

Have a look at the discovery community for an example on how to ping/pong peers:
https://github.com/Tribler/dispersy/blob/devel/discovery/community.py#L671

Maybe we can push this functionality into Dispersy core. By being able to add candidates to a list which should be maintained.

NielsZeilemaker:
@rjruigrok otherwise it seems fine. But, maybe you should have think about those retry=5 when creating the circuits. I'm not sure if it helps the complexity etc to have multiple retry mechanisms.
Personally, I would drop those and only have this on_monitor_downloads loop to retry.

Tribler/community/tunnel/hidden_community.py
rjruigrok: I'll look into that - but problem with pinging them is that it should ping all candidates, no matter if they have circuits or not. Otherwise it leaks information about the initiator of a circuit to the exit node?

NielsZeilemaker: Let's not keep the required_exit then, pinging a shitload of candidates for the fun of it isn't really worth it. The DHT should be able to keep up.

synctext: Indeed, pinging all candidates is not worth it. You are assuming an active, but undocumented adversary:-)

At this point we're trying to get tunnels and hidden seeding working for high bandwidth usage and hopefully reliably. Essential additional security will be added when we move away from randomly selecting/trusting peers to be your rendezvous, introduction point or relays. This is another msc/phd thesis!

rjruigrok: I agree, but the current solution: self._candidates, is it really not worth to keep this thing here? IF the required candidate still exists, its always nicer to use it. There is always a chance that the candidate walker stumbles the same required_exit candidate again after some minutes, or do you think this is just useless when it's going into production?

NielsZeilemaker: I would remove the complexity, if tunnel is gone -> build new tunnel. The peer isn't going to be there.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -186,9 +187,27 @@ def monitor_downloads(self, dslist):
...
+            time_elapsed = (time.time() - self.last_dht_lookup.get(info_hash, 0))
+            force_dht_lookup = time_elapsed >= self.settings.dht_lookup_interval
+
+            # If the introducing circuit does not exist anymore or timed out: Build a new circuit
+            if info_hash in self.infohash_ip_circuits:
+                for circuit_id in self.infohash_ip_circuits[info_hash]:
+                    if circuit_id not in self.my_intro_points:
+                        self.infohash_ip_circuits[info_hash].remove(circuit_id)
+                        self._logger.debug('Recreate the introducing circuit for %s' % info_hash.encode('hex'))
+                        self.create_introduction_point(info_hash, required_exit=None)
```
NielsZeilemaker: I would modify the create_introduction_point method and drop the required_exit argument, as its not preserved while retrying.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -326,10 +345,11 @@ def on_create_e2e(self, messages):
                 relay_circuit.tunnel_data(message.candidate.sock_addr, TUNNEL_PREFIX + message.packet)
             else:
                 self._logger.debug('On create e2e: create rendezvous point')
-                
-                self.create_rendezvous_point(self.hops[message.payload.info_hash],
-                                             lambda rendezvous_point, message=message: 
-                                             self.create_created_e2e(rendezvous_point, message))
+                self.create_rendezvous_point(
```
NielsZeilemaker: Crazy indenting

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: How often is this monitor downloads loop called btw?

rjruigrok: It's in tribler_main, `# Adjust speeds and call TunnelCommunity.monitor_downloads once every 4 seconds`

NielsZeilemaker: So if a introduction point circuit is not created within 2 seconds, we'll start building more than 1 introduction point.
I revert, and put the dictionary back which contains circuits which are being built.

rjruigrok: Nope. It will only get there once in the monitor_downloads, because after the disconnected circuit check the circuit is removed from `self.infohash_ip_circuits`. In my next commit it will be appended by the create_circuit() return value, so it is always containing the actual value when a new introduction circuit is created.
btw, same holds for rendezvous point circuits.

Tribler/community/tunnel/hidden_community.py
rjruigrok: I added some indenting to the lambda thing that's one of the arguments passed to create_rendezvous. Reverted. :wink:

NielsZeilemaker:
@egbertbouman @whirm @rjruigrok seems fine to merge

Tribler/Test/test_tunnel_community.py
```diff
@@ -469,6 +469,10 @@ def load_community(session):
                 if not crypto_enabled:
                     settings.crypto = NoCrypto()
                 settings.become_exitnode = become_exit_node
+
+                # To test recovery: circuit breaks down after each 2 megebaytes
```
whirm: typo

whirm:
Looks fine to me too

Issue #1456, opened 2015-06-05
Tunnel tests don't have captions in their screenshots


NielsZeilemaker:
Please add some captions to the screenshots taken by jenkins, see
http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester_devel/4288/artifact/output//Screenshot-42.png

rjruigrok:
I will incorporate this in the next PR I'm currently working on :+1: (next PR will be distinguishing different types of exitnodes for circuits, exit nodes are only necessary for anon downloading, hidden services should work without any exit enabled circuits).

rjruigrok:
Can be closed

PR #1458, opened 2015-06-05
WIP - Circuits for hidden services don't need to be exit enabled circuits

rjruigrok:
This PR incorporates the following changes:
* Add captions to tunnel tests, see #1456.
* Hidden services related circuits (like RP, IP, etc.) should not require any exit-enabled circuit at all. 
* Only the `CIRCUIT_TYPE_DATA` circuits will look for available exit nodes, they will be only used for downloading over proxies. All other circuit types get a connectable hop as the endpoint. 
* Renaming `required_exit` to `required_endpoint`, it would be confusing because circuits may have exits that are not considered exit node. 
* exit_data() does not refuse data when it is a dispersy message. (used extensively for transferring data from one circuit to another in hidden services protocol)


rjruigrok:
retest this please

whirm:
@rjruigrok It didn't actually pass, I forgot to set the new flag I implemented in process guard to exit with error when the timeout is reached :( http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester-turbo_devel/7/console

rjruigrok:
I was not surprised. My current code could have never passed, there were some errors in it on purpose. :smile: Still have to debug some things before it's ready.

synctext:
Almost there hopefully.. 
```
INFO:OneShotProcessProtocol:[wrap_in_vnc.sh run_n...] ERR: test_simple (Tribler.Test.test_gui_server.TestGUITaskQueue) ... ok
INFO:OneShotProcessProtocol:[wrap_in_vnc.sh run_n...] ERR: test_hidden_services (Tribler.Test.test_hidden_community.TestHiddenCommunity) ... ERROR
INFO:OneShotProcessProtocol:[wrap_in_vnc.sh run_n...] ERR: test_hidden_services_with_exit_nodes (Tribler.Test.test_hidden_community.TestHiddenCommunity) ... ERROR
```

Tribler/community/tunnel/hidden_community.py
```diff
@@ -257,25 +257,26 @@ def dht_callback(info_hash, peers, _):
...
+            # Send a key-request message
+            self._logger.error("Create key request: send key request")
+            cache = self.request_cache.add(KeyRequestCache(self, circuit, sock_addr, info_hash))
+            meta = self.get_meta_message(u'key-request')
+            message = meta.impl(distribution=(self.global_time,), payload=(cache.number, info_hash))
+            circuit.tunnel_data(sock_addr, TUNNEL_PREFIX + message.packet)
+
+        # Create a circuit for getting the keys
+        self._logger.error("Create key request: create a circuit")
+        self.create_circuit(self.hops[info_hash],
```
NielsZeilemaker: Why do we need a new circuit for a key-request. Key-requests arn't part of the protocol right?

Tribler/community/tunnel/main.py
```diff
@@ -304,7 +305,7 @@ def cb(ds):
             # Introduce dispersy port from other main peer to this peer
             line_split = line.split(' ')
             to_introduce = int(line_split[1])
-            self.anon_tunnel.community.add_discovered_candidate(Candidate(('127.0.0.1', to_introduce), tunnel=False))
+            self.anon_tunnel.community.add_discovered_candidate(Candidate((self.anon_tunnel.dispersy._wan_address[0], to_introduce), tunnel=False))
```
NielsZeilemaker: I think this is wrong, Dispersy should be listening on 0.0.0.0. Hence also on 127.0.0.1. The wan_address is guessed, so should be incorrect more often.

Tribler/community/tunnel/main.py
```diff
@@ -143,7 +143,8 @@ def start_community():
 
             self.session.set_anon_proxy_settings(
                 2, ("127.0.0.1", self.session.get_tunnel_community_socks5_listen_ports()))
-
+            if introduce_port:
+                self.community.add_discovered_candidate(Candidate((self.dispersy._wan_address[0], introduce_port), tunnel=False))
```
NielsZeilemaker: Same here, I would use 127.0.0.1

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -219,7 +219,7 @@ def __init__(self, community):
         self.index = -1
 
     def has_options(self, hops):
-        return len(self.community.active_data_circuits(hops)) > 0
+        return len(self.community.socks_circuits()) > 0
```
NielsZeilemaker: No more hops?

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -1169,8 +1174,9 @@ def tunnel_data_to_origin(self, circuit_id, sock_addr, source_address, data):
...
 
     def exit_data(self, circuit_id, sock_addr, destination, data):
-        if not self.become_exitnode():
-            self._logger.error("Dropping data packets, refusing to be an exit node")
+        if not self.become_exitnode() and not TunnelConversion.could_be_dispersy(data):
+            self._logger.error("Dropping data packets on %s, refusing to be an exit node for data. " %
+                               self.circuits[circuit_id].ctype)
         elif circuit_id in self.exit_sockets:
             if not self.exit_sockets[circuit_id].enabled:
                 # We got the correct circuit_id, but from a wrong IP.
```
NielsZeilemaker: These lines are looking mighty dodgy btw, if not enabled -> enable?

Tribler/community/tunnel/main.py
rjruigrok: When running the main.py on e.g. port 8847, this causes dispersy to discover both 127.0.0.1:8847 and later also 10.0.0.25:8847 (or whatever ip). The same candidate is twice in the list of verified candidates and causing crypto problems (same hop used twice in circuit). That's why I'm using the dispersy wan address here.

Tribler/community/tunnel/main.py
rjruigrok: You're right. :smile:  but the main.py is only used for testing purposes, this optional parameter helps to find candidates quicker. 127.0.0.1 causes problems, see other comment.

Tribler/community/tunnel/main.py
NielsZeilemaker: Let's attempt to fix that in Dispersy. This should be easily detectable, as both have the same key. I'll have a go at a fix.

Tribler/community/tunnel/hidden_community.py
rjruigrok: Key requests are not part of the official protocol, but in the original code it requires a circuit with a real exit-node, and uses an exit-data for sending this dispersy message. But for hidden services, we want to avoid the use of exit-enabled circuits as much as possible because they are scarce. With this setup it is part of the protocol with a separate circuit, not requiring any exit node at all.

Tribler/community/tunnel/tunnel_community.py
rjruigrok: You can ignore most of this stuff in the PR - it's just crap for debugging,  :+1: 

Primary goal was to get hidden services running, without the need of any exit-enabled nodes. But after setting `min_circuits=0`, the whole downloading over proxies thing stops to work, resulting in strange errors caused by the socks (somehow requiring circuits to exist). I don't know why it requires those circuits and what type of data is sent through them (apparently it sends data over the circuit without using the exit_data() in tunnelcommunity?). According to @egbertbouman Chris was stuck on this for several months, and 'solved' by waiting until there are at least 4 circuits. I'm not sure whether it is worth to solve the underlying problem, so have to accept the fact it requires at least a number of exiting circuits, even when they're not used. Any downloader will create those circuits anyway when downloading files. What do you think about this?

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: But you're allowing dispersy data right? See exit_data:
```
not TunnelConversion.could_be_dispersy(data)

File: Tribler/community/tunnel/tunnel_community.py (hunk 4 above)
Comment (NielsZeilemaker) [inline Tribler/community/tunnel/tunnel_community.py]: I think he made some circuits to have some ready for downloading. Not having them ready before a download would cause problems in the socks server/libtorrent combo. That's why we would only attempt to start a download if there is a tunnel ready beforehand. For me, that makes sense

File: Tribler/community/tunnel/hidden_community.py (hunk 1 above)
Comment (rjruigrok) [inline Tribler/community/tunnel/hidden_community.py]: Correct. I discovered this method on a later moment, allowing dispersy data seems to be a cleaner option for this, as there are more places in the code where dispersy messages are tunneled to the end of a circuit (especially in the hiddencommunity I see lots of them). I'll have to clean up this code a bit. :)

Comment (rjruigrok) [inline Tribler/community/tunnel/hidden_community.py]: Furthermore, I have to create this new circuit because the circuit returned by the selection strategy yields an exit enabled circuit, which is not necessary just for transferring some keys/dispersy messages. That's why I create a separate circuit here. The whole end-to-end related protocol would not require any exit node at all this way.

Comment (NielsZeilemaker) [inline Tribler/community/tunnel/hidden_community.py]: So, you would rather use a new circuit (which takes several messages to construct, has a chance to fail etc.) than an existing one (which might be a bit overkill, due to having an actual exit_node), to send a single request/response.

Comment (rjruigrok) [inline Tribler/community/tunnel/hidden_community.py]: @NielsZeilemaker that makes sense. As the 4 exit circuits are still there, you're right and there is no need to create an extra one. :+1:

**Comment by rjruigrok** [comment]
retest this please

**Comment by rjruigrok** [comment]
@NielsZeilemaker @whirm @LipuFei 
Someone having any more comments or requests? :)

**Comment by whirm** [comment]
@rjruigrok I don't ;)

**Comment by rjruigrok** [comment]
Pushed a small missing min_circuits fix to the tests, think this was causing the anonymous_download to fail sometimes on turbo.

File: Tribler/Test/test_tunnel_community1.py [hunk 6]
```diff
@@ -0,0 +1,87 @@
...
+# see LICENSE.txt for license information
+import time
+
+# This needs to be imported before anything from tribler so the reactor gets initalied on the right thread
+from Tribler.Test.test_tunnel_base import TestTunnelBase
+
+from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
+
+
+class TestTunnelCommunity1(TestTunnelBase):
```
NielsZeilemaker: test_tunnel_community1.py -> i guess that should be renamed.

Tribler/Test/test_tunnel_community2.py
```diff
@@ -0,0 +1,57 @@
...
+# see LICENSE.txt for license information
+import time
+
+# This needs to be imported before anything from tribler so the reactor gets initalied on the right thread
+from Tribler.Test.test_tunnel_base import TestTunnelBase
+
+from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
+
+
+class TestTunnelCommunity2(TestTunnelBase):
```
NielsZeilemaker: Same here

Tribler/Test/test_tunnel_community3.py
```diff
@@ -0,0 +1,95 @@
...
+# see LICENSE.txt for license information
+import time
+
+# This needs to be imported before anything from tribler so the reactor gets initalied on the right thread
+from Tribler.Test.test_tunnel_base import TestTunnelBase
+
+from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
+
+
+class TestTunnelCommunity3(TestTunnelBase):
```
NielsZeilemaker: ?

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -405,16 +403,10 @@ def _generate_circuit_id(self, neighbour=None):
...
+            num_to_build = num_circuits - sum([1 for _ in self.data_circuits(circuit_length).itervalues()])
+            self._logger.debug("want %d data circuits of length %d", num_to_build, circuit_length)
             for _ in range(num_to_build):
-                try:
-                    if not self.create_circuit(circuit_length):
-                        break
-                except:
-                    self._logger.exception("Error creating circuit while running do_circuits")
-
+                self.create_circuit(circuit_length)
```
NielsZeilemaker: Let's check the return value of create_circuit, if its false there's no need to continue.

Tribler/Test/test_tunnel_community3.py
rjruigrok: Automatic tests are executed in parallel on the turbo jenkins job, by splitting up the tests over multiple files it will speed up the overall time needed for executing all the tests (with a few minutes). Not a nice solution, but it works.

rjruigrok: It's only necessary for the tunnel tests, because most of them take a very long time

NielsZeilemaker: Ok

rjruigrok:
The order of the tests seems to influence the results of the tests, the socks server dies immediately after the `test_anonymous_download` starts. :stuck_out_tongue_closed_eyes: I suspect the `replace_socks` method in the `test_anon_tunnel`. I'll have to fix this thing first.

whirm:
@rjruigrok Maybe we could have a more descriptive name for the test modules instead of numbering them?

rjruigrok:
Tunnel tests are also fixed for parallel running.

synctext:
@whirm @LipuFei @egbertbouman Ready for merging says Rob.. Nice.

whirm:
Great!

NielsZeilemaker:
Seems fine by me

PR #1463, opened 2015-06-08
Bugfix, default anonymous circuit length to 1

rjruigrok:
The current circuit length was set to 0, which gives strange results on adding new .torrents via gui to the downloads list in Tribler. Anonymous downloading can only be disabled by unchecking the box in the saveas dialog.


rjruigrok:
Ready!

PR #1466, opened 2015-06-09
Change Libtorrent into a TaskManager

LipuFei:


LipuFei:
retest this please

LipuFei:
retest this please

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -235,6 +234,10 @@ def load_communities():
         if self.session.get_libtorrent():
             from Tribler.Core.Libtorrent.LibtorrentMgr import LibtorrentMgr
             self.ltmgr = LibtorrentMgr(self.session)
+            self.ltmgr.initialize()
+            # FIXME(lipu): upnp APIs are not exported in libtorrent python-binding.
```
NielsZeilemaker: I know, just added it here to support it whenever they became enabled

Tribler/Main/vwxGUI/channel.py
```diff
@@ -979,11 +979,11 @@ def startDownloadFromUrl(self, url, *args, **kwargs):
 
     def startDownloadFromMagnet(self, url, *args, **kwargs):
         try:
-            return TorrentDef.retrieve_from_magnet(self.guiutility.utility.session, url, self.AddTDef, timeout=300)
+            callback = lambda meta_info: self.AddTDef(TorrentDef.load_from_dict(meta_info))
+            self.guiutility.utility.session.lm.ltmgr.get_metainfo(url, callback, timeout=300)
```
NielsZeilemaker: No return value?

LipuFei: This is async, so no return value.

Tribler/Core/APIImplementation/LaunchManyCore.py
LipuFei: The code is there, which is the line after this `FIXME`. I commented it for the moment. This `FIXME` is just a simple documentation.

Tribler/Main/vwxGUI/channel.py
NielsZeilemaker: return True after the call has been made, and revert the False if the except fires.

LipuFei: Will do.

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -345,7 +319,7 @@ def get_metainfo(self, infohash_or_magnet, callback, timeout=30, timeout_callbac
             elif infohash not in self.metainfo_requests:
                 # Flags = 4 (upload mode), should prevent libtorrent from creating files
                 atp = {'save_path': self.metadata_tmpdir, 'duplicate_is_error': True, 'paused': False,
-                       'auto_managed': False, 'flags': 4}
+                       'auto_managed': False, 'upload_mode': True}
```
NielsZeilemaker: Is this change backwards compatitble?

LipuFei: It is compatible with `0.16.18`.

NielsZeilemaker: Maybe keep both methods? Or is flags incompatible with newer versions of libtorrent?

LipuFei: I checked all the versions since `0.16.18`. `upload_mode` is an alias of flag `0x4` and this alias has been there since `0.16.18`. In one word, flag `0x4` is equivalent to setting `upload_mode` to `True`. So I decide to use the alias instead which is more readable.

NielsZeilemaker: Makes sense, everything else seems fine

Tribler/Core/APIImplementation/LaunchManyCore.py
NielsZeilemaker: I would uncomment the add upnp mapping calls

LipuFei: It doesn't work because the `add_port_mapping()` is not exposed in libtorrent session. That's why I commented it.

NielsZeilemaker: They are exposed right? They are simply not working

LipuFei: No, the funny thing is that libtorrent's documentation says to use this `add_port_mapping()` function. But it only exists in the C++ library, while the python-binding doesn't have it. If I use it with a libtorrent session, I will get a `something doesn't have this attribute...` error.

I guess perhaps they forgot to add it into the python-binding. I have already submitted an issue on libtorrent's bug tracker. Hope they get it fixed soon.

whirm:
@LipuFei ready?

LipuFei:
@NielsZeilemaker @whirm ready

LipuFei:
I will leave a ticket so that we remember to fix uPnP when it's working on libtorrent python.

PR #1467, opened 2015-06-09
Fix unique infohash problem

LipuFei:


LipuFei:
@whirm ready

whirm:
Souldn't we see why are we getting dupes on the first place?

LipuFei:
@whirm They came from the Channel community, but I haven't checked how this happened.

whirm:
```channel/community.py:_disp_on_torrent()``` If we get a batch of messages in which there are repeated torrents we get in trouble. I guess the current solution is fine.

PR #1476, opened 2015-06-15
Fix for parallel running the tunnel tests

rjruigrok:
Each tunnel test will generate its own community identifier on runtime, to prevent the tests from discovering each others candidates.

### DIFF
--- a/Tribler/Test/test_tunnel_base.py
+++ b/Tribler/Test/test_tunnel_base.py
@@ -100,9 +100,6 @@ class TestTunnelBase(TestGuiAsServer):
                     settings.crypto = NoCrypto()
                 settings.become_exitnode = become_exit_node
 
-                # To test circuit recovery: circuit breaks down after each 2 megabytes
-                settings.max_traffic = 2 * 1024 * 1024
-
                 return dispersy.define_auto_load(HiddenTunnelCommunity,
                                                  dispersy_member,
                                                  (session, settings),
@@ -157,27 +154,12 @@ class TestTunnelBase(TestGuiAsServer):
         self.sessions = []
         self.session2 = None
 
-        # Create a global community identifier
-        from Tribler.dispersy.crypto import ECCrypto
-        eccrypto = ECCrypto()
-        ec = eccrypto.generate_key(u'NID_sect571r1')
-        master_key = eccrypto.key_to_bin(ec.pub())
-
-        # Monkey patching the master key for the HiddenTunnelCommunity
-        def test_get_master_members(cls, dispersy):
-            return [dispersy.get_member(public_key=master_key)]
-        self.original_get_master_members = HiddenTunnelCommunity.get_master_members
-        HiddenTunnelCommunity.get_master_members = classmethod(test_get_master_members)
-
     def tearDown(self):
         if self.session2:
             self._shutdown_session(self.session2)
 
         for session in self.sessions:
             self._shutdown_session(session)
-
-        # Undo monkey patching of master key for the HiddenTunnelCommunity
-        HiddenTunnelCommunity.get_master_members = self.original_get_master_members
 
         time.sleep(10)
         TestGuiAsServer.tearDown(self)
--- a/Tribler/Test/test_tunnel_base.py
+++ b/Tribler/Test/test_tunnel_base.py
@@ -11,14 +11,26 @@ from Tribler.Core.simpledefs import dlstatus_strings
 from Tribler.dispersy.candidate import Candidate
 from Tribler.dispersy.util import blockingCallFromThread
 from Tribler.community.tunnel.tunnel_community import TunnelSettings
-from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
 from Tribler.dispersy.crypto import NoCrypto
+from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
 
 
 class TestTunnelBase(TestGuiAsServer):
 
     def startTest(self, callback, min_timeout=5, nr_relays=5, nr_exitnodes=3, crypto_enabled=True, bypass_dht=False):
-        self.getStateDir()  # getStateDir copies the bootstrap file into the statedir
+
+        # Create a global community identifier
+        from Tribler.dispersy.crypto import ECCrypto
+        eccrypto = ECCrypto()
+        ec = eccrypto.generate_key(u'NID_sect571r1')
+        master_key = eccrypto.key_to_bin(ec.pub())
+
+        # Monkey patching the master key for the HiddenTunnelCommunity
+        def test_get_master_members(cls, dispersy):
+            return [dispersy.get_member(public_key=master_key)]
+        HiddenTunnelCommunity.get_master_members = classmethod(test_get_master_members)
+
+        self.getStateDir()   # getStateDir copies the bootstrap file into the statedir
 
         def setup_proxies():
             tunnel_communities = []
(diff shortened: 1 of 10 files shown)


rjruigrok:
@whirm Ready for merge, it's just the torrent search that fails sometimes.

Btw, I renamed one of my tests to achieve a better distribution of the tests. It now runs under 400 seconds. 
See http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester-turbo_devel/165/artifact/output//utimes.png

synctext:
retest this please

synctext:
Is it normal that things pile up like this in the RequestCache within HiddenTunnelCommunity?
```
INFO   \ 1434367620\        dispersy:Dispersy:21644d\ Stopping Dispersy Core..
INFO   \ 1434367620\        dispersy:Dispersy:21724d\ stopping the Dispersy core...
DEBUG  \ 1434367620\        dispersy:Dispersy:21574d\ Unloading <Tribler.community.tunnel.hidden_community.HiddenTunnelCommunity object at 0x7f9588023610> (the reactor has 97 delayed calls scheduled)
DEBUG  \ 1434367620\    requestcache:RequestCache:2324d\ Clearing <Tribler.dispersy.requestcache.RequestCache object at 0x7f9583616dd0> [1]
```

synctext:
Ai
```
DEBUG  \ 1434369102\       community:AllChannelCommunity:12114d\ previous sync was 5.0 seconds ago
DEBUG  \ 1434369102\       community:AllChannelCommunity:14174d\ returning [ 0: 0: 0: 0] None
DEBUG  \ 1434369102\       community:AllChannelCommunity:12194d\ 8164f55c2f828738fa779570e4605a81fec95c9d AllChannelCommunity no candidate to take step
DEBUG  \ 1434369102\  test_as_server:TestTunnelCommunityPositive:2824d\ TestTunnelCommunityPositive.test_anon_download - calling callback, condition was not satisfied in 120 seconds (Anonymous download should be finished in 120 seconds (0.0% downloaded))
DEBUG  \ 1434369103\ LibtorrentDownloadImpl:LibtorrentDownloadImpl:6854d\ Torrent video.avi PROGRESS 0.949382007122 DLSTATE 3 SEEDTIME 0
DEBUG  \ 1434369103\   DownloadState:DownloadState:764d\ DownloadState.__init__: copy from stats 'video.avi'
DEBUG  \ 1434369103\ UserCallbackHandler:UserCallbackHandler:354d\ Session: perform_getstate_usercallback()
```

synctext:
retest this please

rjruigrok:
@synctext the errors from the non-turbo thing are generated while it was not ready yet. (before 13:30), it's not useful to look at results of those tests. Let's wait for the retests if it happens again. :)

whirm:
@rjruigrok Could you do the MMID patching in the setUp() method and undo it on tearDown()?

NielsZeilemaker:
This shouldn't be neccesary, we're manually connecting peers right? And they shouldn't connect to the trackers.

rjruigrok:
@NielsZeilemaker Yep we're manually connecting peers. But are you sure the tests do not connect to the trackers? Problems with the tunnel test started to occur after splitting up the tests into multiple files, running in parallel. It looked like the tunnel tests were sharing each others candidates, messing up the tests. 

It now looks there's still a problem. Ocassionaly, it has circuits, but it does not download. No errors, anyone sees something strange here? 
http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester_devel/4460/testReport/junit/Tribler.Test.test_tunnel_community_positive/TestTunnelCommunityPositive/test_anon_download/

NielsZeilemaker:
We're copying a bootstrap.txt file which contains am incorrect peer ip, disabling the trackers. I'll have a look.

rjruigrok:
@whirm The monkey patching takes place now in the setup/teardown. 
In my previous PR I added a limit of 2 mb to be transferred through the tunnels, to cover the code responsible for resuming the download when a circuit is removed. The tests became more strict after this change, sometimes the download did not have enough time to recover from broken circuits causing the test failures.

rjruigrok:
retest this please

rjruigrok:
retest this please

rjruigrok:
retest this please

rjruigrok:
retest this please

synctext:
retest this please

NielsZeilemaker:
@whirm I can't find the bootstrap.txt file stuff anymore. Any clue if it's still there?

NielsZeilemaker:
Ah, I remember, we're disabling the discovery community. That should suffice

whirm:
@NielsZeilemaker yeah, we removed the bootstrap.txt stuff so we could use it instead for the isolated runs.

NielsZeilemaker:
Should be fine if none of the peers are loading the discovery community. If one does, all others are going to discover the trackers as well

NielsZeilemaker:
Maybe the peer running the Tribler gui is still running the discovery community and finding all of the peers. 
I would make sure it doesn't connect to the trackers using the bootstrap.t the file

whirm:
@NielsZeilemaker I don't think we should go that way as it would break the isolated stuff. Better to make the tunnel tests listen only on the loopback interface.

NielsZeilemaker:
Why would it break?

NielsZeilemaker:
Hmm, we're setting the autoload_discovery to false, hence it also shouldn't load the discovery community in the Tribler gui.

whirm:
@NielsZeilemaker because we are creating one to direct all the tribler instances to the dedicated tracker and it would get overwritten as it was before.

NielsZeilemaker:
@rjruigrok I looked at the failed test and the only thing I found is that there's actually a vod download still being there?
```
DEBUG  \ 1434371359\ LibtorrentDownloadImpl:LibtorrentDownloadImpl:6854d\ Torrent video.avi PROGRESS 0.966255009174 DLSTATE 3 SEEDTIME 0
```
@whirm are we sure this isolated stuff is fully working?

Tribler/Test/test_tunnel_base.py
```diff
@@ -154,24 +157,39 @@ def setUp(self):
         self.sessions = []
         self.session2 = None
 
-    def quit(self):
+        # Create a global community identifier
```
NielsZeilemaker: I would remove this, it shouldn't be neccesary.

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -466,9 +466,9 @@ def create_circuit(self, goal_hops, ctype=CIRCUIT_TYPE_DATA, callback=None, max_
         if not required_endpoint:
             for c in self.dispersy_yield_verified_candidates():
                 pubkey = c.get_member().public_key
+                exit_candidate = self.exit_candidates[pubkey]
```
NielsZeilemaker: Are all verified candidates in this exit_candidates dict?

whirm:
> @whirm are we sure this isolated stuff is fully working?
> It's not connecting/talking to any trackers.

Those you are seeing are not the isolated tests, they will be enabled once we move to the parallel job as the main runner.

Tribler/community/tunnel/tunnel_community.py
rjruigrok: Yep. They're managed by overriding the callback for the dispersy-introduction-request and dispersy-introduction-response messages in the TunnelCommunity, and cleaned up periodically by the do_remove thing.

rjruigrok:
@NielsZeilemaker There is no stale vod download there, it's the tunnel test that creates a torrent based on this file for testing the anon downloading. I assume it's not doing something vod related here, so the tunnels should perform here as they should. It sometimes hangs on 96%, that was exactly the maximum a single circuit was allowed to transfer before breaking down. I removed the custom behaving of the tunnel community, possibly fixing the tests too.

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -475,21 +475,22 @@ def create_circuit(self, goal_hops, ctype=CIRCUIT_TYPE_DATA, callback=None, max_
                 else:
                     self._logger.debug("Try to find a connectable node to set as required_endpoint for this circuit")
                     required_endpoint = (c.sock_addr[0], c.sock_addr[1], pubkey)
-                    if self.candidate_is_connectable(c):
-                        self._logger.debug("Valid required_endpoint found for this circuit, stop looking further")
+                    if self.candidate_is_connectable(c) and not exit_candidate.become_exit:
```
NielsZeilemaker: So only connectable non-exitnodes are considered right?
Maybe add a comment here.

rjruigrok: Added a note. Because exit nodes are scarce, we should save bandwidth. In the current code it's still possible to have an exit as endpoint of a non-exiting circuit, just as a fallback when there are only exiting enabled nodes available.

rjruigrok:
Ready. Tests are robust again.

whirm:
Looks fine to me

NielsZeilemaker:
same here

synctext:
```
Test raised 1 unhandled exceptions, last one was: AssertionError('no channels',)
```

whirm:
@synctext that's unlrelated to this PR

PR #1483, opened 2015-06-18
Fix metadata TFTP

LipuFei:
Fix metadata download through TFTP.

Thumbnails are stored in LevelDb. The key is the SHA1 hash of the thumbnail file.

### DIFF
--- a/Tribler/Core/SessionConfig.py
+++ b/Tribler/Core/SessionConfig.py
@@ -607,6 +607,34 @@ class SessionConfigInterface(object):
         """
         self.sessconfig.set(u'allchannel_community', u'enabled', mode)
 
+    def get_enable_metadata(self):
+        """
+        Gets if to enable metadata.
+        :return: True or False.
+        """
+        return self.sessconfig.get(u'metadata', u'enabled')
+
+    def set_enable_metadata(self, mode):
+        """
+        Sets if to enable metadata.
+        :param mode: True or False.
+        """
+        return self.sessconfig.set(u'metadata', u'enabled', mode)
+
+    def get_metadata_store_dir(self):
+        """
+        Gets the metadata_store directory.
+        :return: The metadata_store directory.
+        """
+        return self.sessconfig.get(u'metadata', u'store_dir')
+
+    def set_metadata_store_dir(self, value):
+        """
+        Sets the metadata_store directory.
+        :param store_dir: The metadata_store directory.
+        """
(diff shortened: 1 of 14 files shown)


LipuFei:
retest this please

LipuFei:
@whirm ready

LipuFei:
@whirm wait. perhaps after the Plyvel PR.

Tribler/Core/SessionConfig.py
```diff
@@ -608,6 +608,48 @@ def set_enable_channel_search(self, mode):
         """
         self.sessconfig.set(u'allchannel_community', u'enabled', mode)
 
+    def get_enable_metadata_community(self):
```
NielsZeilemaker: Is the metadata community making a comeback?

LipuFei: @NielsZeilemaker I will probably just use the Channel community for metadata propagation. This PR is to fix the metadata download via TFTP only so I haven't modified the metadata community part yet.

NielsZeilemaker: But, you're adding these new methods here right? Why did you add them?

LipuFei: Good point... I will change it to something like "enable_metadata" instead.

LipuFei:
@NielsZeilemaker The `enable_metadata_community` option is no more.

NielsZeilemaker:
Seems ready to me

LipuFei:
@NielsZeilemaker @whirm Let's wait for #1488. I will do a rebase after it.

Tribler/Core/leveldbstore.py
```diff
@@ -50,11 +50,11 @@
 # it if it doesn't as it complicates the code.
```
whirm: can you remove this todo entry :D ?  
```
# TODO(emilon): This could be easily abstracted into a generic cached store
```

LipuFei: @whirm Done.

whirm:
that last failure is actually my fault, I introduced a bug in the non-parallel test execution. Fixing it.

whirm:
should be fine now

LipuFei:
@whirm ready when it passes

whirm:
@LipuFei looks fine to me

LipuFei:
@whirm Should be ready then :)

PR #1484, opened 2015-06-18
[WIP] Add Plyvel support to TorrentStore

Darsestheus:
So we tried to go with an adapter like approach, but this became complicated as the behaviour of WriteBatch is different. Does this work with you guys? Using this there is no need for subclasses or changing the calls to TorrentStore. It looks pretty ugly however.

One problem I noticed is with the TFTP handler. This requests the infohash in unicode, in _load_torrent, and py-leveldb supports this but plyvel gives an error on it. As far as I can tell, it should be no problem to do some conversion in the function.

### DIFF
--- a/Tribler/Core/torrentstore.py
+++ b/Tribler/Core/torrentstore.py
@@ -36,7 +36,13 @@
 from collections import MutableMapping
 from itertools import chain
 
-from leveldb import LevelDB, WriteBatch
+try:
+    raise ImportError("Fake import error")
+    from leveldb import LevelDB, WriteBatch
+    LEVELDBPROVIDER = "leveldb"
+except:
+    import plyvel
+    LEVELDBPROVIDER = "plyvel"
 from twisted.internet import reactor
 from twisted.internet.task import LoopingCall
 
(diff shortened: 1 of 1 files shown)


synctext:
add to whitelist

Darsestheus:
Beware of the 'raise ImportError', this was for debugging if you also have py-leveldb installed. Should be removed in the final merge.

synctext:
```raise ImportError("Fake import error")```
At least tell the poor developer something like: LevelDB import failed.

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
test this please

**Comment by Darsestheus** [comment]
retest this please

**Comment by synctext** [comment]
<invite sent>

**Comment by synctext** [comment]
_invite sent_

**Comment by Darsestheus** [comment]
retest this please

**Comment by synctext** [comment]
LOL, please resubmit as new.
Jenkins :-1:

**Comment by whirm** [comment]
ok to test

**Comment by whirm** [comment]
Don't add all this if's everywhere. Just provide an adapter class to replace LevelDB's with the same name that provides the same API.

**Comment by Darsestheus** [comment]
@whirm I have tried that approach but the problem is that in plyvel the WriteBatch is an object that is retrieved from the database, whereas in LevelDB it is first created and then added to the database. The adapter would then have to cache all the results and add them to a new WriteBatch when it knows what database it gets added to. To work around this we will have to modify the TorrentStore anyway to understand both plyvel and py-leveldb.

**Comment by whirm** [comment]
@Darsestheus You are making your life too hard. I told you to just provide a shell class for plyvel to expose the same API than LevelDB. It just seems that you will have to implement a replacement for WriteBatch too. 

So, instead of directly instancing WriteBatch in flush, use a global method ```get_write_batch``` that gets a different definition depending on whether LevelDB import was sucessful or not:

```python
try:
    from leveldb import LevelDB, WriteBatch

    def get_write_batch(_):
        return WriteBatch()

except ImportError:
    from plyvel import DB

    class WriteBatch(object):
    ....
    class LevelDB(object):
    ....

    def get_write_batch(db):
        return WriteBatch(db)

```

That's a one line modification in TorrentStore and two trivial adaptor classes for WriteBatch and LevelDB.

Darsestheus:
Closing this in favor of #1488

PR #1485, opened 2015-06-18
Use requests to get torrent via URL instead of custom code

LipuFei:


whirm:
Maybe better if you use https://twistedmatrix.com/documents/8.2.0/api/twisted.web.client.getPage.html instead, so it's a non-blocking operation.

LipuFei:
I'd like to have non-blocking as well, but GUI wants it to be blocking. So I will still need to have a blocking wrapper function if we have the non-blocking version.

LipuFei:
@whirm The `getPage()` works fine except that I can't get the header encoding. I also tried using twisted agent `twisted.web.client.Agent` according to this documentation:
http://twistedmatrix.com/documents/10.2.0/web/howto/client.html
But it looks like that I need to create a custom `Protocol` to retrieve all the data (see the doc), which I think doesn't worse the effort for this in this case. Also, if the URL is `https`, it seems that we need an additional `ClientContextFactory`.

An ugly work around is to getPage() and for every data we get, we always try to decompress it with gzip. If gzip works, we return the decompressed data, otherwise the original data will be returned.

whirm:
I found this, but I don't think it's worth the trouble as we also need to add the https stuff...
https://stackoverflow.com/questions/5428473/twisted-http-gzip-support

Just leave it as it is.

LipuFei:
@whirm Then it's ready to be merged. Maybe we can leave a ticket as a remainder if we want to change to twisted in the future.

whirm:
@LipuFei ok

PR #1490, opened 2015-06-22
Various small bugfixes

rjruigrok:
* No more bartercast debug logging
* NoCrypto fix missing key

### DIFF
--- a/Tribler/Test/test_hidden_community.py
+++ b/Tribler/Test/test_hidden_community.py
@@ -39,10 +39,10 @@ class TestHiddenCommunity(TestTunnelBase):
             self.Call(1, do_asserts)
 
         def do_progress(download, start_time):
-            self.CallConditional(180,
+            self.CallConditional(40,
                                  lambda: download.get_progress() == 1.0,
                                  lambda: take_screenshot(time.time() - start_time),
-                                 'Hidden services download should be finished in 180 seconds (%.1f%% downloaded)' %
+                                 'Hidden services download should be finished in 40 seconds (%.1f%% downloaded)' %
                                  (download.get_progress() * 100),
                                  on_fail)
 
@@ -81,7 +81,7 @@ class TestHiddenCommunity(TestTunnelBase):
                 community.trsession.lm.mainline_dht.get_peers(info_hash, Id(info_hash), cb_dht, bt_port=port)
             for community in tunnel_communities:
                 community.dht_announce = lambda ih, com = community: dht_announce(ih, com)
-            self.CallConditional(90, dht.is_set, lambda: self.Call(5, lambda: start_download(tf)),
+            self.CallConditional(30, dht.is_set, lambda: self.Call(5, lambda: start_download(tf)),
                                  'Introduction point did not get announced')
 
         self.startTest(setup_seeder, bypass_dht=True)
@@ -129,10 +129,10 @@ class TestHiddenCommunity(TestTunnelBase):
 
             d.set_state_callback(cb, True)
 
-            self.CallConditional(120,
+            self.CallConditional(30,
                                  lambda: d.get_progress() == 1.0 and hs_progress.is_set() and en_progress.is_set(),
                                  lambda: take_screenshot(time.time() - start_time),
-                                 'Hidden services download should be finished in 120s', on_fail)
+                                 'Hidden services download should be finished in 30s', on_fail)
 
         def start_download(tf):
             start_time = time.time()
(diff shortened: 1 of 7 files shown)


Tribler/Core/CacheDB/Notifier.py
```diff
@@ -69,6 +68,7 @@ def remove_observers(self):
             self.observers = []
 
     def notify(self, subject, changeType, obj_id, *args):
+        from Tribler.Core.Utilities.twisted_thread import callInThreadPool
```
whirm: hm, I didn't think of that... wait a sec, I'll move the utility functions out of the twisted_thread module instead.

whirm: Can you rebase once this one is merged? https://github.com/Tribler/tribler/pull/1492

rjruigrok: :+1:

LipuFei: @rjruigrok #1492 has been merged.

rjruigrok:
@whirm Thanks! I rebased. 

PR is ready if it passes.

rjruigrok:
retest this please

whirm:
@rjruigrok it timed out. are the tests taking more than before?

whirm:
@rjruigrok :http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester-turbo_devel/282/artifact/output/test_runners_output/00009.err/*view*/

rjruigrok:
@whirm Not that I know of, did not change anything that could cause a test to take longer. Maybe there was a higher load on the jenkins server?

whirm:
I doubt it very much (the gorilas have either 16 or 24 cores each). But even so, the test shouldn't get stuck there for 15 mins.

rjruigrok:
Can we see which test exactly took so long? I doubt it has to do with the changes in this PR, did this also happen in other PR's too?

whirm:
@rjruigrok I added a link to the log already:

> @rjruigrok :http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester-turbo_devel/282/artifact/output/test_runners_output/00009.err/*view*/

It timed out while running: 

```test_hidden_services_with_exit_nodes (Tribler.Test.test_hidden_community.TestHiddenCommunity) ... ```

rjruigrok:
Yup saw those, but the timeout occured while running `test_hidden_services_with_exit_nodes` does not say that the problem is in this test. Possibly it is caused by one of the earlier test(s). 
Is it possible to add timestamps in the test logging?

whirm:
nope, that's nose printing the info. You can look at non-timed out runs' reports to see how much each test usually takes. But the thing here is that the tests are not timing out after a reasonable amount of time by themselves.

whirm:
@rjruigrok I don't know if it's you who wrote the tests or if you are responsible of fixing them. So tell me if you aren't :)

rjruigrok:
The tests should be timing out themselves and it looks like it's not happening. I'll have a look at this. :)

whirm:
@rjruigrok are you doing it in this branch or it's ready to merge?

rjruigrok:
@whirm I reduced the CallConditional timeouts significantly to around 15s above the time a test usually takes. Maybe to rigid, but this can be adjusted in small amounts if it's necessary. Let's see how it performs after a number of retests

rjruigrok:
retest this please

NielsZeilemaker:
@rjruigrok the CallConditional timeouts don't really impact the runtime. Those will continue if the condition is true, I would attempt to convert the Call methods into a CallConditional. That would speedup the tests.

rjruigrok:
@NielsZeilemaker Hmm good point. There are a number of Calls and sleeps in the code, those can be eliminated completely. 

But as you can see in test 282 it timed out after 15 minutes of execution without any feedback, that should not happen, by having a lower CallConditional's this hopefully will not happen anymore :)

rjruigrok:
retest this please

rjruigrok:
@NielsZeilemaker I thought there were some unnecessary Call methods around, but they actually are there most of the times to get a scenario-like running of the test. for example the negative tests start a download, waits for 30 seconds using a Call(30, ...) call to ensure there is no progress after 30 seconds.

rjruigrok:
Is there still need to have the `test_anon_tunnel` test here? It's just tunneling "42" through a mocked socks. The correct working of circuits is already validated by the `test_anon_download` test. It saves another 57 seconds of testing.. :smiley:

NielsZeilemaker:
I would keep it, I guess its still a valid test.

rjruigrok:
True. 
Then this PR is ready in my opinion. Tests are running stable again.

PR #1492, opened 2015-06-22
Move the twisted utility methods to its own module.

whirm:
To avoid problems when not running the reactor on a separate thread.


whirm:
@LipuFei ready when it passes

PR #1493, opened 2015-06-22
Add timeout to checker_thread.join()

whirm:
To avoid a deadlock.


LipuFei:
@whirm This is not a very good solution. If the thread doesn't exit before we shutdown something else, it may break something such as the DBHandlers. I can have a look to change it to a `LoopingCall` if you want.

whirm:
@LipuFei ok, I'm closing this one then.

NielsZeilemaker:
I like this one, let's see what kind of errors we get whenever this join does not work. And then afterwards, let's see if we can fix those.

PR #1494, opened 2015-06-22
Fully twistify torrent_checker

LipuFei:
@whirm Should be good to go

### DIFF
--- a/Tribler/Core/TorrentChecker/torrent_checker.py
+++ b/Tribler/Core/TorrentChecker/torrent_checker.py
@@ -11,7 +11,7 @@ from Tribler.dispersy.taskmanager import TaskManager, LoopingCall
 from Tribler.dispersy.util import blocking_call_on_reactor_thread, call_on_reactor_thread
 
 from Tribler.Core.simpledefs import NTFY_TORRENTS
-from Tribler.Core.TorrentChecker.session import TRACKER_ACTION_CONNECT, MAX_TRACKER_MULTI_SCRAPE, create_tracker_session
+from Tribler.Core.TorrentChecker.session import TRACKER_ACTION_CONNECT, create_tracker_session
 from Tribler.Core.Utilities.network_utils import InterruptSocket
 
 from .session import FakeDHTSession
--- a/Tribler/Core/TorrentChecker/torrent_checker.py
+++ b/Tribler/Core/TorrentChecker/torrent_checker.py
@@ -2,17 +2,15 @@ from binascii import hexlify
 from collections import deque
 import logging
 import select
-from threading import Thread
 import time
 
 from twisted.internet import reactor
 
-from Tribler.dispersy.taskmanager import TaskManager, LoopingCall
+from Tribler.dispersy.taskmanager import TaskManager
 from Tribler.dispersy.util import blocking_call_on_reactor_thread, call_on_reactor_thread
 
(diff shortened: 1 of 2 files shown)


Tribler/Core/TorrentChecker/torrent_checker.py
```diff
@@ -111,8 +59,8 @@ def initialize(self):
 
         self._reschedule_torrent_select()
 
-        self._checker_thread = TorrentCheckerThread(self)
-        self._checker_thread.start()
+        self.register_task(u"task_check_sockets", reactor.callLater(self._socket_selection_interval,
```
whirm: How come you aren't using a LoopingCall instead?

NielsZeilemaker: The select interval is dynamic

NielsZeilemaker: Oh no, it's not. The other one is. Then I don't know.

LipuFei: I am scheduling the task again in `_task_select_sockets`.

LipuFei: I found that if something goes wrong in `_check_sessions()`, it will not return two proper read_socket and write_socket lists and the thing will break when selecting upon some invalid file descriptor. So I either schedule it when it's ok or I make a LoopingCall in the beginning and when it goes wrong, I cancel the task. Both seem ok so I did the `callLater()` which is faster for me to do.

LipuFei: @whirm @NielsZeilemaker The torrent selection interval is dynamic. This socket selection interval is fixed for the moment. I don't know if 0.3 is an appropriate number. Any idea?

NielsZeilemaker:
I'm not sure I like this new approach. We're going to check the select every 0.X seconds instead of having a thread block on the select call.

NielsZeilemaker:
I would like it better if we could remove the blocking_call decorators. Most don't really add anything.

LipuFei:
OK, I will use the old thread and see if I can remove the blocking_call.

NielsZeilemaker:
Come to think of it, why do most components have a blocking shutdown method? It doesn't make a lot of sense.

NielsZeilemaker:
Look at libtorrent, before any thread could stop the manager.
Since
https://github.com/Tribler/tribler/commit/d9b90d104196f4779b107f49424c887a1e1c503d
the shutdown is blocking on the reactor thread.

I guess it's due to the cancell_all_tasks thingy, but I'd rather make that one threadsafe

LipuFei:
@NielsZeilemaker @whirm I actually think having blocking_call on shutdown makes sense because you may want to wait for something to be fully shutdown before doing something else.

I made another one #1495 without the `blocking_call`, but the TorrentChecking thread is still there. We can decide which one to use.

NielsZeilemaker:
Having a blocking_call is extremely dangerous, like this torrent_checking example. It simply will cause a deadlock.
Maybe the blocking_call itself should giveup at somepoint. That's what we implemented in the blocking_wx thingy.

whirm:
The right thing to do would be to twistify the whole thing. You get a deferred shutdown result and you chain it to the next step of the shutdown process.

PR #1495, opened 2015-06-22
No blocking_call for TorrentChecking thread

LipuFei:
Option 2 for #1494

### DIFF
--- a/Tribler/Core/TorrentChecker/torrent_checker.py
+++ b/Tribler/Core/TorrentChecker/torrent_checker.py
@@ -123,7 +123,7 @@ class TorrentChecker(TaskManager):
         # stop the checking thread first because it can block on the reactor thread
         self._should_stop = True
         self._checker_thread.interrupt()
-        self._checker_thread.join()
+        self._checker_thread.join(10)  # do not block forever, 10 seconds should be way sufficient
         self._checker_thread = None
 
         # it's now safe to block on the reactor thread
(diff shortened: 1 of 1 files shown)


NielsZeilemaker:
I don't like this either, just fix the shutdown method. Don't mess with the rest.

LipuFei:
@whirm @NielsZeilemaker If I was thinking correctly, this should fix the dead lock.

NielsZeilemaker:
This is a lot better, but still if we happen to be on the reactor thread already, then this thread.join might cause a deadlock.

NielsZeilemaker:
It simply needs a timeout value there. You can even set a crazy high timeout, but it needs one.

LipuFei:
@NielsZeilemaker Done

NielsZeilemaker:
And log it as an error:
https://github.com/Tribler/tribler/pull/1493/files#diff-9c4b6475dbed144b806a02887d2e1cc3R132

LipuFei:
retest this please

synctext:
retest this please

LipuFei:
@whirm @NielsZeilemaker Should be ready

LipuFei:
@whirm @NielsZeilemaker And close #1494 if you merge this one

NielsZeilemaker:
We should log if the join failed, like @whirm was doing here https://github.com/Tribler/tribler/pull/1493/files#diff-9c4b6475dbed144b806a02887d2e1cc3R132

PR #1499, opened 2015-06-23
add setup.py for core functionality

mathewvermeer:
this is the setup.py that is used to install Tribler for the shadow-Internet app

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,17 @@
+from distutils.core import setup
+from Tribler.Core.version import version_id
+
+setup(
+    name='libtribler',
+    version=str(version_id),
+    author='Tribler',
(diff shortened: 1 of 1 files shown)


whirm:
Could you call the package libtribler? So when we create the actual Tribler package with the gui everything is ready? Otherwise it looks fine.

setup.py
```diff
@@ -0,0 +1,18 @@
+from distutils.core import setup
+
+execfile('Tribler/Core/version.py')
```
whirm: Why aren't you just importing version_id?

mathewvermeer: @whirm 
this answer kind of scared me away from doing that: http://stackoverflow.com/a/2073599
but since Tribler dependencies are installed manually anyway, it doesn't really matter. i'll change it.

whirm: Ah, but AFAICS that's not an issue in our case as that file is not importing anything at all, just setting a few vars. If the file was importing stuff, that might be a problem.

I don't really care either way, I was just curious :)

whirm:
Looks good. Squash the commits and I'll merge.

mathewvermeer:
@whirm 
done!

PR #1503, opened 2015-06-25
Fix some pydead object problems

LipuFei:


LipuFei:
@whirm ready when it passes

PR #1506, opened 2015-06-25
Misc refactors and fixes

whirm:
A few cosmetic refactors to make the base test classes API look a bit more consistent with twisted and a small fix for a rare issue with operations on the torrent store happening after it's closed.


whirm:
@LipuFei @NielsZeilemaker  ready

PR #1510, opened 2015-06-28
Various hidden services and tunnel related improvements

rjruigrok:
* Notify the ctrl+n debug panel about a bunch of hidden service related activities. 
* Fix for the connectability problem that occured on the key-response message relayed to the downloader. Downloading over hidden services now also works on non-connectable internet connections! 
* Fixed bug where monitor_downloads repeatedly created introduction points while the previous one was not ready yet. 
* Tunnel main.py was breaking circuits after 60 seconds, now it doesn't. 
* Removed stale anon_test.torrent file
* The dht-lookup for finding introduction points takes place over a circuit now, to prevent eavesdropping of info_hash lookups. This is implemented as a new message type in the community. 
* Rendezvous recreation removed, as it was incorrect / not necessary (I was confused by the original Tor design where the RP is located at the downloader side, in Tribler it's on the seeders' side). 
* Peers tab in download overview of the gui shows the circuit-id instead of an encoded ip-address that represents the circuit. 
* GUI: 'bittorrent swarm' image changed to a Darknet cloud near the bittorrent swarm cloud
* One test of hidden servcies uses 3 hops for downloading now
* Changed the community identifier of the tunnelcommunity. 
* Basic peer exchange by appending peers in the key-response from a seeder

Ready if it passes 
.. and if @egbertbouman and @NielsZeilemaker agree :smile:


synctext:
tetest this please

synctext:
retest this please

Tribler/community/tunnel/hidden_community.py
```diff
@@ -294,17 +374,22 @@ def check_key_request(self, messages):
...
         for message in messages:
-            # if we have received this message over a socket, we need to forward it
             if not message.source.startswith(u"circuit_"):
+                # The intropoint receives the message over a socket, and forwards it to the seeder
                 self._logger.debug("On key request: relay key request")
+                cache = self.request_cache.add(KeyRelayCache(self, message.payload.identifier, message.payload.circuit_id))
+                meta = self.get_meta_message(u'key-request')
+                message = meta.impl(distribution=(self.global_time,), payload=(message.payload.circuit_id, cache.number, message.payload.info_hash))
                 relay_circuit = self.intro_point_for[message.payload.info_hash]
-                relay_circuit.tunnel_data(message.candidate.sock_addr, TUNNEL_PREFIX + message.packet)
```
NielsZeilemaker: I'm not sure if this is a correct change. My understanding was that message.candidate.sock_addr was the ip/port of the exit-socket of the previous tunnel.
Hence, if we backpropagate it to that ip/port, it would automatically be forwarded back to the originator.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -312,26 +397,34 @@ def on_key_request(self, messages):
 
     def check_key_response(self, messages):
         for message in messages:
-            if not message.source.startswith(u"circuit_"):
```
NielsZeilemaker: Hence, this assert. The exit-peer does not need to forward it/process it as it receives it over the exit-socket. All packets received by this socket are automatically forwarded back to the originator. And hence the check_key_response method is only called at the originator and not at the exit-peer.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -228,53 +258,103 @@ def monitor_downloads(self, dslist):
...
+        candidate = self.get_candidate(sock_addr, replace=False)
+        if not candidate or candidate.get_category(time.time()) not in (u"walk", u"stumble"):
+            if max_retries > 0:
+                self._logger.debug("Create key request: retry introducing candidate")
+                retry_lambda = lambda i = info_hash, s = sock_addr, r = max_retries - 1: self.create_key_request(i, s, r)
+                self.register_task(retry_lambda, reactor.callLater(5, retry_lambda))
             return False
 
-        # 2. Send a key-request message
-        self._logger.debug("Create key request: send key request")
```
NielsZeilemaker: Similar to before, I wouldn't create a new circuit for a key-request. It doesn't make a lot of sense. Sock_addr is supposed to be connectable, hence we can use any circuit to send it a message. As long as the port received from the dht is the port where Dispersy is running.

Tribler/community/tunnel/hidden_community.py
```diff
@@ -228,53 +258,103 @@ def monitor_downloads(self, dslist):
...
             # cleanup dht_blacklist
             for i in xrange(len(blacklist) - 1, -1, -1):
                 if time.time() - blacklist[i][0] > 60:
                     blacklist.pop(i)
-
             exclude = [rp[2] for rp in self.my_download_points.values()] + [sock_addr for _, sock_addr in blacklist]
-            for peer in set(peers):
+            for peer in peers:
                 if peer not in exclude:
+                    self.add_discovered_candidate(Candidate((peer[0], peer[1]), tunnel=False))
```
NielsZeilemaker: This will cause you to make a direct connection to these peers. That is something we want to avoid right?

Tribler/community/tunnel/hidden_community.py
```diff
@@ -187,28 +224,21 @@ def monitor_downloads(self, dslist):
...
-            time_elapsed = (time.time() - self.last_dht_lookup.get(info_hash, 0))
-            force_dht_lookup = time_elapsed >= self.settings.dht_lookup_interval
-
             # If the introducing circuit does not exist anymore or timed out: Build a new circuit
             if info_hash in self.infohash_ip_circuits:
-                for circuit_id in self.infohash_ip_circuits[info_hash]:
-                    if circuit_id not in self.my_intro_points:
-                        self.infohash_ip_circuits[info_hash].remove(circuit_id)
+                for (circuit_id, time_created) in self.infohash_ip_circuits[info_hash]:
+                    if circuit_id not in self.my_intro_points and time_created < time.time() - 30:
```
NielsZeilemaker: I think this should be an or, (circuit_id not in my_intro) or (time_created < time.time() - 30).
Moreover, the 30 seems a bit low.

NielsZeilemaker:
@rjruigrok have a look at my comments, I'm not sure if the change you made makes sense.
I do like the dht-request stuffs though

Tribler/community/tunnel/hidden_community.py
rjruigrok: I also thought that. But when the key-response reached the intro-point (which should relay the response to the downloader over the originating circuit), it did not use the originating circuit but it sends directly to the downloader. 

Even more: it meant that the sock_addr of the downloader was known by the seeder? Than the whole hidden services would not make sense.

Tribler/community/tunnel/hidden_community.py
rjruigrok: Exactly, but not over the originating circuit, causing connectability problems when the downloader is not connectable

Tribler/community/tunnel/hidden_community.py
rjruigrok: I need the key of the introduction point to create a circuit that has the intropoint as required_endpoint (making it possible for the introduction point to respond on the same circuit), by adding the candidate manually it would introduce the candidate (like the walker does anyway with other candidates). 

Do you know another way to create a circuit between the downloader and intropoint which can be used by the intropoint to respond messages back?

Tribler/community/tunnel/hidden_community.py
rjruigrok: Creating an introduction circuit should not cost more than 30 seconds I think? And if it does it is a slow circuit and will create a second one. I can increase the 30 seconds, but if one circuit generation fails, it takes even longer to create another one. What about 60 seconds?

rjruigrok:
@NielsZeilemaker Although I'm stubornly defending things :smiley: thanks for the comments! 
Most comments are related to the connectability problem between downloader and introduction point that currently causes the whole hidden services thing to not work. What would have been your solution to solve this downloader-connectability problem?

synctext:
@rjruigrok Can you elaborate a bit more on the key problem?
What is the problem with connectability and the ```key_request```, ```key_response``` message?

synctext:
Yeah! it downloads for me now, exclusively from hidden seeders.
tribler_6 5pre_hidden_seeding_works_finally__29june2015

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: That must have been caused by something else then, I would verify this using the request cache object to verify.

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: Ah I get it you create a new introduction point if the previous one did not successfully connected within 30s. Ignore my comment

rjruigrok:
@synctext to elaborate: 
* a downloader retrieves the introduction point from DHT (works fine)
* it used one of the existing circuits to send a key-request to the introduction point as final destination. (works fine) Because it is a circuit that does not end in the introduction point, there is no path between downloader and introduction point. 
* the introduction point relays the key-request to the seeder over the introduction-circuit.
* the seeder responds with keys over the introduction circuit, with the downloader as final destination. (thus the downloaders' identity is known to the seeder?)
* the introduction point does not have a circuit to the downloader, but knows its sock addr, so it directly exits data to this address without using a circuit for this. But the introduction point may not be able to contact the downloader if it is not punctured yet or if the downloader is not connectable at all.

Tribler/community/tunnel/hidden_community.py
rjruigrok: I verified this yesterday, and got the insight that the introduction point does an exit_data with the dispersy key-response directly to the downloader, no circuits were involved there. Downloading over hidden services therefore only worked if the candidates of the downloader and intropoint already knew each other or when no firewall/nat is between them.

NielsZeilemaker: If the source starts with circuit_ the packet is received from a circuit, so not send to the downloader directly. It could use the wrong circuit though.

synctext:
retest this please

Tribler/community/tunnel/hidden_community.py
rjruigrok: It now uses the same circuit as it received the key-request from, I really tried to understand why these circuit_ prefixes in the code did not work out as they should, but the on-key-response did not contain any code responsible for relaying to some circuit. A separate circuit for the keys makes it much easier to understand the code. It gives some overhead - true - but guarantees the message will always arrive. 

One thing that's not clear to me now, but you know better about this stuff then me: how can the intro-point relay a message to the downloader without a circuit between them (as the key-request is 'exited' over an exit-node circuit?)

synctext:
retest this please

synctext:
4 failures with errors including ```TestHiddenCommunity.test_hidden_services_with_exit_nodes - Hidden services download should be finished in 30s - Condition was not satisfied in 30 seconds```

Tribler/community/tunnel/hidden_community.py
```diff
@@ -312,26 +398,34 @@ def on_key_request(self, messages):
...
                 yield DropMessage(message, "invalid key-response identifier")
                 continue
-
             yield message
 
     def on_key_response(self, messages):
         for message in messages:
-            self._logger.debug("On key response: received keys")
-            cache = self.request_cache.pop(u"key-request", message.payload.identifier)
-            self.create_e2e(cache.circuit, cache.sock_addr, cache.info_hash, message.payload.public_key)
```
synctext: @NielsZeilemaker @egbertbouman  Just getting "tiny tutorial" by Rob. According to experiments this piece of code was never executed on non-connectable downloaders. The reason for that is still slightly vague. Has this code ever been confirmed to work? Or only on connectable peers? Did the lack of connectability break the incoming on-key-response? (clearly I'm not understanding any of this, just trying to ensure Rob his thoughts are clearly commented/documented:-)

synctext:
retest this please

synctext:
retest this please

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: I'm going to try, lets imagine Alice is going to request the key of Bob, which has an intro-point at Charlie. Alice is going to use her pre-existing circuit she made with Dean.

```
A ---> D ---> C ---> B (key-request)
A <--- D <--- C <--- B (key-response)
```

But there was no code, which would cause Dean to forward the key-response message back to Alice right? Let me explain how it should have worked.

Charlie is the receiver of the key-request message. He receives this message over its dispersy port, as that's the port he used to announce to the DHT to. But he's receiving this message from the exit-socket of Dean (message.candidate.sock_addr). This sock_addr is put into the message as the source, and the key-request is forwarded to Bob. 

Bob can now reply to Alice, by simply instructing Charlie to exit-data to the sock_addr he received as the source. This is still the ExitSocket that Dean constructed for the circuit he has with Alice. And hence, any packet which is received over this socket, is automatically forwarded to her.

NielsZeilemaker:
@rjruigrok as a reply to your comment:
* a downloader retrieves the introduction point from DHT (works fine)
jup that's how it was implemented

* it used one of the existing circuits to send a key-request to the introduction point as final destination. (works fine) Because it is a circuit that does not end in the introduction point, there is no path between downloader and introduction point.
jup that's how it was implemented

* the introduction point relays the key-request to the seeder over the introduction-circuit.
    the seeder responds with keys over the introduction circuit, with the downloader as final destination. (thus the downloaders' identity is known to the seeder?)
not the downloader, as it's not know by the introduction-point. message.candidate.sock_addr is the ip/port of the exit-circuit choosen when sending a key-request

* the introduction point does not have a circuit to the downloader, but knows its sock addr, so it directly exits data to this address without using a circuit for this. But the introduction point may not be able to contact the downloader if it is not punctured yet or if the downloader is not connectable at all.
yes, it will exit data to this address, but its the address of the exit-socket of the pre-existing circuit. no direct contact to the downloader, as it does not know its ip/port

NielsZeilemaker:
Moreover, there is a check in the check_key_response method which verifies that this key_response was received over a circuit.
As a new check I would add a check to make sure it was received over the correct circuit, and implement that as well for all other check_ methods.

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: The only problem I can come up with is that Charlie is not using the dispersy socket to send the packet back to Dean. That could cause a problem with some Nats. 
I'm not sure how to fix that.

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: That's only due to the change to the key-request right, I would simply revert that

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: I'm pretty sure this is not true

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: I'm thinking now, that it's caused by replying to the exitsocket of the downloader (Dean) with the wrong socket

Tribler/community/tunnel/hidden_community.py
rjruigrok: Yep! I'll revert this (and the rest of the key-circuit modifications). Your
explaination about the sock-addr that's set as source of the keyrequest
message makes sense. I did not realize the possibility of sending messages
to an exit socket of a circuit, I'll have a look tonight whether I can see
another cause why the messages are not coming through using the original
code.

Thanks for your help!

Op maandag 29 juni 2015 heeft Niels Zeilemaker <notifications@github.com>
het volgende geschreven:

NielsZeilemaker: I'm guessing you need to modify the community to exit the  packet over the dispersy socket at the introduction point. Not sure how to do that in a nice manner

Tribler/community/tunnel/hidden_community.py
NielsZeilemaker: Hmm, you're actually onto something here, the funny thing is that this code is running at Charlie. Because you changed from message.candidate.sock_addr to self.dispersy.wan. 
If you keep track of message.candidate.sock_addr somewhere and forward the response there you overcome the National problems. 

I'll send you a pull request

rjruigrok: Thanks @NielsZeilemaker! I'll wait with any changes until further notice. :+1:

rjruigrok:
Thanks @NielsZeilemaker! 
I'll incorporate your idea of routing of messages to 0.0.0.0 directly to the last hop without using its wan address. 

As it breaks the on_key_response, I would propose to split the key-request into two messages, e.g. key-response1 and key-response2. With same payloads, just different on_* processing methods. What do you think about that?

NielsZeilemaker:
I'm not so sure, it might be nice to simply differentiate based on the circuit id. But let's just make it work, then we can improve upon this fix later on.

rjruigrok:
I commented the call to add_discovered_candidate with the peers retrieved from DHT, because it was causing crypto errors. The ip announced is a localhost address (e.g. 127.0.0.1:19022) while dispersy already knows a candidate 95.211.198.140:19022 with same mid). 

We need introduction to retrieve the pubkey of the intropoint to build a circuit, requiring a fix in dispersy (should be done though some moment), so the next PR will be in dispersy. Can you give me a hook in the code where to start? (community.py get_candidate method looks to me the place to fix)?

NielsZeilemaker:
I'm not sure the add discovered candidate call is required, it only due to building a new circuit to the intro point right? 
That stuff should still be reverted in my opinion

rjruigrok:
:+1: Already thought about that too. 
But then the PR you made did not make sense, as it relies on the code that forwards the key-response beyond the intro-point into the exit socket of the requesting circuit. What was your idea about this?

rjruigrok:
And yep it's indeed only required to retrieve the pubkey of the intropoint, so it could create a circuit ending IN the intro-point instead of using the exit-circuit that exit's TO the intro-point. 
Assuming that exiting to the exit-socket is also possible, I would say: let's put the intro-point as the source of the key-request message, so the seeder will respond to the intro-point again, and then use the KeyRelayCache to relay the response back into the correct exiting circuit. 
That differs from the original setup, where the exiting-relay was set as the source, so the intropoint was not involved anymore. 
It still does not explain why it would solve the problems with the NAT, or am I wrong here?

NielsZeilemaker:
It should work without the circuit ending in the intro point, only revert that part and it should be fine

rjruigrok:
@NielsZeilemaker Almost there.. But still a bit stuck here, maybe time to sleep :) 
the self.send_data call crashes on decoding org_address. The rest of the key-request/response already works. Do you have any idea what's going on there? The code looks fine in my opinion, sock_addr is valid and message payload is also correct. But something still breaks the send_data..
```
DEBUG   1435704025       community:HiddenTunnelCommunity:22034d in... 1 key-response messages from {95.211.198.142:53250}
DEBUG   1435704025 hidden_community:HiddenTunnelCommunity:4054d On key response: forward message because received over socket
ERROR   1435704025 hidden_community:HiddenTunnelCommunity:4064d ('127.0.0.1', 60558)
ERROR   1435704025        dispersy:Dispersy:17244d exception during handle_callback for key-response
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester-turbo-isolated_devel/tribler/Tribler/dispersy/dispersy.py", line 1719, in _update
    messages[0].handle_callback(messages)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester-turbo-isolated_devel/tribler/Tribler/community/tunnel/hidden_community.py", line 409, in on_key_response
    self.send_data([Candidate(cache.return_sock_addr, False)], u'data', TUNNEL_PREFIX + relay_message.packet)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester-turbo-isolated_devel/tribler/Tribler/community/tunnel/tunnel_community.py", line 674, in send_data
    circuit_id, _, _, _ = TunnelConversion.decode_data(packet)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester-turbo-isolated_devel/tribler/Tribler/community/tunnel/conversion.py", line 497, in decode_data
    org_address, offset = decode_address(packet, offset)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester-turbo-isolated_devel/tribler/Tribler/community/tunnel/conversion.py", line 490, in decode_address
    port, = unpack_from('!H', packet, offset)
error: unpack_from requires a buffer of at least 2 bytes

**Comment by NielsZeilemaker** [comment]
Hmm, we can't use send_data. Maybe switch to dispersy._send instead. That should work.

**Comment by rjruigrok** [comment]
`dropping forbidden packets to exit socket with circuit_id 2418139042`, looks like it requires the prefix to get beyond the exit socket.

**Comment by rjruigrok** [comment]
It works again. :) 
@NielsZeilemaker found a send_packet method in the tunnelcommunity that essentially does the same as dispersy._send. 

I'll do some cleanup this afternoon, and will have a look at allowing 0.0.0.0 to the endpoint instead of using dispersy wan.

**Comment by rjruigrok** [comment]
@NielsZeilemaker Tried to work out the 0.0.0.0 thing, but am a bit lost in this. Dispersy candidates and socks do not consider 0.0.0.0 as valid, I got the feeling that i'm currently on the wrong track. 
https://github.com/rjruigrok/tribler/commit/4e908ed75170d2370412a49992aa96b42e9b329b

Can you point a good place to start?

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by rjruigrok** [comment]
@synctext Lot of spamming, but tests work again. After setting the hops to 3 it failed because there were not enough non-exit candidates left, so it got never beyond building circuits for the intro/rendezvous point. That's fixed now by increasing the amount for this test.

**Comment by rjruigrok** [comment]
retest this please

**Comment by rjruigrok** [comment]
The 0.0.0.0 change will be incorporated later on in another PR, I'm going to focus a bit more on gathering graphs. 

In my opinion it's ready, but any feedback is welcome!

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
retest this please

**Comment by synctext** [comment]
```Tribler.Test.test_hidden_community.TestHiddenCommunity.test_hidden_services (from nosetests)
Failing for the past 1 build (Since Failed#482 )
Unable to access the X Display, is $DISPLAY set properly?```
http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester-turbo-isolated_devel/482/testReport/junit/Tribler.Test.test_hidden_community/TestHiddenCommunity/test_hidden_services/

**Comment by synctext** [comment]
@NielsZeilemaker  Willing to merge this? Next PR will have additional cleanup.

**Comment by whirm** [comment]
@rjruigrok maybe it would be nice to rebase against devel now that the watchdog branch has been merged so we can see what's causing the wx thread to lock sometimes?

**Comment by rjruigrok** [comment]
@whirm I rebased, let's see where it blocks. Do you want me to retest this multiple times until it yields a wachtdog hit?

**Comment by whirm** [comment]
@rjruigrok  Well, it didn't take that long :)

**Comment by rjruigrok** [comment]
Exactly.. In this case, the failed test was too sharp, I gave it a bit more time. But I see some stacktraces at the bottom of it, so it definitely does something. :)

Could it be caused by this code: 
```
        def compare_progress(lastprogress, download):
            progress = download.get_progress()
            self.assert_(progress == lastprogress,
                         "Expected no progress, but actual progress was progress=%s, lastprogress=%s" % (progress,
                                                                                                         lastprogress))
            self.quit()
```

Does it not always get to quit() if the assert above it is triggered?

**Comment by rjruigrok** [comment]
Which means: each time a tunnel test fails and contains code like that, it will block the thread. And if the tests passes smoothly it will correctly quit, so wx will not block?

**Comment by whirm** [comment]
I haven't looked at the other runs (just the first one after the rebase). There it fails at shutdown because _shutdown_session() is hoarding the main thread for up to a minute waiting for the session to close. I dont know why it only seems to happen in your tests. I'm looking for a solution for the shutdown procedure and then we can try again and see where do we get.

**Comment by rjruigrok** [comment]
So if it takes about a minute to close the session and you solve this problem, it will indirectly improve the runtime of the tunneltests? Cool! :+1:

**Comment by whirm** [comment]
Well, It's waiting for the session to shutdown, I guess it will take as much time to shutdown as before, but we need to keep the wx event loop spinning in the meanwhile, see:  #1526

**Comment by rjruigrok** [comment]
@whirm Less output now after rebase :)
```
test_hidden_services_with_exit_nodes (Tribler.Test.test_hidden_community.TestHiddenCommunity) ... watchog wx thread *******TRIPPED!******, hasn't been set for 10.0152 secs.

*** STACKTRACE - START ***


### ThreadID: 139637084768000 Thread name: WatchDog
Locals by frame, innermost last:
/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Core/Utilities/instrumentation.py:172 print_all_stacks:
/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Core/Utilities/instrumentation.py:145 run:
/usr/lib/python2.7/threading.py:810 __bootstrap_inner:
/usr/lib/python2.7/threading.py:783 __bootstrap:

### ThreadID: 139636699535104 Thread name: PoolThread-twisted.internet.reactor-1
Locals by frame, innermost last:
/usr/lib/python2.7/threading.py:340 wait:
/usr/lib/python2.7/Queue.py:168 get:
/usr/lib/python2.7/dist-packages/twisted/python/threadpool.py:212 _worker:
/usr/lib/python2.7/threading.py:763 run:
/usr/lib/python2.7/threading.py:810 __bootstrap_inner:
/usr/lib/python2.7/threading.py:783 __bootstrap:

### ThreadID: 139637490747136 Thread name: MainThread
Locals by frame, innermost last:
/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Test/test_tunnel_base.py:164 tearDown:
/usr/lib/python2.7/unittest/case.py:358 run:
/usr/lib/python2.7/unittest/case.py:393 __call__:
/usr/lib/python2.7/dist-packages/nose/case.py:151 runTest:
/usr/lib/python2.7/dist-packages/nose/case.py:133 run:
/usr/lib/python2.7/dist-packages/nose/case.py:45 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/core.py:62 run:
/usr/lib/python2.7/dist-packages/nose/core.py:207 runTests:
/usr/lib/python2.7/unittest/main.py:95 __init__:
/usr/lib/python2.7/dist-packages/nose/core.py:121 __init__:
|          env = {'BUILD_URL': 'http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/4811/', 'NOSE_TESTS_TO_RUN': 'Tribler/Test/', 'BUILD_TAG': 'jenkins-GH_Tribler_pull-request-tester_devel-4811', 'TRACKER_CMD': '', 'SHELL': '/bin/bash', 'REMOTE_WORKSPACE_DIR': './', 'ROOT_BUILD_CAUSE': 'GHPRBCAUSE', 'BUILD_NUMBER': '4811', 'JOB_URL': 'http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/', 'PYTHONPATH': '/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/gumby:/home/jenki...
|       config = Config(addPaths=True, args=(), configSection='nosetests', debug=None, debugLog=None, env={}, exclude=None, files=[], firstPackageWins=False, getTestCaseNamesCompat=False, ignoreFiles=[<_sre.SRE_Pattern object at 0x7effe2f25030>, <_sre.SRE_Pattern object at 0x7effe2ee7330>, <_sre.SRE_Pattern object at 0x7effe3022cf0>], ignoreFilesDefaultStrings=['^\\.', '^_', '^setup\\.py$'], include=None, includeExe=False, logStream=<open file '<stderr>', mode 'w' at 0x7effe30691e0>, loggingConfig=None, options=...
/usr/bin/nosetests:9 <module>:

### ThreadID: 139637352249088 Thread name: Twisted
Locals by frame, innermost last:
/usr/lib/python2.7/dist-packages/twisted/internet/epollreactor.py:379 doPoll:
/usr/lib/python2.7/dist-packages/twisted/internet/base.py:1204 mainLoop:
/usr/lib/python2.7/dist-packages/twisted/internet/base.py:1192 run:
/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Core/Utilities/twisted_thread.py:55 _reactor_runner:
/usr/lib/python2.7/threading.py:763 run:
/usr/lib/python2.7/threading.py:810 __bootstrap_inner:
/usr/lib/python2.7/threading.py:783 __bootstrap:

*** STACKTRACE - END ***
```

**Comment by whirm** [comment]
I guess this doesn't help: https://github.com/rjruigrok/tribler/blob/hs-ctrl-n/Tribler/Test/test_tunnel_base.py#L164

**Comment by rjruigrok** [comment]
Hmm indeed... don't know what the effect of removing will be, it's been there for very long. Blame 
https://github.com/Tribler/tribler/commit/998fbe5c8d3e6dae37904ad8668ec5e5ad560258

**Comment by rjruigrok** [comment]
@whirm Blocking wx thread is fixed now! :+1:

**Comment by whirm** [comment]
@rjruigrok well, you've just shaved 10 secs out of the tests :D

**Comment by rjruigrok** [comment]
Haha! :+1:

**Comment by synctext** [comment]
retest this please

**Comment by whirm** [comment]
@rjruigrok In case you haven't seen it:
```
ERROR   1435936592             log:twisted:443  Unhandled Error
  File "/usr/lib/python2.7/threading.py", line 763, in run
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Core/Utilities/twisted_thread.py", line 55, in _reactor_runner
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 1192, in run
    self.mainLoop()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 1201, in mainLoop
    self.runUntilCurrent()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 797, in runUntilCurrent
    f(*a, **kw)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/Main/tribler_main.py", line 463, in define_communities
    kargs=tunnel_kwargs)[0]
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/dispersy/dispersy.py", line 384, in define_auto_load
    community = community_cls.init_community(self, master, my_member, *args, **kargs)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/dispersy/community.py", line 226, in init_community
    community.initialize(*args, **kargs)
  File "/home/jenkins/workspace/GH_Tribler_pull-request-tester_devel/tribler/Tribler/community/tunnel/tunnel_community.py", line 287, in initialize
    self.dispersy.endpoint.listen_to(self.data_prefix, self.on_data)
exceptions.AttributeError: 'NullEndpoint' object has no attribute 'listen_to'
```

rjruigrok:
@whirm Weird. Where does the NullEndpoint came from... In which test did you see this error?

whirm:
@rjruigrok  http://jenkins.tribler.org//job/GH_Tribler_pull-request-tester_devel/4813/testReport/junit/Tribler.Test.test_gui_dialogs/TestGuiDialogs/test_add_save_create_dialog/

NielsZeilemaker:
We need to modify the nullendpoint and add a listen_to method which does a pass. 
We're using the nullendpoint to prevent dispersy to contact the outside world

NielsZeilemaker:
It's in dispersy/endpoint.py

rjruigrok:
@NielsZeilemaker a PR is in the dispersy repo. Only the automated tests are not triggered yet.

whirm:
@rjruigrok merged, you can update the dispersy ref. in this PR.

whirm:
retest this please

whirm:
retest this please

whirm:
@rjruigrok looking good!

whirm:
There's been a timeout, but it's on the gui tests and it seems it's not a blocked thread anyway.

rjruigrok:
Added peer exchange via the key-response message. Just to get more other peers if one peer is found.

rjruigrok:
retest this please

rjruigrok:
I'm not planning any more work in this PR, so it can be considered ready now. :smiley:

rjruigrok:
retest this please

synctext:
OK, if this is merged we'll be building V6.5 RC1!

PR #1515, opened 2015-06-30
New metadata

LipuFei:


LipuFei:
retest this please

LipuFei:
retest this please

LipuFei:
@NielsZeilemaker @whirm ready except that I need to make a new unit test for it.

LipuFei:
@NielsZeilemaker @whirm It's ready. Feel free to merge if you think it's ok.

whirm:
@LipuFei maybe it would be nice to have the tests here too?

LipuFei:
I've made a metadata unit test in `test_my_channel.py`.

LipuFei:
@whirm @NielsZeilemaker Ready

whirm:
Let's wait for 6.5 to be released before merging this.

Tribler/Test/test_my_channel.py
```diff
@@ -125,6 +129,88 @@ def do_page():
...
+        def take_screenshot_quit():
+            self.screenshot('Home')
+            self.quit()
+
+        def check_home():
+            self.CallConditional(20, lambda: len(self.guiUtility.frame.home.aw_panel.list.GetItems()) == 1,
+                                 take_screenshot_quit, "No thumbnail torrent in Home panel")
+
+        def on_channel_torrent_created(subject, events, object_id, data_list):
+            add_metadata = False
```
NielsZeilemaker: I'm not sure about this, there's way too much logic here. I would modify it to read a rss-feed from disk.
Moreover, what happened to the ffmpeg thumbnail creation stuffs? That should still be implemented right, after a download changes to seeding it should attempt to create thumbnails if the torrent is part of the channel a user "owns".

LipuFei: @NielsZeilemaker I first came up with two ideas: one is to start a metadata-injector create a channel and some thumbnails, and then start a normal tribler, subscribe to that channel, and wait for the thumbnails to arrive. Another one is that one: I create a channel myself, add some thumbnails to it, and then I should be able to see them on the home page.

Here I took the second one. Because we no longer expose the metadata functionality to the users, in order to keep it simple, I have to do some of those crazy magic. Otherwise, a good way is to create an API wrapper for channels, but this requires a lot of refactory again.

The problem with the two instance approach is that, the metadata may take a long time to arrive on the normal tribler instance. In my case, when I manually tested the injector, it took at least 5 minutes for me to be able to see the channel which I create with the injector, and another 5 to 10 minutes to get the thumbnails. So, if we make the unit test this way, it might be a longest test ever.

As to the ffmpeg thing, since we are in full control of the metadata, there is no need for ffmpeg any more because we retrieve the thumbnails through certain TV or movie databases.

NielsZeilemaker: We already have a test which uses an rss-feed to create item in a channel. Can't you change the rss stuffs a bit to also accept a local rss.xml file? That should be pretty trivial.

As for the ffmpeg stuffs, not all rss-feeds have valid thumbnails attached to the items. It's a nice fallback to be able to generate thumbnails for files without a thumbnail in the rss/.torrents added from a file.

LipuFei: I made a new channel manager and a rss-parser in `Tribler/Core/Modules`, which support creating multiple channels. That's why I didn't touch the old rss-feed. I will try to figure out a better way to do this.

NielsZeilemaker: The old rss-feed? I would simply copy that unit-test, but specify a local file, feedparser seems to be able to handle it:
https://pythonhosted.org/feedparser/introduction.html#parsing-a-feed-from-a-local-file
Then it will actually test the rss metadata import stuff

LipuFei:
@NielsZeilemaker I made a new metadata test using a rss feed. No more magic.

LipuFei:
@NielsZeilemaker @whirm ready

whirm:
Looks fine to me

Tribler/Main/Utility/GuiDBTuples.py
```diff
@@ -153,8 +154,12 @@ def hasChannel(self):
 
     @cacheProperty
     def metadata(self):
-        # TODO(lipu): fix this to get metadata
-        return {}
+        if not self.metadata_dict:
```
NielsZeilemaker: I would change this to
```
if self.metadata_dict is None:

File: Tribler/Main/vwxGUI/list_details.py [hunk 3]
```diff
@@ -129,6 +129,9 @@ def __init__(self, parent):
 
         self.parent = parent
         self.torrent = Torrent('0', '0', '', 0, 0, 0, 0, 0, None)
```
NielsZeilemaker: And then set metadata_dict to False here

Tribler/Main/vwxGUI/list_details.py
```diff
@@ -522,20 +525,23 @@ def updateDetailsTab(self):
         self.channel.Show(show_channel)
 
         # Toggle thumbnails
-        # FIXME(lipu): fix the thumbnail path to use metadata
-        thumb_dir = os.path.join(u"",
-                                 binascii.hexlify(self.torrent.infohash))
-        thumb_files = [os.path.join(dp, fn) for dp, _, fns in os.walk(thumb_dir)
-                       for fn in fns if os.path.splitext(fn)[1] in THUMBNAIL_FILETYPES]
-        show_thumbnails = bool(thumb_files)
+        if self.torrent.metadata and 'thumb_hash' in self.torrent.metadata:
```
NielsZeilemaker: Then this won't cause the two database calls

Tribler/Main/vwxGUI/list_details.py
```diff
@@ -129,6 +129,9 @@ def __init__(self, parent):
 
         self.parent = parent
         self.torrent = Torrent('0', '0', '', 0, 0, 0, 0, 0, None)
+        self.torrent.torrent_db = self.guiutility.channelsearch_manager.torrent_db
```
NielsZeilemaker: Then you don't need these two lines

NielsZeilemaker:
Otherwise, very nice improvement @LipuFei

Tribler/Main/vwxGUI/list_details.py
LipuFei: I think I added this because some other code was using `torrent.torrent_db` and `torrent.channelcast_db` as well, but I don't remember which part of code any more...

Tribler/Main/vwxGUI/list_details.py
LipuFei: I remember when I was testing using two tribler instances, a injector and a normal tribler, I had this problem that the `modification` messages (the metadata messages) come way slower than the torrents. On the normal tribler instance, I could see the torrents almost immediately after I subscripted to the injector channel, but it took another 20 minutes or more to get those modification messages and show the thumbnails.

It is nice to avoid frequent database I/O here, but I don't know if we do this, the thumbnails will still work as it is now.

I will do some more tests to see if there is any work around.

LipuFei: @NielsZeilemaker I applied the `metadata=False` method and did a few tests with two instances again. So, when I could see two thumbnail torrents on the front page (which always reloads the thumbnail torrents), I switched to my subscribed channel and I could only see one torrent with thumbnail. The other one didn't reload its metadata_dict.

LipuFei: @NielsZeilemaker I found out why. There is a bug in when loading ChannelTorrents on demand: it somehow assigned a wrong channel to this ChannelTorrent. That's why it couldn't fetch its metadata.

LipuFei:
@NielsZeilemaker I've fixed a metadata loading bug and used the `False` method to avoid loading metadata for non-channel torrents. This PR has been sitting here for months so I already forgot what I was doing... Please ignore some of those comments above.

LipuFei:
retest this please

LipuFei:
retest this please

LipuFei:
retest this please

LipuFei:
@NielsZeilemaker @whirm ready

PR #1520, opened 2015-07-01
Remove TorrentStateManager

LipuFei:


LipuFei:
@whirm ready when it passes

PR #1527, opened 2015-07-03
Don't indiscriminately swallow exceptions.

whirm:


whirm:
@LipuFei ready if you agree :)

Issue #1540, opened 2015-07-07
KeyError


brandc:
I was using the down arrow key to preload everything in a channel, and this is what showed up in the log.
```
Traceback (most recent call last):
  File "/home/iron/tmp/tribler/Tribler/Main/vwxGUI/list_body.py", line 1396, in <lambda>
    self.Bind(wx.EVT_MENU, lambda event: self.SelectNextItem(True), id=adownId)
  File "/home/iron/tmp/tribler/Tribler/Main/vwxGUI/__init__.py", line 119, in invoke_func
    return func(*args, **kwargs)
  File "/home/iron/tmp/tribler/Tribler/Main/vwxGUI/list_body.py", line 1301, in SelectNextItem
    select].GetSize()[1] + 1) / self.GetScrollPixelsPerUnit()[1]
KeyError: 'K\r`\xc0p\xd69.\xb7\x98\xb7^\x01\xb8\xa4\xacS\xc7^\x8d'
```

Edited by @whirm: One copy of the stacktrace is enough.

LipuFei:
@whirm I am not sure if I can fix this for `6.5`... Those list problems in the GUI is always a big headache.

whirm:
@LipuFei let's do it for 6.6

whirm:
@brandc thanks for the report!

NielsZeilemaker:
Basically this line:
https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/list_body.py#L1297
needs a call to self.HasItem(select) and then it should be fixed.

devos50:
@whirm is this one fixed?

whirm:
I don't see the check that @NielsZeilemaker mentioned there, so I assume it's not.

PR #1554, opened 2015-07-09
[refactor+new features] A generic channel manager

LipuFei:
This is a subset of PR #1515. After this one is merged, I will make a PR to make metadata working on GUI and make a new unit test for metadata using the RSS feed approach, which is less crazier than #1515.

1. This PR makes a generic `channel_manager` that handles the functionality related to the channels owned by myself, including:
  * creating new channels
  * modifying rss feeds attached to a channel owned by myself
  * rss feed parser that can parse and retrieve metadata
2. Certain old channel functionalities have been replaced with this new `channel_manager`.


LipuFei:
retest this please

LipuFei:
I need to fix this `ListBody` `PyDeadObject` problem now. It's killing me.

LipuFei:
retest this please

LipuFei:
@NielsZeilemaker @whirm The `PyDeadObject` problem should be fix. Ready for review.

Tribler/Core/Modules/channel/channel.py
```diff
@@ -0,0 +1,119 @@
...
+    def shutdown(self):
+        self.cancel_all_pending_tasks()
+        for key, rss_parser in self._rss_feed_dict.iteritems():
+            if rss_parser is not None:
+                rss_parser.shutdown()
+        self._rss_feed_dict = None
+        self._channel_community = None
+        self._session = None
+
+    @call_on_reactor_thread
```
NielsZeilemaker: Why are these methods run on the reactor thread?

Tribler/Core/Modules/channel/channel_manager.py
```diff
@@ -0,0 +1,124 @@
...
+        from Tribler.community.channel.community import ChannelCommunity
+        for community in self.session.lm.dispersy.get_communities():
+            if isinstance(community, ChannelCommunity) and community.master_member and community.master_member.private_key:
+                channel_obj = ChannelObject(self.session, community, is_created=True)
+                channel_obj.initialize()
+                self._channel_list.append(channel_obj)
+
+                self._logger.debug(u"loaded channel '%s', %s", channel_obj.name, hexlify(community.cid))
+
+    @blocking_call_on_reactor_thread
```
NielsZeilemaker: Hmm, is this a good idea?

Tribler/Main/metadata-injector.py
```diff
@@ -31,21 +32,31 @@
 from Tribler.dispersy.util import call_on_reactor_thread
 
 
-def parge_rss_config_file(file_path):
-    rss_list = []
+def parge_channel_config_file(file_path):
```
NielsZeilemaker: parse?

Tribler/Main/vwxGUI/list_body.py
```diff
@@ -763,6 +763,9 @@ def OnCollapse(self, item=None, raise_events=True, from_expand=False):
 
     @warnWxThread
     def OnChange(self, scrollToTop=False):
+        if GUIUtility.getInstance().utility.abcquitting:
```
NielsZeilemaker: Why is this here?

Tribler/Core/Modules/channel/channel.py
LipuFei: @NielsZeilemaker Here we have two functions that can change `rss_feed_dict`: one is create_rss_feed(), which is an API function directly called by others, and the other is this `_on_channel_created()`, which is called by the `Notifier`. I want to make sure that only one guy can do something with the `rss_feed_dict`, so I made them both `call_on_reactor_thread`.

Tribler/Core/Modules/channel/channel_manager.py
LipuFei: @NielsZeilemaker There is no other threads involved here except the reactor thread, so it is safe to do this here.

Tribler/Main/metadata-injector.py
LipuFei: @NielsZeilemaker Nice one! I will fix it.

Tribler/Core/Modules/channel/channel.py
NielsZeilemaker: Put a lock around them, there is no need to push them on the reactor thread.

Tribler/Core/Modules/channel/channel_manager.py
NielsZeilemaker: Is there a need for it?

Tribler/Main/vwxGUI/list_body.py
LipuFei: This is to fix the mysterious `PyDeadObject` problem in `ListBody`. See the first few test results for the backtrace.

This problem happens during the shutdown.  I tracked it down to this `SetupScrolling()`. This is a wx function which schedules `_SetupAfter()` with `wx.CallAfter()`, and this `_SetupAfter()` is the one that raised the `PyDeadObject` error.

The only place I found that calls `SetupScrolling()` is this `OnChannge()` in `ListBody`. Because this new RSS parser I made is too fast (compared to the old one) that in the rss_import test, you can see tens of torrents showing up simultaneously on the "Channel Manage" page, I suspect that this performance improvement might be the issue.

In short, the solution to fix this `PyDeadObject` problem is to not let `ListBody` update itself during the shutdown. So I added this line.

NielsZeilemaker: List_body should be a generic component, no need for a call to guiutility here.
A PyDeadObject can be fixed by doing
```
if self:
```
no need for abcquitting

LipuFei: This is the tricky part. `if self` doesn't work for this problem. The issue is that when this `OnChange()` is called by an wx event and `ListBody` is not dead yet. So `_SetupAfter()` has scheduled but somehow not executed before `ListBody` is dead. After we exit the MainLoop and wake up the MainThread again (for the shutdown splash window), this `_SetupAfter()` got scheduled and `PyDeadObject` appears.

So the only way I found is to use this `GUIUtility.getInstance().utility.abcquitting` to avoid calling `SetupScrolling()`.

Tribler/Core/Modules/channel/channel_manager.py
LipuFei: I will make this one a `call_on_reactor_thread`.

Tribler/Core/Modules/channel/channel.py
LipuFei: I will schedule a task with `self.register_task()`. This way, we can cancel the tasks without having them dangling around.

LipuFei:
@NielsZeilemaker @whirm ready

LipuFei:
retest this please

Tribler/community/channel/community.py
```diff
@@ -726,21 +728,23 @@ def _disp_check_modification(self, messages):
                     if infohash:
                         infohash = str2bin(infohash)
                         logger.debug(
-                            "Incoming swift-thumbnails with infohash %s from %s",
+                            "Incoming metadata-json with infohash %s from %s",
                             infohash.encode("HEX"),
                             message.candidate.sock_addr[0])
 
-                        if not th_handler.has_metadata("thumbs", infohash, thumbnail_subpath):
+                        print "Incoming metadata-json with infohash %s from %s" % (infohash.encode("HEX"),
```
NielsZeilemaker: I guess you should remove this line, otherwise. It's ready right @LipuFei ?

LipuFei: It is. I will remove this line first

NielsZeilemaker:
Go ahead and merge this one. What issues are you guys working on btw?

LipuFei:
@whirm Ready

@NielsZeilemaker Not Many. We have some issues that @synctext needs to check.
#1568 is tricky; I cannot reproduce #1569;
Now I'm working on #1574 and #1575

NielsZeilemaker:
I would create a unit-test for #1568, a special build for #1569 which enables the debug log of the quit methods in mainframe.py. Also #1540 seems like a trivial fix.

Tribler/Main/vwxGUI/channel.py
```diff
@@ -1519,20 +1521,26 @@ def OnChange(self, event):
...
         item = event.GetEventObject()
 
-        self.torrentfeed.deleteURL(item.url, self.channel.id)
-        self.RebuildRssPanel()
+        my_channel_object = self.guiutility.utility.session.lm.channel_manager.get_my_channel(self.channel.id)
+        my_channel_object.remove_rss_feed(item.url)
 
     def OnRefreshRss(self, event):
-        self.torrentfeed.doRefresh()
+        # TODO(lipu): fix this
```
whirm: Forgot this?

whirm:
The rest seems fine

Tribler/Main/vwxGUI/channel.py
LipuFei: This will be fixed by the next PR. I have split this metadata thing into several smaller parts. It's a remainder for myself..

LipuFei:
@whirm That TODO should go actually. I removed it and fixed one more typo. Ready when it passes.

whirm:
Great

Issue #1559, opened 2015-07-13
Don't know how to decrypt incoming message


brandc:
Is this a error a communication error, or a Tribler bug. If the first, does it need to be logged?

```
ERROR   1436826407.74 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 3628865491
ERROR   1436826842.04 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 789349534
ERROR   1436827114.30 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 1509502498
ERROR   1436827114.60 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 1509502498
ERROR   1436827198.96 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 2179541574
ERROR   1436827199.45 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 2179541574
ERROR   1436827269.73 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 1133489124
ERROR   1436827269.98 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 1133489124
ERROR   1436827424.62 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 83676674
ERROR   1436827428.26 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 83676674
ERROR   1436827510.80 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 3746535498
ERROR   1436827749.58 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 3800905826
ERROR   1436827901.57 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 1480436787
ERROR   1436827974.54 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 840835394
ERROR   1436827981.42 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 840835394
ERROR   1436828061.45 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 282010085
ERROR   1436828071.42 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 282010085
ERROR   1436828087.12 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 282010085
ERROR   1436828218.20 tunnel_community:898   Don't know how to decrypt incoming message for circuit_id 4066292712
[...]

PR #1563, opened 2015-07-14
Change error to warning for communication errors in crypto

rjruigrok:
Crypto errors sometimes happen when a circuit breaks (due to security limiters or timeouts) and there are still messages in transit on the circuit. Those messages can't be decrypted because the keys have already been cleaned up. Those errors are useful for debugging but not relevant in production, as they can happen any time.

Change error to warning for communication errors related to encrypting / decrypting. Solves #1559


rjruigrok:
retest this please

PR #1589, opened 2015-08-19
Some fixes

LipuFei:
1. Fixed some log level problem
2. Fixed a download path problem switching from normal download to anonymous seeding


LipuFei:
@whirm ready when it passes

PR #1597, opened 2015-08-20
Fix leveldb file system encoding problem

LipuFei:


LipuFei:
Should be against next...

PR #1663, opened 2015-09-14
Be more informative when dropping an incoming msg

whirm:
When due to unknown circuit ID or hop count value.

### DIFF
--- a/Tribler/community/tunnel/tunnel_community.py
+++ b/Tribler/community/tunnel/tunnel_community.py
@@ -1,44 +1,42 @@
 # Written by Egbert Bouman
 # Based on ProxyCommunity by Chris Tanaskoski and Rutger Plak (crypto)
 
-import time
-import random
 import logging
-
+import random
+import socket
+import time
 from collections import defaultdict
+from cryptography.exceptions import InvalidTag
 
 from twisted.internet import reactor
 from twisted.internet.protocol import DatagramProtocol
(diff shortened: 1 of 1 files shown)


whirm:
@LipuFei ready if it passes

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -1216,22 +1214,35 @@ def crypto_out(self, circuit_id, content, is_data=False):
...
+        if circuit:
+            if len(circuit.hops) > 0:
+                # Remove all the encryption layers
+                layer = 0
+                for hop in self.circuits[circuit_id].hops:
+                    layer += 1
+                    try:
+                        content = self.crypto.decrypt_str(
+                            content, hop.session_keys[ORIGINATOR], hop.session_keys[ORIGINATOR_SALT])
+                    except InvalidTag as e:
```
rjruigrok: @whirm 
Just FYI. The times I saw an InvalidTag error, it was caused by one candidate that was added as hop two or more times after each other in a single circuit (which should not happen anyway), then the InvalidTag error occured while peeling of the encryption layers. 
So if the error occurs again, it might be useful to have a look which hops are participating in the circuit.

whirm: Ok, thanks for the info, I'll add this to my TODO list.

PR #1675, opened 2015-10-01
Typo? Breakage on non-win

quite:


synctext:
@quite   Interesting patch! we don't often get these to-the-point patches.
Can you elaborate a bit more please? Does this make Tribler work on your NetBSD, or Whonix distribution?

quite:
It seems like the constructor, __init__, of LevelDB recently began requiring 1+2 arguments (self, store_dir, state_dir), but this call here was not updated to reflected that. Apparently, it was only implemented and tested on windows -- the call to __init__ before the 'else' does indeed have the required arguments.

whirm:
@quite that's a bug on the failback plyvel wrapper, I'll fix it there (It shouldn't require the state_dir arg as it's not using it anyway) but that code is wrong, I'll refactor the whole thing.

Thanks for the report/PR this and sorry for the delay!

PR #1686, opened 2015-10-16
Handful o'fixes

### LINKED ISSUE #1670: On Windows, opening a torrent file using Tribler will start Tribler.exe in an arbitrary directory
Problem found in #1588

On Windows, opening a torrent file using Tribler will start Tribler.exe in an arbitrary directory. A direct consequence is that all code using relative paths will break. One of them is the "logger.conf" problem.

A solution is to use absolute path for "logger.conf" in `tribler.py`.

whirm:
Closes #1670

### DIFF
--- a/Tribler/Core/Video/utils.py
+++ b/Tribler/Core/Video/utils.py
@@ -116,6 +116,12 @@ def return_feasible_playback_modes():
 
     l = []
     try:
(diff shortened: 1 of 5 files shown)


whirm:
retest this please

whirm:
retest this please

PR #1699, opened 2015-10-29
Removed some unused variables in the ABCApp class

devos50:
It appears that some variables in the ABCApp class are not used anymore and are probably safe to remove.


whirm:
ok to test

whirm:
@devos50 oh,  you should send this to `devel` instead. `next` is frozen for release.

devos50:
@whirm ok I will send it to the devel branch!

whirm:
Great, thx!

PR #1701, opened 2015-10-29
Removed some unused variables in the ABCApp class.

devos50:
I came across some unused variables when inspecting code. These variables are not used in ABCApp class and outside it and can probably be removed.


whirm:
ok to test this

whirm:
Thanks!

PR #1719, opened 2015-11-10
WIP: Wx3 support

devos50:
This PR adds support for the Python Wx3 library. Note that it is still experimental and should be tested on more platforms.

Also fixed a key error on the Darwin platform where the left menu would not collapse when clicking on ‘channels’. 

I also changed the used Python version on OS X to 64 bit and added a link to the README on how to setup a developer environment on OS X.


Tribler/Main/tribler.py
```diff
@@ -27,7 +27,7 @@
     # in the windows and mac distribution, there may be no version available.
     # so select a version only when there is any available.
     if wxversion.getInstalled():
-        wxversion.select("2.8-unicode")
+        wxversion.select("3.0-osx_cocoa")
```
whirm: This will probably break windows, right?

devos50: True, not sure which wxversion I should select on Windows yet so I will change it after I know this.

README.md
```diff
@@ -71,7 +71,7 @@ Now installing the list of dependencies should no longer throw an error.
 
 #### Windows and OSX
 
-Tribler runs on Windows and OSX, but development is only supported on Linux.
```
whirm: I would rather have the documentation on the repo than on the wiki.

Tribler/Main/vwxGUI/list.py
```diff
@@ -2180,7 +2180,7 @@ def __SetData(self):
         self.list.SetData(data_list)
         self.ResizeListItems()
         self.DisableItem(2)
-        if not self.guiutility.frame.videoparentpanel:
+        if not self.guiutility.frame.videoparentpanel and sys.platform != 'darwin':
```
whirm: This is already fixed on devel, you should rebase.

whirm:
Looking good!

whirm:
@devos50 did you notice any performance improvements with that update? I'm curious to see if that will improve things on OSX.

Tribler/Main/tribler.py
whirm: great

README.md
devos50: Shall I add a .md file in the docs/ directory?

devos50:
I haven't compared it yet with the release version on OS X since I'm unable to run the release version somehow. When I open the app, it closes immediately. I will open an issue about it.

whirm:
@devos50 try the latest RC and see if that one works fine, please. We need to make sure it works before we release it :)

devos50:
The latest RC does not work but 666d057 fixes the issue with the images path.

README.md
whirm: Sure, go ahead. 
Move the linux stuff that we already have on the README.md there too.

whirm:
BTW, remember to update `debian/control` with the updated dependencies.

devos50:
Will open a new PR.

PR #1730, opened 2015-11-17
WIP: Pylint fixes

devos50:
In this PR, I've submitted several Pylint fixes. The goal is to conform to the PEP8 rules (see https://www.python.org/dev/peps/pep-0008/ for more information).


whirm:
I wouldn't work on this until the 6.5.X cycle is done. It will make really annoying rebasing the fixes from `next` back to `devel`. 
Once we decide the 6.5.X cycle is done, I would do a unix style newline conversion to the whole source and then a few non-aggressive autopep8 passes. 

Once that gets merged we can start with the manual fixes.

Thanks!

devos50:
We will wait after the 6.5 release before we tackle this. For now, I will close this PR.

PR #1731, opened 2015-11-17
Windows openssl applink fix

devos50:
On Windows, I encountered an issue when setting up a 64-bit environment. The problem occurred when invoking M2Crypto functions. I've explained the issue in more detail on Stack Overflow [here](http://stackoverflow.com/questions/33720087/error-when-importing-m2crypto-in-python-on-windows-x64/33728728#33728728).

One solution for this problem is to recompile Python and to include `openssl/applink.c` in the `Python.c` file. Since recompiling Python turned out to be quite hard (and not very desirable), I believe we can use the workaround as given by the answer on Stack Overflow. By making use of M2Crypto.BIO, we avoid using the openssl applink interface and thus the No OPENSSL_AppLink error.

Note that this problem only occurs on Windows (and I believe in particular 64-bit builds). On Linux and OS X, it is working without any problem.

### DIFF
--- a/Tribler/Core/permid.py
+++ b/Tribler/Core/permid.py
@@ -2,7 +2,7 @@
 # see LICENSE.txt for license information
 import os
 import logging
-from M2Crypto import Rand, EC
+from M2Crypto import Rand, EC, BIO
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


whirm:
Looks good. 
Could you squash the fix commit?

synctext:
impressive!
is this not a mem leak? explicit create membuf, no FreeMemBuf...

whirm:
Squash the commits together.

whirm:
@devos50 could you add some comment on why it's done like this so people don't "optimize" it later and break it again?

devos50:
@whirm I added comments to the top of the file.

whirm:
Great, thanks!

PR #1735, opened 2015-11-19
Refactor tunnel code

NielsZeilemaker:
Overal cleanup of the code. Biggest changed:
* Removed retrying for circuits
* Fixed Extended message (the created message was actually forwarded back to the origin, instead of the last hop). Causing the Extended message never to be created :smile: 
* Tried to simply the crypto stuffs, but failed
* Removed the tunnel_to_end methods

For now I left on all the debug logging. Laurens, have a look at the number of (very small) packets being sent to the same ip/ports during a libtorrent download (e.g. the test_hidden_community). I think adding a buffer, and batch sending those will substantially improve download speeds.

### DIFF
--- a/Tribler/community/tunnel/routing.py
+++ b/Tribler/community/tunnel/routing.py
@@ -3,6 +3,7 @@ import time
 from Tribler.community.tunnel import CIRCUIT_STATE_READY, CIRCUIT_STATE_BROKEN, CIRCUIT_STATE_EXTENDING, \
     CIRCUIT_TYPE_DATA
 from Tribler.dispersy.crypto import LibNaCLPK
+from Tribler.dispersy.candidate import Candidate
 import logging
 
 __author__ = 'chris'
@@ -95,7 +96,14 @@ class Circuit(object):
         @return bool: whether the tunnel request has succeeded, this is in no
          way an acknowledgement of delivery!
         """
-        return self.proxy.tunnel_data_to_end(destination, payload, self)
+
+        self._logger.error("Tunnel data (len %d) to end for circuit %s with ultimate destination %s", len(payload),
+                           self.circuit_id, destination)
+
+        num_bytes = self.proxy.send_data([Candidate(self.first_hop, False)], self.circuit_id, destination, ('0.0.0.0', 0), payload)
+        self.proxy.increase_bytes_sent(self.circuit_id, num_bytes)
+
+        return num_bytes > 0
 
     def destroy(self, reason='unknown'):
         """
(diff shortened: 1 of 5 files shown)


NielsZeilemaker:
retest this please

Tribler/community/tunnel/routing.py
```diff
@@ -95,7 +96,14 @@ def tunnel_data(self, destination, payload):
         @return bool: whether the tunnel request has succeeded, this is in no
          way an acknowledgement of delivery!
         """
-        return self.proxy.tunnel_data_to_end(destination, payload, self)
+
+        self._logger.error("Tunnel data (len %d) to end for circuit %s with ultimate destination %s", len(payload),
```
whirm: We should lower this, any level will be shown if a test fail and if you are debugging the tunnels locally, you can always edit logger.conf to lower this particular logger.

NielsZeilemaker: I know, I left it at error for now to show @lfdversluis know that there are a huge amount of small packets being sent.

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -458,15 +472,12 @@ def do_remove(self):
...
+    def create_circuit(self, goal_hops, ctype=CIRCUIT_TYPE_DATA, callback=None, required_endpoint=None, info_hash=None):
         assert required_endpoint is None or isinstance(required_endpoint, tuple), type(required_endpoint)
         assert required_endpoint is None or len(required_endpoint) == 3, required_endpoint
-        retry_lambda = first_hop = None
+        first_hop = None
 
-        if max_retries > 0:
-            retry_lambda = lambda h = goal_hops, t = ctype, c = callback, r = max_retries - 1, e = required_endpoint: \
-                self.create_circuit(h, t, c, r, e)
+        self.tunnel_logger.warning("Creating a new circuit of length %d", goal_hops)
```
whirm: This one too

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -739,53 +750,92 @@ def check_create(self, messages):
...
+                continue
+            yield message
+
+    def check_destroy(self, messages):
+        for message in messages:
+            circuit_id = message.payload.circuit_id
+            cand_sock_addr = message.candidate.sock_addr
+
+            if circuit_id in self.relay_from_to:
+                pass
```
whirm: Was that meant to be a `continue`?

NielsZeilemaker: I copy/pasted that logic from the original on_destroy method. I was too afraid to touch it, so I simply converted into a pass :smile:

lfdversluis:
I managed to run the test successfully once but didn't saw output in my terminal. Are the logs written to a file somewhere?

Also, I get segmentation faults when running the test:

``nosetests Tribler/Test/test_tunnel_community_positive.py``
Gives ``Segmentation fault (core dumped)``

whirm:
@lfdversluis if a test passes, it's output is dropped by default. Have a look at nose's arguments if you want the output to be always printed.

If you get segfaults, run the tests through gdb so you can inspect the process state.

whirm:
@lfdversluis I'm merging this so you can work from `devel`

PR #1747, opened 2015-11-24
Fixed tribler.sh startup script on OS X

devos50:
To run Tribler from the Tribler.sh script, we first have to add the current working directory to the `PYTHONPATH` variable. Otherwise, the Tribler modules cannot be found.

### DIFF
--- a/tribler.sh
+++ b/tribler.sh
@@ -10,6 +10,9 @@ else
     TRIBLER_SCRIPT=Tribler/Main/tribler_profiler.py
 fi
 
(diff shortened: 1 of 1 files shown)


tribler.sh
```diff
@@ -40,6 +40,9 @@ else
             exit 1
         fi
 
+	PYTHONPATH=.:"$PYTHONPATH"
```
whirm: If it has to be done on both Linux and OSX, just move the Linux one out of the `if`.

whirm:
Great, thanks!

PR #1756, opened 2015-12-01
fixed dialog race condition

lfdversluis:
@whirm Please take a look at this PR. We (also @devos50) think this should give a better structure and solve the race dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ flycheck_*
 memorydump.out
 test_file
 test_file.torrent
+
+# pycharm
+**.idea/
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -6,7 +6,6 @@ import logging
 import os
 import socket
 from binascii import hexlify
-
 from Tribler.Core import NoDispersyRLock
 from Tribler.Core.APIImplementation.LaunchManyCore import TriblerLaunchMany
 from Tribler.Core.CacheDB.Notifier import Notifier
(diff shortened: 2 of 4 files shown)


lfdversluis:
Hmmm apparently your commits were also taken? Git magic!

whirm:
ok to test

lfdversluis:
```
Test FAILed.
No JDK named ‘null’ found
Finished: FAILURE
```

Stuff broke?

whirm:
@lfdversluis yeah, the leaseweb tunnel helpers will be gone for a few days. Maybe it's a good oportunity for you guys to fix the tunnel tests in a way that they actually timeout and fail instead of getting stuck there and making the whole jenkins job time out. :)

lfdversluis:
@whirm nosetests has support for this so it should be an easy fix. How long do you suggest the timeouts should be?

whirm:
@lfdversluis that's what gumby is doing now, and it's not what we want. 
What we do want is that each test is able to fail if there's no progress instead of just wait indefinitely until it gets killed. The test is broken and it should time out by itself like the other tests do.

whirm:
re. timeout duration: I would run it a bunch of times and decide a sensible value from the observed times.

lfdversluis:
@whirm Processes can run functions as targets as well. Then you can kill individual functions, would that work?

whirm:
No. Look at any other test that time out (most of the tunnel tests that work and the download stuff too for instance) and see how it's done. If the tests that get stuck (easy to see by running `nosetests -v ...` ) are doing it properly, then it might be a bug on the tunnel community or even elsewhere.

whirm:
@lfdversluis have a look at some of the failed tests:

https://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/5139/artifact/output/test_runners_output/00000.err/*view*/

https://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/5139/artifact/output/test_runners_output/00002.err/*view*/

whirm:
The second link is not related to this PR.

lfdversluis:
@whirm I am now working on getting timeouts in the tests, but I noticed the prestart() function is of course renamed now since I branched of my fix.

A few other files call prestart as well, so I want to fix this with my new PR that also fixes the test timeout. Is that ok?

whirm:
retest this please

lfdversluis:
retest this please

PR #1771, opened 2015-12-07
Added information about setting the PATH variable

devos50:


whirm:
How come you have to add this paths? I would expect the python installer to leave you with a fully working installation.

devos50:
Because @lfdversluis did not add his Python directory to his `PATH` so we had to fix that.

lfdversluis:
(The installer offers this at the bottom of the list of packages, but it is disabled by default.)

whirm:
@devos50 waiting for the installer comment as talked about in person.

devos50:
Done.

whirm:
Looks great, could you squash it together?

devos50:
@whirm done

PR #1774, opened 2015-12-07
Fixed popup menus in panels on OS X

devos50:
On OS X, the popup menus would not disappear when clicking outside the popup menu. Changing the mouse event fixes this.


whirm:
Does it still work fine on windows/Linux?

devos50:
Yes it is, just tested on Ubuntu and Windows.

whirm:
Great, merging then.

PR #1787, opened 2015-12-13
Added health value inside the bar

pedrofraga:
Attending Software Engineering at FEUP, we were encouraged to add a feature to this project. We decided to add the percentage value inside of the health bar so that it would be easier for the user to compare.


synctext:
OK to test

devos50:
Thanks for the feature addition. I just tested it on Ubuntu and it seems to work fine. The percentage is becomes hard to read when the bar is about half filled. Maybe you could address this issue (i.e. by making the text another, more readable color when the percentage is > 40%)?

Also, I would not place the `esof-docs` directory directly inside the root directory. All documentation should be moved in the `doc` directory. I'm not sure whether your report is suitable for merging inside the repo, this is up to @synctext and @whirm probably.

I really like the `Logical View.png` image since it gives a very good overview of the current packages and dependencies on each other. I suggest this could be added to the `doc` directory. It would be nice if some developer-oriented documentation about this image (together with the images) is written :+1:

pedrofraga:
Since the _esof-docs_ folder contains class reports from the subject we are attending we decided to remove it from this branch. Some updates were done in the health bar, the colour squares doesn't have a border anymore, the squares are completely filled now. We think the text its easier to read that way.

whirm:
ok to test

whirm:
@pedrofraga Thanks for the PR!

Could you please squash all the commits into a single one?

whirm:
I didn't catch the squares thing. I guess it would be better for it to be a config option?

@synctext what do you say?

whirm:
While we wait for @synctext's response, @pedrofraga: could you leave only the commit with your changes?
You should have done a rebase instead of a merge.

synctext:
Commit now only contains minor changes to 2 files. No report anymore. Looks good.

All documentation goes to either the Gitihub Tribler wiki or Docs Dir.

btw rebase needed..

whirm:
@synctext I was expecting an opinion on removing the vertical divisions on the health bar (the bar is plain now with a percent in the center)

whirm:
@synctext gave the OK, so merging.

@pedrofraga thanks for the PR!

PR #1788, opened 2015-12-13
WIP: Changes to the Tribler startup process

devos50:
When working on the CLI client, I noticed that the startup process of Tribler is incredibly complex and a lot of functionality is placed inside the `tribler_main.py` file. I believe this file should be as clean as possible and that starting up the Tribler core should only take several lines of code (to encourage developers to work work with the Tribler core).

After an investigation of this class, I noticed various methods which could be moved to the GUI part of the code (mainly, the `GUIUtility` class). This includes many of the session callback methods which often only manipulate GUI elements.

Moreover, I removed many unused methods and even a file. Also, I propose to remove some very old comments that are not relevant anymore.


whirm:
I would like to have the version checking thing working rather than deleted.
Is anybody up for that?

devos50:
I will gradually split this PR into multiple ones since this one is way too big.

PR #1794, opened 2015-12-14
Update Tribler development on OS X.md

devos50:


whirm:
Looking great!
Could you squash the thing into one commit if you are done updating it?

devos50:
Yeah I will, have one issue left regarding `libsodium.dylib` which cannot be found by `ctypes`. Think this is due to to SIP (https://support.apple.com/nl-nl/HT204899) but I do not want to disable it so I should think about a good solution for this. For now I have a very bad workaround place but this should be changed (see the open .txt file on the Mac Mini).

devos50:
As discussed with @whirm I will file a bug report for this at Apple.

devos50:
Apparently, it is not a bug but a security feature, see https://forums.developer.apple.com/thread/13161 for more information about it. I've added a dedicated section about this to the guide with two solutions: either symlink the `libsodium.dylib` file into the Tribler root directory or disable System Integrity Protection (not recommended).

devos50:
@whirm I'm done with this PR.

whirm:
@devos50 what about the packaged version of Tribler?

PR #1804, opened 2015-12-17
Startup/close window now make use of the event system

devos50:
When starting and closing Tribler, there are several ticks in the window. I've adjusted the code so these ticks are broadcasted by the event system in Tribler. This allows, for instance, a CLI to capture the ticks and display a line of text or something like that.

In order to make this possible, I've added two new events: `NTFY_STARTUP_TICK` and `NTFY_CLOSE_TICK` for ticks in the startup and close window respectively.


Tribler/Core/CacheDB/Notifier.py
```diff
@@ -10,7 +10,7 @@
                                      NTFY_ACTIVITIES, NTFY_REACHABLE, NTFY_CHANNELCAST, NTFY_VOTECAST, NTFY_DISPERSY,
                                      NTFY_TRACKERINFO, NTFY_UPDATE, NTFY_INSERT, NTFY_DELETE, NTFY_TUNNEL,
                                      SIGNAL_ALLCHANNEL_COMMUNITY, SIGNAL_SEARCH_COMMUNITY, SIGNAL_TORRENT,
-                                     SIGNAL_CHANNEL)
+                                     SIGNAL_CHANNEL, NTFY_STARTUP_TICK, NTFY_CLOSE_TICK)
```
lfdversluis: I just reordered these additions to keep the NTFY and SIGNAL stuff separated :D

devos50: Ah good one, thanks! Didn't pay attention to that when adding this.

whirm:
This needs to go to `devel`

devos50:
@whirm I will open a new PR in that case.

PR #1805, opened 2015-12-17
Starting up/closing splash now makes use of the event/callback system of Tribler

devos50:
When starting and closing Tribler, there are several ticks in the window. I've adjusted the code so these ticks are broadcasted by the event system in Tribler. This allows, for instance, a CLI to capture the ticks and display a line of text or something like that.

In order to make this possible, I've added two new events: `NTFY_STARTUP_TICK` and `NTFY_CLOSE_TICK` for ticks in the startup and close window respectively.


Tribler/Main/tribler_main.py
```diff
@@ -290,11 +289,15 @@ def __init__(self, params, installdir, autoload_discovery=True,
             self.ready = True
 
         except Exception as e:
-            if self.splash:
-                self.splash.Destroy()
-
+            self.guiUtility.destroy_startup_splash()
             self.onError(e)
 
+    def startup_tick(self, subject, changetype, objectID, *args):
```
whirm: Could you make those inner methods and prepend `on_` to the names as they are event callbacks?

whirm:
Otherwise looks nice, do that and squash it.

devos50:
@whirm done

PR #1812, opened 2015-12-18
Removed unused statements, changed comments

devos50:


devos50:
retest this please

devos50:
retest this please

whirm:
have you checked if prctlimported is used at all in gumby & co.?

devos50:
Yes, it is not used by Gumby.

Verstuurd vanaf mijn iPhone

whirm:
Also, that statement about the httplib bug is also fixed?

devos50:
That bug was part of M2Crypto 0.16 which is pretty old. Moreover, we are not using the https part of M2Crypto since it was used by the version check system.

whirm:
ok

PR #1825, opened 2015-12-30
Temporarily disable broken tunnel tests

whirm:
It doesn't make sense to waste so much time re-running broken tests. 
When somebody gets around fixing the tunnel community and its tests, they can be enabled again.

Pioneer one torrent is dying and it takes a long time to start downloading, so switching to a different torrent for the test.


devos50:
Seems like the best solution for now, retesting gets really annoying after some time. Nevertheless, they should be fixed.

whirm:
retest this please

PR #1840, opened 2016-01-04
WIP: Issue 1837

lfdversluis:


Tribler/Core/leveldbstore.py
```diff
@@ -83,6 +82,14 @@ def __getitem__(self, key):
         try:
             return self._pending_torrents[key]
         except KeyError:
+            pass
+
+        try:
+            return self._db.Get(key)
+        except:
```
whirm: Only catch the IO error

Tribler/Core/leveldbstore.py
```diff
@@ -83,6 +82,14 @@ def __getitem__(self, key):
...
+            #self.flush()
```
whirm: Let's try to sleep() for a bit? It's as ugly as it can get but let's see if it helps

Tribler/Core/leveldbstore.py
```diff
@@ -83,11 +82,18 @@ def __getitem__(self, key):
         try:
             return self._pending_torrents[key]
         except KeyError:
+            pass
+
+        try:
+            return self._db.Get(key)
+        except IOError:
+            #self.flush()
```
whirm: Let's try to sleep() for a bit? It's as ugly as it can get but let's see if it helps

lfdversluis:
10 second sleep and the error still persists.

whirm:
mph... What about closing and reopening when this happens?

devos50:
@lfdversluis I guess we can close this one since it is fixed by #1853?

lfdversluis:
yep.

Issue #31, opened 2013-03-15
strongest possible reputation system


synctext:
Idea: strongest possible reputation system by combining all existing science.

Problem: our scientists Rahim, Adele, Nitin and Dimitra did not combine their work.
Reason: optimal publication strategy is one clean and simple idea
There are no joint publications, no re-usage of algorithms and little collaboration.

We evaluated in isolation multiple enhancements to make accounting of contributions stronger and incentive compatible:
- max-flow calculation
- delete old information to reduce storage cost
- depletion of often used paths, provides sybil resilience with additional bookkeeping
- introduce negative trust, these negative edges increase sybil attack resilience
- friendship as an edge property to strengthen the trust overlay against sybils
- age as an peer property to strengthen against sybil attack
- biased random walk, taking into account peer and edge properties
- double signed signature (proof-of-work)
- append-only signatures, misreporting attack degrades it to weaker white-washing attack
- running total account of consumed and contributed resources, degrades misreporting attack
- transfer encrypted information, only provide decode key after proof-of-work is signed
- incremental signatures, reduce risk-taking by demanding signatures during transfer
- use n-hop Max-Path to reduce compulational cost
- reason from betweenness centrality node
- use full gossip
## 

Reputation system proposals often assume either:
1. Tamper-proof and scalable storage of information
2. Scalable storage of information

Option number 1 is beyond the state-of-the-art, without assuming an honest central server with infinite capacity. Option number 2 is difficult to achieve. For scalability seasons, the storage layer will always be imperfect and will never offer both instant, accurate and complete recall of information. Thus we need to be able to tolerate limited accuracy or coverage in addition to the usual misreporting, lying and sybil creation.
## 

A reputations system needs to determine (assuming background gossip of trust vectors):
 1 Who to talk to
 2 how to reach them
 3 What to say
 4 How much to trust

Existing work is focused on no.4 mostly, while the others have proven to be
much harder to solve. Nobody identified them as a problems and no
solutions exist. 
Trust calculation is just 50 lines of code, while no. 1-3 is beyond
the state-of-the-art in epidemic protocols.

Missing topics:
3. "what to say", Efficient synchronisation of BarterCast (subgraph) trust vectors

Existing material:
 1. Who to talk to
    Biased random walk
 2. how to reach them
    Churn resiliance, state preservation, live overlay, NAT traversal, etc
 4. How much to trust
   Meulpolder 2008: Max-flow
   Rahim: various sybil-limiters
   Adele: rewards also effort
[...]

PR #1843, opened 2016-01-05
WIP: Multichain

pimveldhuisen:
New pull request for the multichain feature. Supersedes both #1429 and #1784.

ToDo list:
- [ ] fix the re-introduce candidate bug in the gumby experiments.
- [ ] dramatically increase the blocksize from 1 MByte 100MByte and reduce checking interval.
- [ ] disable community by default, no autoload.


whirm:
Why did you squash all commits together?

whirm:
Also, please, don't keep closing and opening PR's for the same stuff, we lose track of the comments and discussions everytime you do.

pimveldhuisen:
There were some merge conflicts in the rebase. Rebasing all 8 commits individually would result in more merge conflicts, including ones that would be reversed by the later commits.

pimveldhuisen:
The other PR was initially meant for jenkins only. Together with the redoing of the rebase it seemed to make sense to make a new PR. But I understand your annoyance. This should be the definitive pull request.

whirm:
It's a pity, @snorberhuis did a pretty decent job at separating concepts on each commit... 
Have you tried with rerere? 
https://git-scm.com/blog/2010/03/08/rerere.html
https://chuva-inc.com/blog/fast-tip-enable-git-rerere-right-now

whirm:
re rebases: just do a `push --force` to the same branch when you are sure it's fine.

pimveldhuisen:
I redid the rebase to include the separate commits.

whirm:
Thats way better, thanks! :)

snorberhuis:
disable community by default, no autoload

I thought I did this.

whirm:
retest this please

pimveldhuisen:
An update on the work in this pull request:
I've commited code that implements the following changes: (see discussion in https://github.com/Tribler/tribler/issues/31)

- The multichain is now non-blocking. This means that when a node sends a signature request, it will first store the incomplete block, and continue it's chain from there. This means a block now has two hashes that refer to it: the one that hashes the complete block (responder point of view) and the one that hashes only the requester part. This makes the multichain somewhat less elegant, but avoids mutual exclusion problems across the network.
- The requesting of blocks is no longer triggered by a threshold, but by the closing of tunnels.
- The multichain blocks no longer contain the dispersy member id (mid) of the interacting nodes, but the full public key. The mid is a sha1 hash of the public key. Since the hashing of the key is irreversible, storing the mid does not allow a third party to verify blocks, which is an important aspect of the multichain philosophy.
- The previous hashes now use sha256 instead of sha1, since the later is no longer considered secure.
- The GENESIS_HASH is now defined as [00...01], the EMPTY_HASH [00...00], instead of the other way around.

The code passes the unit tests (except for crawler tests), and creates valid blocks in an integration test (https://jenkins.tribler.org/job/Experiment_multichain_100/21/). However the, multichain blocks still do not seem to include all the traffic, i.e. not all MBs are transcribed. This seems to me to be the most major open issue with the multichain right now. 
Another thing that has to be included is the new crawler. @Captain-Coder has been working on this, and we have yet to combine our code.
This branch will (obviously) also still need some squashing of commits.

Captain-Coder:
retest this please

whirm:
add to whitelist

pimveldhuisen:
retest this please

whirm:
retest this please

whirm:
retest this please

pimveldhuisen:
retest this please

whirm:
retest this please

pimveldhuisen:
@NielsZeilemaker Would you care to take a look at the code for this pull request? Since you provided a lot of input on the previous multichain PR, I would love to hear you thoughts on this, and improve code quality where necessary. Note that we would prefer to merge this sooner rather than later, meaning we are aiming in the first place for working with benevolent users. Security checks against attackers are a future step. Of course, the code that is already in here should be high quality.

NielsZeilemaker:
Hi @pimveldhuisen, I'll have a look maybe on Sunday.

synctext:
retest this please

pimveldhuisen:
retest this please

synctext:
@NielsZeilemaker If you have a bit of time.. 
We're aiming to test this code on an exit node to do accounting for upcoming V6.6 release.
Exit nodes are moving 5 TByte/day, so quite a stress & scalability test.

pimveldhuisen:
retest this please

Tribler/Test/test_multichain_community.py
```diff
@@ -0,0 +1,544 @@
...
+from Tribler.community.tunnel.tunnel_community import TunnelExitSocket
+
+from Tribler.Test.test_as_server import BaseTestCase
+
+from Tribler.dispersy.tests.dispersytestclass import DispersyTestFunc
+from Tribler.dispersy.util import blocking_call_on_reactor_thread
+from Tribler.dispersy.tests.debugcommunity.node import DebugNode
+
+
+class TestMultiChainCommunity(DispersyTestFunc):
```
devos50: I would rather make a subclass of `AbstractServer` since this test is part of the Tribler test suite. `AbstractServer` contains the watchdog and additional checks to verify whether the tests leave a clean reactor/leave no open sockets.

NielsZeilemaker:
Hi Guys,

So I had a quick look, and have some comments. Its seems fine (unitl now :smile:), but I would change one thing. I don't like the custom synchronization stuffs you made (the crawler). Basically, what I think you like to do is get all messages of a particular peer. I would simply use the normal bloom filter sync but modify the _select_and_fix method to only return messages made by the peer itself.

For a pointer have a look at:
https://github.com/NielsZeilemaker/tribler/blob/good_old/Tribler/community/privatesocial/community.py#L79

Then you would be able to get rid of all the crawl messages. But you'll need to change the distribution to fullsyncdist.

Also, If you switch to fullsyncdist, dispersy will store the message for you. And hence, also both pubkeys/signatures. Therefore I would remove the pubkeys, and signatures from the messages and add a foreign key to the message.id to be able to load the dispersy message on demand.

Have a look at here:
https://github.com/Tribler/tribler/blob/devel/Tribler/community/channel/community.py#L390
and here
https://github.com/Tribler/tribler/blob/devel/Tribler/community/channel/community.py#L429

Basically, we're storing the dispersy packet_id, and loading the message whenever we need it. So no need to store the signatures etc ourselves..

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,433 @@
...
+
+        from Tribler.Core.simpledefs import NTFY_TUNNEL, NTFY_REMOVE
+        Session.get_instance().add_observer(self.on_tunnel_remove, NTFY_TUNNEL, [NTFY_REMOVE])
+
+        self._private_key = self.my_member.private_key
+        self._public_key = self.my_member.public_key
+        self.persistence = MultiChainDB(self.dispersy, self.dispersy.working_directory)
+
+        # No response is expected yet.
+        self.expected_response = None
```
NielsZeilemaker: I'm guessing this one isn't needed anymore

pimotte:
Hi @NielsZeilemaker,

While I would probably do it the way you're suggesting if I were to write the whole of this from scratch, I'd like to understand the implications of switching before starting to re-engineer the entire thing.

The current crawler approach provides a certain flexibility in requesting specific (contiguous) blocks. Would using dispersy imply that we always need to request all the messages of a specific peer? 
At the moment we are mostly using crawling for data collection, but at some point we will want to move to a paradigm in which every peer checks a couple of blocks from their peers, in order to get distributed error checking. Do you think the overhead of this use would be manageable if we have a fullsyncdist community? I'm not sure if I get this correctly, but does using fullsyncdist imply that all messages are distributed to everyone, or is that mitigated by modifying _select_and_fix?

I'm also going to ask a question that neither of us are happy with: What specific advantages does switching the crawler to dispersy have, given that the code already exists as is? I get that it's logical looking at responsibilities and such, but if there is little reason to re-write it other than that, I would like to take that into consideration.

Captain-Coder:
Hi @NielsZeilemaker, thanks for having a look at the code! I agree that the crawler could be more fancy and make better use of what dispersy has to offer, however it is a quit'n'dirty way for us to get some data in. That is the main goal at this time.

From what I understand about dispersy it is not designed for mutable messages, but the trouble is that multichain blocks are mutable (the responder adds some data). So equating a block to a single dispersy message is not that easy. (Atleast it requires a mapping in between) From what I understand @snorberhuis already ran into this problem. I have been working on a refactor (["halves"](https://github.com/Captain-Coder/tribler/tree/halves/Tribler/community/multichain)) that takes the half-signed-as-default to the next level and makes the blocks (and thus dispersy messages) immutable.

On the slightly more technical side, we have two databases at the moment, dispersy and multichain (it has a separate one). So adding any kind of FK constraints would mean that the multichain tables need to be in the dispersy database (no cross database FK's in sqlite), thus making multichain a component of Dispersy. I am uncertain if this is the right place for the multichain community.

Tribler/community/multichain/community.py
pimveldhuisen: True, fixed with f0e9ecc

Tribler/Test/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore nosetest cover html output
+cover/
```
whirm: Better put it on the main .gitignore

Tribler/Test/test_multichain_conversion.py
```diff
@@ -0,0 +1,297 @@
+import logging
+from hashlib import sha1
+from struct import unpack
+
+from Tribler.Test.test_multichain_utilities import TestBlock, MultiChainTestCase
+from Tribler.community.multichain.conversion import MultiChainConversion, split_function, signature_format, \
```
whirm: We use parens for multiline imports

Tribler/community/multichain/.gitignore
```diff
@@ -0,0 +1 @@
+.Tribler-*
```
whirm: If your test is writing stuff here it's broken. Use the statedir path on the base test class (which is a temporary one that gets cleaned up after the test finishes)

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,429 @@
...
+            a. True, if we accept this message
+            b. False, if not (because of inconsistencies in the payload)
+            :param request The original message as send by this node
+            :param response The response message received
+            :param modified (bool) True if the message was modified
+        """
+        if not response:
+            self.logger.info("Timeout received for signature request.")
+            # Unpack the message from the cache object and store a half-signed record.
+        #    self.persist_signature_response(request.request.payload.message)
```
whirm: This needs to be removed or aligned properly

Tribler/Test/.gitignore
pimveldhuisen: Fixed with 0f188f3

Tribler/Test/test_multichain_conversion.py
pimveldhuisen: Fixed with 9a18fbf

Tribler/community/multichain/database.py
```diff
@@ -0,0 +1,376 @@
...
+        return (db_result[0], db_result[1]) if db_result[0] is not None and db_result[1] is not None \
+            else (-1, -1)
+
+    def open(self, initial_statements=True, prepare_visioning=True):
+        return super(MultiChainDB, self).open(initial_statements, prepare_visioning)
+
+    def close(self, commit=True):
+        return super(MultiChainDB, self).close(commit)
+
+    def check_database(self, database_version):
```
whirm: We've got all the DB format upgrade stuff in the same place (Tribler/Core/Upgrade). This stuff should be moved there.

Tribler/community/multichain/database.py
```diff
@@ -0,0 +1,376 @@
...
+
+        return LATEST_DB_VERSION
+
+
+class DatabaseBlock:
+    """ DataClass for a multichain block. """
+
+    def __init__(self, data):
+        """ Create a block from data """
+        """ Common part """
```
whirm: use comments

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -609,12 +630,23 @@ def remove_relay(self, circuit_id, additional_info='', destroy=False, got_destro
                 self.tunnel_logger.error("Could not remove relay %d %s", circuit_id, additional_info)
 
     def remove_exit_socket(self, circuit_id, additional_info='', destroy=False):
+
+        self.tunnel_logger.error("MULTICHAIN: Removing exit socket")
```
whirm: Update the log calls to have the proper loglevel

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -609,12 +630,23 @@ def remove_relay(self, circuit_id, additional_info='', destroy=False, got_destro
...
             if destroy:
                 self.destroy_exit_socket(circuit_id)
 
             # Close socket
             exit_socket = self.exit_sockets.pop(circuit_id)
+            if self.notifier:
+                peer = (exit_socket.sock_addr[0], exit_socket.sock_addr[1])
+                candidate = self.get_candidate(peer)
+                if candidate:
+                    from Tribler.Core.simpledefs import NTFY_TUNNEL, NTFY_REMOVE
```
whirm: How come this import is here?

pimveldhuisen: This style of 'just in time imports' was used elsewhere in the file, see for example line 946 of this file. (https://github.com/pimveldhuisen/tribler/blame/multichain/Tribler/community/tunnel/tunnel_community.py#L946) I can refactor all imports to move them to the top if you think that is a good idea?

NielsZeilemaker: This is done to allow the community to run without the Tribler code being present.

Tribler/community/tunnel/tunnel_community.py
pimveldhuisen: Fixed with 96dffa7

Tribler/community/multichain/database.py
pimveldhuisen: Fixed with 96dffa7

Tribler/community/multichain/database.py
pimveldhuisen: This function implements the abstract check_database function, from the dispersy Database class (dispersy/database.py), of which this class is a subclass. AFAIK this function therefore should be present here. I think I would be good to do actual database conversion stuff in (Tribler/Core/Upgrade) once we need that feature.

pimveldhuisen:
Retest this please

Tribler/community/tunnel/tunnel_community.py
whirm: Thx

Tribler/community/multichain/database.py
whirm: Ok, Have in mind that any stuff that goes into the _Tribler_ DB needs to be upgraded there.

whirm:
Looking at the coverage diff I see you aren't testing for Tribler to be able to start when the Multichain community is enabled?

Do you plan on increasing the coverage for this PR?

pimveldhuisen:
@whirm I added such a test. I ensured that it worked by starting Tribler myself, and running gumby experiments, but indeed a unit test would be better. Do you think this (41bce0d) is a proper way to do it? 

I don't plan on adding more tests, unless you have ideas on additional tests that would ensure more functionality.

whirm:
I would start Tribler itself (TestGUIAsServer)

whirm:
Also, add a newline at the end of your files

pimveldhuisen:
Retest this please

pimveldhuisen:
Ready to merge

whirm:
Have you guys looked at the coverage? There's some stuff that it probably would be a good idea to cover.

Tribler/Test/test_multichain_community.py
```diff
@@ -0,0 +1,514 @@
...
+
+    def test_get_next_total_no_block(self):
+        # Arrange
+        node, other, = self.create_nodes(2)
+        other.send_identity(node)
+        up_previous, down_previous = 5, 5
+
+        target_other = self._create_target(node, other)
+        node.call(node.community.publish_signature_request_message, target_other, up_previous, down_previous)
+        """ Create a block"""
```
whirm: Use normal comments everywhere

Tribler/Test/test_multichain_conversion.py
```diff
@@ -0,0 +1,297 @@
...
+        super(TestConversion, self).__init__(*args, **kwargs)
+        self.community = TestCommunity()
+
+    def test_encoding_decoding_signature(self):
+        """
+        Test if a responder can send a signature message.
+        This only contains requester and responder data.
+        """
+        # Arrange
+        converter = MultiChainConversion(self.community)
```
whirm: this stuf that is done in all the tests should be moved to setUp/tearDown

Tribler/Test/test_multichain_database.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def setUp(self, **kwargs):
+        super(TestDatabase, self).setUp()
+        path = os.path.join(self.getStateDir(), DATABASE_DIRECTORY)
+        if not os.path.exists(path):
+            os.makedirs(path)
+
+    def test_add_block(self):
+        # Arrange
+        db = MultiChainDB(None, self.getStateDir())
```
whirm: ditto

Tribler/Test/test_multichain_startup.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestMultichainStartup(TestAsServer):
+
+    """Start a Tribler session and ensure it does not crash."""
+
+    def __init__(self, *argv, **kwargs):
+        super(TestMultichainStartup, self).__init__(*argv, **kwargs)
+
+    def test_multichain_startup(self):
+        sleep(5)
```
whirm: ?

Tribler/Test/test_multichain_startup.py
```diff
@@ -0,0 +1,47 @@
...
+from Tribler.dispersy.util import blocking_call_on_reactor_thread
+
+from time import sleep
+
+
+class TestMultichainStartup(TestAsServer):
+
+    """Start a Tribler session and ensure it does not crash."""
+
+    def __init__(self, *argv, **kwargs):
```
whirm: This can be removed

Tribler/Test/test_multichain_startup.py
```diff
@@ -0,0 +1,47 @@
...
+            dispersy.define_auto_load(MultiChainCommunityCrawler, dispersy_member, (session, settings), load=True)
+        else:
+            dispersy.define_auto_load(MultiChainCommunity, dispersy_member, (session, settings), load=True)
+
+    def setUp(self):
+        super(TestMultichainStartup, self).setUp()
+        self.dispersy = self.session.get_dispersy_instance()
+        self.load_communities(self.session, self.dispersy, True)
+
+    def tearDown(self):
```
whirm: This can be removed

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,427 @@
...
+        Session.get_instance().add_observer(self.on_tunnel_remove, NTFY_TUNNEL, [NTFY_REMOVE])
+
+        self._private_key = self.my_member.private_key
+        self._public_key = self.my_member.public_key
+        self.persistence = MultiChainDB(self.dispersy, self.dispersy.working_directory)
+
+        # No response is expected yet.
+        self.expected_response = None
+
+    def initialize(self, a=None, b=None):
```
whirm: Is that actually needed?

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,427 @@
...
+                        .impl(authentication=(self.my_member,),
+                              distribution=(self.claim_global_time(),),
+                              destination=(candidate,),
+                              payload=block.to_payload()) for block in blocks]
+            self.dispersy.store_update_forward(messages, False, False, True)
+            if len(blocks) > 1:
+                # we sent more than 1 block. Send a resumption token so the other side knows it should continue crawling
+                # last_block = blocks[-1]
+                # resumption_number = last_block.sequence_number_requster if
+                # last_block.mid_requester == self._mid else last_block.sequence_number_responder
```
whirm: Are you keeping this commented out code there for a reason?

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,427 @@
...
+                message = self.get_meta_message(CRAWL_RESUME).impl(authentication=(self.my_member,),
+                                                                   distribution=(self.claim_global_time(),),
+                                                                   destination=(candidate,),
+                                                                   # payload=(resumption_number))
+                                                                   payload=())
+                self.dispersy.store_update_forward([message], False, False, True)
+        else:
+            # This is slightly worrying since the last block should always be returned.
+            # Or rather, the other side is requesting blocks starting from a point in the future.
+            self.logger.info("Crawler: No blocks")
```
whirm: If it's worrying would it make sense to raise the log level for this?

Tribler/Test/test_multichain_utilities.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def generate_block_payload(self):
+        return self.generate_requester() + self._generate_responder() + [self.public_key_requester,
+                                                                         self.signature_requester,
+                                                                         self.public_key_responder,
+                                                                         self.signature_responder]
+
+
+class MultiChainTestCase(AbstractServer):
+    def __init__(self, *args, **kwargs):
```
whirm: This one and the next can be removed too

Tribler/community/multichain/community.py
whirm: This expected_response attr doesn't seem to be used at all

Tribler/community/multichain/community.py
```diff
@@ -0,0 +1,427 @@
...
+                              "RelayRoute or TunnelExitSocket")
+            raise TypeError("Got a tunnel remove event for an object that is not a Circuit,"
+                            " RelayRoute or TunnelExitSocket")
+
+        if isinstance(bytes_up, int) and isinstance(bytes_down, int):
+            if bytes_up > MEGA_DIVIDER or bytes_down > MEGA_DIVIDER:
+                # Tie breaker to prevent both parties from requesting
+                if self._public_key > candidate.get_member().public_key:
+                    self.schedule_block(candidate, bytes_up, bytes_down)
+                # else:
```
whirm: same here

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,226 @@
...
+        :param data: ByteStream containing the message.
+        :return: (offset, SIGNATURE.impl)
+        """
+        if len(data) < offset + signature_size:
+            raise DropPacket("Unable to decode the payload")
+
+        values = unpack_from(signature_format, data, offset)
+        offset += signature_size
+
+        return \
```
whirm: why? :D

Tribler/community/multichain/conversion.py
```diff
@@ -0,0 +1,226 @@
...
+        :param data: ByteStream containing the message.
+        :return: (offset, CrawlResponse.impl)
+        """
+        if len(data) < offset + crawl_response_size:
+            raise DropPacket("Unable to decode the payload")
+
+        values = unpack_from(crawl_response_format, data, offset)
+        offset += crawl_response_size
+
+        return \
```
whirm: ditto

whirm:
I've found mostly cosmetic/style stuff.

Please convert all the string comments and remove commented out code that you don't really want there.

Tribler/Test/test_multichain_startup.py
pimveldhuisen: This file loads the multichain community with the Tribler core. It will only catch pretty obvious bugs, but this does actually increase coverage.

whirm: Please, wait for the community to be ready in a decent way. Start it, see that it's running and stop it. Avoid sleeps like the plage.

Tribler/community/multichain/community.py
pimveldhuisen: This was fixed by 7767760 I think, but it doesn't show because you commented on one line above the actual problem (?)

Tribler/community/multichain/community.py
pimveldhuisen: @Captain-Coder ?

Tribler/community/multichain/community.py
pimveldhuisen: @Captain-Coder ?

pimveldhuisen:
Retest this please

pimveldhuisen:
Regarding coverage, @Captain-Coder could you make a test for on_introduction_response(self, messages)? 
@whirm I would say the rest does not really need coverage?

Tribler/Test/test_multichain_community.py
```diff
@@ -0,0 +1,514 @@
...
+        crawler.give_message(block_response, node)
+
+        # Assert
+        self.assertTrue(self.assertBlocksInDatabase(node, 2))
+        self.assertTrue(self.assertBlocksInDatabase(crawler, 2))
+        self.assertTrue(self.assertBlocksAreEqual(node, crawler))
+
+
+    @blocking_call_on_reactor_thread
+    def assertBlocksInDatabase(self, node, amount):
```
whirm: Make all these assert functions really assert instead of returning a bool that has to be asserted elsewhere.

whirm:
I'd cover what you have left multichain/community.py and tribler_main.py

whirm:
Disable the multichain community in TestAsServer.setupPresession() so the tests are not sensitive to changes in the default settings.

Tribler/community/multichain/community.py
whirm: There's a comment saying that it's worrying that/when this happens, shouldn't it show on the logs in a higher level?

Tribler/community/multichain/community.py
whirm: The linked commit doesn't remove this commented out code and the code is still there. Do you really want this commented out stuff to stay?

whirm: The linked commit doesn't remove this commented out code and the code is still there. Do you really want this commented out stuff to stay?

pimveldhuisen: Ah, I see which code you mean now. I think the pseudo-code here clarifies the TODO. I can delete it if you want, but I would say it has a purpose here.

Tribler/community/multichain/community.py
Captain-Coder: There is some commented code (in more places that this), that allows resumption tokens to indicate where to resume. I'm still debating if this choice should be made by the crawler or the crawlee. For now it is fine to remove this code, I can look it up if i have to.

Tribler/community/multichain/community.py
Captain-Coder: The worrying part is why the crawler is requesting this from the crawlee, there is no reason for the crawlee (where this code runs) to worry about this. So I would say that for the expected use cases of this log message, Info is an appropriate level.

whirm:
@pimveldhuisen don't take all of the pylint warnings seriously yet. We have to tune it's settings (I'm not enforcing it ATM). Suggestions are welcome BTW.

pimveldhuisen:
Retest this please

whirm:
@pimveldhuisen have you looked at the test results before retesting? 
https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/754/testReport/

pimveldhuisen:
Retest this please

whirm:
@pimveldhuisen what failed on the previous run?

pimveldhuisen:
It timed out, I'm not sure why. Might have to do that I pushed twice just after each other, which triggered two builds running simultaneous?

devos50:
@pimveldhuisen @whirm 

`test_download_torrent_from_url (Tribler.Test.API.test_download.TestDownload) ... `

This test keeps hanging sometimes. I'm investigating right now, only managed to reproduce it once on the debian container but after placing print statements to see where it hangs, it does not trigger anymore.

whirm:
@pimveldhuisen two builds simultaneously should be fine. @devos50 thx!

devos50:
Nice to see that it's merged, good work! 👍 

@whirm 6.6 release very soon now?

whirm:
nope, we need to fix the issues on the 6.5.x milestone so we can release 6.5.2 to free up `next` for 6.6 stabilization.

synctext:
nice work team!

lfdversluis:
Grats on the merge!

PR #1865, opened 2016-01-12
Update Tribler development on Windows.md

lfdversluis:
- Removed build tools as they do not install required distributions
- Added dependency walker explanation in case something is not installed by Visual Studio (if someone deselected something)


whirm:
Please, add the link to the dependencies instead of telling people to use google.

lfdversluis:
@whirm That varies what the user is missing, in our case we missed two of them. If more are required in the future, we continuously have to edit this file. The point is: MVS2015 should have all of them shipped.

synctext:
you oncountered two cases of missing DLLs, however only one is listed in this commit. As Elric indicated, list them and approach to fix/download location.

whirm:
@lfdversluis exactly, every time the dependencies change, we need to update the related documentation as we have done until now :)

lfdversluis:
@whirm ready to merge if you agree

whirm:
Squash it and I'll merge

lfdversluis:
@whirm This can be merged btw

Issue #1862, opened 2016-01-12
Clean up the mac license file so the OS X .dmg opens faster


devos50:
The license file seems to be too long which causes the .dmg file to take longer to load (about 3 seconds). I've already reduced the amount of loading time by using an external script to inject the EULA in the .dmg file.

The file loads even faster if we make the license agreement shorter. @synctext can you take a look at this? The file I'm using right now is https://github.com/devos50/tribler/blob/builder_osx_64bit/Tribler/Main/Build/Mac/mac_license.txt.

synctext:
if it improves speed use something like this 3-line version:

This software is released under the LGPL license by Delft University of Technology. Included third party libraries and software modules may have different licences. For details see the included file tribler/licence.txt or http://github.com/tribler/tribler.. etc.

PR #1867, opened 2016-01-13
Changed the way the old builder adds the license to the .dmg file

devos50:
We're now adding the EULA to the .dmg file using the `eulagise` script. I've installed this script on the old OS X builder. I've tested this script using the new Mac Mini. I'm pull requesting this against next so we can use this new EULA for the 6.5 release.

### DIFF
--- a/mac/makedistmac.sh
+++ b/mac/makedistmac.sh
@@ -121,10 +121,7 @@ hdiutil convert dist/temp/rw.dmg -format UDZO -imagekey zlib-level=9 -o dist/$AP
 rm -f dist/temp/rw.dmg
 
 # add EULA
-hdiutil unflatten dist/$APPNAME.dmg
-DeRez -useDF $LIBRARYNAME/Main/Build/Mac/SLAResources.rsrc > dist/temp/sla.r
-Rez -a dist/temp/sla.r -o dist/$APPNAME.dmg
-hdiutil flatten dist/$APPNAME.dmg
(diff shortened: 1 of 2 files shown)


mac/makedistmac.sh
```diff
@@ -121,10 +121,7 @@ hdiutil convert dist/temp/rw.dmg -format UDZO -imagekey zlib-level=9 -o dist/$AP
 rm -f dist/temp/rw.dmg
 
 # add EULA
-hdiutil unflatten dist/$APPNAME.dmg
-DeRez -useDF $LIBRARYNAME/Main/Build/Mac/SLAResources.rsrc > dist/temp/sla.r
-Rez -a dist/temp/sla.r -o dist/$APPNAME.dmg
-hdiutil flatten dist/$APPNAME.dmg
+eulagise --license $LIBRARYNAME/Main/Build/license.txt --target dist/$APPNAME.dmg
```
whirm: Could you update the mac documentation 
to add this dependency?

devos50: This is only relevant for the builder. Is it necessary to write about this in the mac documentation?

whirm: Yes, we need to have up to date documentation on how to set the builders for win/osx/linux

devos50: I will create a separate document for that (see #1870).

whirm:
Ok, squash it and I'll merge if Johan is OK with the contents of license.txt

devos50:
The contents was created by Johan himself, see #1862.

devos50:
@whirm squashed.

PR #1868, opened 2016-01-13
Removed packages that are no longer used by Tribler

devos50:


whirm:
Update `debian/control` and check that both the windows and osx build scripts won't break if this is missing.

devos50:
@whirm the OS X build script was not using the `gmpy` library as dependency. The Windows build also succeeds.

whirm:
@devos50 you missed `Tribler/Main/Build/Win32/setuptribler.py`

devos50:
That one is removed by the Windows builder PR of @lfdversluis, shall I also remove it in this PR?

whirm:
Yeah, let's do all of it at once.

devos50:
done.

lfdversluis:
Looks like I'm going to get a conflict :-1:

synctext:
You removed 1 package: python-gmpy
I vaguely recall that this was an optional package which dramatically improved the performance of crypto operations.

devos50:
@synctext it is not imported anywhere (also not in Dispersy) so this dependency is not used.

I also recall that It was used to speed up arithmetic operations but apparently it got removed somewhere in the development process. Not sure about the reason why though.

synctext:
OK, thnx for quickly checking.

whirm:
I've checked and no dependency of Tribler recommends this either. If I remember correctly that was used in @NielsZeilemaker's custom-made crypto stuff that was later replaced with python-cryptography

PR #1903, opened 2016-01-28
Fix double starting reactor

Captain-Coder:
In some scenarios the gumby scenario runner already has the reactor running. When a dispersy session gets started through the full tribler core, it would try to start the reactor again, which causes an exception and a stack trace.

### DIFF
--- a/Tribler/Core/Utilities/twisted_thread.py
+++ b/Tribler/Core/Utilities/twisted_thread.py
@@ -52,7 +52,8 @@ def threaded_reactor():
 
         def _reactor_runner():
             reactor.suggestThreadPoolSize(1)
-            reactor.run(installSignalHandlers=False)
+            if not reactor.running:
+                reactor.run(installSignalHandlers=False)
 
         _twisted_thread = Thread(target=_reactor_runner, name="Twisted")
         _twisted_thread.setDaemon(True)


whirm:
This probably means the reactor is imported too late, could you show me where the error pops up?

Captain-Coder:
If you run the gumby multichain experiment multichain_integrated_single.conf (Beware of setting the proper scenario file in the environment variable). Without this patch it results in the following stacktrace:
```
ERR: Exception in thread Twisted:
ERR: Traceback (most recent call last):
ERR:   File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
ERR:     self.run()
ERR:   File "/usr/lib/python2.7/threading.py", line 754, in run
ERR:     self.__target(*self.__args, **self.__kwargs)
ERR:   File "/home/ewoutb/gumby_workspace/tribler/Tribler/Core/Utilities/twisted_thread.py", line 55, in _reactor_runner
ERR:     reactor.run(installSignalHandlers=False)
ERR:   File "/usr/lib/python2.7/site-packages/twisted/internet/base.py", line 1193, in run
ERR:     self.startRunning(installSignalHandlers=installSignalHandlers)
ERR:   File "/usr/lib/python2.7/site-packages/twisted/internet/base.py", line 1173, in startRunning
ERR:     ReactorBase.startRunning(self)
ERR:   File "/usr/lib/python2.7/site-packages/twisted/internet/base.py", line 682, in startRunning
ERR:     raise error.ReactorAlreadyRunning()
ERR: ReactorAlreadyRunning
```

whirm:
@Captain-Coder this means twisted_thread.reactor is being imported too late. It needs to be imported before anything else that is twisted related.

whirm:
@Captain-Coder closing this as it will never get merged (It would hide the case where the non-threaded reactor gets installed first, and we don't want that).

PR #1917, opened 2016-02-03
Add Arch Linux dependencies

Anth0rx:
I just managed to find out which packages are needed for Arch Linux (they have slightly different package names) and thought it would be nice if they could be found in here.


whirm:
Perfect, thanks!

PR #1921, opened 2016-02-06
Changed a print to log

lfdversluis:


lfdversluis:
Lucky test run :D

synctext:
add to whitelist

whirm:
@lfdversluis: We want this to be printed to stderr, otherwise we will never see it, as the tests will time out because the thread is stuck and nose will never get the chance to flush the log buffers.

whirm:
@synctext ignore the bot, it's the WIP QA job that keeps asking this for PRs to branches the job is not assigned to for some reason.

PR #1923, opened 2016-02-07
Filesize widget fix

devos50:
On OS X, the filesize widget did not appear. This seems to be due to a bug in wx (see http://stackoverflow.com/questions/35201245/size-of-panel-in-panel-using-wxpython-incorrect/). So now, I pass the size in the call to `super`. Moreover, the return value of the `GetClientRect` method always returns a (0, 0, 0, 0) rect so I removed this call. The y-position of the up/down arrow images are now dependent on the `self.slider_size` attribute.


whirm:
Great, thanks!

PR #1925, opened 2016-02-08
Added conditional statement for Cygwin environment

### LINKED ISSUE #1906: Update tribler.sh so Tribler can run from a Cygwin environment
Currently, when running `tribler.sh` in a Cygwin environment, it will just terminate. We can easily change it so it will be able to run in Cygwin.

devos50:
Fixes #1906

### DIFF
--- a/tribler.sh
+++ b/tribler.sh
@@ -27,7 +27,8 @@ if [ "$UNAME" = "Linux" ]; then
     }
 
     python2.7 $TRIBLER_SCRIPT
-
+elif [ ! -z `uname -s | grep CYGWIN_NT` ]; then
+    python $TRIBLER_SCRIPT
 else
     if [ "$UNAME" = "Darwin" ]; then
 


tribler.sh
```diff
@@ -27,7 +27,8 @@ if [ "$UNAME" = "Linux" ]; then
     }
 
     python2.7 $TRIBLER_SCRIPT
-
+elif [[ "$UNAME" = CYGWIN_NT* ]]; then
+    python $TRIBLER_SCRIPT
```
whirm: That's not working for me.

I tested with:
```
[ HELLO = HEL* ] && echo "lel"
[ HELLO = "HEL*" ] && echo "lel"
[[ HELLO = "HEL*" ]] && echo "lel"
```
I would go for something like: ```[ ! -z `uname -s|grep ^Linux` ] && echo lel```

devos50: @whirm in the 3rd example you're checking against the string "HEL*" I think. Removing the quotes around `HEL*` works for me.

devos50: However, that double square bracket thingy does not work with all versions of bash, I will rewrite it to use `grep` which is more universal.

whirm: None of them work for me, I just was trying random stuff, never seen that syntax before either.

devos50:
@whirm I've rewritten the condition to use `grep`.

tribler.sh
```diff
@@ -27,7 +27,8 @@ if [ "$UNAME" = "Linux" ]; then
     }
 
     python2.7 $TRIBLER_SCRIPT
-
+elif [ ! -z `uname -s | grep CYGWIN_NT*` ]; then
```
whirm: Remove the *, that means any amount of T's at the end (or will expand to a file name if there's any matching that name on $PWD.

devos50:
@whirm hmm to much used to use `*` as wildcard. I removed the `*`.

PR #1930, opened 2016-02-09
Fixes #1928

### LINKED ISSUE #1928: Tests are defining torrent piece length not as a power of two
When running the tests on OS X using libtorrent 1.0.8 (with assertions enabled and debug symbols), I get an assertion: https://jenkins.tribler.org/job/Test_macmini_unittests/68/artifact/output/test_runners_output/1.err/*view*/. When running the `test_vod.py` test suite, the problem is that the piece length is not defined as a power of two but instead, has a value of 10.

According to the BitTorrent specifications, the piece size of a torrent is usually a power of two (see https://wiki.theory.org/BitTorrentSpecification - ctrl+f on piece size). Even `libtorrent` has an assertion built-in to test for a power-of-two block size: https://github.com/arvidn/libtorrent/blob/master/src/torrent.cpp#L125.

While these tests will probably pass when I recompile libtorrent without assertions enabled, I feel we should fix this by changing the piece size to a power of two (for instance, 16). Any suggestions on this one?

devos50:


whirm:
retest this please

whirm:
reetest this please

PR #1937, opened 2016-02-10
Changed height of expanding panel below button

devos50:
We noticed an assertion during the test runs where an assertion came up after starting to stream a video file. This was due to the panel below the video player button in the left menu which was too high. The height of this panel caused the popup that shows up when adding a torrent to be squashed, resulting in an assertion.


devos50:
retest this please

whirm:
@devos50 could you explain the reason of the change on the commit message? Otherwise we have no record of this modification being done to fix a bug.

PR #1956, opened 2016-02-18
LibtorrentMgr and channel_rss import the correct reactor

Captain-Coder:
The LibtorrentMgr.py imported the incorrect reactor. This caused exceptions in gumby experiments that use the full tribler core as dispersy source. I searched for other uses of the twisted_thread, and found and fixed another one that is probably wrong.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -1,6 +1,5 @@
 # Written by Egbert Bouman
 import binascii
-import os
 import logging
 import tempfile
 import threading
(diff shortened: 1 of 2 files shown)


whirm:
Please, keep the pep8 import ordering: https://www.python.org/dev/peps/pep-0008/#imports

Captain-Coder:
You are refering to placing the "general" 3rd party libs before local lib imports right? See if this is better.

whirm:
```
 Imports should be grouped in the following order:

    standard library imports
    related third party imports
    local application/library specific imports

You should put a blank line between each group of imports. 
```

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -9,10 +9,10 @@
 from binascii import hexlify
 from copy import deepcopy
 from shutil import rmtree
+from twisted.internet import reactor
```
whirm: this should go one line down

Also, any decent code editor could fix that for you without even thinking :)

Captain-Coder:
I removed the redundant os import while I was at it :P

whirm:
add to whitelist

whirm:
retest this please

PR #1961, opened 2016-02-19
WIP: Created first endpoint of the Tribler API

devos50:
This is a first attempt to create the Tribler API that eventually will be a big part of libtribler.

The `TriblerAPI` class is the root resource object that receives all requests. These requests are internally dispatched to child resources, depending on the url. `EventRequestHandler` is such a child resource and handles and returns all important events in libtribler.

Checking the disk space was originally done in `tribler_main` (by the `guiservthread_free_space_check` method). Now, this check is performed by the `FreeSpaceChecker` class which schedules a `LoopingCall`. When the free disk space has been determined, this free space is broadcasted using the notifier in a Tribler session. The API observes this event and when the free disk space has been determined, it is sent to the client that is connected to the HTTP server.


Tribler/Core/Modules/freespace_checker.py
```diff
@@ -0,0 +1,22 @@
...
+
+class FreeSpaceChecker(TaskManager):
+
+    def __init__(self, session):
+        super(FreeSpaceChecker, self).__init__()
+
+        self.session = session
+        self.check_path = DefaultDownloadStartupConfig.getInstance().get_dest_dir()
+
+        self.free_space_check_task = self.register_task("check free disk space ", LoopingCall(self.check_free_space))
```
whirm: I would have a start() and stop() method on all these components so that instancing them doesn't mean they will be active immediately.

devos50: Good point, will add them in a moment.

Dmole:
Nice start, I look forward to a headless version.

devos50:
retest this please

Tribler/Core/triblerapi.py
```diff
@@ -0,0 +1,309 @@
...
+from twisted.web import resource
+from Tribler.Core.CacheDB.db_objects import Channel
+from Tribler.Core.DownloadState import DownloadState
+from Tribler.Core.simpledefs import NTFY_FREE_SPACE, NTFY_INSERT, NTFY_CHANNELCAST, DOWNLOAD, UPLOAD, NTFY_TORRENTS
+
+
+class TriblerAPI(resource.Resource):
+
+    '''
+    This class implements an HTTP API that can be used by external processes to control the Tribler Core.
```
lfdversluis: a* http api

whirm:
retest this please

whirm:
^ testing a fix for the pylint check

devos50:
I'm first going to focus on the GUI side and create a fake endpoint (for testing purposes). After that, I will continue to work on the API that is shipped with the Tribler core so I will open a new PR then.

PR #1966, opened 2016-02-20
Refactored str2unicode method

devos50:
After having a discussion on IRC (#python channel), I got several suggestions about refactoring the `str2unicode ` method so I made the following changes:
1) Catching a more specific exception (`UnicodeDecodeError`)
2) Using True/False values instead of 0/1
3) Removed  `unicode-escape` encoding from the list since `iso-8859-1` will be able to decode any byte string
4) Due to point 3, the `unicode(s, sys.getdefaultencoding(), errors='replace')` statement is not needed anymore.

@whirm we might use Chardet (https://pypi.python.org/pypi/chardet) to do the encoding/decoding of strings?

### DIFF
--- a/Tribler/Core/Utilities/unicode.py
+++ b/Tribler/Core/Utilities/unicode.py
@@ -30,22 +30,14 @@ def bin2unicode(bin, possible_encoding='utf_8'):
 
 def str2unicode(s):
     try:
-        s = unicode(s)
-    except:
-        flag = 0
-        for encoding in [sys.getfilesystemencoding(), 'utf_8', 'iso-8859-1', 'unicode-escape']:
+        return unicode(s)
+    except UnicodeDecodeError:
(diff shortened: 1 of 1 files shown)


devos50:
retest this please

devos50:
retest this please

Tribler/Core/Utilities/unicode.py
```diff
@@ -31,20 +31,17 @@ def bin2unicode(bin, possible_encoding='utf_8'):
 def str2unicode(s):
     try:
         s = unicode(s)
```
whirm: why don't you just replace all `s = ...`  with  `return ...` and remove `flag`?

devos50: @whirm done!

whirm:
retest this please

PR #1969, opened 2016-02-21
Removed check for length in info metadata

devos50:
These lines will never trigger.


lfdversluis:
These were the lines that I found the other week :p

PR #1973, opened 2016-02-22
Fixed parameter type passed to _test_port

devos50:
Currently when the socket type is hard-coded (for instance, `tcp` or `udp`), a string is being passed to the `_test_port` method, triggering the assertion which expects something else. Instead, the `_sock_type` parameter should be passed.

### DIFF
--- a/Tribler/Core/Utilities/network_utils.py
+++ b/Tribler/Core/Utilities/network_utils.py
@@ -60,7 +60,7 @@ def check_random_port(port, socket_type="all"):
         if _test_port(_family, socket.SOCK_DGRAM, port):
             is_port_working = _test_port(_family, socket.SOCK_STREAM, port)
     else:
-        is_port_working = _test_port(_family, socket_type, port)
+        is_port_working = _test_port(_family, _sock_type, port)
 
     return is_port_working
 


lfdversluis:
@devos50 A test should be added to make sure this bug doesn't trigger again.

devos50:
@lfdversluis I have one in my unit tests PR, that's how I discovered this one :+1:

lfdversluis:
@devos50 awesome!

whirm:
retest this please

whirm:
BTW, if you have tests for one fix, just ship them on the same PR.

devos50:
@whirm I have tests for this fix in my unit tests branch.

whirm:
@devos50 I got that, but you can cherry pick that test to this branch, the next time you rebase against devel, the duplicated commit will be gone automatically, so don't worry about that.

devos50:
@whirm ah ok, then we have both the fix and the test in one PR which is a bit more structured (actually, having tests for each fix should be a standard rule as @lfdversluis already suggested).

whirm:
That would be great, more so if the line is not already covered already.

As soon as I'm confident with the new test runner I'll start enforcing the second case

PR #1977, opened 2016-02-22
Add issue template

whirm:


devos50:
This is going to be really helpful :+1:

whirm:
Apparently the pylint comparator gets confused if but it gets confused if the PR branch is not rebased on top of the target branch.

PR #1978, opened 2016-02-23
Various GUI improvements on OS X

devos50:
I made various changes to improve the GUI on OS X:
-some tooltips are now hidden on OS X (does not work when the tooltip is refreshed)
-the artwork panel on the homepage is now hidden
-on OS X, all `wx.SearchCtrl` instances are replaced by `wx.TextCtrl` (there is a bug in wxWidgets 3.0 on OS X when using a SearchCtrl)
-the elements on the home page have been changed a bit
-the stream button when hovering over a torrent has been hidden (since streaming does not work on OS X)


whirm:
Could you add TODO entries on all the hacks you add so we can find it quickly in case we have to fix it propery for some reason?

devos50:
@whirm I didn't use real hacks yet, I only disabled some things on OS X. Want me to add TODOs to describe what I disabled and why?

whirm:
Just the ones that cripple the gui in any way (like disabling the tooltips)

devos50:
@whirm I added TODO entries where necessary.

devos50:
retest this please

devos50:
@whirm I think this one is ready?

devos50:
It seems that there is no method on `wx.Tooltip` that indicates whether a specific tooltip is shown currently: http://docs.wxwidgets.org/stable/classwx_tool_tip.html

PR #1980, opened 2016-02-24
Removed unused info in __init__ of the Tribler core

devos50:
It seems that these declarations are not used anywhere in the project and I don't see any use for them. Please correct me if I'm wrong if they actually have a purpose.


devos50:
retest this please

whirm:
That's from when we where using the python implementation of bittorrent

PR #1990, opened 2016-02-26
Updated deprecated calls in libtorrent

### LINKED ISSUE #1478: [reminder] fix uPnP when libtorrent python binding has those functions exposed
According to libtorrent's documentation, we need to use `add_port_mapping()` in a libtorrent session to add uPnP port mapping. But this `add_port_mapping()` is not exposed in the libtorrent session yet on python side. So we need to fix this when it gets exposed.

devos50:
In order to increase the stability of Tribler and the tests, the deprecated calls to libtorrent in Tribler (and there are many) should be replaced by newer calls.

To find these deprecated methods, I compiled libtorrent without deprecated methods and ran Tribler and the tests to see where it crashes.

I also added the uPnP port mapping call since this is exposed in later versions of libtorrent. This fixes #1478.


devos50:
During the rewriting and investigation of the libtorrent API, I found out that the `save_path` in the `torrent_handle` object has been deprecated. I asked on `#libtorrent` what the preferred method is to get this kind of information. This is `handle.status().save_path`, however, this attribute was not exposed by the Python bindings. Arvid created a PR for this field: https://github.com/arvidn/libtorrent/pull/503.

For now, I will compile libtorrent on OS X with this patch. The PR will be included in libtorrent 1.0.9. I will also make a conditional check to see whether this attribute is available in the `torrent_status` object. If so, this one will be used. This means that we will remain backward compatible with older versions of libtorrent.

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
@whirm I'm finished with this PR. More tests are passing on OS X currently. Can you review it please?

whirm:
Right away!

BTW, I think I've fixed the Coverage job to show the right diff. I guess we can start enforcing it once the OSX tester is linked to the PR job.

whirm:
re exposing new API stuff to the python bindings. Maybe you could talk yo arvid to get the DHT API exposed? that would mean having a better/maintained DHT implementation and allow us to drop the pymdht submodule.

whirm:
I'll forward you to the right Issue.

whirm:
@devos50 what's the oldest version of libtorrent you have tested this with?

devos50:
@whirm exposing the DHT api is something I talked to with @synctext. Would be a very nice addition and should not be that hard to do, but we should discuss this with Arvid.

I believe the oldest version I tested with is 1.0.7 on Debian.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -628,6 +630,11 @@ def move_storage(self, new_dir):
     @checkHandleAndSynchronize()
     def get_save_path(self):
         if not isinstance(self.tdef, TorrentDefNoMetainfo):
+            # torrent_handle.save_path() is deprecated in newer versions of Libtorrent. We should use
+            # self.handle.status().save_path to query the save path of a torrent. However, this attribute
+            # is only included in libtorrent 1.0.9+
+            if hasattr(self.handle.status(), 'save_path'):
```
whirm: it's nicer to obtain the status object and use it twice

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -78,11 +77,6 @@ def shutdown(self):
 
         self.get_session().stop_upnp()
 
-        # Save DHT state
```
whirm: I assume the dht state is stored in the resume_data now?

whirm:
Do you have an LTS Ubuntu around to see if the shipped libtorrent there will still work with these changes?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
devos50: What do you mean with that exactly?

Tribler/Core/Libtorrent/LibtorrentMgr.py
devos50: Hmm I'm not sure where the DHT state is saved now, I only know that it has been deprecated. Let me have a look in the documentation

devos50:
@whirm I don't have an Ubuntu installation (anymore). I can compile an older version of libtorrent on OS X if you wish to test it?

whirm:
I guess it should still work with Vivid? (15.04) so that would be libtorrent 0.16.18: 

http://packages.ubuntu.com/search?keywords=python-libtorrent&searchon=names&suite=all&section=all

@synctext any saying on this?

whirm:
@devos50 also, update `debian/control` with the minimum version required.

lfdversluis:
While you are at it: https://github.com/Tribler/tribler/blob/bd3a972a088bbe51f67395edca92f2692b4e5353/Tribler/Core/APIImplementation/LaunchManyCore.py#L232 ?

whirm:
@lfdversluis good one!

devos50:
@lfdversluis nice, I will take a look at it!

Libtorrent 0.16.18 is failing unfortunately. The problem is that the `set_settings` method on a `session` expects a `session_settings` object while that one is deprecated. We can place an if-statement to check this but this can become messy since `set_settings` is used all over the place. Any suggestions?

whirm:
@devos50 a quick grep gives only 4 results?

```
LibtorrentDownloadImpl.py:488:                self.ltmgr.get_session().set_settings(settings)
LibtorrentDownloadImpl.py:497:                self.ltmgr.get_session().set_settings(settings)
LibtorrentMgr.py:123:        ltsession.set_settings(settings)
LibtorrentMgr.py:198:            ltsession.set_settings(settings)
```

whirm:
vivid seems to be the first ubuntu that packs python-wxgtk3.0, so it would be really nice to support it.

whirm:
retest this please

whirm:
^ that's for some modifications I did to the test runner, not related to this PR

whirm:
retest this please

whirm:
Ok, the css is fixed on the diffs

Tribler/Core/Libtorrent/LibtorrentMgr.py
devos50: @whirm the DHT state is part of the libtorrent state since 0.15. I've changed it in commit 267d66a4ada606d6443bf08f13f538d663a9df58.

devos50:
@whirm I've updated this PR and tested it with libtorrent 0.16.18. Downloading is working on this version and also on newer versions of libtorrent.

whirm:
Great!

whirm:
@devos50 does the coverage diff make sense to you?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
whirm: status=self.handle.status()
...

devos50:
retest this please

devos50:
@whirm I fixed the uPnP mapping call. I made it conditional so I first check whether this method exists in a libtorrent session. If so, the method is called.

devos50:
@lfdversluis thanks!

whirm:
@devos50 great!

whirm:
Don't forget to update `debian/control` with the latest supported libtorrent version

devos50:
@whirm done and done

devos50:
retest this please

ardhipoetra:
Sorry for bumping merged PR.

In [libtorrent documentation](http://www.libtorrent.org/reference-Core.html#save_resume_data%28%29), we have to consider alert of `save_resume_data` to actually get the resume data. I don't see that implemented in this PR.

synctext:
can you explain a bit more?

ardhipoetra:
In previous (deprecated) call, `write_resume_data` is synchronously (blocked call) return the so-called resume data. 

In new call, `save_resume_data` is returning immediately (in my case : `None`). After resume data is ready, it will put an alert `save_resume_data_alert` (or `save_resume_data_alert_failed` if failed) which contains dictionary-format of resume data.

devos50:
@ardhipoetra can you create a separate issue for this?

PR #1999, opened 2016-03-03
WIP: Small benchmark code and making the first function with inline callbacks in allchannel community

lfdversluis:


lfdversluis:
@whirm [This function](https://github.com/Tribler/tribler/blob/devel/Tribler/community/allchannel/community.py#L643) has the `@property` decorator, but makes use of the `_cacheTorrents` function I just decorated with inlinecallbacks. I think mixing @property and @inlineCallback is a bad idea so I propose removed the property decorator and make it a normal function where args are passed to. 

Is that ok with you?

lfdversluis:
@whirm Also, is the the most simplest benchmark you were referring to? Without the @inlineCallbacks it took 110 ms and now about 35ms (even with yielding in that loop)

whirm:
Ok, but name it getCachedTorrents or something similar

whirm:
use a bigger number than 5 to get a better idea of the performance difference

lfdversluis:
@whirm Why are the tests passing on Linux? I thought you said that it was checked that no exceptions occured and if there are, it would let the build fail? The logs are being spammed (as expected) with errors now.

lfdversluis:
retest this please

lfdversluis:
This work will be replaced with StormDBManager in the near future.

PR #2006, opened 2016-03-07
Fixed bug in adding nodes to torrent and added test for it

devos50:


devos50:
@whirm all the unfixed coverage changes are stacked in the next PR if we don't solve them apparently.

Tribler/Test/Core/test_torrent_utils.py
```diff
@@ -23,6 +23,12 @@ def test_create_torrent_one_file(self):
     def test_create_torrent_one_file_2(self):
         create_torrent_file([os.path.join(self.TORRENT_DATA_DIR, self.FILE2_NAME)], {})
 
+    def test_create_torrent_with_nodes(self):
+        params = self.get_params()
+        params["nodes"] = [("127.0.0.1", 1234)]
+        create_torrent_file([os.path.join(self.TORRENT_DATA_DIR, self.FILE1_NAME)],
+                            params, self.created_torrent_one_file)
```
whirm: Shouldn't you be checking that the file gets actually created etc?

devos50: This is checked in the callback `self.created_torrent_one_file`.

PR #2007, opened 2016-03-07
Written some more unit tests for the configuration files

devos50:
Tribler is using various kinds of configuration files. I've written some basic unit tests for the general configuration file parser, the session configuration file and the download configuration file. Each of the tested files now has a coverage of > 90%.

### DIFF
--- a/Tribler/Test/Core/test_downloadconfig.py
+++ b/Tribler/Test/Core/test_downloadconfig.py
@@ -1,0 +1,84 @@
+import os
+from tempfile import mkdtemp
+
+import shutil
+from nose.tools import raises
+
+from Tribler.Core.DownloadConfig import DownloadConfigInterface, DownloadStartupConfig, get_default_dest_dir, \
+    get_default_dscfg_filename
+from Tribler.Core.Utilities.configparser import CallbackConfigParser
+from Tribler.Core.simpledefs import DLMODE_VOD, UPLOAD, DOWNLOAD
+from Tribler.Test.Core.base_test import TriblerCoreTest
+
+
+class TestConfigParser(TriblerCoreTest):
+
+    FILE_DIR = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
+    CONFIG_FILES_DIR = os.path.abspath(os.path.join(FILE_DIR, u"data/config_files/"))
+
+    def test_downloadconfig(self):
+        dlconf = CallbackConfigParser()
+        dlconf.add_section('downloadconfig')
+        dlconf.set('downloadconfig', 'hops', 5)
+        dlcfg = DownloadConfigInterface(dlconf)
+
+        temp_dir = mkdtemp(suffix="_tribler_test_session")
+        self.assertIsInstance(dlcfg.get_dest_dir(), unicode)
+        dlcfg.set_dest_dir(temp_dir)
+        self.assertEqual(dlcfg.get_dest_dir(), temp_dir)
+
+        dlcfg.set_corrected_filename("foobar")
+        self.assertEqual(dlcfg.get_corrected_filename(), "foobar")
+
+        dlcfg.set_mode(1)
+        self.assertEqual(dlcfg.get_mode(), 1)
+
+        dlcfg.set_hops(4)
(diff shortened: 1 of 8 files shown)


whirm:
Cool, next step on this would be to converge into a single config sytem, I like ConfigObj a lot (we use it on Gumby)

devos50:
retest this please

devos50:
retest this please

whirm:
got stuck at test_download_torrent_from_file (Tribler.Test.API.test_download.TestDownload) ...

devos50:
@whirm was about to test the health of the torrent ;)

whirm:
It shouldn't get stuck nonetheless, it should just fail if the test doesn't succeed in a predetermined amount of time.

devos50:
retest this please

lfdversluis:
@devos50 if you would need a way to time it out, let the function return a deferred. nosetests has a method to wait for a deferred to resolve and you can specify a timeout, see more here http://nose.readthedocs.org/en/latest/api/twistedtools.html

Tribler/Test/Core/test_downloadconfig.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(dlcfg_copy.get_hops(), 0)
+
+    def test_downloadstartupconfig_copy(self):
+        dlcfg = DownloadStartupConfig()
+        dlcfg_copy = dlcfg.copy()
+
+        self.assertEqual(dlcfg_copy.get_hops(), 0)
+
+    def test_startup_download_save_load(self):
+        temp_dir = mkdtemp(suffix="_tribler_test_session")
```
whirm: Avoid creating temporary files in the tests, if they fail they will never get cleaned up. Instead create new dirs inside of `self.session_base_dir` which is already created and cleaned up by the base classes even if the test fails.

devos50:
Timeout is a good solution but I'm thinking about making `test_download_torrent_from_file` (and the other ones) a little different. Instead of checking whether the torrent starts downloading (i.e. libtorrent receives incoming bytes), maybe we should only check whether the torrent is correctly added to libtorrent instead of waiting for incoming data. This way, we are not dependent on the health of the torrent. @lfdversluis @whirm any suggestions on this one?

whirm:
we still need to get data on the streaming test

Tribler/Test/Core/test_downloadconfig.py
devos50: Using `self.session_base_dir` assumes that I'm starting a session. Maybe it's easier to just clean up the temp_dir in the `tearDown` method in `TriblerCoreTest`?

devos50:
With the streaming test, do you mean the test that starts the video player and starts playing night of the living dead?

whirm:
yes

whirm:
I've got a seeder for that one, so it should be fine, but in no case it should be stuck, even if for some reason the download doesn't work.

devos50:
retest this please

whirm:
I see now that your tests aren't using the base test class the rest are using. So all the checks done during teardown aren't done on yours?

devos50:
No, these tests are meant to be way more simple than the tests that are already there.

whirm:
what happens if any of your tests leaves a dirty reactor?

devos50:
I just made a very simple test by invoking `callLater` and immediately returning. It seems that nothing happens (no error is being thrown or something similar).

whirm:
so that's bad :) That's why we have all these checks, you need to be _really_ careful of not leaving anything behind when finishing a test or when it fails you will be N tests further away and will bring madness upon whoever tries to debug that stuff.

It's OK if you need to split the base class into two so it doesn't do more than it should in your tests, but let's keep these checks across the board even if it's overkill for many of these tests. Better safe than crazy.

devos50:
Ok, I will refactor all the tests so they have these checks included (let them inherit from `AbstractServer`).

whirm:
Feel free to add any extra checks you deem necessary (such as that the reactor hasn't been stopped :))

whirm:
@devos50 should I merge this one? I guess you will send a new PR for the refactor.

devos50:
@whirm yes please

PR #2008, opened 2016-03-07
Test improvements

devos50:
I made two improvements to the test suite:
-changed the keyword for the channel search to `de`. This is way more common than `tv` (I checked this in my own local Tribler database - `tv` has 25 hits where `de` has 300 hits). I've performed this test several times on my machine and I didn't have a single failure.
-the test that starts a download using the `start_download_from_uri` method in session now only checks whether a download has started.

### DIFF
--- a/Tribler/Test/API/test_download.py
+++ b/Tribler/Test/API/test_download.py
@@ -1,7 +1,7 @@
 from binascii import hexlify
 import logging
 import threading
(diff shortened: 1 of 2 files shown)


devos50:
Now the play download test is failing again... this will be the next fix on my list.

devos50:
retest this please

PR #2017, opened 2016-03-09
Remove plyvel and plyvel related code

### LINKED ISSUE #2005: Plyvel support can be removed from Tribler
@brussee Fixed python-levedb and sent a recipe to python4android upstream's so Tribler can use the same bindings across all supported OSs.

@devos50 just to confirm: OSX is not using plyvel either, right?

lfdversluis:
Plyvel is no longer a dependency. Closes #2005


lfdversluis:
@whirm I just noticed that the docs on arch linux were changed. Apparently leveldb is not on Pacman (Arch Linux), but is available [here](https://aur.archlinux.org/packages/python2-leveldb-svn/). I do not have any experience with Arch Linux, so what is the best practice here?

whirm:
Is plyvel on arch?

lfdversluis:
Yes, plyvel is available (on pacman).

whirm:
Then let's not remove plyvel support then.

whirm:
Thanks for noticing BTW, would have been a shitty move for them :D

lfdversluis:
Yes, currently the only leveldb package is in the user repository. I'll close this PR

PR #2021, opened 2016-03-10
Refactored the creation of a torrent file

devos50:
On OS X, an assertion was thrown when using the `libtorrent.add_files` method. After some searching, I found https://github.com/StorjOld/storjtorrent/blob/master/storjtorrent/storjtorrent.py which was working very well for me. I took their code as guideline and refactored our code to create a torrent.


devos50:
retest this please

devos50:
@whirm I'm finished with this one. Can you review it?

PR #2025, opened 2016-03-11
Free space actually deletes items

lfdversluis:
This PR fixes the bug where the freespace method does not delete anything. 

Fixes **1** bug
Added **1** (property) method for lint issues (enhancement). This can be used by other classes also accessing the protected member `_connection`.
Adds **2**  and modifies **1** tests.

### DIFF
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -601,6 +601,16 @@ class Session(SessionConfigInterface):
             raise OperationNotEnabledByConfigurationException("torrent_store is not enabled")
         self.lm.torrent_store.put(hexlify(infohash), data)
 
+    def delete_collected_torrent(self, infohash):
+        """
+        Deletes the given torrent from the torrent_store database.
+        :param infohash: The given infohash binary.
+        """
+        if not self.get_torrent_store():
+            raise OperationNotEnabledByConfigurationException("torrent_store is not enabled")
+
+        del self.lm.torrent_store[hexlify(infohash)]
+
     def search_remote_torrents(self, keywords):
         """
         Searches for remote torrents through SearchCommunity with the given keywords.
--- a/Tribler/Test/Core/test_session.py
+++ b/Tribler/Test/Core/test_session.py
@@ -1,0 +1,41 @@
+from binascii import hexlify
+
+from nose.tools import raises
+
+from Tribler.Core.Session import Session
+from Tribler.Core.SessionConfig import SessionStartupConfig
+from Tribler.Core.exceptions import OperationNotEnabledByConfigurationException
+from Tribler.Core.leveldbstore import LevelDbStore
+from Tribler.Test.Core.base_test import TriblerCoreTest
+
+
+class testSession(TriblerCoreTest):
+
+    @raises(OperationNotEnabledByConfigurationException)
+    def test_torrent_store_not_enabled(self):
+        config = SessionStartupConfig()
+        config.set_torrent_store(False)
+        session = Session(config, ignore_singleton=True)
+        session.delete_collected_torrent(None)
+
+    def test_torrent_store_delete(self):
+        config = SessionStartupConfig()
+        config.set_torrent_store(True)
+        session = Session(config, ignore_singleton=True)
+        # Manually set the torrent store as we don't want to start the session.
+        session.lm.torrent_store = LevelDbStore(session.get_torrent_store_dir())
+        session.lm.torrent_store[hexlify("fakehash")] = "Something"
+        self.assertEqual("Something", session.lm.torrent_store[hexlify("fakehash")])
+        session.delete_collected_torrent("fakehash")
+
+        raised_key_error = False
+        # This structure is needed because if we add a @raises above the test, we cannot close the DB
+        # resulting in a dirty reactor.
+        try:
+            self.assertRaises(KeyError,session.lm.torrent_store[hexlify("fakehash")])
+        except KeyError:
+            raised_key_error = True
+        finally:
+            session.lm.torrent_store.close()
+
+        self.assertTrue(raised_key_error)
--- a/Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
+++ b/Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
@@ -5,6 +5,7 @@ from Tribler.Category.Category import Category
 from Tribler.Core.CacheDB.SqliteCacheDBHandler import TorrentDBHandler, MyPreferenceDBHandler
 from Tribler.Core.CacheDB.sqlitecachedb import str2bin
 from Tribler.Core.TorrentDef import TorrentDef
+from Tribler.Core.leveldbstore import LevelDbStore
 from Tribler.Test.Core.test_sqlitecachedbhandler import AbstractDB
 from Tribler.Test.test_as_server import TESTS_DATA_DIR
 from Tribler.dispersy.util import blocking_call_on_reactor_thread
@@ -33,6 +34,11 @@ class TestTorrentFullSessionDBHandler(AbstractDB):
 
 
 class TestTorrentDBHandler(AbstractDB):
+
+    def setUpPreSession(self):
+        super(TestTorrentDBHandler, self).setUpPreSession()
+        self.config.set_megacache(True)
+        self.config.set_torrent_store(True)
 
     def setUp(self):
         super(TestTorrentDBHandler, self).setUp()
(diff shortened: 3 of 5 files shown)


lfdversluis:
"Tests on Mac — An error occurred during the tests on Linux. " ?

lfdversluis:
Build was aborted
Aborted by anonymous

lfdversluis:
retest this please

lfdversluis:
@whirm Why are the tests behaving so weird on this PR?

devos50:
@lfdversluis I just figured out that the tests on Mac are failing (https://jenkins.tribler.org/job/GH_Tribler_PR_tests_mac/16/testReport/junit/Tribler.Test.Core.test_sqlitecachedbhandler/TestTorrentDBHandler/test_freeSpace/) which cause the Tests on Linux job to abort immediately and report a failure.

lfdversluis:
@devos50 wow, so that message "an error occured during the tests on linux" wasn't copy pasta. Thanks for letting me know.

lfdversluis:
@devos50 Thinking about it.. it makes even less sense because the current code should also break then (update vs delete on the same table).

NielsZeilemaker:
It's not the same table, torrent vs torrenttracker

lfdversluis:
@NielsZeilemaker Ah yes, my bad. Do you remember why this function was disabled in the past?

NielsZeilemaker:
Deleting from the database doesn't make a lot of sense, you want to get rid of the torrent file. Also looking at the comment below, keeping the foreign key available for other tables makes sense. 
Especially the channels have a link to the torrentid

lfdversluis:
@NielsZeilemaker I am not too familiar with the current logic. Would deleting the file in `torrent_file_name` (in the Torrent table), lead to the deletion of the correct .torrent file?

NielsZeilemaker:
Not sure, but it would cause the torrent to not show up in the collectedtorrent view,  and hence it wouldn't be send to others etc. 
The newish torrent database has been created after I left

lfdversluis:
@NielsZeilemaker Scratch that last, I just found an upgrade "replace torrent_file_name column with is_collected column" in `_upgrade_25_to_26`. Looks like solving this requires knowledge of the details. This will take some time I guess..

whirm:
The torrent store is a key values store, the key is the infohash an the value the raw contents of the .torrent file.

If is_collected is true, it means we've got the torrent on the store, if it's false it means it's not. So if you set is_collected to false when cleaning, you first need to delete that entry from the torrent store and a quick grep through Tribler doesn't show a single place where the torrents are removed from the store.

I also think we could store way more than 50.000 torrents in levelDB if that improves the user experience.

devos50:
You can easily delete a torrent from the leveldb torrent store by calling `del session.lm.torrent_store[infohash]`. However, maybe it is better to create a separate method in `Session` (called `delete_collected_torrent` or something similar) and use that.

I also think we can safely increase the limit of collected torrents to 100K or even more. [This](http://highscalability.com/blog/2011/8/10/leveldb-fast-and-lightweight-keyvalue-database-from-the-auth.html) post talks about scalable leveldb databases with more than a million entries:

- _Search performance is still O(lg N) with a very large branching factor (so the constant factor is small and number of seeks should be <= 10 even for gigantic databases).
- One early user found performance degraded at around 200 million keys._

Maybe @lfdversluis can do a performance measurement regarding leveldb? Should be easy to setup, we can just fill a database with many torrent files and perform queries on it :)

lfdversluis:
@devos50 I wouldn't do too many benchmarks now. I would first focus on issues and then on improvements including benchmarks to show it's indeed improving.

NielsZeilemaker:
@devos50 what do you want to achieve by increasing the limit to 100k?

lfdversluis:
The torrent store is enabled in the tests, yet they report it is disabled somehow. Any idea what could be wrong @devos50 @whirm ?

lfdversluis:
retest this please

lfdversluis:
@whirm Ready to be reviewed.

Tribler/Core/Session.py
```diff
@@ -601,6 +601,20 @@ def save_collected_torrent(self, infohash, data):
             raise OperationNotEnabledByConfigurationException("torrent_store is not enabled")
         self.lm.torrent_store.put(hexlify(infohash), data)
 
+    def delete_collected_torrent(self, infohash):
+        """
+        Deletes the given torrent into the torrent_store database.
```
whirm: s/into/from/

Tribler/Core/Session.py
```diff
@@ -601,6 +601,20 @@ def save_collected_torrent(self, infohash, data):
...
+        :param infohash: The given infohash binary.
+        """
+        if not self.get_torrent_store():
+            raise OperationNotEnabledByConfigurationException("torrent_store is not enabled")
```
whirm: Try to make lines shorter than 120

whirm: Not sure why pylint is not catching this. Did I forget to configure this?

Tribler/Core/Session.py
```diff
@@ -601,6 +601,20 @@ def save_collected_torrent(self, infohash, data):
...
+
+        if self.lm.torrent_store is not None:
+            del self.lm.torrent_store[hexlify(infohash)]
+        else:
+            self._logger.error("Torrent store is enabled but could not delete infohash %s" % infohash)
+            raise RuntimeError("Torrent store is enabled but None")
```
whirm: No need for this kind of checks, it will fail anyway. just leave the del statement on this chunk.

lfdversluis:
@whirm looks like it

Tribler/Test/Core/test_session.py
```diff
@@ -0,0 +1,23 @@
...
+        config.set_torrent_store(False)
+        session = Session(config, ignore_singleton=True)
+        session.delete_collected_torrent(None)
+
+    @raises(RuntimeError)
+    def test_torrent_store_none(self):
+        config = SessionStartupConfig()
+        config.set_torrent_store(True)
+        session = Session(config, ignore_singleton=True)
+        session.delete_collected_torrent(None)
```
whirm: Missing EOL

whirm: This should be caught by the test job too

Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
```diff
@@ -224,10 +230,14 @@ def test_getCollectedTorrentHashes(self):
 
     @blocking_call_on_reactor_thread
     def test_freeSpace(self):
+        # Manually set the torrent store because register is not called.
+        self.session.lm.torrent_store = LevelDbStore(self.session.get_torrent_store_dir())
+        # print self.session.lm.torrent_store
```
whirm: leftover debug stuff

Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
```diff
@@ -224,10 +230,14 @@ def test_getCollectedTorrentHashes(self):
...
         old_res = self.tdb.getNumberCollectedTorrents()
         self.tdb.freeSpace(20)
         res = self.tdb.getNumberCollectedTorrents()
-        self.assertEqual(old_res, res)
+        self.session.lm.torrent_store.close()
+        self.assertLess(res, old_res)
```
whirm: Maybe you should check that the amount left is the right one, not just less than before.

whirm:
Looks fine

Tribler/Core/Session.py
lfdversluis: The line is less than 120 according to pycharm.

lfdversluis:
@whirm Ready to merge.

lfdversluis:
retest this please

whirm:
retest this please

lfdversluis:
`TestLibtorrentDownload.test_playdownload - streaming did not start - Condition was not satisfied in 60 seconds`

lfdversluis:
retest this please

lfdversluis:
`TestLibtorrentDownload.test_downloadfromurl - no download progress - Condition was not satisfied in 30 seconds`

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
@whirm Ready for review.

whirm:
Looks good, squash the last commit with the one that introduced what you are fixing (get used to do fixup commits) and I'll merge.

lfdversluis:
@whirm Squashed. Tests on mac are failing because something is up with the tester on mac. (see other PRs).

lfdversluis:
retest this please

PR #2026, opened 2016-03-12
Improvements of sqlitecachedbhandler tests

devos50:
`sqlitecachedbhandler.py` is probably the largest file that can be found in Tribler. In this PR, I improved the tests for this file. Since `sqlitecachedbhandler.py` is concerned with separate database handlers, I created separate files to test each database handler on its own.

I also did some other changes. I noticed that we were using quite big databases for the unit tests (around 10MB). They were zipped into a `.tar.gz` file. However, 99% of this database was unused during the unit tests. The tests for the upgrader only needed the version field of the database. So I replaced these databases with way smaller databases with some dummy data.

The coverage of the `Tribler/Core` module is around 50% now when running all tests in `Tribler/Test/Core`.


devos50:
retest this please

devos50:
retest this please

whirm:
Regarding the smaller database: once that file got committed to git, it will take the same space even if you delete the file. I'd rather keep it and use it for future tests (for instance the database upgrade code is basically untouched by any test). So if you don't see any noticeable change in the test execution time, I'd rather drop that change.

Tribler/Test/Core/test_sqlitecachedb_upgrade.py
```diff
@@ -34,27 +43,24 @@ def tearDown(self):
...
 
         db_migrator = DBUpgrader(self.session, self.sqlitedb, torrent_store=MockTorrentStore())
         self.assertRaises(VersionNoLongerSupportedError, db_migrator.start_migrate)
 
-    def test_upgrade_from_17(self):
-        pass
-        # TODO(emilon): Implement that one and 18 22 23
-        # assert sqlitedb.version == LATEST_DB_VERSION, "Database didn't get upgraded to latest version (%s != %s)" % (
-        #     sqlitedb.version, LATEST_DB_VERSION)
+    def test_upgrade_17_to_latest(self):
```
whirm: I see you implemented that, so never mind my previous comment about that.

Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
```diff
@@ -0,0 +1,256 @@
...
+        comments = 'something not inside'
+        assert m_comment.find(comments) == -1
+
+        m_trackers = self.tdb.getTrackerListByInfohash(m_infohash)
+        assert len(m_trackers) == 8
+        assert 'http://tpb.tracker.thepiratebay.org/announce' in m_trackers, m_trackers
+
+        s_torrent = self.tdb.getTorrent(s_infohash)
+        m_torrent = self.tdb.getTorrent(m_infohash)
+        assert s_torrent['name'] == 'Tribler_4.1.7_src.zip', s_torrent['name']
```
whirm: Now that we are at it, could you replace the plain asserts with nose's ones?

whirm:
I don't know what you added into the DB, but I think having a "real" Tribler DB is a good Idea for the upgrade tests.

Tribler/Test/Core/test_sqlitecachedb_upgrade.py
devos50: @whirm I can place the file back. Do you think it's ok if I use the smaller database for the unit tests then? Maybe we can use the bigger databases for other unit tests in that case.

devos50:
@whirm I can write a second that that uses the 'old' bigger database I removed.

Tribler/Test/Core/test_sqlitecachedbhandler_torrents.py
devos50: Yeah I will do that.

whirm:
What would be the use of the smaller one then? Wouldn't be bigger one supersede it?

devos50:
@whirm yeah it would have no use then. I will copy the rows I created myself in the smaller database into the bigger database and use that database instead.

devos50:
@whirm I replaced all built-in asserts with the assertions of nose. Moreover, I placed the bigger database back again and used that one for the unit tests. I noticed that the tar extract script was extracting to the tests directory so I changed that. I found out that I could replace the whole script with only a few lines of code. Now I'm first copying the `tar.gz` file to the session directory and I'm extracting it there.

devos50:
@whirm I moved the `import tarfile` to the top and skipped the copying of the archive to the session directory.

devos50:
retest this please

devos50:
retest this please

whirm:
Nice, I'll merge if everything works then

devos50:
retest this please

whirm:
re this: https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/342/artifact/output/test_runners_output/00010.err/*view*/

I guess we will have to create a PR with debug prints all over in `_test_port`, `check_random_port` and `get_random_port` and make it fail.

whirm:
retest this please

devos50:
retest this please

devos50:
@whirm tests are finally passing...

PR #2028, opened 2016-03-14
Changed parameters of channel search test + removed blocking call to lookup trackers

devos50:
The channel search test is still failing because it cannot find channels (though a lot less than before). I increased the number of peers to have before the search can be executed. I also increased the callback time for the check whether there are channel results.

Still, these GUI tests needs a major refactoring but that's something for the longer term. Hopefully this change will increase the stability of the channel search test.

### DIFF
--- a/Tribler/Core/Utilities/tracker_utils.py
+++ b/Tribler/Core/Utilities/tracker_utils.py
@@ -112,10 +112,4 @@ def parse_tracker_url(tracker_url):
     else:
         raise RuntimeError(u'No port number for UDP tracker URL.')
 
-    # TODO(emilon): A URL parser shoudn't have a blocking host resolution on it.
-    try:
-        hostname = socket.gethostbyname(hostname)
-    except:
-        raise RuntimeError(u'Cannot resolve tracker URL.')
-
     return tracker_type, (hostname, port), announce_page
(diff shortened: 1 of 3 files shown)


whirm:
I'll retest this a few times to see how things improve

whirm:
retest this please

devos50:
retest this please

whirm:
`DEBUG   1457968154   network_utils:Tribler.Core.Utilities.network_utils:88   Port test failed (port=52842, family=2, type=1): [Errno 98] Address already in use`

https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/321/artifact/output/test_runners_output/00002.err/*view*/

whirm:
I don't think testlibtorrentdownload should have the tunnels enabled?

whirm:
retest this please

devos50:
retest this please

devos50:
@whirm That `parse_tracker_url` method was being used at two places (and in the tests), namely in `TorrentChecker/session.py`. There, the Udp or Http connection to a tracker will be opened, however, it is not needed to resolve the hostname in the `parse_tracker_url` method (this will be done during the connection anyway). I also do not see any strange side effects when running the tests.

PR #2040, opened 2016-03-16
Fix watchdog and tests

whirm:


whirm:
retest this please

whirm:
retest this please

whirm:
retest this please

whirm:
retest this please

devos50:
@whirm I'm investigating the issue, it seems to get stuck on the invocation of the `join` method in WatchDog.

whirm:
@devos50 https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/381/artifact/output/test_runners_output/00001.err/*view*/

devos50:
@whirm the `self.should_stop` attribute is not correctly set to `True`, preventing the while-loop in the `run` method of the WatchDog from stopping. Maybe we are calling the `join` method from the wrong thread?

whirm:
If the should_stop wasn't properly set, how come it stops looping?

The join() is called from the TearDown() method on test_as_server, and it can be called from any thread even several times.

devos50:
@whirm that's weird, I added a print statement in the while-loop and it keeps spamming that print statement when it gets stuck.

whirm:
Look at the link I pasted, I enabled debugging and it stops printing the debug stuff

whirm:
Well, it actually never starts

whirm:
I guess the test ends too quickly

whirm:
in any case, if it the bool wasn't being set properly, it should keep printing the debug info

devos50:
@whirm yeah indeed, I now see the print in the while loop and it stops spamming. It also says `stopping watchdog` so the join is done.

devos50:
@whirm I see you pushed a fix, I will checkout your branch and test whether it also works for me by running the particular test that is failing several times :)

whirm:
The join is not done, if You saw "Stopped watchdog" it would be done

https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/384/testReport/junit/Tribler.Test.Core.test_utilities/TriblerCoreTestUtilities/test_valid_torrent_file_nodes_invalid_length/

devos50:
@whirm lol confused stopped with stopping -.-

whirm:
it's really weird, because the same thread that's setting the should_stop attr by calling join, still sees the value as false

whirm:
I think I know what's going on

devos50:
```
ERROR: test_a_decode_false (Tribler.Test.Core.test_encoding.TriblerCoreTestUnicode)
  File "/home/martijn/tribler/Tribler/Test/test_as_server.py", line 134, in tearDown
    self.watchdog.join(2)
  File "/home/martijn/tribler/Tribler/Core/Utilities/instrumentation.py", line 110, in join
    with self._synchronized_lock:
AttributeError: 'WatchDog' object has no attribute '_synchronized_lock'
```

whirm:
I think I fixed it

whirm:
Apparently there's a race between the thread actually starting and calling join() before that actually happens, so _reset_state() (first line of run()) was being called AFTER join() was called in the main thread.

devos50:
@whirm that makes sense! I'm running the tests now on your branch and it seems to work very well. Had no failures yet.

whirm:
retest this please

whirm:
yay

PR #2041, opened 2016-03-17
Fixed and improved plyvel tests

devos50:
The plyvel tests were not actually using the plyvel wrapper. To fix this, I defined a global variable  `USE_LEVELDB` that can be changed by the tests and during runtime.

I also added a test to test the Put method in leveldb directly. We are not using the Put method directly since we are batching operations.


devos50:
retest this please

devos50:
retest this please

whirm:
Why aren't you just overriding `openStore` on `TestLevelDBStore_plyvel` instead of adding that flag and all those `if``s?

devos50:
@whirm that was indeed my initial idea to solve this. However, this is only solving one part of the problem since we also have to deal with the fact that the signature of the `WriteBatch` class is different for plyvel and leveldb (the `WriteBatch` used by plyvel expects one parameter being passed while the leveldb `WriteBatch` does not require any).

Before, we were solving this with the `try except` construction. The problem with this approach is that it is begin executed immediately when the file is imported. Moreover, it always gives priority to leveldb, which means that it will choose the `WriteBatch` used by leveldb, no matter what. This is part of the explanation why these plyvel tests were using leveldb instead of plyvel. By creating a file-wide attribute (`USE_LEVELDB`), I have the flexibility to change this parameter when running the plyvel tests.

Another issue I can across is that when I subclass the `TestLeveldbStore` class, is that the tests are executed twice. This is behaviour of nose: the tests are executed one time if the `TestLeveldbStore` is imported and another time when a subclass is being made. This is why I removed the subclass implementation and added a (seemingly useless) import.

I'm pretty sure there might be a better construction so if you or someone else has a better suggestion on how to do this, please let met know :)

whirm:
I guess we need to have a superclass with __test__ = False and two subclasses, one for each leveldb module that sets the pimped TorrentStore and __test__ to True. 

See: http://nose.readthedocs.org/en/latest/finding_tests.html

The pimped torrent store replaces the reactor with a Clock and sets the right levelDB module for the test. This way we force plyvel on one and python-leveldb on the other and we avoid adding cruft to the non-testing code.

whirm:
Let me know if I'm missing something from your explanation.

devos50:
@whirm didn't know about the `test` attribute, that is nice to use! In that case, we have a clear structure for the tests.

The only problem (?) left I think is deciding which `WriteBatch` we should use. If I revert my changes, the leveldb `WriteBatch` is always used, even if plyvel is installed. Any suggestions for that?

whirm:
Just have a `_writebatch` attr like `_leveldb` so we can override it from the test

devos50:
@whirm I've implemented your suggestion. Can you take a look at it?

Tribler/Test/Core/test_leveldb_store.py
```diff
@@ -48,14 +48,44 @@
...
+class ClockedAbstractLevelDBStore(LevelDbStore):
     _reactor = Clock()
 
 
-class TestLevelDbStore(BaseTestCase):
+class ClockedLevelDBStore(ClockedAbstractLevelDBStore):
+    from leveldb import LevelDB
+    _leveldb = LevelDB
+
+    def __init__(self, store_dir):
```
whirm: You can drop __init__  and get_write_batch if you set _writebatch as a static attribute like _leveldb is.
Also keep both modules and let the plyvel one import what's needed from the other one.

devos50:
@whirm this should work.

Tribler/Core/leveldbstore.py
```diff
@@ -77,6 +86,7 @@ def __init__(self, store_dir):
         self._writeback_lc = self.register_task("flush cache ", LoopingCall(self.flush))
         self._writeback_lc.clock = self._reactor
         self._writeback_lc.start(WRITEBACK_PERIOD)
+        print self._db
```
whirm: you missed this :)

PR #2046, opened 2016-03-19
Unit tests improvements

### LINKED ISSUE #1729: Increase code coverage by running unit tests on OS X and Windows too
Currently, the code coverage covers 48% of the conditional statements. This number can be increased by also running unit tests on OS X and Windows since many of the statements are only executed on specific platforms. At the end of the build process, the code coverage reports can be merged.

### LINKED ISSUE #1801: Create tests that checks whether upgrading of older Tribler versions is working
We should have tests that test whether upgrading the database to a newer version is working. This can be done by placing old and/or corrupt .Tribler directory in the tests and invoke the upgrader.

### LINKED ISSUE #1811: Mark unit tests as failed if an exception occurred
Currently, when an exception is thrown, the test still continues to run. We should add a Twisted log observer to check whether there are errors and if so, mark the test as failed.

### LINKED ISSUE #1944: Create local webserver and add tests to test torrent and rss downloading
In PR #1942 we changed remote torrents to local ones because one of the dependencies dropped and a lot test broke.

We should set up a webserver of our own to host some torrent files that will always be available, or at least under our control to fix to also test if using remote downloading works using additional tests.

### LINKED ISSUE #2022: Stack trace being thrown sometimes when executing test_channel_search
Sometimes during the tests, the following stack trace is displayed:

```
watchog wx thread *******TRIPPED!******, hasn't been set for 12.0160 secs.

*** STACKTRACE - START ***


### ThreadID: 139818506487552 Thread name: WatchDog
Locals by frame, innermost last:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/Utilities/instrumentation.py:175 print_all_stacks:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/Utilities/instrumentation.py:148 run:
/usr/lib/python2.7/threading.py:801 __bootstrap_inner:
/usr/lib/python2.7/threading.py:774 __bootstrap:

### ThreadID: 139818489702144 Thread name: PoolThread-twisted.internet.reactor-0
Locals by frame, innermost last:
/usr/lib/python2.7/threading.py:371 wait:
/usr/lib/python2.7/Queue.py:182 get:
/usr/lib/python2.7/dist-packages/twisted/_threads/_threadworker.py:45 work:
/usr/lib/python2.7/threading.py:754 run:
/usr/lib/python2.7/threading.py:801 __bootstrap_inner:
/usr/lib/python2.7/threading.py:774 __bootstrap:

### ThreadID: 139818852628224 Thread name: Twisted
Locals by frame, innermost last:
/usr/lib/python2.7/dist-packages/twisted/internet/fdesc.py:94 readFromFD:
/usr/lib/python2.7/dist-packages/twisted/internet/posixbase.py:145 doRead:
/usr/lib/python2.7/dist-packages/twisted/internet/posixbase.py:610 _doReadOrWrite:
/usr/lib/python2.7/dist-packages/twisted/python/context.py:83 callWithContext:
/usr/lib/python2.7/dist-packages/twisted/python/context.py:118 callWithContext:
/usr/lib/python2.7/dist-packages/twisted/python/log.py:84 callWithContext:
/usr/lib/python2.7/dist-packages/twisted/python/log.py:101 callWithLogger:
/usr/lib/python2.7/dist-packages/twisted/internet/epollreactor.py:396 doPoll:
/usr/lib/python2.7/dist-packages/twisted/internet/base.py:1206 mainLoop:
/usr/lib/python2.7/dist-packages/twisted/internet/base.py:1194 run:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/Utilities/twisted_thread.py:55 _reactor_runner:
/usr/lib/python2.7/threading.py:754 run:
/usr/lib/python2.7/threading.py:801 __bootstrap_inner:
/usr/lib/python2.7/threading.py:774 __bootstrap:

### ThreadID: 139818960881408 Thread name: MainThread
Locals by frame, innermost last:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_misc.py:1324 Start:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_core.py:16817 Start:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_core.py:16801 __init__:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_as_server.py:373 wx_watchdog_keepalive:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_core.py:16872 Notify:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_misc.py:1367 Notify:
/usr/lib/python2.7/dist-packages/wx-3.0-gtk2/wx/_core.py:8370 Yield:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Main/tribler_main.py:920 OnExit:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Main/vwxGUI/__init__.py:122 invoke_func:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Main/tribler_main.py:1029 run:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_as_server.py:450 startTest:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_remote_search.py:27 startTest:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_remote_search.py:84 test_channel_search:
/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_as_server.py:69 check:
/usr/lib/python2.7/unittest/case.py:329 run:
/usr/lib/python2.7/unittest/case.py:393 __call__:
/usr/lib/python2.7/dist-packages/nose/case.py:151 runTest:
/usr/lib/python2.7/dist-packages/nose/case.py:133 run:
/usr/lib/python2.7/dist-packages/nose/case.py:45 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/suite.py:224 run:
/usr/lib/python2.7/dist-packages/nose/suite.py:177 __call__:
/usr/lib/python2.7/dist-packages/nose/core.py:62 run:
/usr/lib/python2.7/dist-packages/nose/core.py:207 runTests:
/usr/lib/python2.7/unittest/main.py:95 __init__:
/usr/lib/python2.7/dist-packages/nose/core.py:121 __init__:
|          env = {'BUILD_URL': 'https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/265/', 'NOSE_TESTS_TO_RUN': 'Tribler/Test/', 'BUILD_TAG': 'jenkins-GH_Tribler_PR_tests_linux-265', 'ghprbAuthorRepoGitUrl': 'https://github.com/devos50/tribler.git', 'TRACKER_CMD': '', 'SHELL': '/bin/bash', 'REMOTE_WORKSPACE_DIR': './', 'ROOT_BUILD_CAUSE': 'GHPRBCAUSE', 'BUILD_NUMBER': '265', 'XDG_RUNTIME_DIR': '/run/user/1001', 'JOB_URL': 'https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/', 'PYTHONPATH': '/home/je...
|       config = Config(addPaths=True, args=(), configSection='nosetests', debug=None, debugLog=None, env={}, exclude=None, files=[], firstPackageWins=False, getTestCaseNamesCompat=False, ignoreFiles=[<_sre.SRE_Pattern object at 0x7f2a236efcb0>, <_sre.SRE_Pattern object at 0x7f2a23647cb0>, <_sre.SRE_Pattern object at 0x7f2a2378adb0>], ignoreFilesDefaultStrings=['^\\.', '^_', '^setup\\.py$'], include=None, includeExe=False, logStream=<open file '<stderr>', mode 'w' at 0x7f2a237d11e0>, loggingConfig=None, options=...
/usr/bin/nosetests:9 <module>:

*** STACKTRACE - END ***
```

### LINKED ISSUE #2044: win32regchecker is not covered at all
Was looking in the coverage, there is no test for `Tribler/Core/Utilities/win32regchecker.py`. @devos50 Maybe you can add this in your unit tests for windows PR?

### LINKED ISSUE #2081: Exception when running tests on Windows
When running tests on the win64 builder, sometimes the following error is raised:

```
ERROR: test_watchdog_print_all_stacks (Tribler.Test.Core.test_instrumentation.TriblerCoreTestWatchDog)
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Test\test_as_server.py", line 71, in check
    result = fun(*argv, **kwargs)
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Test\Core\test_instrumentation.py", line 110, in test_watchdog_print_all_stacks
    self.watchdog.print_all_stacks()
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Core\Utilities\instrumentation.py", line 175, in print_all_stacks
    self.printe("\n### ThreadID: %s Thread name: %s" % (thread_id, thread_names[thread_id]))
KeyError: 868
```

During the past months, I've seen the same error once or twice, also on other platforms. It seems like the threads returned by `threading.enumerate()` is not always the same as the threads given by `sys._current_frames().items()`.

@whirm Is it an idea to add a check to see whether the `thread_id` key exists in `thread_names`? If it doesn't, at least the tests won't fail.

devos50:
This PR contains the following fixes/improvements to the unit tests:
-Some tests gave unexpected results when running on Windows so I had to fix them.
-Cleaned up `win32regchecker.py` by removing unused methods and written a test for it. When the new GUI goes live, this class (together with all video player utilities) can be removed.
-When closing, sometimes a stack trace would show. This only happens in the yield methods during shutdown which might sometimes take a while to execute. When during this period, the wx watchdog event is not set, watchdog trips. To prevent this from happening, I unregister the wx watchdog event earlier.
-Added some tests for the thread pool manager.
-The torrent upgrader classes do not require a session anymore as parameter. This makes it easier to unit test the upgrader.
-Written tests for the torrent upgrade process from Tribler 6.3 -> 6.4 and 6.4 -> 6.5. I also written test to test `upgrade.py`, the class that is responsible for the whole upgrade process. I've covered all files regarding upgrading for more than 90% so I think the upgrade process is currently well-tested.
-Written a local HTTP file server, used to serve torrents in the tests. This should be used instead of external network URLs since they might be unreliable. I've rewritten some tests to make use of this local HTTP server.

Fixes #1729, fixes #2044, fixes #1811, fixes #2022, fixes #1944, fixes #1801, fixes #2081. 

Note: the win32 en win64 unit tests are currently testing this branch. This should be changed as soon this PR is merged into devel.

Note for @whirm: the diff coverage tools is complaining about uncovered files in the `Tribler.Test` module. This can be explained by the fact that we do not cover `Tribler.Test` during the nosetests (we do not pass the `--cover-tests` option to nose). So the diff tool reports a lower coverage. We can either ignore test coverage in the the tests files or we can pass `--cover-tests` to nosetests so we include coverage of the test files. I prefer the first options since I think it is not very beneficial to also cover the tests files.


devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

synctext:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

synctext:
retest this please

devos50:
@whirm ready for review

whirm:
> -Plyvel tests are not executed on Windows. This is because Plyvel is complicated to compile from source on 32-bit.

Could you elaborate on that?

whirm:
> Note for @whirm: the diff coverage tools is complaining about uncovered files in the Tribler.Test module. This can be explained by the fact that we do not cover Tribler.Test during the nosetests (we do not pass the --cover-tests option to nose). So the diff tool reports a lower coverage. We can either ignore test coverage in the the tests files or we can pass --cover-tests to nosetests so we include coverage of the test files. I prefer the first options since I think it is not very beneficial to also cover the tests files.

Why do you think not reporting the coverage on the tests is better? We would have "false positive" failures, but it's better than the other way around.

devos50:
> Why do you think not reporting the coverage on the tests is better? We would have "false positive" failures, but it's better than the other way around.

Yeah, on second thought it is better to include the tests themselves. For instance, if I create a test with an if-statement to run some specific code on a platform, we should also make sure that the conditional is fully covered.

Tribler/Test/API/test_download.py
```diff
@@ -33,11 +36,20 @@ def setUpPostSession(self):
...
+        shutil.copyfile(TORRENT_FILE, os.path.join(files_path, 'ubuntu.torrent'))
+        file_server_port = get_random_port()
+        self.setUpFileServer(file_server_port, files_path)
+
+        d = self.session.start_download_from_uri('http://localhost:%s/ubuntu.torrent' % file_server_port)
         self._logger.debug("Download started: %s", d)
         d.set_state_callback(self.downloader_state_callback)
         assert self.downloading_event.wait(60)
 
+        self.file_server.stopListening()
```
whirm: You should be using setUp/tearDown for this kind of stuff. If the test fails, you will leak the file_server.

Tribler/Test/Core/Upgrade/test_torrent_upgrade_63_64.py
```diff
@@ -0,0 +1,118 @@
...
+from Tribler.Test.test_libtorrent_download import TORRENT_FILE
+
+
+class AbstractTorrentUpgrade63to64(TriblerCoreTest):
+
+    FILE_DIR = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
+    DB_DATA_DIR = os.path.abspath(os.path.join(FILE_DIR, u"../data/upgrade_databases/"))
+
+    def write_data_to_file(self, file_name):
+        file = open(file_name, 'w')
```
whirm: totally cosmetic remark: you can use with open(...) as file: to make it nicer :)

Tribler/Test/Core/test_install_dir.py
```diff
@@ -7,7 +8,9 @@ class TriblerCoreTestInstallDir(TriblerCoreTest):
 
     def test_install_dir(self):
         install_dir = determine_install_dir()
-        print install_dir
-        self.assertIsInstance(install_dir, str)
+        if sys.platform == "win32":
+            self.assertIsInstance(install_dir, unicode)
+        else:
+            self.assertIsInstance(install_dir, str)
```
whirm: second arg can be a tuple of types, use that instead.

```
 assertIsInstance(obj, cls, msg=None)¶
 assertNotIsInstance(obj, cls, msg=None)¶

 Test that obj is (or is not) an instance of cls (which can be a class or a tuple of classes, as supported by
 isinstance()). To check for the exact type, use assertIs(type(obj), cls).
```

Tribler/Test/Core/test_sessionconfig.py
```diff
@@ -26,7 +27,10 @@ def test_session_config_init(self):
         sci.set_state_dir(self.session_base_dir)
         self.assertEqual(sci.get_state_dir(), self.session_base_dir)
 
-        self.assertIsInstance(sci.get_install_dir(), str)
```
whirm: idem

Tribler/Test/Core/test_leveldb_store_plyvel.py
```diff
@@ -1,13 +1,17 @@
...
 from Tribler.Core.leveldbstore import get_write_batch_plyvel
 from Tribler.Test.Core.test_leveldb_store import ClockedAbstractLevelDBStore, AbstractTestLevelDBStore
 
 
-class ClockedPlyvelStore(ClockedAbstractLevelDBStore):
-    from Tribler.Core.plyveladapter import LevelDB
-    _leveldb = LevelDB
-    _writebatch = get_write_batch_plyvel
+# TODO Martijn: compile and enable Plyvel tests on Windows
+if sys.platform != "win32":
```
whirm: Just use `unittests.skipIf`


```

unittest.skipIf(condition, reason)¶

 Skip the decorated test if condition is true.
```

Tribler/Test/Core/test_threadpoolmanager.py
```diff
@@ -0,0 +1,45 @@
...
+    def callback_func(self):
+        self.callback_called = True
+
+    def test_check_task_name(self):
+        self.assertEqual(self.tpm._check_task_name("test"), "test")
+        self.assertEqual(self.tpm._check_task_name(None), "threadpool_manager 1")
+
+    def test_add_task(self):
+        self.tpm.add_task(self.callback_func, 0.0)
+        sleep(0.1)
```
whirm: Use a deferred with a timeout decorator, let's try not to add more sleeps if we can avoid it.

Tribler/Test/test_as_server.py
```diff
@@ -465,6 +477,8 @@ def callLater(self, seconds, callback):
                 callback()
 
     def quit(self):
+        # Martijn: stop wx thread watchdog event sooner to prevent errors when yielding in the GUI shutdown
+        self.watchdog.unregister_event("wx thread")
```
whirm: This means that if something really freezes while shutting down, we will never see the stacktrace to help us debug it.

Tribler/Test/test_libtorrent_download.py
```diff
@@ -70,27 +70,35 @@ def do_downloadfromfile():
...
             self.guiUtility.showLibrary()
-            self.frame.startDownloadFromArg(TORRENT_R, self.getDestDir())
+            self.frame.startDownloadFromArg(r'http://localhost:%s/ubuntu.torrent' % self.file_server_port,
+                                            self.getDestDir())
 
-            self.CallConditional(30, lambda: self.session.get_download(infohash), download_object_ready,
+            self.CallConditional(10, lambda: self.session.get_download(TORRENT_FILE_INFOHASH), download_object_ready,
                                  'do_downloadfromurl() failed')
 
+        # Setup file server to serve torrent file
```
whirm: shouldn't this block be an auxiliar function so it can be reused?

Tribler/Test/test_my_channel.py
```diff
@@ -94,7 +111,7 @@ def do_add_torrent(torrentfilename):
             managefiles = self.managechannel.fileslist
             manager = managefiles.GetManager()
             manager.startDownload(torrentfilename, fixtorrent=True)
-            manager.startDownloadFromUrl(TORRENT_R, fixtorrent=True)
+            manager.startDownloadFromUrl(r'http://localhost:7286/video.torrent', fixtorrent=True)
```
whirm: fixed port?

devos50:
@whirm managed to compile and plyvel on 32-bit :D. The last time I tried, I had issues with the 32-bit compiler of VS2010 but I just tried to compile plyvel with [this](https://www.microsoft.com/en-us/download/details.aspx?id=44266) compiler and it worked. I will remove the first commit.

whirm:
> Update: plyvel tests are working now on both Windows 32-bit and 64-bit.

That's great! :)

whirm:
Waiting for the rest of comments to be addressed.

devos50:
retest this please

Tribler/Test/test_as_server.py
devos50: @whirm I agree but during a `wx.Yield`, the wx watchdog event is not set. This is problematic if the yield takes a long time (> 10 seconds). I've only seen the tripping happening when closing Tribler during the tests but the tests themselves do not fail if it trips. It's also really hard to reproduce this issue with the tests. It only happens in around 5% of the test cases and mostly on the tests that takes a long time (i.e. channel search).

Can you think of a better solution for this problem? Maybe we can ignore the stack trace for now. Nevertheless, I will remove the commit with the code above since this is not really a good solution for now.

Tribler/Test/test_libtorrent_download.py
devos50: In this case not. The files to be served by the file server can differ for every test so this part we cannot extract into a separate method. Also note that the `setUpFileServer` method can be reused.

Tribler/Test/test_my_channel.py
devos50: This is done so the URLs in the `test_rss.xml` file are correct. We could dynamically replace these ports in `test_rss.xml` when running the test?

devos50:
@whirm ready again

whirm:
https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/692/artifact/output/test_runners_output/00005.err/*view*/

devos50:
@whirm that's because of the new Twisted log observer in this PR :), without it the error would be unnoticed. I will investigate it now.

whirm:
https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/692/artifact/output/test_runners_output/00008.err/*view*/

I don't get why that one happened though...

whirm:
I'm fixing the gumby test runner BTW, if one test fails, the report generation is aborted with the current script.

whirm:
Ok, should be fixed now

devos50:
retest this please

devos50:
@whirm I just pushed a fix for the error we got.

devos50:
retest this please

devos50:
@whirm I created some code to setup a local seeder (since I want to seed the video that is being downloaded in `test_playdownload` locally). This works fine when I created a headless session, however, it does not seem to work when I use it in combination with wx.

I was thinking of another solution: create a small script that starts up a Tribler seeder and run it as a subprocess. This way, we have no interference between session objects. Do you agree on this?

devos50:
retest this please

devos50:
retest this please

devos50:
I created some code to setup a local seeder. I've also refactored some tests to make use of this local seeder (also the video playback test which appears to be stable now!).

@whirm can you review my 5 latest commits? I'm not planning to add more stuff to this PR.

Also, the line that returns `Unknown` when printing all stack seems very hard to cover with a test unless I make `get_thread_name` a class method.

devos50:
retest this please

devos50:
Will test a few more times to see whether tests are more stable

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
5 successful runs in a row

lfdversluis:
But will he 6th succeed?

lfdversluis:
But there is stuff left to cover!

devos50:
retest this please

devos50:
Mac has issues...

devos50:
retest this please

synctext:
+1
nice thesis contribution

devos50:
@lfdversluis 6th run also succeeded!

@whirm ready for review again (last 5 commits)

lfdversluis:
@devos50 But what about the not covered code? Is the diff tool off?

devos50:
@lfdversluis nope, the missing lines in instrumentations are very hard to cover (since they are checking that the tests are not raising Twisted errors), however, I was expecting the lines in `test_win32regchecker.py` to be covered now (at least they should be covered on Windows). Something is not going well when combining coverage xml files I guess. I will take a look.

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
This is weird, I enabled test coverage reporting on all PR builders and only on Linux, the `Tribler/Test` package is almost completely covered. On Mac and Windows, only a small part of the files in `Tribler/Test` is covered. It seems that the `coverage.xml` file is overwritten every time a test runner finishes but I'm unable to find the root cause of this and the reason why it's working on Linux and not on other platforms.

@whirm is there anything wrong in the way the coverage files are generated and combined? The command on Windows/OS X is started [here](https://github.com/Tribler/gumby/blob/devel/scripts/run_nosetests_for_jenkins.sh#L155) and on Linux, it is started [here](https://github.com/Tribler/gumby/blob/devel/scripts/run_nosetests_for_jenkins.sh#L127). Can you please take a look to see if there are any errors here?

whirm:
@devos50 well, you are running only the `Core` tests, so isn't that expected?

devos50:
@whirm oh I'm sorry, I meant `Tribler/Test/Core` ;)

whirm:
I guess you will have to check the individual XML files, I've looked at the mac tests only and the script looks fine to me.

devos50:
There are no individual xml files, only one big `coverage.xml` file. This morning I ran the tests isolated on the mac so I could investigate further but there are no xml files related to coverage (also no `.coverage` files).

whirm:
Ohh, sorry, I wasn't paying attention. Yes theres an obvious bug: you have an extra newline :D 

154 and 155 should be the same line. otherwise you are just setting a local variable on 154 and then calling the test runner which won't see the variable.

devos50:
@whirm thanks! I was expecting that the script could find the `COVERAGE_FILE` variable but apparently, it doesn't. I will open a PR to fix this.

whirm:
@devos50 it will find it if you `export` it otherwise it will stay a local variable and not be inherited by the shell's sub-processes. In this case it's an ephemeral variable so just prepend it to the command so you don't pollute your local namespace.

devos50:
@whirm yeah, didn't know that. I fixed it in https://github.com/Tribler/gumby/pull/265 :)

devos50:
retest this please

devos50:
retest this please

whirm:
Nice!

devos50:
@whirm can you review my last 5 commits so we can finish this PR? :)

whirm:
I'm trying to :D 

I'm thinking about limiting the PR size, these huge PRs are going to be the end of me...

Tribler/Test/Core/Upgrade/upgrade_base.py
```diff
@@ -0,0 +1,74 @@
...
+
+    FILE_DIR = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
+    DATABASES_DIR = os.path.abspath(os.path.join(FILE_DIR, u"../data/upgrade_databases/"))
+
+    def write_data_to_file(self, file_name):
+        with open(file_name, 'w') as file:
+            file.write("lorem ipsum")
+            file.close()
+
+    def setUpPreSession(self):
```
whirm: I don't like having copies of these functions around.

@pimveldhuisen is updating the base one, and you are not getting the changes here... Is there a good reason not to use TestAsServer?

Tribler/Test/test_as_server.py
```diff
@@ -149,6 +153,16 @@ def tearDown(self, annotate=True):
             self.watchdog.print_all_stacks()
             raise RuntimeError("Couldn't stop the WatchDog")
 
+        if self.file_server:
+            deferred = self.file_server.stopListening()
```
whirm: Quick note so you guys know: addCallback/errBack returns the deferred, so these three lines can be reduced to:

```
return self.file_server.stopListening().addCallback(self.checkReactor)
```

Tribler/Test/Core/test_sqlitecachedbhandler.py
```diff
@@ -59,7 +59,6 @@ def setUp(self):
         self.sqlitedb.initialize()
         self.session.sqlite_db = self.sqlitedb
 
-    @blocking_call_on_reactor_thread
     def tearDown(self):
         self.sqlitedb.close()
```
whirm: Are all these threadsafe? (DB operations are done on the Twisted thread usually)

Tribler/Test/Core/Upgrade/upgrade_base.py
devos50: Ah apparently I don't need the `setUpPreSession` method at all since I'm only creating a `Session` object, not starting it. I will remove the method.

Tribler/Test/test_as_server.py
lfdversluis: Don't forget that the `stoplistening()` may return None if there wasn't a connection, so wrap it in a `maybeDeferred`.

whirm: good one

Tribler/Test/Core/test_sqlitecachedbhandler.py
devos50: The `blocking_call_on_reactor_thread` was a mistake and shouldn't be there. Closing the sqlite database is apparently something that can happen on the main thread: we are also doing it when closing Tribler (both when running from command line and when running the tests).

Tribler/Test/test_as_server.py
devos50: @lfdversluis thanks!

devos50:
@whirm implemented the feedback

Tribler/Test/Core/test_sqlitecachedbhandler.py
whirm: ok

devos50:
retest this please

Tribler/Test/test_my_channel.py
whirm: yes, or breakage will ensue when/if the port is busy (concurrent test runs, etc.)

whirm:
@devos50 https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/758/

devos50:
@whirm fixed, I wasn't waiting for the session to be started completely. We should really refactor the startup method when we remove the GUI from the current code...

devos50:
Why is jenkins testing a completely different PR?

devos50:
retest this please

devos50:
retest this please

whirm:
@devos50 where? It seems to be running the right one:

`GitHub pull request #2046 of commit da125a7bb98ba46127101fbb360ba00810a1068d, no merge conflicts.`

whirm:
@devos50 BTW, when in reviewing phase, leave the fixup commits until the review is done, otherwise it makes it hard fo find the latest changes that need reviewing.

Tribler/Test/test_as_server.py
```diff
@@ -130,13 +145,28 @@ def tearDown(self, annotate=True):
...
         self.watchdog.join(2)
         if self.watchdog.is_alive():
             self._logger.critical("The WatchDog didn't stop!")
             self.watchdog.print_all_stacks()
             raise RuntimeError("Couldn't stop the WatchDog")
 
+        if self.file_server:
+            return maybeDeferred(self.file_server.stopListening).addCallback(self.checkReactor)
+        else:
+            deferred = Deferred().addCallback(self.checkReactor)
```
whirm: you can do the same than in 164

Tribler/Test/test_my_channel.py
whirm: That's not yet fixed

Tribler/Test/test_as_server.py
devos50: does `callback(...)` also return the deferred?

Tribler/Test/test_my_channel.py
devos50: Oh oops, forgot this one :/, will fix it immediately

Tribler/Test/test_as_server.py
whirm: Nope, I didn't read properly.
Then it's even simpler: you  are adding a callback to a newly created deferred and then you fire it immediatelly which triggers the added callback and then return the deferred.
You can just do:

```
return succeed(self.checkReactor())
```

devos50: Nice trick! (I read in the docs that it requires a `callable` as parameter, so it has to be `succeeded(self.checkReactor)`).

devos50:
@whirm can I push my fixes?

lfdversluis:
@devos50 just use git add <files> and then git commit --fixup <sha of the commit you are fixing code of that made those changes> and you will do a fixup commit that will order itself \o/

see http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

devos50:
@lfdversluis Yeah I really have to get used to these using these fixup commits, thanks for the link.

whirm:
yes, of course, just remember to use fixups so I can just review those instead of having to scan through the whole codebase :)

devos50:
@whirm done, can you take a look?

devos50:
Tests are failing, I'm investigating why

devos50:
@whirm seems to work fine now. See my last two fixup commits

whirm:
looks great, squash and i'll merge

devos50:
@whirm done

devos50:
@whirm there are still issues with the tests, I have to fix them first :(

whirm:
ok

devos50:
retest this please

devos50:
I'm unable to reproduce this locally so I have to rely on Jenkins to find the issue. It appears that `test_download_from_file` hangs but I'm not sure why this happens...

whirm:
I saw the same failure in one of @lfdversluis tests, I don't think it's related to your PR

devos50:
Hmm ok then, in that case, can you merge it please? :) I will create an issue for it then

devos50:
Again tests do not seem to be stable, same error I experienced before, I will take a look

devos50:
@whirm found the issue I think, hold on :D

whirm:
This PR grew way too much. You should make smaller ones from now on.

devos50:
@whirm I totally agree with you and I will absolutely create way smaller PRs in the future. Then it will also be more manageable.

Regarding the test failure, I'm pretty sure it is due to a race condition here: https://github.com/Tribler/tribler/blob/devel/Tribler/Main/tribler_main.py#L615. The download is added again with safe seeding enabled just after shutdown. I will disable safe seeding on the download test to prevent this from happen. Still, we need a proper fix for this TODO.

whirm:
If this is called after shutdown, it means the threadpool (not twisted's but Tribler's) is not being stopped correctly and the delayed calls registered on it are not canceled.

devos50:
@whirm it is being stopped but too late. The `torrent_db` has been closed already and then the torrents gets readded to the downloads, performing a method invocation on `torrent_db` which has been closed, causing the error.

We could cancel all tasks on the thread pool earlier to prevent this error but I'm not sure what the consequences are and at this point, I rather do that in a separate PR. So for now, I propose to disable safe seeding on the `test_libtorrent_download` tests. Do you agree with this?

whirm:
My opinion (I've been thinking about doing that myself for a while now) is that Tribler's threadpool should be just gone,

I refactored it to be a shim around twisted's threadpool and TaskManager a while ago to be able to remove a bunch of unnecessary NIH code without breaking everything that used it. The next step should be to just remove it and refactor every component that uses it to use its own TaskManager instance to control its tasks. This way you completely avoid the issues we are having now.

Go ahead and disable safe seeding (and create a new ticket describing the issue at hand).

devos50:
@whirm done (see last commit) and tests are passing.

whirm:
At last! :D  The PR  needs some squashing though, there's stuff modified in the initial commits and then fixed in later commits.

devos50:
@whirm merged together the relevant commits (mostly commits regarding the upgrader tests), should be ok now.

PR #2052, opened 2016-03-24
Fix tunnel helper permissions problem

mmetro:
Unless you've given your tunnel_helper user/group special permissions, it can't create the /var/lib/tunnel_helper directory.


whirm:
That directory is the actual home of the `tunnel_server` user.
There's some WIP ansible roles to deploy tunnel helpers:  https://github.com/whirm/ansible_anontunnel_helper which will create/install everything needed. We need to find a place for all these and refer to them in the documentation.

Thanks!

mmetro:
Ok, makes sense.  I was creating the user and placing the unit manually

PR #2054, opened 2016-03-25
Refactor general excepts

lfdversluis:
A lot of general try excepts are in the wx code which I will not trace because 1. this will be deleted soon 2. most of them are impossible to find qua what they can raise.

Some code chunks are too big or calling C libraries for which no doc is enclosed.

### DIFF
--- a/Tribler/Category/Category.py
+++ b/Tribler/Category/Category.py
@@ -4,6 +4,7 @@
 import os
 import re
 import logging
+from ConfigParser import MissingSectionHeaderError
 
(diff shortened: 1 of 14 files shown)


devos50:
Nice. Note that the almost everything in `Tribler/Main` (except for `tribler_main.py`, `tribler.py` and maybe some other files) will be gone when the new GUI goes live.

lfdversluis:
Woops. Apparently I had a typo in the branch name. I close and reopen this PR with the right branch name as my git is pushing to that one.

PR #2062, opened 2016-03-31
Fixed a memory issue when running the tunnel helpers

### LINKED ISSUE #2000: Memory leak on the tunnel helper
Current `devel` tunnel helper is leaking memory.

Apparently it only happens if exit mode is enabled, (either that or the lower amount of users on that kind of helper is not making the leak so obvious).

UPDATE: the non-exit tunnel helpers are also leaking, just at about half the rate than the others.

### LINKED ISSUE #2083: Session is not closing when running the tunnel helpers
When running the tunnel helpers (`python Tribler/community/tunnel/main.py`), the session is not closed when I quit the application by typing `q`. It keeps spamming and ultimately crashes:

```
INFO    1460452115.55            main:185   ONEXIT Waiting for Session to shutdown, will wait for 18 more seconds
INFO    1460452116.55            main:185   ONEXIT Waiting for Session to shutdown, will wait for 17 more seconds
INFO    1460452117.55            main:185   ONEXIT Waiting for Session to shutdown, will wait for 16 more seconds
INFO    1460452118.55            main:185   ONEXIT Waiting for Session to shutdown, will wait for 15 more seconds
INFO    1460452119.55            main:185   ONEXIT Waiting for Session to shutdown, will wait for 14 more seconds
INFO    1460452120.56            main:185   ONEXIT Waiting for Session to shutdown, will wait for 13 more seconds
INFO    1460452121.56            main:185   ONEXIT Waiting for Session to shutdown, will wait for 12 more seconds
INFO    1460452122.56            main:185   ONEXIT Waiting for Session to shutdown, will wait for 11 more seconds
INFO    1460452123.56            main:185   ONEXIT Waiting for Session to shutdown, will wait for 10 more seconds
```

devos50:
After a day of memory investigation, I found the root cause of the memory usage problem in the tunnel helpers when they are running for a long time.

The problem is that the Twisted log observer is intercepting lower-level Twisted events. When Twisted starts, a `LimitedHistoryLogObserver` is created, however, the name is very misleading since the observed history is not limited at all! An entry is written to the log, every time something interesting happens in the Tunnel community (for instance, the reactor starts or stops listening on a specific port or the `TunnelExitSocket` protocol starts).

Searching on the internet, brought me to this issue: [here](http://twistedmatrix.com/trac/ticket/7841). A workaround has been implemented [here](https://github.com/plq/neurons/blob/417836a729706f2d1449109ccbb7d4a510227050/neurons/daemon/config.py#L623). I basically took their approach with some minor changes. For instance, I don't replay the logs in `LimitedHistoryLogObserver`.

The fact that the memory is increasing faster when the community is running as exit node can easily be explained since we are opening more tunnels thus writing more log entries.

If you’re interested in how I found this exactly, I’ve written a blog post about it [here](http://blog.code-up.nl/finding-python-memory-peaks-using-meliae/).

I also fixed another issue in the tunnel helper main script that prevented the session from closing, fixing #2083.

Fixes #2000, fixes #2083

### DIFF
--- a/Tribler/community/tunnel/main.py
+++ b/Tribler/community/tunnel/main.py
@@ -11,6 +11,7 @@ from collections import defaultdict, deque
 
 from twisted.internet.task import LoopingCall
 from twisted.internet.stdio import StandardIO
+from twisted.logger import globalLogPublisher
 from twisted.protocols.basic import LineReceiver
 from twisted.internet.threads import blockingCallFromThread
 
@@ -26,6 +27,8 @@ from Tribler.community.tunnel.hidden_community import HiddenTunnelCommunity
 import logging.config
 from Tribler.Core.simpledefs import dlstatus_strings
 from Tribler.dispersy.candidate import Candidate
+from Tribler.dispersy.tool.clean_observers import clean_twisted_observers
+
 logging.config.fileConfig("logger.conf")
 logger = logging.getLogger('TunnelMain')
 
@@ -84,6 +87,8 @@ class Tunnel(object):
         for k in self.crawl_message.keys():
             if now - 3600 > self.crawl_message[k]['time']:
                 self.crawl_message.pop(k)
+
+        clean_twisted_observers(globalLogPublisher)
 
     def build_history(self):
         self.history_stats.append(self.get_stats())
--- a/Tribler/community/tunnel/main.py
+++ b/Tribler/community/tunnel/main.py
@@ -62,6 +62,7 @@ class Tunnel(object):
         Tunnel.__single = self
 
         self.settings = settings
+        self.should_run = True
         self.crawl_keypair_filename = crawl_keypair_filename
         self.dispersy_port = dispersy_port
         self.crawl_data = defaultdict(lambda: [])
@@ -163,9 +164,6 @@ class Tunnel(object):
         return (4.0, [])
 
     def stop(self):
-        self.session.lm.threadpool.call(0, self._stop)
-
-    def _stop(self):
         if self.clean_messages_lc:
             self.clean_messages_lc.stop()
             self.clean_messages_lc = None
@@ -328,8 +326,7 @@ class LineHandler(LineReceiver):
             anon_tunnel.session.lm.threadpool.call(0, start_download)
 
         elif line == 'q':
-            anon_tunnel.stop()
-            return
+            anon_tunnel.should_run = False
 
         elif line == 'r':
             print "circuit\t\t\tdirection\tcircuit\t\t\tTraffic (MB)"
@@ -399,8 +396,10 @@ def main(argv):
         cherrypy.config.update({'server.socket_host': '0.0.0.0', 'server.socket_port': args.json})
         cherrypy.quickstart(tunnel)
     else:
-        while True:
+        while tunnel.should_run:
             time.sleep(1)
+
+    tunnel.stop()
 
 if __name__ == "__main__":
     main(sys.argv[1:])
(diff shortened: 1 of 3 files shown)


devos50:
@whirm ready for review

whirm:
I'd say this should be done as part of the startup process for any process (not just the tunnel helper) The tracker service will probably be affected too.

whirm:
I would move it to dispersy so the Tracker twistd plugin can use it too and call it from all the twistd plugins in both dispersy and Tribler and from the logger setup function in Tribler.

devos50:
@whirm that's ok, I can move it to Dispersy. Can you point out what's the best place to have this code in Dispersy?

whirm:
I guess in `tool/`... dispersy itself won't be using it directly, only the dispersy tracker twistd plugin.

devos50:
So we have this code in two places if I understand it correctly? In the tunnel helpers and the tracker twistd plugin in Dispersy?

whirm:
No, the twistd plugin will be importing and calling the function, and as Tribler depends on dispersy, anything that needs to do the same on the Tribler repo can import and call the function too.

devos50:
@whirm shall I open a new PR for calling the relevant code in dispersy?

whirm:
nah, do it in this one

whirm:
hm, you forgot to move the test to dispersy

devos50:
@whirm see https://github.com/Tribler/dispersy/pull/466

devos50:
@whirm I've updated this PR with the relevant changes.

Tribler/community/tunnel/main.py
```diff
@@ -11,6 +11,7 @@
 
 from twisted.internet.task import LoopingCall
 from twisted.internet.stdio import StandardIO
+from twisted.logger import globalLogPublisher
```
whirm: This should be gone

devos50: @whirm `globalLogPublisher` is passed as root log object to the `clean_twisted_observers` method.

Tribler/dispersy
```diff
@@ -1 +1 @@
-Subproject commit d056b388836cec227f6752683045534325b5b5e4
+Subproject commit 67c22ed23f0e079d775cbe6df05d907b1282fc24
```
whirm: I merged the test, so that should be 57ef0d210d98f2d9e41fca01fb48ef355efacf3a

Tribler/community/tunnel/main.py
```diff
@@ -85,6 +88,8 @@ def clean_messages(self):
             if now - 3600 > self.crawl_message[k]['time']:
                 self.crawl_message.pop(k)
 
+        clean_twisted_observers(globalLogPublisher)
```
whirm: Oh, I just noticed that you keep passing the type to the function, it's a bit weird, but let's not send yet another PR to dispersy for that.

devos50:
@whirm I will write a test tomorrow that starts and closes the tunnel helpers to see whether they are working. This also makes sure that the uncovered code is covered.

whirm:
Nice!

synctext:
Twisted has bad behavior or bug if it keeps 65k lines in memory. Even if you overwrite it.

Is an upstream fix possible? Lots of time?

devos50:
@whirm ready for another review. Also, should I really cover that single line in the `LineReceiver` protocol?

whirm:
@devos50 it's fine. I'll review it ASAP

PR #2063, opened 2016-04-01
Nsis fix uninstall location

### LINKED ISSUE #1664: Issue: During installation of Tribler after having a prior biuld installed it says it'll delete from a default directory even though that's not where Tribler is installed.
Basically, on my W10 computer, i have Tribler installed in a folder under disk D somewhere. When i try to install a build it would say it's going to delete the old from from "C:\Program Files (x86)\Tribler" Which is not the location, or even close. I'm not sure which effects this could have on the program, if another installation is made too. It's says it's going to remove rc-5(installing a more recent build) which is my current one, and i'm not sure if it shows it wrong or not. I'd just delete my installation from the correct folder manually but i'm not sure if there are other files that should stay in the folder or if there are files elsewhere(like in the roaming folder) that are supposed to be affected, and probably are. Since you can't choose where it'll uninstall you kind of are stuck. 

Also the folder it says it'll delete from does not exist. And after running the remover of it, the folder remains in D which means nothing has changed, and i can still use Tribler even though i just ran the remover which popped up after trying to install the newer build. 

There is a simple way to circumvent, and that's simply using the uninstall file in the folder itself which will target the right place.

lfdversluis:
Will now read the uninstallstring in the register on windows and use that location, rather than the $INSTDIR of the installer which is always the default location at that point.

Fixes #1664

### DIFF
--- a/Tribler/Main/Build/Win/tribler.nsi
+++ b/Tribler/Main/Build/Win/tribler.nsi
@@ -365,7 +365,8 @@ Function .onInit
 
   uninst:
     ClearErrors
-    ExecWait '$R0 _?=$INSTDIR' ;Do not copy the uninstaller to a temp file
+    ReadRegStr $UNINSTDIR HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT}" "UninstallString"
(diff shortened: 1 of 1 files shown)


lfdversluis:
@whirm Ready to review/merge

whirm:
Great!

Issue #1038, opened 2014-12-19
Enhancement Request - Watch Folder


priceaj:
Please implement a watch folder function i.e. the client polls a folder for torrent files and adds when found, so that tribler can interface with automated torrent downloaders.

The variables set to the download i.e. Download Location/Anonymity etc should be set from within the settings interface and applied to the torrents found in the watch folder automatically

Dmole:
+1, also note you can use 
inoticoming/incron on linux,
FindFirstChangeNotification/FileSystemWatcher on windows,
and kqueue/FSEvents on OS X.

Transcend01:
+1 trhis is a must have.

Mr-Obelix:
(almost)unusable without.
+1

whirm:
@synctext apparently there's a lot of interest in this. Should we implement this for 6.6?

Transcend01:
It would be great for a lot of people. There a lot of program that 's use black hole or watch folder. It would be really great if you guys could imlement this for 6.6.

Mr-Obelix:
+1 for 6.6!

synctext:
@Mr-Obelix  @Transcend01 @Dmole @k4n30 @priceaj 
Would this design work for you?

_A new settings option is added "watch folder for .torrent files and automatically download them".
This feature can be enabled with a toggle box. Additional setting is the "watch folder" location. The anonymity setting is used from other anon settings menu._

@whirm if this would lot take more then an afternoon, it seems well spend time. 
Several fans of this feature clearly.

Transcend01:
This sounds great !!! I think a lot of people will be looking forward to the new version.

k4n30:
Sounds fantastic :)

Transcend01:
When can we expect the 6.6 version or it's first beta's ?

whirm:
@Transcend01 I'm planning on releasing V6.5.1 (bugfix release) really soon and shortly after V6.6-alpha1 to receive some early feedback on the migration to WX3 on the GUI side. The alpha will still not have this issue's feature implemented yet probably, but it should come shortly after.

Transcend01:
@whirm thx for your fast reply.

devos50:
@whirm I think this is a very nice addition in the new Qt GUI.

whirm:
That's not really gui related, the only gui related part would be the dir selector on the settings panel and a checkbox to enable the feature.

Dmole:
[...]

PR #2065, opened 2016-04-02
Implemented the watch folder functionality

devos50:
As requested in #1038, I've implemented the watch folder functionality.

I've created a separate subclass of `TaskManager` that polls the watch folder functionality every two minutes and adds downloads of (new) torrent files it finds in this directory. I'm aware that there are several directory watching libraries available in Python (for instance, [watchdog](https://pypi.python.org/pypi/watchdog)) but I don't want to add another dependency to Tribler (but this is open for discussion).

The watch folder can be enabled in settings -> general and the path of the watch directory can be adjusted.

### DIFF
--- a/Tribler/Core/Modules/watch_folder.py
+++ b/Tribler/Core/Modules/watch_folder.py
@@ -1,8 +1,10 @@
+import hashlib
 import logging
 import os
 from twisted.internet.task import LoopingCall
 from Tribler.Core.TorrentDef import TorrentDef
 from Tribler.Core.Utilities.utilities import fix_torrent
+from Tribler.Core.simpledefs import NTFY_WATCH_CORRUPT_FOLDER, NTFY_INSERT
 from Tribler.dispersy.taskmanager import TaskManager
 
 
@@ -18,7 +20,8 @@ class WatchFolder(TaskManager):
         self.session = session
 
     def start(self):
-        self.register_task("check watch dir", LoopingCall(self.check_watch_folder)).start(WATCH_FOLDER_CHECK_INTERVAL)
+        self.register_task("check watch folder", LoopingCall(self.check_watch_folder))\
+            .start(WATCH_FOLDER_CHECK_INTERVAL, now=False)
 
     def stop(self):
         self.cancel_all_pending_tasks()
@@ -34,6 +37,9 @@ class WatchFolder(TaskManager):
 
                 torrent_data = fix_torrent(os.path.join(root, name))
                 if not torrent_data:  # torrent appears to be corrupt
+                    os.rename(os.path.join(root, name), os.path.join(root, name + ".corrupt"))
+                    self._logger.warning("Watch folder - corrupt torrent file %s", name)
+                    self.session.notifier.notify(NTFY_WATCH_CORRUPT_FOLDER, NTFY_INSERT, None, name)
                     continue
 
                 tdef = TorrentDef.load_from_memory(torrent_data)
--- a/Tribler/Core/Modules/watch_folder.py
+++ b/Tribler/Core/Modules/watch_folder.py
@@ -1,0 +1,44 @@
+import logging
+import os
+from twisted.internet.task import LoopingCall
+from Tribler.Core.TorrentDef import TorrentDef
+from Tribler.Core.Utilities.utilities import fix_torrent
+from Tribler.dispersy.taskmanager import TaskManager
+
+
+WATCH_FOLDER_CHECK_INTERVAL = 10
+
+
+class WatchFolder(TaskManager):
+
+    def __init__(self, session):
+        super(WatchFolder, self).__init__()
+
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+    def start(self):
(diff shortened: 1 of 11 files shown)


devos50:
@whirm ready for review (I don't think it's worth the time to hunt this last coverage in the GUI code)

Tribler/Core/defaults.py
```diff
@@ -137,6 +137,10 @@
 sessdefaults['upgrader'] = OrderedDict()
 sessdefaults['upgrader']['enabled'] = True
 
+# Watch folder config
```
whirm: If you alter the config, you need to update the config version number

Tribler/Core/Modules/watch_folder.py
```diff
@@ -0,0 +1,46 @@
...
+            return
+
+        for root, _, files in os.walk(self.session.get_watch_folder_path()):
+            for name in files:
+                if not name.endswith(u".torrent"):
+                    continue
+
+                torrent_data = fix_torrent(os.path.join(root, name))
+                if not torrent_data:  # torrent appears to be corrupt
+                    continue
```
whirm: We should log this case and maybe even show a notification?

Tribler/Core/Modules/watch_folder.py
```diff
@@ -0,0 +1,46 @@
+import logging
+import os
+from twisted.internet.task import LoopingCall
+from Tribler.Core.TorrentDef import TorrentDef
+from Tribler.Core.Utilities.utilities import fix_torrent
+from Tribler.dispersy.taskmanager import TaskManager
+
+
+WATCH_FOLDER_CHECK_INTERVAL = 120
```
whirm: Isn't two minutes a lot? It's not like this is an expensive operation. Maybe make it 10 secs?

devos50:
@whirm a dialog is now shown in the GUI if a corrupt torrent file is found in the watch folder. I maintain a list of sha256 hashes of the files to make sure that the corrupt torrent is only reported once in the logs and to the user (otherwise, the user will be spammed with dialogs). I also decreased the looping call interval time to 15 seconds and increased the config version.

whirm:
@devos50 what about adding the `.corrupt` extension to corrupt torrent files instead of keeping a list? this way they will be ignored even across restarts.

devos50:
retest this please

devos50:
@whirm good suggestion, the .corrupt extension is now added!

devos50:
retest this please

devos50:
@whirm this one is also ready, not planning on increasing wx coverage.

Tribler/Core/Modules/watch_folder.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, session):
+        super(WatchFolder, self).__init__()
+
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+        self.watch_folder_check_task = None
+
+    def start(self):
+        self.watch_folder_check_task = self.register_task("check watch folder", LoopingCall(self.check_watch_folder))
```
whirm: You don't need to keep a reference to this task, just register_task(foo).start(bar) and keep it simple

Tribler/Core/simpledefs.py
```diff
@@ -87,6 +87,7 @@
 NTFY_ACTIVITIES = 'activities'  # an activity was set (peer met/dns resolved)
 NTFY_REACHABLE = 'reachable'  # the Session is reachable from the Internet
 NTFY_DISPERSY = 'dispersy'  # an notification regarding dispersy
+NTFY_CORRUPT_TORRENT = 'corrupt_torrent'  # a corrupt torrent has been found in the watch folder
```
whirm: I would call this NTFY_WATCH_CORRUPT_FOLDER to avoid name clashes in the future

Tribler/Main/vwxGUI/settingsDialog.py
```diff
@@ -341,6 +350,15 @@ def BrowseClicked(self, event=None):
         else:
             pass
 
+    def BrowseWatchFolderClicked(self, event=None):
+        dlg = wx.DirDialog(None, "Choose watch folder", style=wx.DEFAULT_DIALOG_STYLE)
+        dlg.SetPath(self.utility.session.get_watch_folder_path())
+        if dlg.ShowModal() == wx.ID_OK:
+            self._wf_location_ctrl.SetForegroundColour(wx.BLACK)
+            self._wf_location_ctrl.SetValue(dlg.GetPath())
+        else:
```
whirm: This can go away

Tribler/Main/vwxGUI/settingsDialog.py
```diff
@@ -448,6 +466,23 @@ def __create_s1(self, tree_root, sizer):
         gp_s2_sizer.Add(self._default_anonymous_label, 0, wx.EXPAND)
         gp_s2_sizer.Add(self._default_anonymity_dialog, 0, wx.EXPAND)
 
+        # Watch folder
```
whirm: I would add a description about what a watch folder is

whirm:
Other than the remarks, looks great to me!

devos50:
@whirm ready for another review

PR #2069, opened 2016-04-04
Reorganize data files for tests

### LINKED ISSUE #1984: We have the same video file four times in the Test directory
We have the same video four times in the tests directory:
- `Tribler/Test/data/video.avi`
- `Tribler/Test/API/video.avi`
- `Tribler/Test/API/video2.avi`
- `Tribler/Test/API/contentdir/video.avi`

We should rewrite these tests so this file is only available once. If we need to use the file in another location, we should make use of temporary directories.

pimveldhuisen:
Move data files to the Test/data directory and remove duplicate files.
This fixes #1984


devos50:
@pimveldhuisen there are some issues with coverage regarding files in the `Tribler/Test` module (the tests themselves are not covered by the coverage tool so any change in the tests will be marked as uncovered). I will discuss this with @whirm when he's back.

lfdversluis:
Wait, we are not going to cover our tests with tests that test the tests? :laughing: Yeah it should be excluded. Although.. our testing infrastructure currently depends on testAsServer et al. oh well we'll discuss it.

pimveldhuisen:
I think the fact that the coverage report includes the tests should in itself not be a problem, as the tests should themselves be run. Here however I changed a line in a test that was (before and after the PR) set to be skipped, which means it is not run, thus not covered.

whirm:
@pimveldhuisen it's fine if a line it's not covered for a good reason.

whirm:
Looks great!

PR #2070, opened 2016-04-04
Update files() call on torrent_info

ardhipoetra:
Since libtorrent 1.0.9 (or maybe earlier), function `files()` on `torrent_info` object returned a `file_storage` object, instead of dictionary. I think it's safe to update the function usage to a new one.

This PR will fix that problem.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -236,6 +236,7 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
             torrentinfo = lt.torrent_info(metainfo)
 
             self.orig_files = [file_entry.path.decode('utf-8') for file_entry in torrentinfo.files()]
+
             is_multifile = len(self.orig_files) > 1
             commonprefix = os.path.commonprefix(self.orig_files) if is_multifile else ''
             swarmname = commonprefix.partition(os.path.sep)[0]
@@ -598,7 +599,14 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
                     filepriorities.append(0)
                     new_path = os.path.join(unwanteddir, '%s%d' % (hexlify(self.tdef.get_infohash()), index))
 
-                cur_path = get_info_from_handle(self.handle).files()[index].path.decode('utf-8')
+                torrent_storage = get_info_from_handle(self.handle).files()
+
+                # TODO(ardhi) : as from 1.0, files returning file_storage (lazy-iterable)
+                if hasattr(lt, 'file_storage') and isinstance(torrent_storage, lt.file_storage):
+                    cur_path = torrent_storage.at(index).path.decode('utf-8')
+                else:
+                    cur_path = get_info_from_handle(self.handle).files()[index].path.decode('utf-8')
+
                 if cur_path != new_path:
                     if not os.path.exists(unwanteddir_abs) and unwanteddir in new_path:
                         try:
@@ -949,6 +957,7 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
         """
 
         dest_files = []
+
         for index, file_entry in enumerate(get_info_from_handle(self.handle).files()):
             if self.handle.file_priority(index) > 0:
                 filename = file_entry.path


lfdversluis:
@ardhipoetra check which version of libtorrent are currently shipped with debian and ubuntu. If those versions also return this `torrent_info`  object it is great. If not then the standard repos are not yet up-to-date and as we rely on those during installation it will lead to a crash.

ardhipoetra:
@lfdversluis, the libtorrent version of ubuntu in 14.04 is (very) old : 0.16.13. Should we stick on this version?

Btw, I reported some backward compatibility in the libtorrent part as seen in arvidn/libtorrent#582 and arvidn/libtorrent#590. Hopefully they will handle those incompatibilities :)

synctext:
Source Package: libtorrent-rasterbar (1.0.7-1build1) [universe]
in upcoming 16.04 LTS of Ubuntu: http://packages.ubuntu.com/source/xenial/libtorrent-rasterbar

lfdversluis:
Even in 1.0.7 the torrent_info file is not introduced afaik. If this is patched backwards compatible we can do this, else you have to make a check that handles both versions.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -235,7 +235,12 @@ def network_create_engine_wrapper(self, lm_network_engine_wrapper_created_callba
             metainfo = self.tdef.get_metainfo()
             torrentinfo = lt.torrent_info(metainfo)
 
-            self.orig_files = [file_entry.path.decode('utf-8') for file_entry in torrentinfo.files()]
+            try:
```
whirm: Use isinstance instead of a try except block.

Also add a TODO note about the version that changes this so we know when the workaround can be removed.

Tribler/dispersy
```diff
@@ -1 +1 @@
-Subproject commit d056b388836cec227f6752683045534325b5b5e4
+Subproject commit ab39486c5572c1e613ea5d8a5fcac9d61d56f802
```
whirm: Why are you updating the dispersy commit?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -598,7 +602,15 @@ def set_selected_files(self, selected_files=None):
                     filepriorities.append(0)
                     new_path = os.path.join(unwanteddir, '%s%d' % (hexlify(self.tdef.get_infohash()), index))
 
-                cur_path = get_info_from_handle(self.handle).files()[index].path.decode('utf-8')
+                cur_path = None
```
whirm: This is redundant

Tribler/dispersy
ardhipoetra: ouch! I don't know :/

any way to revert it back?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -460,7 +461,10 @@ def on_metadata_received_alert(self, alert):
                 self.metadata["announce"] = trackers[0]
 
         self.tdef = TorrentDef.load_from_dict(self.metadata)
+
+        # files() is iterable
```
whirm: Please, don't add white-space modifications or irrelevant commits to the PR, it will cause unnecessary merge conflicts to others.

Tribler/dispersy
whirm: Sure, go back to the right commit, in the submodule and commit. When you squash everything together, it will be gone.

whirm:
Other than the existing remarks, it looks fine. Fix those and squash into a single commit and I'll merge it.

whirm:
You missed the two merge commits

ardhipoetra:
@whirm I already squashed all the commits. I wonder why I still get some error in `coverage_diff`and `pylint. I don't understand the error message mentioned in [coverage report](https://jenkins.tribler.org/job/GH_Tribler_PR_coverage_check/572//artifact/coverage_report.html) and [quality report](https://jenkins.tribler.org/job/GH_Tribler_PR_pylint_check/678//artifact/output/quality_report.html)

devos50:
@ardhipoetra these are new PR checkers, recently created. The `diff_coverage` error means that you have modified lines that are not covered by any tests. You should write a (unit) test to cover the line in violation.

The pylint checkers checks for PEP8 guideline violations introduced by your PR. In your case, you have a comment containing `todo` which we should not have. Hope this helps you!

lfdversluis:
@devos50 I think that TODO is fine, like @whirm mentioned. It allows us to keep track of when we can safely remove the version check once all repos contain a suitable version of libtorrent so that we no longer need to workaround methods not being present.

But overall you should fix PEP8 violations so our code will become better in quality!
Also @ardhipoetra note that sometimes when you modify a line in code that is not yours it may be also marked as not covered. In this case it's old code that never was tested and it should be covered. I and others already found bugs in old code that have been around there for years. Time to find them.

whirm:
@ardhipoetra, @devos50: @lfdversluis is spot on on his explanation :)

whirm:
@ardhipoetra as testing the non-covered line would imply having both old and recent versions of libtorrent around, I'll merge it if you confirm you have tested that part of the code with them.

ardhipoetra:
Thanks for the input, @devos50 @lfdversluis! I do agree that `TODO` tags is beneficial in its own way, maybe there are some other place that need to be fixed, similar to this case.

@whirm, I have tested this build both in `0.16.13.0` (the _stable_ one in the 14.04 repository), and `1.1.0.0` (the most recent one). And both work.

PR #2071, opened 2016-04-04
Fixed wrong status when downloading

### LINKED ISSUE #1702: v6.5.0-rc6 Wrong status is displayed for a file being downloaded
I see `Status: Torrent file is being downloaded from the DHT (10.2%).`

It should be `Downloading (10.2%)`

### LINKED ISSUE #2089: Downloads are not cleaned properly
While investigating why the download status is broken, I found that while having `0` downloads, I get the following lines:

```
length of magnetlist: 0
self.torrent.magnetstatus: None
self.torrent.magnetstatus: None
```

Where the `self.torrent.magnetstatus: None` is in the `UpdateStatus` function in the `list_details.py` file. There should not be a function call as there are no downloads present.

lfdversluis:
The problem is that `self.torrent.magnetstatus` will always be **not** None when adding from a magnet link, even when the .torrent has been downloaded. This will trigger the first if statement and set the status wrong. So now the DL status takes priority over the magnetstatus.

I also tried removing the `if self.torrent.magnetstatus` check which seems to work, but as I do not understand the logic behind [this undocumented function](https://github.com/Tribler/tribler/blob/devel/Tribler/Main/Utility/GuiDBTuples.py#L215) I decided to go with the current approach. Confirmed by removing + data and readding. Now changes immediately.

# edit
Thinking about it, it makes no sense. It switched to seeding just fine but it should not if that magnetstatus list is not None. I'll delete this magnetstatus object from the if statement as I observed no issue with removing it.

Fixes #1702, Fixes #2089 
Fixes **2** bugs
Removed some duplicate code and unused variables (enhancement)

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -122,7 +122,6 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
         self.dlstate = DLSTATUS_WAITING4HASHCHECK
         self.length = 0
         self.progress = 0.0
-        self.bufferprogress = 0.0
         self.curspeeds = {DOWNLOAD: 0.0, UPLOAD: 0.0}  # bytes/s
         self.all_time_upload = 0.0
         self.all_time_download = 0.0
--- a/Tribler/Main/vwxGUI/SearchGridManager.py
+++ b/Tribler/Main/vwxGUI/SearchGridManager.py
@@ -579,6 +579,7 @@ class LibraryManager(object):
         return self.wantpeers
 
     def magnet_started(self, infohash):
+        # [time, amount of peers, ????]
         self.magnetlist[infohash] = [long(time()), 0, 0]
 
     def magnet_got_peers(self, infohash, total_peers):
@@ -612,6 +613,12 @@ class LibraryManager(object):
             self.gui_callback.remove(callback)
 
     def set_want_peers(self, hashes, enable=True):
+        """
+        Sets whether or now we want hashes, appending or removing them
+        from the wantpeers list.
+        :param hashes: A list of hashes we want to add or remove from the wantpeers list
+        :param enable: A boolean indicating if they should be removed or added
+        """
         if not enable:
             for h in hashes:
                 if h in self.wantpeers:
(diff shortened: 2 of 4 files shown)


lfdversluis:
retest this please

lfdversluis:
@whirm ready when you are

whirm:
retest this please

whirm:
@lfdversluis a quick glance shows me that it contains a tuple of (timestamp, peers, somethingelse) and it gets updated via some events. So I would rather have it fixed properly.

lfdversluis:
The object never becomes none, so that first if statement will always trigger. From my observation the object is also never changed/updated.

whirm:
@lfdversluis Then we should check why it's not updated. I'd guess it used to and that's why it's broken now?

lfdversluis:
I think it's ok not to include the magnetstatus here. The `LibtorrentDownloadImpl` class handles the case of when the metadata is received in its `on_metadata_received_alert` function and sets the DLSTATE accordingly. Also, the metadata query is fetched in the `get_metainfo` function in `LibtorrentMgr`.

lfdversluis:
As I see little use of the magnetstatus object, it is used in several places so I am not fully confident we can delete it yet. But I am quite certain that removing it in the list_details can be done.

lfdversluis:
I also still havent been able to find out what that third item in the list is supposed to be. But I do know this stuff comes from the SVN era

lfdversluis:
I think the third item is never assigned at all anymore.

whirm:
Ok, let me know when it's ready

lfdversluis:
@whirm Since the prints were messing up, I squashed it to 1 commit. Want me to hunk it into several (2 for each bug and 1 for the general refactor thingies) ?

whirm:
Leave it this time, but from now use DROPME commits for prints and suchlike so this doesn't happen.

lfdversluis:
huh does the git commit --dropme exists?

lfdversluis:
At one point I had some refactoring things combined with prints, so dropping that commit wasn't an option. Next time I'll commit the print(s), then commit again

lfdversluis:
```
watchog twisted thread *******TRIPPED!******, hasn't been set for 10.0145 secs.

*** STACKTRACE - START ***


### ThreadID: 140543141082880 Thread name: torrent_checker
```

The torrentchecker tripped.

lfdversluis:
Also `Exception: Test raised 1 unhandled exceptions, last one was: AssertionError('no channels matching de',)`

lfdversluis:
retest this please

whirm:
@lfdversluis no, you prepend your commit message (which only contains throw-away debug stuff) with DROPME and you kill those commits when doing the final rebase.

lfdversluis:
@whirm Oh ok. I actually had one of them, but I had refactoring and debug stuff in 1 commit.

lfdversluis:
```
Tribler.Test.test_i2i_download.TestI2IDownload.test_i2i_running (from nosetests)
Error Message
Did not even create a session
```

lfdversluis:
retest this please

Issue #14, opened 2013-03-08
New version check from Tribler


whirm:
We should add a way to check if there's a newer Tribler release. At least for when it crashes so we don't get reports for possibly fixed bugs.

whirm:
@synctext, @LipuFei I think we should start updating the version index again what do you guys say if we fix it for 6.5?

NielsZeilemaker:
If you're going to change it, please upgrade it to https

ThomasdenH:
Wouldn't it make sense to have an 'official' Tribler channel that uploaded all versions of Tribler? This way the updating would also be impossible to shut down.

whirm:
@ThomasdenH Yeah, that would be cool. But it would take more work too...

devos50:
Some work has been done on this issue in #2072 and it will be completed when the new GUI goes live in #2153.

PR #2072, opened 2016-04-05
Added a version check manager

devos50:
Github has a nice API endpoint that can be used to fetch the latest release: `https://api.github.com/repos/tribler/tribler/releases`. This gives the latest stable release.

I propose to add a version check manager that can schedule version checks in the future. This can for instance be called when the GUI starts.

I'm not really sure about the best way to integrate this into the GUI but we can discuss this. This is just a quick implementation of a version check manager that should replace `version_check.py` in `Main/Utility`.

This PR contains work on #14.

### DIFF
--- a/Tribler/Core/Modules/versioncheck_manager.py
+++ b/Tribler/Core/Modules/versioncheck_manager.py
@@ -1,0 +1,53 @@
+from distutils.version import LooseVersion
+import json
+import logging
+from twisted.internet.error import ConnectError, DNSLookupError
+from twisted.internet.task import LoopingCall
+from twisted.web.client import Agent, readBody
+from twisted.web.error import SchemeNotSupported
+from twisted.web.http_headers import Headers
+from Tribler.Core.Utilities.twisted_thread import reactor
+from Tribler.Core.simpledefs import NTFY_INSERT, NTFY_NEW_VERSION
+from Tribler.Core.version import version_id
+from Tribler.dispersy.taskmanager import TaskManager
+
+
+VERSION_CHECK_URL = 'https://api.github.com/repos/tribler/tribler/releases/latest'
+
+
+class VersionCheckManager(TaskManager):
+
+    def __init__(self, session):
+        super(VersionCheckManager, self).__init__()
+
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+    def start(self, interval):
+        self.register_task("tribler version check", LoopingCall(self.check_new_version)).start(interval)
+
+    def stop(self):
+        self.cancel_all_pending_tasks()
+
+    def check_new_version(self):
+
+        def parse_body(body):
+            if body is None:
+                return
+            version = json.loads(body)['name'][1:]
+            if LooseVersion(version) > LooseVersion(version_id):
+                self.session.notifier.notify(NTFY_NEW_VERSION, NTFY_INSERT, None, version)
+
+        def on_request_error(error):
+            error.trap(SchemeNotSupported, ConnectError, DNSLookupError)
+            self._logger.error("Error when performing version check request: %s", error)
+
+        def parse_response(response):
+            if response.code == 200:
+                return readBody(response)
+            else:
+                self._logger.warning("Got response code %s when performing version check request", response.code)
+
+        agent = Agent(reactor)
+        return agent.request('GET', VERSION_CHECK_URL, Headers({'User-Agent': ['Tribler ' + version_id]}), None)\
+            .addCallback(parse_response).addCallback(parse_body).addErrback(on_request_error)
--- a/Tribler/Main/Dialogs/NewVersionDialog.py
+++ b/Tribler/Main/Dialogs/NewVersionDialog.py
@@ -1,0 +1,58 @@
+import wx
+from Tribler.Main.vwxGUI.GuiUtility import GUIUtility
+from Tribler.Main.vwxGUI.widgets import _set_font
+
+
+class NewVersionDialog(wx.Dialog):
+
+    def __init__(self, version, parent, name, msg_bold='', msg='', title='', center_on_frame=True):
+        wx.Dialog.__init__(self, parent=parent, size=(475, 210), name=name)
+
+        self.version = version
+        self.SetTitle(title)
+        messageSizer = wx.BoxSizer(wx.VERTICAL)
+
+        if msg_bold:
+            messageText1 = wx.StaticText(self, label=msg_bold)
+            _set_font(messageText1, fontweight=wx.FONTWEIGHT_BOLD)
+            messageSizer.Add(messageText1, 1, wx.EXPAND)
+        if msg:
+            messageText2 = wx.StaticText(self, label=msg)
+            messageSizer.Add(messageText2, 1, wx.EXPAND | wx.TOP, 10 if msg_bold else 0)
+
+        bodySizer = wx.BoxSizer(wx.HORIZONTAL)
+        bodySizer.Add(wx.StaticBitmap(self, -1, wx.ArtProvider.GetBitmap(
+            wx.ART_INFORMATION, wx.ART_CMN_DIALOG)), 0, wx.ALIGN_TOP | wx.RIGHT, 25)
+        bodySizer.Add(messageSizer, 1, wx.EXPAND)
+
+        buttonSizer = wx.StdDialogButtonSizer()
+        okButton = wx.Button(self, wx.ID_OK, label='Ok')
+        okButton.Bind(wx.EVT_BUTTON, self.OnOk)
+        laterButton = wx.Button(self, id=wx.ID_CANCEL, label='Later')
+        laterButton.Bind(wx.EVT_BUTTON, self.OnLater)
+        ignoreButton = wx.Button(self, id=wx.ID_IGNORE, label='Ignore')
+        ignoreButton.Bind(wx.EVT_BUTTON, self.OnIgnore)
+        buttonSizer.Add(ignoreButton)
+        buttonSizer.Add(laterButton)
+        buttonSizer.Add(okButton)
+        buttonSizer.Realize()
+
+        mainSizer = wx.BoxSizer(wx.VERTICAL)
+        mainSizer.Add(bodySizer, 1, wx.EXPAND | wx.ALL, 10)
+        mainSizer.Add(buttonSizer, 0, wx.ALIGN_RIGHT | wx.ALL, 10)
+        self.SetSizerAndFit(mainSizer)
+        if center_on_frame:
+            x, y, w, h = GUIUtility.getInstance().frame.GetScreenRect()
+            self.SetPosition((x + ((w - self.GetSize().x) / 2), y + ((h - self.GetSize().y) / 2)))
+
+    def OnOk(self, event):
+        import webbrowser
+        webbrowser.open("https://tribler.org")
+        self.EndModal(wx.ID_OK)
+
+    def OnLater(self, event):
+        self.EndModal(wx.ID_CANCEL)
+
+    def OnIgnore(self, event):
+        GUIUtility.getInstance().utility.write_config('last_reported_version', str(self.version))
+        self.EndModal(wx.ID_IGNORE)
--- a/Tribler/Test/Core/Modules/test_versioncheck.py
+++ b/Tribler/Test/Core/Modules/test_versioncheck.py
@@ -1,0 +1,72 @@
+import json
+from twisted.internet.defer import Deferred, maybeDeferred
+from twisted.web import server, resource
+from Tribler.Core.Modules import versioncheck_manager
+from Tribler.Core.Utilities.network_utils import get_random_port
+from Tribler.Core.Utilities.twisted_thread import reactor, deferred
+from Tribler.Test.test_as_server import TestAsServer
+from Tribler.dispersy.util import call_on_reactor_thread
+
+
+class VersionResource(resource.Resource):
+
+    isLeaf = True
+
+    def __init__(self, response, response_code):
+        resource.Resource.__init__(self)
+        self.response = response
+        self.response_code = response_code
+
+    def render_GET(self, request):
+        request.setResponseCode(self.response_code)
+        return self.response
+
+
+class TestVersionCheck(TestAsServer):
+
+    def setUp(self):
+        self.port = get_random_port()
+        self.server = None
+        self.should_call_new_version_callback = False
+        self.new_version_called = False
+        versioncheck_manager.VERSION_CHECK_URL = 'http://localhost:%s' % self.port
+        super(TestVersionCheck, self).setUp()
+
+        self.session.notifier.notify = self.notifier_callback
+
+    def notifier_callback(self, subject, changeType, obj_id, *args):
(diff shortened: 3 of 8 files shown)


Tribler/Core/Modules/versioncheck_manager.py
```diff
@@ -0,0 +1,39 @@
...
+            version = json.loads(body)['name'][1:]
+            if LooseVersion(version) > LooseVersion(version_id):
+                self.session.notifier.notify(NTFY_NEW_VERSION, NTFY_INSERT, None, None)
+
+        def parse_response(response):
+            d = readBody(response)
+            d.addCallback(parse_body)
+
+        agent = Agent(reactor)
+        d = agent.request('GET', VERSION_CHECK_URL, Headers({'User-Agent': ['Tribler']}), None)
```
whirm: maybe the version should be added?

Tribler/Core/Modules/versioncheck_manager.py
```diff
@@ -0,0 +1,39 @@
...
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+    def schedule_version_check(self, delay):
+        self.register_task("tribler version check", reactor.callLater(delay, self.check_new_version))
+
+    def check_new_version(self):
+
+        def parse_body(body):
+            version = json.loads(body)['name'][1:]
```
whirm: We don't want a prerelease to be notified to the user.

devos50: @whirm the API endpoint does always give the latest full stable release according to the Github API docs.

whirm: Great, I used the one on your comment without noticing it wasn't the same than on the code and saw the prereleases there too.

devos50:
@whirm I've updated my PR so the status is sent with the request.

I have two questions for you:
-do you want the version check in 6.6 or shall we integrate it in the new GUI?
-I am trying to write some tests for the version check manager but our reactor is not working correctly for me. Can you point out what I'm doing wrong in the following small script:  https://gist.github.com/devos50/8ec03f7d7b98d75dcf5134482ae5e71c ? When using the 'normal' reactor, it works but when I use our threaded reactor, every GET request times out.

whirm:
We could have this merged in 6.6, I think it's a great addition. 

Have you checked if this version comparator does the right thing with RCx, BETAx, etc?

I would add a config option to make it optional so people that don't want Tribler to "call home" can disable it.

devos50:
@whirm yeah, version comparison with info after a dash is working and that's why I'm using a `LooseVersion` object.

whirm:
but does it consider beta > alpha and suchlike?

devos50:
I'm not sure about that but do we need that? I guess stable versions have a easy versioning structure (i.e. `6.6` vs `6.6-RC5`) and we are not comparing unstable versions against each other.

whirm:
what if the user is running a prerelease?

devos50:
@whirm then the following should happen:

```
>>> from distutils.version import LooseVersion
>>> LooseVersion('6.5-RC7') < LooseVersion('6.6')
True
>>> LooseVersion('6.5-RC7') < LooseVersion('6.5')
False
```

Note that Github is only giving stable releases so at least one operand of our comparison is always a stable version.

whirm:
Great!

devos50:
@whirm nice to see that the pylint checker is working now!

I would like to disable the following checks:
- `C0103: (invalid-name)`: sometimes, we override methods from frameworks that do not confirm to this (i.e. the `render_GET` method is an invalid name according to Pylint).
- `W0613: (unused-argument)`: often, we cannot use a passed if we are overriding a method.

I'm also not sure whether is is necessary to write a docstring for every method. I think this might be unnecessary so I suggest to also disable `C0111 (missing-docstring)`.

Also, this PR is ready for review.

lfdversluis:
I disagree with the docstring. I am now investigating the wrong download status and continuously run into functions which are not documented at all and lack comments as well. I am jumping all over Tribler's code base to figure out what they are even trying to do.

devos50:
We have many small methods which are self-explanatory and not necessarily need comments in my opinion. Unfortunately, there is no easy way to distinguish an easy method from a complex method.

whirm:
Yeah, we have to tune pylintrc before enforcing it.

Let's discuss our ideas/opinions in a separate issue though.

devos50:
Yeah that's fine. I will fix the most obvious errors in this case then and create a separate issue for the pylint checker so we can discuss it.

Tribler/Test/Core/Modules/test_versioncheck.py
```diff
@@ -0,0 +1,74 @@
...
+        self.session.notifier.add_observer(self.new_version_callback, NTFY_NEW_VERSION, [NTFY_INSERT])
+
+    def new_version_callback(self, subject, changeType, objectID, *args):
+        self.new_version_called = True
+
+    def listen(self, response, response_code):
+        site = server.Site(VersionResource(response, response_code))
+        self.server = reactor.listenTCP(self.port, site)
+
+    def setup_version_server(self, response, response_code=200):
```
whirm: You can just decorate the function with @call_on_reactor thread

Tribler/Core/Modules/versioncheck_manager.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, session):
+        super(VersionCheckManager, self).__init__()
+
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+        self.version_check_task = None
+
+    def start(self, interval):
+        self.version_check_task = self.register_task("tribler version check", LoopingCall(self.check_new_version))
```
whirm: minor thing: You don't really need to keep a persistent reference to the version check task as it's registered on the TaskManager

Tribler/Test/Core/Modules/test_versioncheck.py
```diff
@@ -0,0 +1,74 @@
...
+        reactor.callFromThread(self.listen, response, response_code)
+
+    def assert_new_version_called(self):
+        self.assertTrue(self.new_version_called == self.should_call_new_version_callback)
+        stop_deferred = self.server.stopListening()
+        stop_deferred.chainDeferred(self.test_deferred)
+
+    def start_version_checker(self):
+        self.session.lm.version_check_manager.start(10000)
+        reactor.callLater(0.5, self.assert_new_version_called)
```
whirm: Don't use delayed calls for this, each test is adding half a second of runtime and it can also break if the system where the tests run are under load.

See how to simulate the passing of time on the leveldb store tests.

whirm: I've just checked and you don't even need to pass time, you just need to listen for the signal to be emitted to pass the test.

devos50: Which signal do you mean exactly?

Tribler/Main/Dialogs/NewVersionDialog.py
```diff
@@ -0,0 +1,58 @@
...
+        mainSizer.Add(bodySizer, 1, wx.EXPAND | wx.ALL, 10)
+        mainSizer.Add(buttonSizer, 0, wx.ALIGN_RIGHT | wx.ALL, 10)
+        self.SetSizerAndFit(mainSizer)
+        if center_on_frame:
+            x, y, w, h = GUIUtility.getInstance().frame.GetScreenRect()
+            self.SetPosition((x + ((w - self.GetSize().x) / 2), y + ((h - self.GetSize().y) / 2)))
+
+    def OnOk(self, event):
+        import webbrowser
+        webbrowser.open("http://tribler.org")
```
whirm: https

Tribler/Test/Core/Modules/test_versioncheck.py
devos50: In some cases, I'm checking for a signal not to be emitted. That's why I wait 0.5 seconds. Do you have a suggestion about the best way to do this instead of using sleep/callLater?

whirm: In that case you need to know that the function that checks for the version has finished and the signal hasn't been emited. In this case you could make the check function return the deferred it creates. Which is usually a good idea as it will prevent the loopingCall from start the countdown to the next call until that Deferred fires. (In this case it's not a huge issue as it fires once every 24h, but it's a good custom to have.) 
Once you can obtain this deferred, you just need to attach your checker function to it. and return it so nose waits for the deferred to resolve before finishing the test.

devos50: Ok I think I understand, thanks!

I'm trying to implement it but I run into an issue. In `versioncheck_manager.py`, I have a chain of methods, namely performing the request, reading the body and parsing the response. The deferred I'm creating triggers when the first step (performing the request) is done but I want to perform an action when parsing the body is done. What's the best way to do this using deferreds? I'm thinking of creating an attribute on the class (`self.parse_body_deferred` or something like that, that fires when parsing the body is done. I can use that deferred in the tests but I'm not sure whether that is the best way to do it.

whirm: you can chain several callbacks to the same deferred, the value the deferred gets fired with becomes the first argument of the first callback function, the value returned by that function becomes the first argument of the second one and so on.

so if make_a_request() returns a deferred for the request result: `make_a_request(url).addCallback(parse_request_body).addCallback(do_something_else)` will pass the return value of parse_request_body to do_something_else.

devos50: Okay I solved this problem, I've returned a deferred that waits until everything is done (using callback chains) but now I have another issue: I have to wait for the notification to be fired. This notification is broadcasted in a separate thread pool. How should we solve this without `time.sleep`?

devos50: @whirm in the `test_notifier.py` we are using `time.sleep` to wait for the notification btw

whirm: You can listen for the signal to be fired or highack the function that would fire that signal.

re test_notifier.py we should improve that then. Can you create an issue?

devos50: @whirm I will create an issue. How can I wait non-blocking for the signal then?

whirm: The test needs to block whie it waits for the signal to fire.

you can have an inner method that fires a deferred, hook the method to the signal and return the deferred so the test blocks until it fires.

devos50: Ok that makes sense but what if the signal should not be fired? In that case, I should still assert that the signal has not been fired somehow

whirm: then you need to assert that the update check has finished but the signal hasn't been fired. 
Just use a dummy notifyer (self.session.notifier) or directly override self.session.notifier.notify.

devos50:
retest this please

devos50:
@whirm done, see my last 2 fixup commits. By removing the `reactor.callLater` call, the duration of all version check manager tests went down from around 10 seconds to 6.5 seconds on average :+1:

Tribler/Core/Modules/versioncheck_manager.py
```diff
@@ -0,0 +1,48 @@
...
+        def parse_response(response):
+            if response.code == 200:
+                return readBody(response)
+            else:
+                self._logger.warning("Got response code %s when performing version check request", response.code)
+
+        agent = Agent(reactor)
+        return agent.request('GET', VERSION_CHECK_URL, Headers({'User-Agent': ['Tribler ' + version_id]}), None)\
+            .addCallback(parse_response)\
+            .addCallback(parse_body)
```
whirm: You are not dealing with readBody's deferred failing:
https://twistedmatrix.com/trac/browser/tags/releases/twisted-14.0.1/twisted/web/client.py#L1959

whirm:
Awesome!

Looks great! Remove the commented out stuff on the tests and deal with my last remark and I'll merge.

Tribler/Core/Modules/versioncheck_manager.py
lfdversluis: @devos50 for this, look at the `addCallbacks`(note the s at the end) function. Adds both an calllback and errback in one go.

devos50:
@whirm something like this (see last commit)?

Tribler/Core/Modules/versioncheck_manager.py
```diff
@@ -44,5 +47,5 @@ def parse_response(response):
 
         agent = Agent(reactor)
         return agent.request('GET', VERSION_CHECK_URL, Headers({'User-Agent': ['Tribler ' + version_id]}), None)\
-            .addCallback(parse_response)\
-            .addCallback(parse_body)
+            .addCallbacks(parse_response, on_request_error)\
+            .addCallbacks(parse_body, on_request_error)
```
whirm: two commentse here:

 - The errbacks are right next to the callbacks, so the first one (line 50) will be called when the request fails. the econd one (line 51) will trigger when parse_response fails. and if parse_body fails, nothing will catch the failure.

 - on_request_error is the equivalent of a catch all try/except. Not only that, but it does nothing to attempt to deal with the situation, so let's say the request fails. Instead of  `parse_response`, `on_request_error` will be called, which will log an entry and return `None`, as `None` is not a failure, it will be passed to `parse_body` which will crash and burn.

Have a look at how the `Deferred` callback chain works and trap specific (expected) failures only.

devos50:
@whirm see my last two fixup commits. Is this ok?

whirm:
Yeah, looks fine.

devos50:
@whirm ok I will squash!

devos50:
@whirm squashed

Tribler/Test/Core/Modules/test_versioncheck.py
```diff
@@ -0,0 +1,78 @@
...
+        site = server.Site(VersionResource(response, response_code))
+        self.server = reactor.listenTCP(self.port, site)
+
+    def assert_new_version_called(self, _):
+        self.assertTrue(self.new_version_called == self.should_call_new_version_callback)
+        stop_deferred = self.server.stopListening()
+        stop_deferred.chainDeferred(self.test_deferred)
+
+    def start_version_checker(self):
+        self.session.lm.version_check_manager.check_new_version().addCallback(self.assert_new_version_called)
```
lfdversluis: Why do you not return the `self.stopListening()` Deferred (also this can return none > use a maybeDeferred!)in assert_new_version_called and chain this to the `check_newVersion()` deferred?

Then you can return `self.start_version_checker()`instead of `test_deferred` which saves using a field. @whirm likes to avoid fields in tests. The test_deferred can then be removed I think.

devos50: @lfdversluis thanks!

devos50:
retest this please

devos50:
Download tests hanging again

devos50:
retest this please

devos50:
Hangs again

devos50:
retest this please

PR #2077, opened 2016-04-10
Fixed flags when adding a torrent file

devos50:
When adding or loading a torrent file in Tribler, the `auto_managed` settings (and some other settings) were not automatically set in a libtorrent without deprecated methods, making it behave very weird when trying to pause or resume a torrent. This PR should fix this.


whirm:
Does this work on all supported libtorrent versions?

devos50:
@whirm yes, it works on the lowest version of libtorrent we support (0.16.18).

PR #2085, opened 2016-04-12
READY: Now cleaning the reactor after a test has left delayed calls behind

### LINKED ISSUE #2075: Idea: clean the reactor after every tests to prevent cascading error
@devos50 and I were discussing the current situation with the tests and thought of the following. Perform the check if the reactor is dirty, if it is print all the stuff like we do now. But in addition **clean** the the reactor so that the next tests won't show the dirty reactor issue if it doesn't leave a dirty reactor behind.
This keeps the overview of which test is the culprit (well it is the first that has this issue obviously) and keeps your local terminal more clean when running tests locally.

There is no real downside to this as the PR tester will fail anyway with the first occurrence.

Thoughts?

lfdversluis:
Fixes #2075 

Prevent from cascading delayed calls in other tests (enhancement).
Added a check which tests for left readers/writers in the reactor (enhancement)

@whirm I ran this with the Core tests (unittests) and it seems to work fine, no errors. Should a test be written for this case? It's hard because you are in the teardown after all.

### DIFF
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -5,6 +5,7 @@
 # Initialize x11 threads before doing anything X11 related.
 import threading
 
+from twisted.internet import interfaces
 from twisted.internet.base import BasePort
 from twisted.internet.defer import Deferred, maybeDeferred, succeed
 from twisted.web.server import Site
@@ -138,8 +139,18 @@ class AbstractServer(BaseTestCase):
             for dc in delayed_calls:
                 self._logger.error(">     %s" % dc)
                 dc.cancel()
+
+        # This is the same check as in the _cleanReactor method of Twisted's Trial
+        selectable_strings = []
+        for sel in reactor.removeAll():
+            if interfaces.IProcessTransport.providedBy(sel):
+                self._logger.error("Sending kill signal to %s", repr(sel))
+                sel.signalProcess('KILL')
+            selectable_strings.append(repr(sel))
+
         self.assertFalse(delayed_calls, "The reactor was dirty when tearing down the test")
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
+        self.assertFalse(selectable_strings, "The reactor had readers/writers left")
 
         # Check whether we have closed all the sockets
         open_readers = reactor.getReaders()
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -137,6 +137,7 @@ class AbstractServer(BaseTestCase):
             self._logger.error("The reactor was dirty:")
             for dc in delayed_calls:
                 self._logger.error(">     %s" % dc)
+                dc.cancel()
         self.assertFalse(delayed_calls, "The reactor was dirty when tearing down the test")
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
 


Tribler/Test/test_as_server.py
```diff
@@ -125,6 +125,11 @@ def tearDown(self, annotate=True):
         self.assertFalse(delayed_calls, "The reactor was dirty when tearing down the test")
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
 
+        # Clean the delayed calls to prevent them cascading into the next test case
+        for delayed_call in reactor.getDelayedCalls():
```
whirm: log everything you do here

Tribler/Test/test_as_server.py
```diff
@@ -125,6 +125,11 @@ def tearDown(self, annotate=True):
...
+            delayed_call.cancel()
+        reactor.removeAll()
```
whirm: Log everything this returns

whirm:
No need for testing but I want it logged

lfdversluis:
@whirm ready to merge.

Tribler/Test/test_as_server.py
whirm: I see you haven't even tested it. there's an assertion at line 125 that makes it impossible to reach your code. 

Just add the cancel call after line 124.

Please, test everything you send.

Tribler/Test/test_as_server.py
```diff
@@ -125,6 +125,12 @@ def tearDown(self, annotate=True):
         self.assertFalse(delayed_calls, "The reactor was dirty when tearing down the test")
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
 
+        # Clean the delayed calls to prevent them cascading into the next test case
+        for delayed_call in reactor.getDelayedCalls():
+            self._logger.warn("Cancelling delayed call: %s", repr(delayed_call))
+            delayed_call.cancel()
+        self._logger.warn("Removed all readers and writers from the reactor: %s", repr(reactor.removeAll()))
```
whirm: This will always log a warning even if there's not readers/writers and will not fail the test.

lfdversluis: @whirm No, the failing is done by the two assertFalse above it, right?

lfdversluis:
I just found out that checking for readers with `reactor.getReaders` will always fail because there are objects such as `[<twisted.internet.posixbase._UnixWaker object at 0x7f37d72cb210>]` present. @whirm Do you agree to skip these?

devos50:
@lfdversluis why is the `reactor.getReaders` test failing with a `_UnixWaker`? `_UnixWaker` is no subclass of `BasePort`:

```
>>> from twisted.internet.posixbase import _UnixWaker
>>> from twisted.internet.base import BasePort
>>> isinstance(_UnixWaker(None), BasePort)
False
```

Am I missing something here?

lfdversluis:
@devos50 There is no getReaders test. I added a check in the teardown of testasserver that the reactor has no readers present, but it does almost always, so I printed this reader and found that object to be a _UnixWaker.

Tribler/Test/test_as_server.py
```diff
@@ -125,6 +125,18 @@ def tearDown(self, annotate=True):
...
 
+        # Clean the delayed calls to prevent them cascading into the next test case
+        for delayed_call in reactor.getDelayedCalls():
+            self._logger.warn("Cancelling delayed call: %s", repr(delayed_call))
+            delayed_call.cancel()
+
+        # Check if there are writers present
+        self.assertFalse(reactor. getWriters(), "The reactor has writers left when tearing down the test")
+        for writer in reactor. getWriters():
+            self._logger.warn("Removing reader from the reactor: %s", repr(writer))
```
devos50: Shouldn't this be: `Removing writer from the reactor: %s` ?

Tribler/Test/test_as_server.py
```diff
@@ -125,6 +125,18 @@ def tearDown(self, annotate=True):
...
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
 
+        # Clean the delayed calls to prevent them cascading into the next test case
+        for delayed_call in reactor.getDelayedCalls():
+            self._logger.warn("Cancelling delayed call: %s", repr(delayed_call))
+            delayed_call.cancel()
+
+        # Check if there are writers present
+        self.assertFalse(reactor. getWriters(), "The reactor has writers left when tearing down the test")
+        for writer in reactor. getWriters():
```
devos50: There's a space between the dot and getWriters

Tribler/Test/test_as_server.py
lfdversluis: Really? I just ran the /Core test suite just fine. But what you probably are referring to is that there needs to be an instance with actually a delayed call which probably did not happen because all tests passed.

devos50:
@lfdversluis do you mean this check? https://github.com/lfdversluis/tribler/blob/clean-reactor-after-dirtiness/Tribler/Test/test_as_server.py#L129

Tribler/Test/test_as_server.py
whirm: I don't think I'm undertanding your comment: if an assertion fails, it raises a AssertionError exception and the function execution aborts, anything behind it will be skipped.

lfdversluis:
@devos50 yup

whirm:
@lfdversluis look at the checks the trial test runner does after a test to see if the reactor is dirty

lfdversluis:
@whirm "The trial test runner" ? What do you mean by that?

whirm:
Twisted has its own test runner (like nose), it's called Trial

lfdversluis:
@whirm Ohh Trial, yeah I've read about. I thought you were referring to some Tribler code. I will check.

lfdversluis:
http://twistedmatrix.com/documents/12.0.0/core/howto/testing.html#auto3 apparently readers and writers need not be closed. I will remove this code now.

whirm:
@lfdversluis 
```
Similarly, sockets created during a test should be closed by the end of the test. This applies to both listening ports and client connections. So, calls to reactor.listenTCP (and listenUNIX, and so on) return IListeningPorts, and these should be cleaned up before a test ends by calling their stopListening method. Calls to reactor.connectTCP return IConnectors, which should be cleaned up by calling their disconnect method. Trial will warn about unclosed sockets.
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Also, look at postClassCleanup() and _cleanReactor() in twisted.trial.util.

Maybe the issue you are having is related to not stopping the threadpool before?

lfdversluis:
@whirm I don't have an issue? so what is "the issue" you are referring to? With this PR I am trying to clean the reactor as per #2075. This way only 1 test will have the "left dirty reactor behind" error and not all subsequent ones, if they in fact did not do anything wrong.

whirm:
I was talking about you getting stuff returned when calling reactor.removeAll() and deciding to remove that part of the code.

lfdversluis:
@whirm Oh I see. When looking at the function you mention, they do use a removeAll() and loop over it, aggregating the results. However, when I added that piece of code almost all unit tests were failing. So either we are doing stuff horribly wrong, or objects like that `UnixWaker` being present is actually not our code wrongdoing and we will get a lot of false positives.

whirm:
As I said: 

> Maybe the issue you are having is related to not stopping the threadpool before?

If you read the class's docs:

```
twisted.internet.posixbase._UnixWaker(_FDWaker) class documentation

Part of twisted.internet.posixbase (View In Hierarchy) 

Known subclasses: twisted.internet._glibbase.GlibWaker

This class provides a simple interface to wake up the event loop.

This is used by threads or signals to wake up the event loop.
```

lfdversluis:
@whirm I am sorry for the confusion I may have caused. Running the check I just pushed now yesterday caused almost all tests to fail in Test/Core. Now they are fine (somehow). 

I assume you did mean this check right? I left our the `cleanPending` as we are essentially already doing that ourselves.

Tribler/Test/test_as_server.py
```diff
@@ -126,6 +128,17 @@ def tearDown(self, annotate=True):
         self.assertFalse(delayed_calls, "The reactor was dirty when tearing down the test")
         self.assertFalse(Session.has_instance(), 'A session instance is still present when tearing down the test')
 
+        # This is the same check as in the _cleanReactor method of Twisted's Trial
```
whirm: Any cleanup should be done before the assertions.

lfdversluis: ok!

lfdversluis:
Haha a fixup that is fixed up (had an enter too much there)

lfdversluis:
retest this please

whirm:
https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/779/artifact/output/test_runners_output/00010.err/*view*/

whirm:
How come that test got stuck?

lfdversluis:
No idea, lately the tests seem unstable again.

whirm:
@lfdversluis one of @devos50 ran into the same issue

lfdversluis:
`est_download_torrent_from_url (Tribler.Test.API.test_download.TestDownload) ... /home/jenkins/workspace/GH_Tribler_PR_tests_linux/gumby/scripts/wrap_in_temp_home.sh: line 34: 37896 Terminated`

lfdversluis:
retest this please

whirm:
@lfdversluis that's a time out, you should check why it timed out.

https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/810/artifact/output/test_runners_output/00011.err/*view*/

`test_download_torrent_from_url` is getting stuck for some reason.

@devos50 apparently there's a bug in the new tests?

lfdversluis:
@whirm besides the hanging test(s). I think all is well with this PR.

whirm:
Great!

PR #2086, opened 2016-04-12
Fixed a Python crash when passing None to load_state of ltsession

devos50:
On Windows 32-bit, Python seems to crash when `None` is passed to the `load_state` method of a libtorrent session object. This is probably because of a corrupted `lt.state` file. The crash does not happen on Windows 64-bit, OS X and Ubuntu. The crash is easy to reproduce with the following Python commands:

```
>>> import libtorrent as lt
>>> sess = lt.session()
>>> sess.load_state(None)
```

My solution is to not load the session state when the result of the `bdecode` method is `None`. The state file will be overridden with correct data when Tribler shuts down so there is no need to remove a corrupt `lt.state` file as soon as we detect that it is corrupted.

I've also written some tests for `LibtorrentMgr`. For these, I make use of a fake Tribler session where only a subset of the methods are implemented so I don't have to setup a complete session. Using a mocked object is not possible without introducing another dependency (since `unittest.mock` is not part of the Python library in Python 2.7).


devos50:
@whirm ready for review

whirm:
@devos50 note that I don't mind at all adding dependencies to the testing environment as long as they are packaged for Debian/Ubuntu.

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -154,8 +154,9 @@ def create_session(self, hops=0):
             if listen_port != ltsession.listen_port():
                 self.trsession.set_listen_port_runtime(ltsession.listen_port())
             try:
-                lt_state = open(os.path.join(self.trsession.get_state_dir(), LTSTATE_FILENAME)).read()
-                ltsession.load_state(lt.bdecode(lt_state))
+                lt_state = lt.bdecode(open(os.path.join(self.trsession.get_state_dir(), LTSTATE_FILENAME)).read())
+                if lt_state is not None:
+                    ltsession.load_state(lt_state)
```
whirm: Maybe we should log the fact that the lt_state file was corrupt

whirm:
Looks fine, add the log entry and I'll merge

devos50:
@whirm done

PR #2088, opened 2016-04-14
Rename .pylintrc to pylintrc

whirm:
As apparently the documentation is wrong and it won't look for a config
file with this name.


whirm:
Let's not do that for now.

Issue #2056, opened 2016-03-27
Windows 10 installation: error running after installation


Koris:
After installing Tribler, I get an error 'Ordinal not found' when I effectively want to start it.

- [ x ] I have read CONTRIBUTING.md

- [ x ] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
6.5.1

##### Operating system and version: #####
Windows 10

##### Expected behavior: #####
Up running

##### Actual behavior: #####
Error : the ordinal 1 could not be located in the dynamic link library C:\Program Files (x86)\Tribler\IPHLPAPI.dll

##### Steps to reproduce the behavior: #####
Install and run Tribler

##### Relevant log file output: #####

Traceback (most recent call last):
  File "tribler_exe.py", line 99, in <module>
  File "Tribler\Main\tribler.pyo", line 47, in <module>
    
  File "Tribler\Core\Utilities\twisted_thread.pyo", line 8, in <module>
    
  File "twisted\internet\reactor.pyo", line 39, in <module>
    
  File "twisted\internet\selectreactor.pyo", line 200, in install
    
  File "twisted\internet\selectreactor.pyo", line 74, in __init__
    
  File "twisted\internet\base.pyo", line 498, in __init__
    
  File "twisted\internet\posixbase.pyo", line 289, in installWaker
    
  File "twisted\internet\posixbase.pyo", line 79, in __init__
    
  File "socket.pyo", line 191, in __init__
    
socket.error: [Errno 10106] The requested service provider could not be loaded or initialized

whirm:
Thanks for the report!

lfdversluis:
@Koris Could you try to replace this dll with the `IPHLPAPI.dll` in C:/Windows/System32`. Also check your network driver is up to date.

Koris:
I have copied the dll and it works now! Thanks!

lfdversluis:
@Koris Great! Feel free to close the issue :+1:

whirm:
I don't think this can be considered fixed.

@lfdversluis is that DLL shipped in windows by default?

lfdversluis:
@whirm yes it is default in the windows32 folder

whirm:
so why are we shipping it?

lfdversluis:
@whirm Good question. First time I've seen this error being reported and had to search a bit to find it was default in win32

lfdversluis:
[...]

PR #2100, opened 2016-04-17
READY: Added some dlls to the exclude list when building Tribler on Windows

lfdversluis:
Most likely fixes the case of #2056. I have tested the installation on proxmox 144 and tribler starts with no (visible) problems.

based on: http://stackoverflow.com/questions/20930173/how-to-include-components-of-psutil-in-py2exe-that-py2exe-cant-find

Fixes **1** bug.

### DIFF
--- a/Tribler/Main/Build/Win/setuptribler.py
+++ b/Tribler/Main/Build/Win/setuptribler.py
@@ -6,6 +6,7 @@ import os
 try:
     import py2exe.mf as modulefinder
     import win32com
+
     for p in win32com.__path__[1:]:
         modulefinder.AddPackagePath("win32com", p)
     for extra in ["win32com.shell"]:
@@ -15,7 +16,6 @@ try:
             modulefinder.AddPackagePath(extra, p)
 except ImportError:
     pass
-
 
 from distutils.core import setup
 import py2exe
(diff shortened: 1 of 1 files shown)


lfdversluis:
TestLibtorrentDownload.test_playdownload - streaming did not start - Condition was not satisfied in 60 second

lfdversluis:
retest this please

lfdversluis:
@devos50 Please review!

Tribler/Main/Build/Win/setuptribler.py
```diff
@@ -59,12 +59,15 @@
     # (Disabling bundle_files for now -- apparently causes some issues with Win98)
     # options = {"py2exe": {"bundle_files": 1}},
     # zipfile = None,
+    # py2exe includes some dlls which it should not see:
+    # http://stackoverflow.com/questions/20930173/how-to-include-components-of-psutil-in-py2exe-that-py2exe-cant-find - IPHLPAPI.dll was our issue.
+    # http://www.py2exe.org/index.cgi/OverridingCriteraForIncludingDlls
```
devos50: Maybe we should give a link to the Github issue here (#2056)?

devos50:
I think the Pylint issues are easy to fix here.

Otherwise, looks fine!

Tribler/Main/Build/Win/setuptribler.py
```diff
@@ -59,12 +59,15 @@
...
     options={"py2exe": {"packages": packages,
                         "optimize": 2,
                         "skip_archive": True,
                         "dist_dir": os.path.join("dist", "installdir"),
-                        "dll_excludes": ["mswsock.dll", 
-										"MSVCR90.DLL"]
+                        "dll_excludes": ["mswsock.dll", "MSVCR90.dll", "msvcr71.dll",
```
devos50: I just noticed that `MSVCR90.dll` is excluded. Is `MSVCR90.dll` shipped with py2exe? If I remember correctly, we are also providing a Microsoft Visual C++ redistributable installer with the NSIS installer (and I thought that was executed after the installation finished).

Also, why do we exclude both `MSVCR90.dll` and `msvcr71.dll` (`msvcr71.dll` is a rather old version)?

lfdversluis: @devos50 The `MSVCR90` was an issue of the past. I've looked at what py2exe packs and what it does is packing all MSVC* it can find. the msvcr71 is old (windows XP), but if it present it will also be packed. So I added it there as per the SO answer.

devos50: @lfdversluis does this mean that we don't have to ship the Visual C++ redistributable (since it is already part of py2exe)? Or are the .dll files not enough?

lfdversluis: What do you mean by not enough?

PR #2114, opened 2016-04-20
Disabled select all functionalities on wx versions older than 3

devos50:
`wx.ACCEL_RAW_CTRL` is not implemented in wx2.8. We should still support wx2.8 so users on Ubuntu 14.04 and older can Tribler.


devos50:
@pimveldhuisen it should work now for wx2.8 ;)

PR #2121, opened 2016-04-21
Ready: Created the HTTP API classes and written a first endpoint to fetch information about your channel

devos50:
In this PR, I created the classes that are setting up the Tribler HTTP API. The API can be enabled or disabled by using the session configuration file. Also, changing the port is possible. By default, the API listens on port 8085 (we should fix this port so the new Qt GUI knows which port to use) but it's also possible to assign a random port to the API as I do in the unit tests.

The first endpoint of the API is to fetch the overview of your channel. This returns the name, identifier (hex-encoded) and description of your channel. The API will be gradually improved and expanded in separate PRs after this one has been merged. I will first focus on the endpoints related to your channel (getting/adding torrents, playlists and rss-feeds) as requested by @brussee.

I've also written a document describing the API and available endpoints in more detail. This document might grow over time so at one point, we might have to split it into separate files. The structure of the API documentations resembles the [Twitch API documentation](https://github.com/justintv/Twitch-API) which I think has a clear structure.

I've added docstrings where I think they might increase the understandability of the code.

Also, this PR got a little larger than I initially hoped but I think it's very hard to split this since everything I did is necessary to create the base for the API.

### DIFF
--- a/Tribler/Core/SessionConfig.py
+++ b/Tribler/Core/SessionConfig.py
@@ -730,6 +730,37 @@ class SessionConfigInterface(object):
         return self.sessconfig.get(u'watch_folder', u'watch_folder_dir')
 
     #
+    # API
+    #
+    def set_http_api_enabled(self, http_api_enabled):
+        """
+        Sets whether the HTTP API is enabled.
+        :param http_api_enabled: True or False.
+        """
+        return self.sessconfig.set(u'http_api', u'enabled', http_api_enabled)
+
+    def get_http_api_enabled(self):
+        """
+        Returns whether the HTTP API is enabled.
+        :return: A boolean indicating whether the HTTP API is enabled.
+        """
+        return self.sessconfig.get(u'http_api', u'enabled')
+
+    def set_http_api_port(self, http_api_port):
+        """
+        Sets the HTTP API listen port.
+        :param http_api_port: An integer, indicating the port where the HTTP API should listen on.
+        """
+        return self.sessconfig.set(u'http_api', u'port', http_api_port)
+
+    def get_http_api_port(self):
+        """
+        Returns the HTTP API listen port.
+        :return: An integer indicating the port where the HTPT API listens on.
+        """
+        return self._obtain_port(u'http_api', u'port')
+
+    #
     # Static methods
     #
     @staticmethod
--- a/Tribler/Core/defaults.py
+++ b/Tribler/Core/defaults.py
@@ -37,8 +37,9 @@ DEFAULTPORT = 7760
 #  Version 10: BarterCommunity settings added (disabled by default)
 #  Version 11: Added a default whether we should upgrade or not.
 #  Version 12: Added watch folder options.
+#  Version 13: Added HTTP API options.
 
-SESSDEFAULTS_VERSION = 12
+SESSDEFAULTS_VERSION = 13
 sessdefaults = OrderedDict()
 
 # General Tribler settings
@@ -146,6 +147,11 @@ sessdefaults['upgrader']['enabled'] = True
 sessdefaults['watch_folder'] = OrderedDict()
 sessdefaults['watch_folder']['enabled'] = False
 sessdefaults['watch_folder']['watch_folder_dir'] = None
+
+# HTTP API config
+sessdefaults['http_api'] = OrderedDict()
+sessdefaults['http_api']['enabled'] = False
+sessdefaults['http_api']['port'] = -1
 
 #
 # BT per download opts
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -260,6 +260,7 @@ class TestAsServer(AbstractServer):
         self.config.set_videoplayer(False)
         self.config.set_enable_metadata(False)
         self.config.set_upgrader_enabled(False)
+        self.config.set_http_api_enabled(False)
 
     def tearDown(self):
         self.annotate(self._testMethodName, start=False)
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -17,6 +17,7 @@ from twisted.internet import reactor
 from Tribler.Core.APIImplementation.threadpoolmanager import ThreadPoolManager
 from Tribler.Core.CacheDB.sqlitecachedb import forceDBThread
 from Tribler.Core.DownloadConfig import DownloadStartupConfig
+from Tribler.Core.Modules.restapi.rest_manager import RESTManager
 from Tribler.Core.Modules.search_manager import SearchManager
 from Tribler.Core.Modules.versioncheck_manager import VersionCheckManager
 from Tribler.Core.Modules.watch_folder import WatchFolder
@@ -72,6 +73,7 @@ class TriblerLaunchMany(TaskManager):
         self.metadata_store = None
         self.rtorrent_handler = None
         self.tftp_handler = None
+        self.api_manager = None
         self.watch_folder = None
         self.version_check_manager = None
 
@@ -248,6 +250,10 @@ class TriblerLaunchMany(TaskManager):
 
         if self.rtorrent_handler:
             self.rtorrent_handler.initialize()
+
+        if self.session.get_http_api_enabled():
+            self.api_manager = RESTManager(self.session)
+            self.api_manager.start()
 
         if self.session.get_watch_folder_enabled():
             self.watch_folder = WatchFolder(self.session)
@@ -675,6 +681,10 @@ class TriblerLaunchMany(TaskManager):
         if self.torrent_store is not None:
             yield self.torrent_store.close()
             self.torrent_store = None
+
+        if self.api_manager is not None:
+            yield self.api_manager.stop()
+            self.api_manager = None
 
         if self.watch_folder is not None:
             yield self.watch_folder.stop()
--- a/Tribler/Core/Modules/restapi/__init__.py
+++ b/Tribler/Core/Modules/restapi/__init__.py
@@ -1,0 +1,3 @@
+"""
+This package contains code for the Tribler HTTP API.
+"""
--- a/Tribler/Core/Modules/restapi/my_channel_endpoint.py
+++ b/Tribler/Core/Modules/restapi/my_channel_endpoint.py
@@ -1,0 +1,57 @@
+import json
+from twisted.web import server, resource
+from Tribler.Core.simpledefs import NTFY_CHANNELCAST
+
+
+class MyChannelEndpoint(resource.Resource):
+    """
+    This endpoint is reponsible for handing all requests regarding your channel such as getting and updating
+    torrents, playlists and rss-feeds.
+    """
+
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+        self.session = session
+
+    def getChild(self, path, request):
+        if path == "overview":
+            return MyChannelOverviewEndpoint(self.session)
+
+
+class MyChannelOverviewEndpoint(resource.Resource):
+    """
+    This endpoint returns a 404 HTTP response if you have not created a channel (yet).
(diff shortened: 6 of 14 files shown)


devos50:
retest this please

devos50:
@lfdversluis can you review this PR please?

devos50:
Hmm I was a little too quick with this PR and I was thinking about two minor modifications:
-the response type of the request should be text/json
-I want to add the configuration type of the channel to the response (open/semi-open/closed), this is displayed on the settings page of your channel

I will implement this tomorrow morning

devos50:
The mac slave went offline again during the build

devos50:
retest this please

devos50:
I only added `text/json` as response content header. It turns out that getting the channel state (open/semi-open/closed) is more complicated than I thought and requires a Dispersy instance so I will leave this for a separate PR. @lfdversluis can you review?

Tribler/Core/Modules/api/api_manager.py
```diff
@@ -0,0 +1,25 @@
...
+    This class is responsible for managing the startup and closing of the Tribler HTTP API.
+    """
+
+    def __init__(self, session):
+        super(APIManager, self).__init__()
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+        self.site = None
+
+    def start(self):
```
lfdversluis: doc missing

Tribler/Core/Modules/api/api_manager.py
```diff
@@ -0,0 +1,25 @@
...
+        self.site = reactor.listenTCP(self.session.get_http_api_port(),
+                                      server.Site(RootEndpoint(self.session)))
+
+    def stop(self):
```
lfdversluis: doc missing

Tribler/Core/Modules/api/my_channel_endpoint.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    This endpoint is reponsible for handing all requests regarding your channel such as getting and updating
+    torrents, playlists and rss-feeds.
+    """
+
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+        self.session = session
+
+    def getChild(self, path, request):
```
lfdversluis: doc missing (I know it's builtin but I think it can use some doc?)

Tribler/Core/Modules/api/my_channel_endpoint.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+    """
+
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+        self.session = session
+        self.channel_db_handler = self.session.open_dbhandler(NTFY_CHANNELCAST)
+
+    def return_404(self, request):
```
lfdversluis: doc missing

Tribler/Core/Modules/api/my_channel_endpoint.py
```diff
@@ -0,0 +1,54 @@
...
+        request.setResponseCode(404)
+        request.finish()
+
+    def render_GET(self, request):
```
lfdversluis: doc missing

Tribler/Core/SessionConfig.py
```diff
@@ -730,6 +730,37 @@ def get_watch_folder_path(self):
...
+        :return: A boolean indicating whether the HTTP API is enabled.
+        """
+        return self.sessconfig.get(u'http_api', u'enabled')
+
+    def set_http_api_port(self, http_api_port):
+        """
+        Sets the HTTP API listen port.
+        :param http_api_port: An integer, indicating the port where the HTTP API should listen on.
+        """
+        return self.sessconfig.set(u'http_api', u'port', http_api_port)
```
lfdversluis: Hmmm maybe make getRandomPort so that even though this port is not in use (keep a global list or something and add this to it), it will not use this port if the http_api_port is enabled?

Tribler/Test/Core/Modules/Api/base_api_test.py
```diff
@@ -0,0 +1,47 @@
...
+        self.config.set_http_api_port(get_random_port())
+        self.config.set_megacache(True)
+
+    def parse_body(self, body):
+        if body is not None:
+            self.assertDictEqual(json.loads(body), self.expected_response_json)
+
+    def parse_response(self, response):
+        self.assertEqual(response.code, self.expected_response_code)
+        if response.code == 200:
```
lfdversluis: No need for this right? If the assert fails the code below will never be executed. This the if and whole else branch can be deleted.

Tribler/Test/test_as_server.py
```diff
@@ -518,6 +520,9 @@ def startTest(self, callback, min_callback_delay=5, autoload_discovery=True,
         # Always start testing from the same dir (repo root)
         os.chdir(os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..')))
 
+        # Override default HTTP API enabled to disable API in tests
+        sessdefaults['http_api']['enabled'] = False
```
lfdversluis: Can't you use your set function here?

devos50: Unfortunately, I cannot manipulate the `SessionConfig` object that is used when the GUI is started in tests. That's why I'm overwriting the defaults instead.

Tribler/Test/Core/Modules/Api/base_api_test.py
devos50: I think it's needed since it makes only sense to continue and call `readBody` when the response code is 200.

Tribler/Core/SessionConfig.py
devos50: I'm not really sure what you mean by this. Can you explain your idea in more detail?

Tribler/Test/Core/Modules/Api/base_api_test.py
lfdversluis: self.assertEqual(response.code, self.expected_response_code) does that for you?

Tribler/Core/SessionConfig.py
lfdversluis: getRandomPort fetches generates a random number and checks if this port is in use. Suppose now that your http api port is enabled, so you are going to use e.g. port 8081. Now getRandom port checks by chance 8081 and sees that it is not in use, so it will use that one. The moment you want to use 8081 for your http you get an exception.

My idea is to create a list of numbers (socks server, http api port, libtorrent port, etc.) that may be used i.e. they are enabled in the config. getRandomport is then not allowed to take any of these numbers as they may be used.

Tribler/Test/Core/Modules/Api/base_api_test.py
devos50: Sometimes I'm expecting a response code other than 404. In these cases, I don't want `readBody` to be called.

lfdversluis: self.expected_response_code is always 200 right?

devos50: Not always, see for instance one of my tests (`test_my_channel_overview_endpoint_no_my_channel`)

lfdversluis: Oh i see the structure now. Ok my bad

Tribler/Core/SessionConfig.py
devos50: Ah, so we make some port numbers reserved. That sounds like a good idea, especially if we are going to use more hardcoded ports. I would create a separate issue for this.

devos50:
@whirm can you review this PR?

Tribler/Core/defaults.py
```diff
@@ -147,6 +148,11 @@
 sessdefaults['watch_folder']['enabled'] = False
 sessdefaults['watch_folder']['watch_folder_dir'] = None
 
+# HTTP API config
+sessdefaults['http_api'] = OrderedDict()
+sessdefaults['http_api']['enabled'] = True
+sessdefaults['http_api']['port'] = 8081
```
whirm: We can't have hardcoded ports on the config file, it breaks multiple Tribler instances running on the same machine. (Used in experiments, tests and such).

Tribler/Test/test_as_server.py
whirm: Why aren't you doing this on `setupPresession()`?

doc/Tribler API.md
```diff
@@ -0,0 +1,31 @@
+# Tribler API
```
whirm: Let's call it REST API, it's what it really is and it will avoid confusion with the core API.

I think the API package should be renamed to something else too.

Tribler/Test/test_as_server.py
devos50: @whirm `setupPreSession` is not called when running the GUI tests.

doc/Tribler API.md
devos50: Ok that's maybe better indeed. Do you agree if I rename the `api` package to `restapi`?

Tribler/Core/defaults.py
devos50: Ok I will change it so it will always use a random port.

doc/Tribler API.md
whirm: sounds good to me

devos50:
@whirm see my last three fixup commits.

Tribler/Core/Modules/restapi/api_manager.py
```diff
@@ -0,0 +1,31 @@
+import logging
+from twisted.internet.defer import maybeDeferred
+from twisted.web import server
+from Tribler.Core.Modules.restapi.root_endpoint import RootEndpoint
+from Tribler.Core.Utilities.twisted_thread import reactor
+from Tribler.dispersy.taskmanager import TaskManager
+
+
+class APIManager(TaskManager):
```
whirm: RESTManager?

whirm:
Looks good

devos50:
@whirm I've rebased, updated the shutdown procedure of the `RESTManager` and added a small test in `test_sessionconfig.py`. If it looks ok I will squash it.

brussee:
@devos50: Please add your new classes to Setup.py (to make it work on Android)

whirm:
@devos50 @brussee is right :)

devos50:
@brussee @whirm which Setup.py do you mean?

devos50:
@brussee @whirm I've added the new module to `setup.py`

whirm:
looks good, squash!

devos50:
Got an error when running tests, see #2134

devos50:
retest this please

devos50:
@whirm ready

Tribler/Core/defaults.py
```diff
@@ -147,6 +148,11 @@
 sessdefaults['watch_folder']['enabled'] = False
 sessdefaults['watch_folder']['watch_folder_dir'] = None
 
+# HTTP API config
+sessdefaults['http_api'] = OrderedDict()
+sessdefaults['http_api']['enabled'] = True
```
whirm: Sorry for noticing now, but I don't think we want this enabled by default yet...

devos50:
@whirm ok np, I've disabled it by default (see my last two fixup commits). If ok, I will squash again.

Tribler/Test/test_as_server.py
```diff
@@ -260,7 +260,6 @@ def setUpPreSession(self):
         self.config.set_videoplayer(False)
         self.config.set_enable_metadata(False)
         self.config.set_upgrader_enabled(False)
-        self.config.set_http_api_enabled(False)
```
whirm: leave these lines, we really want this to stay disabled on the tests even when the REST iface gets enabled by default in the future. (in tunnel/main.py too)

devos50:
@whirm done

whirm:
Ok then! squash it!

devos50:
@whirm done

PR #2128, opened 2016-04-25
Removed lock in get method of configparser

devos50:
The tests in `Tribler/Test/API/test_download.py` occasionally got stuck. It turned out that is caused by locks in the `threading` and `logging` framework which seems to contain a bug (http://bugs.python.org/issue6721).

The location where the tests were hanging is always the `get` method of `configparser.py`. There is a lock here but no thread critical code as far as I can see so I propose to remove the lock. I tested the code below and the tests do not get stuck anymore (been running them for a while in several terminals).

I will run the unit tests several times.

### DIFF
--- a/Tribler/Core/Utilities/configparser.py
+++ b/Tribler/Core/Utilities/configparser.py
@@ -35,15 +35,14 @@ class CallbackConfigParser(RawConfigParser):
             RawConfigParser.set(self, section, option, new_value)
 
     def get(self, section, option, literal_eval=True):
-        with self.lock:
-            value = RawConfigParser.get(self, section, option) if RawConfigParser.has_option(
-                self, section, option) else None
-            if literal_eval:
-                try:
-                    value = ast.literal_eval(value)
-                except:
-                    pass
-            return value
+        value = RawConfigParser.get(self, section, option) if RawConfigParser.has_option(
+            self, section, option) else None
+        if literal_eval:
+            try:
+                value = ast.literal_eval(value)
+            except (ValueError, SyntaxError):
+                pass
+        return value
 
     def copy(self):
         with self.lock:


Tribler/Core/Utilities/configparser.py
```diff
@@ -35,15 +35,14 @@ def set(self, section, option, new_value):
...
-                    value = ast.literal_eval(value)
-                except:
-                    pass
-            return value
+        value = RawConfigParser.get(self, section, option) if RawConfigParser.has_option(
+            self, section, option) else None
+        if literal_eval:
+            try:
+                value = ast.literal_eval(value)
+            except (ValueError, SyntaxError):
```
whirm: Are we getting syntax errors while running the tests?

devos50: @whirm yes we are so I found out that we should also catch `SyntaxError`.

devos50:
retest this please

devos50:
retest this please

devos50:
Tests are going well, three in a row. @lfdversluis can you do a quick review?

Tribler/Core/Utilities/configparser.py
whirm: what's triggering those?

devos50: @whirm it seems to be the state dir in the config file:

```
ERROR: test_aa_hasPeer (Tribler.Test.Core.test_sqlitecachedbhandler_peers.TestSqlitePeerDBHandler)
  File "/Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/tribler/Tribler/Test/Core/test_sqlitecachedbhandler_peers.py", line 13, in setUp
    super(TestSqlitePeerDBHandler, self).setUp()
  File "/Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/tribler/Tribler/Test/Core/test_sqlitecachedbhandler.py", line 50, in setUp
    self.session = Session(self.config, ignore_singleton=True)
  File "/Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/tribler/Tribler/Core/Session.py", line 82, in __init__
    state_dir = scfg.get_state_dir()
  File "/Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/tribler/Tribler/Core/SessionConfig.py", line 137, in get_state_dir
    in_config_path = self.sessconfig.get(u'general', u'state_dir')
  File "/Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/tribler/Tribler/Core/Utilities/configparser.py", line 42, in get
    value = ast.literal_eval(value)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
SyntaxError: invalid syntax (<unknown>, line 1)
DEBUG   1461579265         _legacy:twisted:154  PythonLoggingObserver hooked up
```

lfdversluis:
Removed lock **and** resolved yet another general try except? Nice.
Interesting that we need a syntaxerror except though. Looks like Tribler code is doing stuff wrong?

devos50:
I investigated it and it seems that `ast` has problems with evaluating strings like `/var/folders/r8/d63xy0dx62l8n8mdgpy8dxw00000gn/T/tmpqXsHWf_tribler_test_session/dot.Tribler` (the state dir in our configuration file). When an exception is being caught, we just take the value as string and continue. While this does not lead to errors, it is a bit ugly. Hopefully this will be better when we use `ConfigObj`.

Quick example:

```
>>> import ast
>>> ast.literal_eval('/var/folders/r8/d63xy0dx62l8n8mdgpy8dxw00000gn/T/tmpqXsHWf_tribler_test_session/dot.Tribler')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    /var/folders/r8/d63xy0dx62l8n8mdgpy8dxw00000gn/T/tmpqXsHWf_tribler_test_session/dot.Tribler
    ^
SyntaxError: invalid syntax`
```

@whirm this PR is ready for merge ;)

whirm:
I'm not sure about the ramifications of removing this lock, what else is using this lock and what for? 

If the GUI is not using this extensively, maybe it could be decorated with the @blocking_call_on_reactor_thread?

devos50:
@whirm the lock is used in all other method in `configparser.py` but I think it's not necessary in the `get` method since there is no critical code. Even if two threads are accessing this method, it should work.

Also, the `get` method is used everywhere in Tribler (since it's called every time we are fetching something from a config file) so I don't think it's a good idea to make it `blocking_call_on_reactor_thread` since that might have more consequences than removing the lock.

lfdversluis:
Hmm only if multiple threads attempt to read/write to this file simultaneously it can go wrong... if a file is being opened we could use "rb" and "wb" ?

devos50:
@lfdversluis true but we are not reading from a file in the `get` method. This is done in the `read_file` and `write_file` methods.

devos50:
@lfdversluis also, we are already using the `rb` and `wb` flags when reading/writing config files.

whirm:
I haven't, but have you checked the case where a default value is not set and the getter sets the default value?

devos50:
@whirm if the `get` method is called with a non-existent section/option, `None` is returned so no I/O is performed.

PR #2135, opened 2016-04-25
MultiChain true halves

Captain-Coder:
This PR is a possible next iteration of multichain. The first iteration of multichain made me sad when I wrote the validator. I wanted to:
- Remove the distinction between requester and responder. When writing the validation code you have to constantly worry if you are the requester or responder, and in the database code too we see all the UNION statements. Even though the requester and responder are nearly symmetrical.
- "Not repeat myself", some functions used many arguments, and packing values into and out of tuples and buffers was happening all over the place.

So I started with these to changes and continued their effects throughout multichain and cleaned everything up. The benefits to this refactor are:
- Immutable messages
- Immutable database records
- Removal of the signature timeout
- A single wire and signing format instead of a bunch of them
- A 40% reduction in code size for multichain. (Not counting the validator, since that is new functionality)
- Maintainability improvement. A lot of methods in the community where just calling one another in a long chain without a clear reason for their separation. Quite a bit of code became redundant, and queries where shortend and simplified.
- All block operations (such as compute hash, signature, creation, packing and validation) are now together in one place (block.py) 
- Reduced dependence on dispersy since we no longer need dispersy to handle double signing messages. (Because we're not asking, but tellin'!)
- If the validator sais there is no complete (or not enough) information, we delay the sign block message and start a crawl.

Making the requester and responder "symmetric" also causes a semantic shift. We are no longer asking the peer to pretty please sign our message and tack on his stuff, but we're telling him "Hey I signed this". The peer can sign the same values on his chain (including a link to our block) and tell us about the result "Hey I signed this" (or decide not to sign). See how they are symmetrical? This removes the original blocking issues in a way similar to how the current halfsigned-as-a-default does it without all the downside of having two identities for one (DB) record.

Please check the validator code and try to think of edge cases you think are not covered/tested/falidated/flagged. It is _very_ important that we get this right.

I know there are some further refinements possible but I was not sure enough to make them happen:
- Dispersy would no longer need to sign the message since multichain does that itself. Blocks are tamper proof, so why should we have dispersy resign the message?

### DIFF
--- a/Tribler/Test/Community/Multichain/test_multichain_community.py
+++ b/Tribler/Test/Community/Multichain/test_multichain_community.py
@@ -261,10 +261,10 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         block.up += 10
         block.total_up = block.up
         signature_request = node.community.get_meta_message(HALF_BLOCK).impl(
-                    authentication=(node.community.my_member,),
-                    distribution=(node.community.claim_global_time(),),
-                    destination=(target_other,),
-                    payload=(block,))
+            authentication=(node.community.my_member,),
+            distribution=(node.community.claim_global_time(),),
+            destination=(target_other,),
+            payload=(block,))
         other.give_message(signature_request, node)
 
         self.assertBlocksInDatabase(other, 0)
@@ -272,8 +272,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
 
         # Assert
         with self.assertRaises(ValueError):
-            "No signature responses should have been sent"
-            _, block_responses = node.receive_message(names=[HALF_BLOCK])
+            # No signature responses should have been sent
+            _, _ = node.receive_message(names=[HALF_BLOCK])
 
     def test_block_values(self):
         """
--- a/Tribler/Test/Community/Multichain/test_multichain_community.py
+++ b/Tribler/Test/Community/Multichain/test_multichain_community.py
@@ -224,8 +224,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         _, signature_response = node.receive_message(names=[HALF_BLOCK]).next()
         node.give_message(signature_response, node)
         # Assert
-        self.assertBlocksInDatabase(other, 1)
-        self.assertBlocksInDatabase(node, 1)
+        self.assertBlocksInDatabase(other, 2)
+        self.assertBlocksInDatabase(node, 2)
         self.assertBlocksAreEqual(node, other)
 
         block = node.call(node.community.persistence.get_latest, node.community._public_key)
@@ -297,8 +297,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         _, block_response = crawler.receive_message(names=[FULL_BLOCK]).next()
         crawler.give_message(block_response, other)
         # Assert
-        self.assertBlocksInDatabase(node, 1)
-        self.assertBlocksInDatabase(crawler, 1)
+        self.assertBlocksInDatabase(node, 2)
+        self.assertBlocksInDatabase(crawler, 2)
         self.assertBlocksAreEqual(node, crawler)
 
     def test_request_block_halfsigned(self):
@@ -351,8 +351,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         _, block_response = crawler.receive_message(names=[FULL_BLOCK]).next()
         crawler.give_message(block_response, other)
         # Assert
-        self.assertBlocksInDatabase(node, 1)
-        self.assertBlocksInDatabase(crawler, 1)
+        self.assertBlocksInDatabase(node, 2)
+        self.assertBlocksInDatabase(crawler, 2)
         self.assertBlocksAreEqual(node, crawler)
 
     def test_crawler_no_block(self):
@@ -406,8 +406,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         _, block_response = crawler.receive_message(names=[FULL_BLOCK]).next()
         crawler.give_message(block_response, node)
         # Assert
-        self.assertBlocksInDatabase(node, 1)
-        self.assertBlocksInDatabase(crawler, 1)
+        self.assertBlocksInDatabase(node, 2)
+        self.assertBlocksInDatabase(crawler, 2)
         self.assertBlocksAreEqual(node, crawler)
 
     def test_crawl_batch(self):
@@ -451,8 +451,8 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         crawler.give_message(block_response, node)
 
         # Assert
-        self.assertBlocksInDatabase(node, 2)
-        self.assertBlocksInDatabase(crawler, 2)
+        self.assertBlocksInDatabase(node, 4)
+        self.assertBlocksInDatabase(crawler, 4)
         self.assertBlocksAreEqual(node, crawler)
 
     def test_crawler_on_introduction_received(self):
@@ -515,7 +515,7 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
 
     @blocking_call_on_reactor_thread
     def assertBlocksInDatabase(self, node, amount):
-        assert len(node.community.persistence.get_all_hash_requester()) == amount
+        assert node.community.persistence.execute(u"SELECT COUNT(*) FROM multi_chain").fetchone()[0] == amount
 
     @blocking_call_on_reactor_thread
     def assertBlocksAreEqual(self, node, other):
--- a/Tribler/Test/Community/Multichain/test_multichain_community.py
+++ b/Tribler/Test/Community/Multichain/test_multichain_community.py
@@ -72,17 +72,17 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         # Since there is a tie breaker for requests, exactly one of the nodes should send a signature request
         failures = 0
         try:
+            _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
+            node.give_message(signature_request, other)
             _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
             other.give_message(signature_request, node)
-            _, signature_response = node.receive_message(names=[HALF_BLOCK]).next()
-            node.give_message(signature_response, node)
         except StopIteration:
             failures += 1
         try:
+            _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
+            other.give_message(signature_request, node)
             _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
             node.give_message(signature_request, other)
-            _, signature_response = other.receive_message(names=[HALF_BLOCK]).next()
-            other.give_message(signature_response, other)
         except StopIteration:
             failures += 1
         self.assertEquals(failures, 1)
@@ -111,17 +111,17 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         # Since there is a tie breaker for requests, exactly one of the nodes should send a signature request
         failures = 0
         try:
+            _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
+            node.give_message(signature_request, other)
             _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
             other.give_message(signature_request, node)
-            _, signature_response = node.receive_message(names=[HALF_BLOCK]).next()
-            node.give_message(signature_response, node)
         except StopIteration:
             failures += 1
         try:
+            _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
+            other.give_message(signature_request, node)
             _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
             node.give_message(signature_request, other)
-            _, signature_response = other.receive_message(names=[HALF_BLOCK]).next()
-            other.give_message(signature_response, other)
         except StopIteration:
             failures += 1
         self.assertEquals(failures, 1)
@@ -151,17 +151,17 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         # Since there is a tie breaker for requests, exactly one of the nodes should send a signature request
         failures = 0
         try:
+            _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
+            node.give_message(signature_request, other)
             _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
             other.give_message(signature_request, node)
-            _, signature_response = node.receive_message(names=[HALF_BLOCK]).next()
-            node.give_message(signature_response, node)
         except StopIteration:
             failures += 1
         try:
+            _, signature_request = other.receive_message(names=[HALF_BLOCK]).next()
+            other.give_message(signature_request, node)
             _, signature_request = node.receive_message(names=[HALF_BLOCK]).next()
             node.give_message(signature_request, other)
-            _, signature_response = other.receive_message(names=[HALF_BLOCK]).next()
-            other.give_message(signature_response, other)
         except StopIteration:
             failures += 1
         self.assertEquals(failures, 1)
@@ -560,7 +560,7 @@ class TestMultiChainCommunity(AbstractServer, DispersyTestFunc):
         other.send_identity(node)
         target_other = self._create_target(node, other)
         # Create a (halfsigned) block
-        node.call(node.community.publish_signature_request_message, target_other, 10, 5)
+        node.call(node.community.sign_block, target_other, 10, 5)
         # Get statistics
         statistics = node.community.get_statistics()
         assert isinstance(statistics, dict), type(statistics)
--- a/Tribler/Test/Community/Multichain/test_multichain_community.py
+++ b/Tribler/Test/Community/Multichain/test_multichain_community.py
@@ -8,17 +8,17 @@ from twisted.internet.defer import inlineCallbacks
 
 from Tribler.Core.Session import Session
 
-from Tribler.community.multichain.community import (MultiChainCommunity, MultiChainCommunityCrawler, SIGNED, HALF_BLOCK,
-                                                    FULL_BLOCK, CRAWL, RESUME)
+from Tribler.Test.Community.Multichain.test_multichain_utilities import MultiChainTestCase
+
+from Tribler.community.multichain.community import MultiChainCommunity, MultiChainCommunityCrawler, HALF_BLOCK, CRAWL
 from Tribler.community.multichain.block import MultiChainBlock, GENESIS_SEQ
 
 from Tribler.community.tunnel.routing import Circuit, RelayRoute
 from Tribler.community.tunnel.tunnel_community import TunnelExitSocket
 from Tribler.Test.test_as_server import AbstractServer
 
-from Tribler.Test.test_multichain_utilities import MultiChainTestCase
-
 from Tribler.dispersy.message import DelayPacketByMissingMember
+
 from Tribler.dispersy.tests.dispersytestclass import DispersyTestFunc
 from Tribler.dispersy.util import blocking_call_on_reactor_thread
 from Tribler.dispersy.tests.debugcommunity.node import DebugNode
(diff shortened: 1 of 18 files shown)


Tribler/community/multichain/block.py
```diff
@@ -0,0 +1,282 @@
...
+        link = database.get_linked(self)
+        prev_blk = (database.get_blocks_until(self.public_key, self.sequence_number - 1, limit=1) or [None])[0]
+        next_blk = (database.get_blocks_since(self.public_key, self.sequence_number + 1, limit=1) or [None])[0]
+
+        # Step 2: determine the maximum validation level
+        if not prev_blk and not next_blk:
+            if self.sequence_number != GENESIS_SEQ and self.previous_hash != GENESIS_ID:
+                # No blocks found, there is no info to base on
+                err("No blocks are known for this member before or after the queried sequence number")
+                result[0] = "no-info"
```
pimotte: This does not match one of the options in the docblock. Should this be no-data, or should the docblock be altered?

Tribler/community/multichain/block.py
```diff
@@ -0,0 +1,282 @@
...
+            if self.total_down + next_blk.down > next_blk.total_down:
+                err("Total down is higher than expected compared to the next block")
+            if next_blk.sequence_number == self.sequence_number + 1 and next_blk.previous_hash != self.hash:
+                err("Next hash is not equal to the hash id of the block")
+
+        # TODO: Detect fraud? But should it go here?
+        # Fraud detection might be easier if you inspect the database for all public key's in one go. (Or rather it
+        # might lead to a lot of double work if you do the fraud detection for each PK individually)
+        # Also if this method returns "invalid" for fraud cases, then those blocks cannot be kept in the database for
+        # future reference or as fraud proofs.
```
pimotte: Maybe we want to distinguish between different kinds of invalidity. Or we might actually want to treat "insane" blocks the same as "invalid because something else was in the db" and in the future be harsh about both?

Tribler/community/multichain/block.py
```diff
@@ -0,0 +1,282 @@
...
+    def validate(self, database):
+        """
+        Validates this block against what is known in the database
+        :param database: the database to check against
+        :return: "valid" if the block does not violate any rules,
+                 "partial-next" if the block does not violate any rules, but there are missing blocks in the future
+                 "partial-prev" if the block does not violate any rules, but there are missing blocks in the past
+                 "partial" if the block does not violate any rules, but there are missing blocks on either side.
+                 "no-data" if there is not enough information known about the block to validate
+                 "invalid" if the block violate any of the rules,
```
pimotte: Maybe extract these validation results to constants, since we check for them elsewhere as well?

pimotte:
General question: It appears you have simplified the bytes/MB storage, but now we are storing the number of bytes in a 32-bit integer, right? Do we want to take this opportunity to switch to 64-bit too?

Tribler/community/multichain/block.py
Captain-Coder: I deliberately choose one level of invalidity. Considering how little I know about the exact implementation details of dispersy, even though we depend on it, I'm assuming future programmers on Tribler might be as clueless as me regarding the finer points of MultiChain blocks. However interpreting invalidity correctly is extremely important. Thus keeping the contract/API as simple as possible is a concern. With one level, you only need one check which you can't mess up.

This is also the reason why the validator returns messages about exactly what makes a block invalid. If there are places where you really want to distinguish about why it is different, you can use the validation strings for this. Perhaps we should make constants of these too (so code checking for specific conditions doesn't break if the actual string is updated).

Tribler/community/multichain/block.py
Captain-Coder: Hummm nice semantic point to consider here? Are we missing data or info? I choose no-info for now.

Captain-Coder:
I see no advantage to counting in MB's over counting in bytes, so I took that out. It's just a nuisance when programming (you have to constantly be aware of what unit you are working with), thus it only serves to introduce errors. Especially since the total counters are already 64 bit when encoded and can store upto 16 exabyte when counting bytes.

The delta up/down of a block is 32 bits when encoded in wire format. In terms of python you can create blocks with any number that python accepts but they might not sign or convert to network format since the numbers can be too big. So the question is how easy it would be to overflow in real situations. Assuming a 1Gbit connection, that is in the range of 93MiB/s. If a hidden tunnel is saturating that, and is open for 10 mins, we have 54GiB total trafic that multichain will try to sign. So yes it is easy to saturate that counter, and I should upgrade it to 64bit.

I'm warry of the alternative (keeping the megabyte divider), since throughout CS history people have said "oh that is enough bits, we'll never fill that" and then ppl went ahead and filled them. Perhaps it is inevitable that advancing coputer tech runs into this sort of boundries, but by spending a few bytes now we delay that problem a whole lot.

Captain-Coder:
retest this please

Captain-Coder:
retest this please

Captain-Coder:
retest this please

lfdversluis:
Should be visible in the runner's output

Captain-Coder:
retest this please, Jenkins had weird results on the windows testers

devos50:
@Captain-Coder win32 rebooted somehow, should work now again

devos50:
retest this please

Captain-Coder:
Thanks, that worked. I'm getting the expected results now.

synctext:
Side-by-side performance graphs would be cool. Creating blocks at full CPU speed. Thesis story that your cleaner code still retains the 10000 signs per second...

whirm:
Is this PR ready? Is the rest of the multichain group happy with it? has it been reviewed by somebody? I can't see on where it got set to READY on my history? :confused:

Captain-Coder:
@pimveldhuisen care to do the honours?

(/me expects much ribbon cutting and clapping)

whirm:
And don't forget the fireworks

pimveldhuisen:
I really like this change to the multichain, it adds a lot of elegance, the code quality seems very high and it is appropriately tested. There are two that I think deserve some disscusion still.

First there is the change from MB's to bytes. The size increase seems fine, 64 extra bits on a existing total of 2544 bits is only an 2.5% increase. However, the choice for MB's also had a semantic meaning I think; this is the level of accuracy the multichain aims to work on. It should be exactly this amount of MB's. This is especially important since both parties must agree on the exact amount to sign, what do you do if the count is a 100 bytes off? On the other hand, it seems possible to just count accurately and match the exact amount. 64 bits seems also enough to keep track of total up and down; with an 1Gbit/s connection that is active 24/7 this should be enough for roughly 4000 years. (2^64 / ( 1024^3/8 ) / 3600 / 24 / 365 = 4358 ) With the growth of internet speed this will be less, but we should be fine here.
For the sequence numbers, we should be good for 800 years, assuming a block every 10 minutes on 100 parallel connections.  (2^(4*8) / (6 * 100 * 24 * 365) =  817)
So overall switching to byte based counting seems a good idea? This could ofcourse be a seperate PR, but it might be cleaner to include it here?

The there is the issue of symmetry. While the data format is now symmetrical, the protocol / message sequence still is not fully symmetrical.  There still is a clear difference between the requester and the responder, where the requester sends a message first, and the responder then makes a block that links to that block. It might be possible to avoid even this asymmetry by using the tunnel_id as a link. This might seem like a loss of generality, but in fact by definition multichain describes some previous interaction. This interaction would often have some way to uniquely describe it. For example, in a regular BitTorrent transfer, you could hash the concatenation of the hashes of the transferred blocks. (Even more generally, you can always hash the data the was transferred) . This does however require a clear moment in time when the data is signed.

Neither issue is strictly related to this PR, but it might be nice to make this the last compatibility breaking change for the foreseeable future. What are your thoughts?

Captain-Coder:
I removed the MB's thinking mostly from a code perspective. It simplifies things if you don't have to worry about the scale of unit you are working with. I see your point about the scale that multichain is intended to work at. Which makes me think about the byte accounting on the (hidden)tunnel community, which also includes control messaging. This can lead to a few hundred bytes difference up and down (for example the tunnel extend message that is not counted correctly). To keep things simple for the programmer & future improvements, I propose we keep the byte counting. For the accuracy, I propose we round down to KB's, that should give us ample smudge-factor for tunnels counting things oddly.

My main objective was to make it "symmetrical" in terms of point of view, ie: a block always represents the point of view of it's signer. This eliminates all the "am I requester or responder?" branches while programming. It is true that someone still has to initiate the signing and another reacts. It would simplify things if we could do away with that. The fully symmetrical schemes that you propose only work if you want to account single (uniquely identifiable) interactions. Aggregating balance for an hour/day/week is not possible in such a scenario. Since this might well be a direction for future multichain development, I think that at this time we should not restrict multichain to signing unique interactions.

synctext:
@qstokkink  Would you have time to review this code? 
Please be precise, as this database structure hopefully will evolve towards the basic layer for trusted peer discovery (2017 ish). Thnx

qstokkink:
I have assembled all of the notes of my review in this post, for easier future reference. I have split the review up into high-level and low-level feedback, as the former might be my oversight.

### High-level feedback
In the case of double or counter-sign fraud detection, how does the chain **recover** when valid blocks are shared with a node after it has received and stored one or more fraudulent ones? It seems like it doesn't.

### Low-level feedback
Functionally, I did not spot any typos, wrong-variable-references (adding bytes-up to bytes-down or something) and the like. This seems to be all good.

As for code style/structure, I have the following comments:
 1. I would refactor [MultiChainBlock.validate](https://github.com/Captain-Coder/tribler/blob/e6ce5a49dbbbeac8b75129af1c5a5e8ce06f0512/Tribler/community/multichain/block.py#L77) into multiple functions. This is for both readability and maintainability of the code. Also, since this is a pretty important function (and therefore will probably trigger lots of errors in the future, with experimental use of this code) I would add more in-depth commentary to the steps. More-or-less how it is done in [Dispersy's bloomfilter](https://github.com/Tribler/dispersy/blob/devel/bloomfilter.py).
 2. I would wrap [VALID, PARTIAL, PARTIAL_NEXT, PARTIAL_PREV, NO_INFO and INVALID](https://github.com/Captain-Coder/tribler/blob/e6ce5a49dbbbeac8b75129af1c5a5e8ce06f0512/Tribler/community/multichain/block.py#L19) in an enumerator class, because that is what they are: enumeration values. You can then move the commentary from [the return value of MultiChainBlock.validate](https://github.com/Captain-Coder/tribler/blob/e6ce5a49dbbbeac8b75129af1c5a5e8ce06f0512/Tribler/community/multichain/block.py#L81) into this class, this keeps it cleaner.
 3. I would split up [database.py](https://github.com/Captain-Coder/tribler/blob/e6ce5a49dbbbeac8b75129af1c5a5e8ce06f0512/Tribler/community/multichain/database.py) into a database and a database upgrader, this is also in line with the rest of Tribler. You might even consider moving the upgrade portion to [Tribler/Core/Upgrade](https://github.com/Tribler/tribler/tree/devel/Tribler/Core/Upgrade). This is mainly for making the code future-proof.
 4. Finally, why did you insert a line into [Tribler/Test/test_as_server.py](https://github.com/Captain-Coder/tribler/blob/e6ce5a49dbbbeac8b75129af1c5a5e8ce06f0512/Tribler/Test/test_as_server.py#L50)?

To conclude, answering the first post i.t.t.: having Dispersy use its member authentication _could_ be useful when deciding to ban members based on Dispersy member id instead of IP, regarding accountability. You should be safe to switch to NoAuthentication however; this should also boost your performance.

Captain-Coder:
Thanks for the review!

So about the fraud detection. It doesn't matter what we do with the offending blocks since we are gonna ignore/ban the offender anyway and just disregard his blocks (@synctext banned me from implementing this -.- ). To the rest of the world blocks from the counter parties of the fraudulent node will just look like half-blocks that never got signed, they have no value. So I would say it is pretty much self recovering?

On the more technical side I was planning to do the following (until I was stopped):

- We need a place to store fraud proofs. We don't want to soil the main multi_chain table with crap, and in the future we might well decide to prune the MC table in which case we don't want to prune fraud proofs. So we should probably put it in a separate table. Validate() can construct and insert this proof.
- Create a message to transport fraud proofs, this will take two blocks and possibly an indication of the type of fraud and proof.
- Whenever we send blocks to someone, we first send any fraud proofs for the involved identities. (Elric mentioned a similar system that is currently in dispersy for fraud with the lamport clock)
- Whenever we use blocks locally to determine a rating for someone, we ofcourse first filter out blocks from the fraudulent node.

As for the low level stuff:

1. I've been thinking about splitting it up too, and pylint keeps giving me flack about this. I agree on this in principle, and I will add commentary. However, splitting up the method makes me worry that in the future, someone will come along and without much thought perform only a partial validation. The consecuences of misusing the validator can be severe. Because python doesn't have the true "private" concept from OO, it's hard to enforce proper use of the validator if it is broken up. Perhaps I'm being overly future minded or pessimistic about future programmer competence though.
2. Yup, seems like a good plan.
3. Seems like a prudent idea, being compatible with the rest of Tribler :) But right now our upgrade is so simple (a single line), it might be a bit too much future engineering to get this in place now. I'll discuss this with the team.
4. o.O  ... I have no idea how that empty line ended up there. I'll take it out.

Tribler/community/multichain/block.py
```diff
@@ -78,68 +71,82 @@ def validate(self, database):
...
-            elif prev_blk.sequence_number != self.sequence_number - 1:
-                result[0] = PARTIAL_PREV
-            elif next_blk.sequence_number != self.sequence_number + 1:
-                result[0] = PARTIAL_NEXT
+            if is_prev_gap and is_next_gap:
+                result[0] = ValidationResult.partial
+            elif is_prev_gap:
+                result[0] = ValidationResult.partial_previous
+            elif is_next_gap:
+                result[0] = ValidationResult.partial_next()
```
qstokkink: Lose the parentheses. Everything else looking good :+1:

Captain-Coder: hehehe, I pushed accidentally. I wasn't done yet -.- This and a bunch of other small bits where all caught when runninng the unit tests.

Captain-Coder:
Alright, I think I processed all the comments. Even the coverage diff check passes now. ^.^ Does anyone want to do a final review before we set this to ready?

lfdversluis:
@Captain-Coder Setting it to ready will make @devos50 or @whirm review it anyway :p

devos50:
I see various pylint issues that could easily be fixed (bad whitespace, else if used, bad builtin etc). Can you fix these? After that's done, I will review and merge 👍

Captain-Coder:
Except the white space ones, it's all very deliberate.

- The "else if" is used for readability in validate(). There are 4 cases depending on the existence of next/previous blocks, in the last case where both exist there is further testing to be done. It makes no sense to promote these checks to the level of the 4 main cases. Thus I feel that else with an indented if block is more readable here.
- The bare except has a comment next to it that also explains why it is bare
- The bad builtin is from a map(), that according to pylint should be a list comprehension. But for the life of me I can't figure out how to write a list comprehension that takes parameters from two lists. (Yeah I can use the zip()-builtin to combine the lists in a list of tuples first, but that is just trading one builtin for another.)
- I fixed the whitespace ones but apparently forgot to push.
- I got rid of the no-member one while I was at it.

devos50:
@Captain-Coder as we discussed yesterday, please make sure that the commit history is clean enough (so no fixes on fixes). If you're done, please tag me and I will continue the review.

Also, do you want to fix the bugs that you discovered in this PR or in a separate one?

synctext:
retest this please

synctext:
Do we know why Mac was failing?

devos50:
@synctext it is not rebased (yet) so some test suite fixes are not in this PR.

devos50:
@Captain-Coder please rebase and mark this PR as ready when you fixed the bug in the tests.

Also, apparently I've already reviewed this PR so I will skim through it and focus on the newest code (regarding the API endpoints).

devos50:
We should make sure that https://github.com/Tribler/dispersy/issues/536 is fixed before merging this.

Captain-Coder:
I've rebased to the most recent devel branch and fixed the unit tests (no more sleep() calls now). Regarding the Tribler/dispersy#536 issue: it's reusing the master member that is the fault here. Since this version of multichain is clearly incompatible with the previous version, I also updated the master member id for this version of multichain. That should atleast avoid users from having this issue.

Tribler/Test/Community/Multichain/test_multichain_community.py
```diff
@@ -558,34 +512,100 @@ def test_get_statistics_with_previous_block(self):
...
+
+    @staticmethod
+    def crawl_node(crawler, crawlee, target_to_crawlee, sequence_number=None):
+        crawler.call(crawler.community.send_crawl_request,
+                     target_to_crawlee, crawlee.my_member.public_key, sequence_number)
+        crawlee.give_message(crawlee.receive_message(names=[CRAWL]).next()[1], crawler)
+        TestMultiChainCommunity.transport_halfblocks(crawlee, crawler)
+
+    @staticmethod
+    def transport_halfblocks(source, sink):
```
devos50: I find the `sink` as variable name quite vague; please use `destination` or something similar instead.

Tribler/Test/Community/Multichain/test_multichain_community.py
```diff
@@ -558,34 +512,100 @@ def test_get_statistics_with_previous_block(self):
...
+    def clean_database(node):
+        node.community.persistence.execute(u"DELETE FROM multi_chain;")
 
+    @blocking_call_on_reactor_thread
+    @inlineCallbacks
     def create_nodes(self, *args, **kwargs):
-        return super(TestMultiChainCommunity, self).create_nodes(*args, community_class=MultiChainCommunity,
+        nodes = yield super(TestMultiChainCommunity, self).create_nodes(*args, community_class=MultiChainCommunity,
                                                                  memory_database=False, **kwargs)
+        for x in nodes:
```
devos50: Please use something more descriptive than `x` and `y`.

Tribler/Test/Community/Multichain/test_multichain_database.py
```diff
@@ -106,138 +66,79 @@ def test_contains_block_id_positive(self):
         # Act
         self.db.add_block(self.block1)
         # Assert
-        self.assertTrue(self.db.contains(self.block1.hash_requester))
+        assert self.db.contains(self.block1)
```
devos50: Please use `assertTrue` instead

Tribler/Test/Community/Multichain/test_multichain_database.py
```diff
@@ -106,138 +66,79 @@ def test_contains_block_id_positive(self):
...
-        # Make sure that there is a responder block with a lower sequence number.
-        # To test that it will look for both responder and requester.
-        self.db.add_block(self.block1)
-        self.block2.public_key_responder = self.block1.public_key_requester
-        self.block2.sequence_number_responder = self.block1.sequence_number_requester - 5
-        self.db.add_block(self.block2)
-        # Act & Assert
-        self.assertEquals(self.db.get_latest_sequence_number(self.block1.public_key_requester),
-                          self.block1.sequence_number_requester)
+        assert not self.db.contains(self.block1)
```
devos50: Please use `assertFalse` instead

Tribler/community/multichain/community.py
```diff
@@ -58,340 +75,221 @@ def initialize(self, tribler_session=None):
...
-            b. False, if not (because of inconsistencies in the payload)
-            :param request The original message as send by this node
-            :param response The response message received
-            :param modified (bool) True if the message was modified
-        """
-        if not response:
-            self.logger.info("Timeout received for signature request.")
-            return False
+        # NOTE to the future: This method reads from the database, increments and then writes back. If in some future
+        # this method is allowed to execute in parallel, be sure to lock from before .create upto after .add_block
```
devos50: upto -> up to

Tribler/community/multichain/community.py
```diff
@@ -400,7 +298,7 @@ def unload_community(self):
     def on_tunnel_remove(self, subject, change_type, tunnel, candidate):
         """
         Handler for the remove event of a tunnel. This function will attempt to create a block for the amounts that
-        were transferred using the tunnel.
+        where transferred using the tunnel.
```
devos50: where -> were

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

kc1212:
Hi @Captain-Coder, I will mentioned True Halves in my thesis. Do you have some work which I can cite, or should I just cite this PR?

Captain-Coder:
I have nothing published (or in citable form) yet. @pimveldhuisen has a true halves section in his thesis i think. Perhaps you can cite that. Otherwise feel free to referrence this thread but i don't know how scientificly relevant it would be in your thesis.

pimveldhuisen:
@kc1212 My thesis indeed describes the Multichain protocol, including to some extent the true halves.
Feel free to cite it: https://repository.tudelft.nl/islandora/object/uuid%3A0bd2fbdf-bdde-4c6f-8a96-c42077bb2d49?collection=education

PR #2141, opened 2016-04-27
Added Tribler twistd plugin

devos50:
This PR adds a twistd plugin so Tribler can be started using the `twistd` command. To run the plugin, `cd` to the root folder of Tribler (not the `Tribler` directory) and execute: `PYTHONPATH=twisted twistd -n tribler`. This will run the twistd daemon.

Note that this PR requires https://github.com/Tribler/dispersy/pull/474 to work (otherwise, Tribler hangs during startup).

@whirm Should I cover the code in this PR with a test? I'm not sure how I should write a test for this plugin.

### DIFF
--- a/twisted/twisted/plugins/tribler_plugin.py
+++ b/twisted/twisted/plugins/tribler_plugin.py
@@ -1,0 +1,91 @@
+"""
+This twistd plugin enables to start Tribler headless using the twistd command.
+"""
+import logging
+import os
+import signal
+from time import sleep
+
+from twisted.application.service import MultiService, IServiceMaker
+from twisted.conch import manhole_tap
+from twisted.internet import reactor
+from twisted.plugin import IPlugin
+from twisted.python import usage
+from twisted.python.log import msg
+from zope.interface import implements
+
+from Tribler.Core.Session import Session
+from Tribler.Core.SessionConfig import SessionStartupConfig
+
+
+class Options(usage.Options):
+    optParameters = [
+        ["manhole", "m", 0, "Enable manhole telnet service listening at the specified port", int],
+    ]
+
+
+class TriblerServiceMaker(object):
+    implements(IServiceMaker, IPlugin)
+    tapname = "tribler"
+    description = "Tribler twistd plugin, starts Tribler as a service"
+    options = Options
+
+    def __init__(self):
+        """
+        Initialize the variables of the TriblerServiceMaker and the logger.
+        """
+        self.session = None
+        self._stopping = False
+
+    def start_tribler(self):
+        """
+        Main method to startup Tribler.
+        """
+
+        def signal_handler(sig, _):
+            msg("Received shut down signal %s" % sig)
+            if not self._stopping:
+                self._stopping = True
+                self.session.shutdown()
+                msg("Tribler shut down")
+                reactor.stop()
+        signal.signal(signal.SIGINT, signal_handler)
+        signal.signal(signal.SIGTERM, signal_handler)
+
+        msg("Starting Tribler")
+
+        config = SessionStartupConfig()
+        config.set_http_api_enabled(True)
+        self.session = Session(config)
+        upgrader = self.session.prestart()
+        if upgrader.failed:
+            msg("The upgrader failed: .Tribler directory backed up, aborting")
+            reactor.addSystemEventTrigger('after', 'shutdown', os._exit, 1)
+            reactor.stop()
+        else:
+            self.session.start()
+            msg("Tribler started")
+
+    def makeService(self, options):
+        """
+        Construct a Tribler service.
+        """
+        tribler_service = MultiService()
+        tribler_service.setName("Tribler")
+
+        manhole_namespace = {}
+        if options["manhole"]:
+            port = options["manhole"]
+            manhole = manhole_tap.makeService({
+                'namespace': manhole_namespace,
+                'telnetPort': 'tcp:%d:interface=127.0.0.1' % port,
+                'sshPort': None,
+                'passwd': os.path.join(os.path.dirname(__file__), 'passwd'),
+            })
+            tribler_service.addService(manhole)
+
+        reactor.callWhenRunning(self.start_tribler)
+
+        return tribler_service
+
+service_maker = TriblerServiceMaker()


devos50:
@lfdversluis can you review?

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+
+        def signal_handler(sig, _):
+            msg("Received shut down signal %s" % sig)
+            if not self._stopping:
+                self._stopping = True
+                try:
+                    self.session.shutdown()
+                    msg("Tribler shut down")
+                except Exception, exc:
```
lfdversluis: general exception, can you replace this with a tuple?

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,94 @@
...
+            port = options["manhole"]
+            manhole = manhole_tap.makeService({
+                'namespace': manhole_namespace,
+                'telnetPort': 'tcp:%d:interface=127.0.0.1' % port,
+                'sshPort': None,
+                'passwd': os.path.join(os.path.dirname(__file__), 'passwd'),
+            })
+            tribler_service.addService(manhole)
+
+        reactor.exitCode = 0
```
lfdversluis: Why this statement?

twisted/twisted/plugins/tribler_plugin.py
devos50: I've removed the try/catch block. We are not catching exceptions during Tribler shutdown in other files (`tribler_main.py`, `test_as_server.py`, `tunnel/main.py` etc).

twisted/twisted/plugins/tribler_plugin.py
devos50: I copied most of the file from the Dispersy twistd plugin and this line got copied over. I removed it, tested the shutdown procedure and it seems that the statement is indeed not necessary.

devos50:
@whirm can you review?

whirm:
re. tests: Have a look at how the twisted guys test their stuff, maybe you can get some ideas.

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+class Options(usage.Options):
+    optParameters = [
+        ["manhole", "m", 0, "Enable manhole telnet service listening at the specified port", int],
+    ]
+
+
+class TriblerServiceMaker(object):
+    implements(IServiceMaker, IPlugin)
+    tapname = "tribler"
+    description = "Tribler twistd plugin, starts Tribler as a twistd process"
```
whirm: "as a service"

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+    options = Options
+
+    def __init__(self):
+        """
+        Initialize the variables of the TriblerServiceMaker and the logger.
+        """
+        self.session = None
+        self._stopping = False
+        self._logger = logging.getLogger(self.__class__.__name__)
```
whirm: This is not being used

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+        signal.signal(signal.SIGINT, signal_handler)
+        signal.signal(signal.SIGTERM, signal_handler)
+
+        msg("Starting Tribler")
+
+        config = SessionStartupConfig()
+        config.set_http_api_enabled(True)
+        self.session = Session(config)
+        upgrader = self.session.prestart()
+        while not upgrader.is_done:
```
whirm: upgrader.run() is blocking, so this while loop does nothing. Also, you are not checking if the upgrade failed.

Another issue we probably have is that the notifications are going to pile up until the upgrade process is finished (AFAIR notify schedules the notifiation stuff to be called on the thread pool which will be scheduling stuff to be called on the reactor thread, which (as the upgrade is blocking ATM) will be busy until the upgrade finishes and then all the notifications the upgrader generated will be delivered (to no one until you implement it on the REST interface). So remember that we need to "twistify" the upgrade code before implementing the REST notifications for it.

whirm: re. not checking if the upgrade failed: I know the upgrader will move the .Tribler dir away and start from scratch, but do we really want this to happen on a headless server?

devos50: If the upgrader fails, I think we need to do something to make sure Tribler starts up. Starting with a fresh .Tribler directory is the best choice probably, whether we are running headless or not.

Are there any other advantages of not starting from scratch when running headless?

whirm: The difference is that when it's running headless, the user won't know if he doesn't check the log file immediately after starting the service. I guess we should make it fail so the user notices.

devos50:
@whirm see my last fixup commit where I made the plugin fail if upgrading fails. I also updated the dispersy pointer.

Also, I'm unable to find tests in the Twisted source code for their shipped plugins, only for their core functionality. Also, it lives outside Tribler and it's only a utility script to startup Tribler so I don't think it really needs a dedicated test case.

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+            port = options["manhole"]
+            manhole = manhole_tap.makeService({
+                'namespace': manhole_namespace,
+                'telnetPort': 'tcp:%d:interface=127.0.0.1' % port,
+                'sshPort': None,
+                'passwd': os.path.join(os.path.dirname(__file__), 'passwd'),
+            })
+            tribler_service.addService(manhole)
+
+        reactor.callWhenRunning(self.start_tribler)
```
whirm: I'm thinking this is not needed for a normal reactor, you should just do everything here as the reactor will be running.

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+
+        msg("Starting Tribler")
+
+        config = SessionStartupConfig()
+        config.set_http_api_enabled(True)
+        self.session = Session(config)
+        upgrader = self.session.prestart()
+        if upgrader.failed:
+            msg("The upgrader failed: .Tribler directory backed up, aborting")
+            reactor.stop()
```
whirm: You shouldn't stop the reactor, that's twistd's job.

If raising an exception once you have inlined this with `makeService()` doesn't make twistd exit with nonzero status, then you could do something like this:

```python
reactor.addSystemEventTrigger('after', 'shutdown', os._exit, EXIT_CODE)
```

twisted/twisted/plugins/tribler_plugin.py
devos50: The reactor is not running when I schedule the `self.start_tribler` call (tested with `reactor.running`).

devos50:
@whirm see my last fixup commit. I've also updated the dispersy pointer to include the netifaces fix.

whirm:
Ok, squash it

devos50:
retest this please

devos50:
(anon tunnel tests timeout)

devos50:
@whirm ready

PR #2142, opened 2016-04-27
READY: Implemented endpoint to fetch Tribler settings

devos50:
In this PR, I've implemented the next endpoint: getting the settings present in Tribler. The fact that we are using three different configuration files makes it harder to return a single JSON dictionary with all the settings. Nevertheless, this endpoint returns a single JSON dictionary that should contain all settings needed for the new GUI.

I also made some minor changes to other files. For instance, I created a static method `get_literal_value` in `configparser.py` and I've added a method in that class to return a configuration file as dictionary. Moreover, I've written a test to check the new settings endpoint and extended the documentation.

### DIFF
--- a/Tribler/Core/Modules/restapi/root_endpoint.py
+++ b/Tribler/Core/Modules/restapi/root_endpoint.py
@@ -1,5 +1,6 @@
 from twisted.web import resource
 from Tribler.Core.Modules.restapi.my_channel_endpoint import MyChannelEndpoint
+from Tribler.Core.Modules.restapi.settings_endpoint import SettingsEndpoint
 
 
 class RootEndpoint(resource.Resource):
@@ -14,3 +15,6 @@ class RootEndpoint(resource.Resource):
 
         self.my_channel_endpoint = MyChannelEndpoint(self.session)
         self.putChild("mychannel", self.my_channel_endpoint)
+
+        self.settings_endpoint = SettingsEndpoint(self.session)
+        self.putChild("settings", self.settings_endpoint)
--- a/Tribler/Core/Modules/restapi/settings_endpoint.py
+++ b/Tribler/Core/Modules/restapi/settings_endpoint.py
@@ -1,0 +1,66 @@
+import json
+import os
+
+from twisted.web import resource
+
+from Tribler.Core.Utilities.configparser import CallbackConfigParser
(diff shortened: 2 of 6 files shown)


devos50:
@lfdversluis ready for review

whirm:
Don't you think it would be better to leave this one out until we refactor the config stuff so it's not insane anymore? 

Do you need to access the settings right now?

devos50:
@whirm I don't think it's insane (only around 40 lines of code to return the dictionary I need) and I'm using it for the new GUI already so I would like to have this endpoint implemented :)

whirm:
What's insane is the current configuration system in Tribler, not your code.

devos50:
@whirm yeah it's absolutely insane and there are some duplicates in the config (and probably unused config options) but for now we have to deal with it. It was not that hard to actually return a single JSON dictionary and I'm sure we can do it in way less code once we have a proper configuration system.

Nevertheless, I would like to have this endpoint being part of the code. Also, I'm not sure when exactly I'm going to rewrite the configuration system (maybe around the time a 6.6 alpha is released?) and my main focus currently is the new GUI and the API in Tribler.

whirm:
I don't like that you are mixing persisted values with runtime values with no way to tell them apart. I would rather have the configuration endpoint deal with the actual config (as in getting/setting -1 for random ports and suchlike) and have something else to obtain runtime values.

whirm:
Also, saving this config would persist the randomized value.

devos50:
@whirm what do you think about a separate endpoint to fetch the ports?

whirm:
What about a generic interface for runtime variables?

devos50:
@whirm that's ok but I'm not sure about the name for the endpoint. Do you have a suggestion?

whirm:
`variables`? Maybe having events trigger whenever those variable change could be enough?

devos50:
@whirm variables sounds ok. I will create a separate PR for it and remove the assignment in this PR.

devos50:
@whirm see my last fixup commit.

whirm:
Ok, squash it!

devos50:
retest this please

devos50:
(that was due to a connection error on the mac)

devos50:
@whirm ready (I also updated the dispersy pointer to include the netifaces fix)

PR #2162, opened 2016-05-04
Moved multichain tests to Tribler/Test/Community/Multichain directory

devos50:
I think they belong in this directory instead of the root directory of the tests :)


devos50:
@whirm I'm done with this PR.

PR #2168, opened 2016-05-06
Fix NSIS running twice

### LINKED ISSUE #2110: Installer runs twice
While running the new builder installer I noticed that Tribler (and possibly also the old installer) runs the installation and uninstallation section twice. This is confusing and has to be investigated :)

lfdversluis:
So after quite a while I managed to get rid of the double running problem. The problem was in the UAC plugin and after having a nice chat with the author Anders (nice guy by the way :D) he told me to sacrifice the check box at the last page and get rid of the UAC plugin.

So I got rid of the UAC and kept the check box anyway that now calls the desktop shortcut that is created during installation and works perfect. Tested on proxmox 144 and works splendid.
While I was at it, I formatted the file because it was 2, 3 and 4 space indented mixed with tabs.
Also added a bit of comments to let future generations know why decisions were made and what it does.

Oh... ignore most fixup commits as they feature tries by me to either identify the root cause of this bug, get to know how stuff works, debug items or simply to test things.

Fixes **1** blocker bug
Comments and formatting improved (enhancement)
Got rid of UAC plugin (enhancement)

Fixes #2110

### DIFF
--- a/Tribler/Main/Build/Win/tribler.nsi
+++ b/Tribler/Main/Build/Win/tribler.nsi
@@ -1,5 +1,6 @@
 !define PRODUCT "Tribler"
-; Laurens, 2016-03-14: The __GIT__ string will be replaced by update_version_from_git.py with the current version of the build.
+; Laurens, 2016-03-14: The __GIT__ string will be replaced by update_version_from_git.py 
+; with the current version of the build.
 !define VERSION "__GIT__"
 ; Laurens, 2016-03-14: The _x86 will be replaced by _x64 if needed in update_version_from_git.py
 !define BITVERSION "x86"
@@ -8,6 +9,8 @@
 !include "FileFunc.nsh"
 !include "nsProcess.nsh"
 
+; Laurens, 2016-04-06: We are going to possibly touch the C:/ drive (by default we do) and it's restricted areas,
+; so we need admin permission to do so (for most devices).
 RequestExecutionLevel admin
 
 ;--------------------------------
@@ -18,10 +21,12 @@ Name "${PRODUCT} ${VERSION}"
 OutFile "${PRODUCT}_${VERSION}_${BITVERSION}.exe"
 
 ;Folder selection page. 
-; Laurens, 2016-03-14: Note that $PROGRAMFILES will be replaced by $PROGRAMFILES64 if needed from update_version_from_git.py.
+; Laurens, 2016-03-14: Note that $PROGRAMFILES will be replaced by $PROGRAMFILES64 
+; if the 64 bit argument is passed to update_version_from_git.py.
 InstallDir "$PROGRAMFILES\${PRODUCT}"
 
-; Laurens, 2016-03-15: No silent mode for the installer and uninstaller because this will disbale the init functions being called.
+; Laurens, 2016-03-15: No silent mode for the installer and uninstaller because 
+; this will disbale the init functions being called.
 SilentInstall normal
 SilentUnInstall normal
 
@@ -45,14 +50,14 @@ BrandingText "${PRODUCT}"
 ; Tribler running check - shared function
 !macro RUNMACRO un
   Function ${un}checkrunning
-	DetailPrint "Checking if Tribler is not running..."
-	checkRunning:
-		${nsProcess::FindProcess} "tribler.exe" $r0
-    		StrCmp $r0 0 0 notRunning
-    		MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "${PRODUCT} is running, please close it so the (un)installation can proceed." /SD IDCANCEL IDRETRY checkRunning
-    		Abort
+     DetailPrint "Checking if Tribler is not running..."
+     checkRunning:
+          ${nsProcess::FindProcess} "tribler.exe" $r0
+                StrCmp $r0 0 0 notRunning
+                MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "${PRODUCT} is running, please close it so the (un)installation can proceed." /SD IDCANCEL IDRETRY checkRunning
+                Abort
 
-  	notRunning:
+        notRunning:
   FunctionEnd
 !macroend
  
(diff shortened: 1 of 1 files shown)


lfdversluis:
@devos50 can you review?

Tribler/Main/Build/Win/tribler.nsi
```diff
@@ -110,179 +109,178 @@ LangString DESC_SecDefaultMagnet ${LANG_ENGLISH} "Associate magnet links with ${
...
+    ; Now writing to KHEY_LOCAL_MACHINE only -- remove references to uninstall from current user
+    DeleteRegKey HKEY_CURRENT_USER "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT}"
+    ; Remove old error log if present
+    Delete "$INSTDIR\tribler.exe.log"
+
+    WriteUninstaller "$INSTDIR\Uninstall.exe"
+
+    ; Add an application to the firewall exception list - All Networks - All IP Version - Enabled
+    SimpleFC::AddApplication "Tribler" "$INSTDIR\${PRODUCT}.exe" 0 2 "" 1
+    ; Pop $0 ; return error(1)/success(0)
```
devos50: If we are not using this code, maybe we should just remove it? I saw some more occurrences of commented-out code.

Tribler/Main/Build/Win/tribler.nsi
```diff
@@ -325,72 +322,39 @@ SectionEnd
...
-Function un.onInit
-	SetShellVarContext all
-    	!insertmacro Init "uninstaller"
 FunctionEnd
 
 Function PageFinishRun
-	!insertmacro UAC_AsUser_ExecShell "" "$INSTDIR\tribler.exe" "" "" ""
+    ; This function is invoked if the user leaves the checkbox checked at the last page (do you wish to run...)
+    ; It calls the freshly created desktop shortcut
+    ExecShell "" "$DESKTOP\${PRODUCT}.lnk"
```
devos50: Is a desktop shortcut always created? Is it possible to run `tribler.exe` immediately from the installed directory?

devos50:
Nice to see this one solved! Really annoying to have a bug in a plugin that we are using...

In the future, can you squash the commits first before I review? That makes it a little easier for me to review by commit (so I can distinguish better between bug fixes and formatting fixes).

Tribler/Main/Build/Win/tribler.nsi
lfdversluis: yes the shortcut is always created. As far as I am aware not, as it was ran with admin permission previously.

lfdversluis:
@whirm prefers to leave the fixups, at least when a review has happened. Before it may be ok to squash

Tribler/Main/Build/Win/tribler.nsi
lfdversluis: Legacy stuff. I guess it can go now, yes.

synctext:
admin level access level... Does that alter the Windows install experience? 

noce fix!

lfdversluis:
@synctext No, currently users will also see the "require access" screen because we are touching the restricted C:/ folders. The situation will only improve now as previously two processes ran (which you also saw in the taskmanager), now it's down to one :)

whirm:
> @whirm prefers to leave the fixups, at least when a review has happened. Before it may be ok to squash

I prefer to have fixup commits while I'm reviewing. If you want to be extra nice to each other, you guys can do a squash before @devos50 / @lfdversluis reviews and another before I review.

lfdversluis:
@whirm squashed, ready for your review.

Tribler/Main/Build/Win/tribler.nsi
```diff
@@ -1,16 +1,15 @@
 !define PRODUCT "Tribler"
-; Laurens, 2016-03-14: The __GIT__ string will be replaced by update_version_from_git.py
+; Laurens, 2016-03-14: The __GIT__ string will be replaced by update_version_from_git.py with the current version of the build.
```
whirm: Please, 120 cols if possible.

Tribler/Main/Build/Win/tribler.nsi
```diff
@@ -110,179 +109,175 @@ LangString DESC_SecDefaultMagnet ${LANG_ENGLISH} "Associate magnet links with ${
 ;Installer Sections
 
 Section "!Main EXE" SecMain
- SectionIn RO
```
whirm: Please, no reformatting for untouched code

whirm:
Have you tested it on win 7, 8, 10?

lfdversluis:
Only on win10, do we don't have proxmox machines running 7 and 8. I can test it on my win 7, but then 8 is still untested. So either I need to install a VM or a proxmox with 8 :D

Tribler/Main/Build/Win/tribler.nsi
lfdversluis: So I need to undo all the identation mess fixes for lines I didn't modify? :/

whirm:
> So I got rid of the UAC and kept the check box anyway that now calls the desktop shortcut that is created during installation and works perfect.

Does it run Tribler as Admin?

lfdversluis:
Good question, have to investigate that one. I will do this in the evening today.

lfdversluis:
It ran under my user, so I assume user rights, couldn't see a setting admin vs user permission though. Tested on win 7 so only win 8 to verify.

lfdversluis:
@whirm I removed the long comment lines > 120 chars. The refactoring is in a separate commit as discussed yesterday. Do you have a win-8 that can test the last build of May 6? You should see some alertboxes if you install, those are gone now.

whirm:
> It ran under my user, so I assume user rights, couldn't see a setting admin vs user permission though. Tested on win 7 so only win 8 to verify.

Look at the process manager, if Tribler gets executed as admin when the install finishes, that would be really bad, and that's probably what's happening if the author told you about losing the checkbox at the end.

lfdversluis:
Found the option to enable UAC in the manager. It is indeed running as admin. I propose to let the checkbox go then as Anders argued.

whirm:
Ok

lfdversluis:
Removed the run page, tested it on win 7 and 10

lfdversluis:
@whirm Could you test the installer on a win 8 machine? You can grab it here: https://jenkins.tribler.org/job/Builder_64_laurens/

whirm:
I don't have a windows 8. only vista and 10.

whirm:
I don't have a windows 8. only vista and 10.

lfdversluis:
https://jenkins.tribler.org/job/pers/job/Builder_64_laurens/ now

lfdversluis:
Works on win 8

whirm:
Great!, can you squash it?

PR #2173, opened 2016-05-08
Created a new process checker

### LINKED ISSUE #2148: SingleInstanceChecker is not properly implemented
It even detects pylint processes as Tribler's.

It should be reimplemented by writing a `lock` file containing the current process PID to the state dir. This would eliminate false positives and several Tribler processes using different state dirs could coexist without need of disabling the single instance check.

Note that the current implementation makes some tests randomly fail when mistaking pylint for a Tribler process:

```
Tribler.Test.test_i2i_download.TestI2IDownload.test_i2i_running (from nosetests)
Failing for the past 1 build (Since Failed#960 )
Took 5.1 sec.
Error Message

Did not even create a session
DEBUG   1461689740  test_as_server:TestI2IDownload:559  waiting for session instance
INFO    1461689740 SingleInstanceChecker:SingleInstanceChecker:38   Linux cmd returned 16651 pylint
16652 pylint
16659 pylint
16660 pylint
16661 pylint
16662 pylint
16663 pylint
16664 pylint
16665 pylint
16666 pylint
INFO    1461689740    tribler_main:Tribler.Main.tribler_main:1046 Client shutting down. Detected another instance.
INFO    1461689740    tribler_main:Tribler.Main.tribler_main:1066 Client shutting down. Sleeping for a few seconds to allow other threads to finish
```

devos50:
I've implemented a new checker that checks whether a Tribler process is already running in the specific state dir. This means that it is now possible to run several Tribler instances besides each other (under the condition they are using a separate state dir).

The method I'm using makes use of a lock file. When Tribler is in use, a `.lock` file should be written away, containing the PID of the running Tribler process. Other Tribler processes that are starting, are checking for this file and the PID inside the file to determine whether we have another Tribler running. I removed the `SingleInstanceChecker` class since we are not using that file anymore.

During the review, please check whether no inconsistent states can occur (i.e. the checker reporting that another instance is already running while there isn't and vice versa). I think this implementation works but maybe I forgot a specific case or something like that.

Fixes #2148


devos50:
```
ERROR: test_rss_import (Tribler.Test.test_my_channel.TestMyChannel)
  File "/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/test_as_server.py", line 643, in tearDown
    assert boolean, reason
AssertionError: TestMyChannel.test_rss_import - Channel did not have torrents - Condition was not satisfied in 60 seconds
-------------------- >> begin captured stdout << ---------------------
```

devos50:
retest this please

devos50:
@lfdversluis ready for review. Please ignore the two lines on the code coverage since that code is executed in a separate process (and nose does not recognize code executed by subprocesses).

Tribler/Core/Modules/process_checker.py
```diff
@@ -0,0 +1,72 @@
...
+    This class contains code to check whether a Tribler process is already running.
+    """
+
+    @staticmethod
+    def is_pid_running(pid):
+        """
+        Check whether a given process ID is currently running.
+        """
+        try:
+            os.kill(pid, 0)
```
lfdversluis: You check if the pid is running by killing your process?

Tribler/Core/Modules/process_checker.py
```diff
@@ -0,0 +1,72 @@
...
+        check the PID that is written inside the lock file.
+        """
+        self.already_running = False
+
+        self.statedir = SessionStartupConfig().get_state_dir()
+        self.lock_file_path = os.path.join(self.statedir, LOCK_FILE_NAME)
+
+        if os.path.exists(self.lock_file_path):
+            file_pid = self.get_pid_from_lock_file()
+            if file_pid == str(os.getpid()):
```
lfdversluis: maybe the python `is` is nicer?

Tribler/Main/tribler_main.py
```diff
@@ -1030,11 +1030,12 @@ def run(params=[""], autoload_discovery=True, use_torrent_search=True, use_chann
             params = sys.argv[1:]
     try:
         # Create single instance semaphore
-        single_instance_checker = SingleInstanceChecker("tribler")
+        process_checker = ProcessChecker()
+        print process_checker.already_running
```
lfdversluis: remove print

lfdversluis:
process the comments and I am happy as it is :+1:

Tribler/Core/Modules/process_checker.py
devos50: In this case, we're checking for value equality and not object equality so I think it's safer to use `==`. See for instance the following example:

```
>>> p = ''.join(['a', 'b'])
>>> p is 'ab'
False
>>> p == 'ab'
True
```

lfdversluis: :+1:

devos50:
@whirm processed feedback, ready for your review 👍

Tribler/Core/Modules/process_checker.py
```diff
@@ -0,0 +1,74 @@
...
+        os.unlink(self.lock_file_path)
+
+    def get_pid_from_lock_file(self):
+        """
+        Returns the PID from the lock file.
+        """
+        with open(self.lock_file_path, 'rb') as lock_file:
+            return lock_file.read()
+
+    def __init__(self):
```
whirm: Move the constructor to the beginning of the class

Tribler/Core/Modules/process_checker.py
```diff
@@ -0,0 +1,74 @@
+import os
+
+from Tribler.Core.SessionConfig import SessionStartupConfig
+
+
+LOCK_FILE_NAME = 'tribler.lock'
```
whirm: I wouldn't make this a constant, soon we will have the gui using this too.

whirm: Now that I think of it, QApp has some facility to check for it. So it may not be needed after all.

twisted/twisted/plugins/tribler_plugin.py
```diff
@@ -49,19 +55,24 @@ def signal_handler(sig, _):
...
                 msg("Tribler shut down")
                 reactor.stop()
+                self.process_checker.remove_lock_file()
         signal.signal(signal.SIGINT, signal_handler)
         signal.signal(signal.SIGTERM, signal_handler)
 
+        # Check if we are already running a Tribler instance
+        self.process_checker = ProcessChecker()
+        if self.process_checker.already_running:
+            self.shutdown_process("A Tribler instance is already running")
```
whirm: "Another Tribler instance is using statedir %s"

Tribler/Core/Modules/process_checker.py
devos50: I think I don't really understand how the GUI is checking for another running process. I was thinking that the GUI starts up Tribler and Tribler uses the `ProcessChecker` to see if there already is a running process. If so, this is reported to the GUI and the GUI gives a warning or something like that. Are you thinking of another approach?

whirm: I was referring to the GUI itself, so it doesn't start twice.

devos50:
@whirm see my last fixup commit.

whirm:
`
def process_func():
    while True:
        pass
`
...

whirm:
Remove that and squash it

devos50:
Channel search test:

```
self._cursor_table[thread_name] = self._connection.cursor()
exceptions.AttributeError: 'NoneType' object has no attribute 'cursor'
```

Seen this one several times before.

devos50:
retest this please

devos50:
```
test_ffsearch (Tribler.Test.test_remote_search.TestMixedRemoteSearch) ... Unhandled Error
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/Utilities/twisted_thread.py", line 55, in _reactor_runner
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 1194, in run
    self.mainLoop()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 1203, in mainLoop
    self.runUntilCurrent()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 798, in runUntilCurrent
    f(*a, **kw)
  File "/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/Modules/tracker_manager.py", line 98, in update_tracker_info
    tracker_info = self._tracker_dict[tracker_url]
exceptions.TypeError: 'NoneType' object has no attribute '__getitem__'
```

devos50:
retest this please

devos50:
Same error as above

devos50:
retest this please

devos50:
@whirm ready. I will try to fix the errors in the tests asap.

Also, I'm unable to remove the code you proposed since that method is used in the subprocess. The code coverage tool does not recognise the code is that it runs inside a subprocess.

whirm:
I see.

PR #2200, opened 2016-05-12
fix typo includeDispsersy to includeDispersy

ardhipoetra:
Bugs me for a while. This fixes the typo in those which affected. Is there any other place using this method?


whirm:
I'd say it's fine. There's an non covered modification, but it's on the gui, so it's fine.

PR #2221, opened 2016-05-19
WIP: Multichain option

pimotte:
This adds an option to the script that starts the tunnel helper. In addition, it fixes a bug such that if the multichain is enabled when not running the entirety of Tribler, the community is actually started.


lfdversluis:
add to whitelist

PR #2223, opened 2016-05-20
READY: added endpoint to resume, stop and remove downloads

devos50:
I added three functionalities of the downloads in one PR since they are not very hard to understand.

### DIFF
--- a/Tribler/Core/Modules/restapi/downloads_endpoint.py
+++ b/Tribler/Core/Modules/restapi/downloads_endpoint.py
@@ -1,20 +1,37 @@
 import json
 
-from twisted.web import resource
+from twisted.web import http, resource
 from Tribler.Core.Libtorrent.LibtorrentDownloadImpl import LibtorrentStatisticsResponse
 
 from Tribler.Core.simpledefs import DOWNLOAD, UPLOAD, dlstatus_strings
 
 
-class DownloadsEndpoint(resource.Resource):
+class DownloadBaseEndpoint(resource.Resource):
     """
-    This endpoint is responsible for all requests regarding downloads. Examples include getting all downloads,
-    starting, pausing and stopping downloads.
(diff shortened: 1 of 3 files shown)


devos50:
@lfdversluis ready for review

Tribler/Core/Modules/restapi/downloads_endpoint.py
```diff
@@ -90,3 +107,87 @@ def render_GET(self, request):
...
+    """
+
+    def __init__(self, session, infohash):
+        DownloadBaseEndpoint.__init__(self, session)
+        self.infohash = infohash
+
+    def render_DELETE(self, request):
+        parameters = http.parse_qs(request.content.read(), 1)
+
+        if 'remove_data' not in parameters or len(parameters['remove_data']) == 0:
```
lfdversluis: Is the 0 check necessary?

lfdversluis:
One comment given, looks fine.

Tribler/Core/Modules/restapi/downloads_endpoint.py
devos50: I've added the `len` check since `parameters['remove_data']` turned out to be an array and I was not sure whether this list is always filled with data. If it's not, an exception would be thrown when trying to access an empty list. So I added it to make 100% sure that we actually have the data.

I just tried make the list empty using some curl statements but it's always filled, even in the following command: `curl -X PUT --data "name=&description=" http://localhost:8085/mychannel`. @whirm do you think the check should be removed?

devos50:
@whirm can you review (and also answer the question on Laurens' comment)? Thanks!

whirm:
retest this please

whirm:
retest this please

devos50:
retest this please

whirm:
retest this please

devos50:
I will reopen this PR at a later point when I have some other endpoints ready ;)

PR #2246, opened 2016-06-01
Introduced chardet, the fullpath is now decoded with chardet

### LINKED ISSUE #2236: Paths fo torrents are decoded as utf-8 which may lead to an error
Problem is in line ` File "/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Core/APIImplementation/maketorrent.py", line 246, in pathlist2filename
    return fullpath.decode('utf-8')`

It will lead to a `exceptions.UnicodeDecodeError: 'utf8' codec can't decode byte 0xf1 in position 62: invalid continuation byte` if it's not utf-8. Many alphabets will cause a crash thus. Should be encoded and decoded as unicode/utf-16 or so I guess.

lfdversluis:
Fixes **1** bug.
Adds **2** tests.

Fixed #2236


lfdversluis:
Naturally all fails. Looks like we have some installing to do on the PR machines.

lfdversluis:
@whirm can you install chardet on the build machines?

lfdversluis:
I have installed chardet on the win 32 and 64 bit builders. So mac and bbq left.

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
`Tribler.Test.Core.Modules.RestApi.test_downloads_endpoint.TestDownloadsEndpoint.test_get_downloads timeout expired before end of test (20.000000 s.)`

lfdversluis:
retest this please

lfdversluis:
retest this please

lfdversluis:
same issue as above

lfdversluis:
retest this please

lfdversluis:
again the one above. Only on mac too. @devos50 any idea?

lfdversluis:
retest this please

devos50:
When running

```
nosetests Tribler/Test/Core/Modules/RestApi/test_downloads_endpoint.py:TestDownloadsEndpoint.test_get_downloads -v
```

It takes around 11.5 seconds on `devel`. When running the tests in your PR however, the run time has increased to around 16 seconds. These 16 seconds are probably just enough to cause a timeout since the Mac Mini is a bit slower than my laptop.

Maybe you can check whether the increase in run time also occurred on Linux/Windows and try to investigate what's causing this?

lfdversluis:
@devos50 that would mean chardet is incredibly slow?

lfdversluis:
@devos50 on the container you exact command takes 5.043 seconds.

lfdversluis:
And with chardet 9.066 seconds. That is quite a significant increase.

lfdversluis:
@devos50 can you review?

devos50:
I think the `APIImplementation` module is a bit strange location for `maketorrent.py`. I'm not sure if you agree but if so, maybe you can take this PR as opportunity to move it to a better location?

Tribler/Test/Core/APIImplementation/test_maketorrent.py
```diff
@@ -0,0 +1,19 @@
+# coding=utf-8
+import os
+
+from Tribler.Core.APIImplementation.maketorrent import pathlist2filename
+from Tribler.Test.test_as_server import BaseTestCase
+
+
+class TestMakeTorrent(BaseTestCase):
```
devos50: Maybe it's better to derive from `AbstractServer` (so we have the reactor checks and a temporary directory)?

devos50:
Other than that, I think it looks fine.

lfdversluis:
I agree about the place and moving, do you have any suggestions?

devos50:
@lfdversluis maybe to `Core/` (where `TorrentDef.py` is located)? Or maybe it's better to create a separate module for torrent-file related stuff and put `make_torrent.py` and `TorrentDef.py` in there.

lfdversluis:
I went for utilities as most of them are utility methods. I guess that makes sense in this case?

lfdversluis:
Ah pycharm move does a pep-8 on the imports :/

whirm:
Looks good, remove the import shuffling and squash it

lfdversluis:
@whirm Squashed and ready.

PR #2252, opened 2016-06-02
Fixes #2188, added test cases for byte formatting methods

### LINKED ISSUE #2188: "Balance:" does not collapse to megabytes unlike "Total:".

b60d3b126ddfcafc2a939c32ca6b9a97a962f561 devel

Debian

`Total: 94.87 MB   Balance: -80.71 MB`

##### Actual behavior: #####

`Total: 94.87 MB   Balance: -80712.54 KB`

Captain-Coder:
Test cases just test the current behaviour. Refactoring size_format might be usefull after Qt transition.

### DIFF
--- a/Tribler/Main/Utility/utility.py
+++ b/Tribler/Main/Utility/utility.py
@@ -87,17 +87,18 @@ class Utility(object):
 def speed_format(s):
     text = ''
     if s is not None:
-        if s < 102400:
+        s_abs = abs(s)
+        if s_abs < 102400:
             text = '%2.1f KB/s' % (s / 1024.0)
-        elif s < 1022797:
+        elif s_abs < 1022797:
             text = '%d KB/s' % (s // 1024)
-        elif s < 104857600:
+        elif s_abs < 104857600:
             text = '%2.1f MB/s' % (s / 1048576.0)
-        elif s < 1047527425:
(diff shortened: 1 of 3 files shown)


devos50:
@Captain-Coder Please not that this method will be removed when the wx GUI can be removed since formatting speed strings will be a task of the Qt GUI. See https://github.com/devos50/tribler/blob/qt-gui/TriblerGUI/utilities.py#L7 for my implementation. Feedback or improvements on that method would be appreciated :)

Also, regarding #2188, currently it is not worth to fix most of the open wx-related issues. We should however still take them in considerations since some of the GUI bugs might be caused by an error in the core of Tribler. These issues should be fixed however.

whirm:
ready?

Captain-Coder:
@devos50, I already figured I shouldn't fix WX issues, that is why I chose an issue without the WX tag on it. I was unsure if this general util thing was wx specific or not. Also doing this makes me look like i'm doing something :) 

@whirm, yes it is ready as far as I can tell, even the CI seems to agree that it is good to go. (That's a first for me)

whirm:
@Captain-Coder remember to set the issue to READY or I'll probalby ignore it (I clicked on this one by chance).

Issue #2251, opened 2016-06-02
Alert for save_resume_data


ardhipoetra:
I noticed that there are not any alert for receiving `save_resume_data`. I have mentioned this in closed PR https://github.com/Tribler/tribler/pull/1990. 

A bit explanation : 
In previous (deprecated) call, `write_resume_data` is synchronously (blocked call) returned the so-called resume data. 

In new call, `save_resume_data` is returned immediately (in my case : `None`). After resume data is ready, it will report an alert `save_resume_data_alert` (or `save_resume_data_alert_failed` if failed) which contains dictionary-format of resume data.

whirm:
So download state is not being stored to disk ever?

ardhipoetra:
Yes, I think so.

User can still resume the torrent downloading because libtorrent recheck download target.

whirm:
Well, that explains several reports about downloads been checked every time Tribler started.

devos50:
Yeah it's quite annoying to have downloads rechecked every time on startup. Should be an easy one to fix however, just create a new alert type and process it in `LibtorrentDownloadImpl` by writing the resume data to the `.state` config file :)

lfdversluis:
Nice find @ardhipoetra This bug was described here: #2199 and should be fixed I guess by your fix 👍

ardhipoetra:
This is fixed, right?

whirm:
Isn't that fixed already?

ardhipoetra:
Can I close this?

whirm:
If it's fixed, sure.

Issue #2199, opened 2016-05-12
[devel] Tribler re-checks all downloads on starup


vi:
b60d3b126ddfcafc2a939c32ca6b9a97a962f561 devel

After I close Tribler and start it again, it re-checks all the downloads.
I think it should be only recheck if there may be issues (e.g. filesystem modification date changed or some peer rejected chunks from the file)

lfdversluis:
@whirm I think this was the original report. I think it can be closed too with #2265 ?

PR #2265, opened 2016-06-07
Alert for resume data

ardhipoetra:
This PR create alert for `save_resume_data`. I used `Deferred` on `resume_data`. This deferred object will call `lm.start_download` as callback (I don't find other use beside this). Resume data will be stored in `LibtorrentDownloadImpl` attribute if available.

The result is, `engineresumedata` is now available in disk-saved state. This to a certain extent, solves #2251. However, I can't get fastresume in libtorrent working, even with my own small program. So #2199 is not solved yet.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -761,11 +761,11 @@ class TriblerLaunchMany(TaskManager):
 
     def save_download_pstate(self, infohash, pstate):
         """ Called by network thread """
-        basename = binascii.hexlify(infohash) + '.state'
-        filename = os.path.join(self.session.get_downloads_pstate_dir(), basename)
 
-        self._logger.debug("tlm: network checkpointing: to file %s", filename)
-        pstate.write_file(filename)
+        self.downloads[infohash].pstate_for_restart = pstate
+
+        self.register_task("save_pstate %f" % timemod.clock(),
+                           self.downloads[infohash].save_resume_data())
 
     def load_download_pstate(self, filename):
         """ Called by any thread """
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -5,10 +5,10 @@ import sys
 import time
 from binascii import hexlify
 from traceback import print_exc
-from twisted.internet.defer import Deferred
 
 import libtorrent as lt
 from twisted.internet import defer
+from twisted.internet.defer import Deferred, CancelledError
 
 from Tribler.Core import NoDispersyRLock
 from Tribler.Core.DownloadConfig import DownloadStartupConfig, DownloadConfigInterface
@@ -22,7 +22,6 @@ from Tribler.Core.simpledefs import (DLSTATUS_WAITING4HASHCHECK, DLSTATUS_HASHCH
                                      DLSTATUS_DOWNLOADING, DLSTATUS_SEEDING, DLSTATUS_ALLOCATING_DISKSPACE,
                                      DLSTATUS_CIRCUITS, DLSTATUS_STOPPED, DLMODE_VOD, DLSTATUS_STOPPED_ON_ERROR,
                                      UPLOAD, DOWNLOAD, DLMODE_NORMAL, PERSISTENTSTATE_CURRENTVERSION, dlstatus_strings)
-
 
 if sys.platform == "win32":
     try:
@@ -146,6 +145,8 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
         self.askmoreinfo = False
 
         self.correctedinfoname = u""
+
+        self.deferreds_resume = []
 
     def __str__(self):
         return "LibtorrentDownloadImpl <name: '%s' hops: %d>" % (self.correctedinfoname, self.get_hops())
@@ -442,12 +443,42 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
             self._logger.debug("LibtorrentDownloadImpl: alert %s with message %s", alert_type, alert)
 
         alert_types = ('tracker_reply_alert', 'tracker_error_alert', 'tracker_warning_alert', 'metadata_received_alert',
-                       'file_renamed_alert', 'performance_alert', 'torrent_checked_alert', 'torrent_finished_alert')
+                       'file_renamed_alert', 'performance_alert', 'torrent_checked_alert', 'torrent_finished_alert',
+                       'save_resume_data_alert')
 
         if alert_type in alert_types:
             getattr(self, 'on_' + alert_type)(alert)
         else:
             self.update_lt_stats()
+
+    def on_save_resume_data_alert(self, alert):
+        """
+        alert to handle save_resume_data_alert
+        it will assign stored resume data in an attribute,
+            and write it to a file
+        """
+        resume_data = alert.resume_data
+
+        if not self.pstate_for_restart:
+            self.pstate_for_restart = self.network_get_persistent_state()
+
+        self.pstate_for_restart.set('state', 'engineresumedata', resume_data)
+        self._logger.debug("%s get resume data %s", hexlify(resume_data['info-hash']), resume_data)
+
+        # save it to file
+        basename = hexlify(resume_data['info-hash']) + '.state'
+        filename = os.path.join(self.session.get_downloads_pstate_dir(), basename)
+
+        self._logger.debug("tlm: network checkpointing: to file %s", filename)
+
+        self.pstate_for_restart.write_file(filename)
+
+        # fire callback for all deferreds_resume
+        for deferred_r in self.deferreds_resume:
+            deferred_r.callback(resume_data)
+
+        # empties the deferred list
+        self.deferreds_resume = []
 
     def on_tracker_reply_alert(self, alert):
         self.tracker_status[alert.url] = [alert.num_peers, 'Working']
@@ -706,6 +737,26 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
         with self.dllock:
             return self.curspeeds[dir]
 
+    def _on_resume_err(self, failure):
+        failure.trap(CancelledError)
+        self._logger.error("Resume data cancelled")
+
+    @waitForHandleAndSynchronize()
+    def save_resume_data(self):
+        """
+        method to save resume data.
+        """
+        # only call libtorrent save resume in the first call
+        if not self.deferreds_resume:
+            self.handle.save_resume_data()
+
+        defer_resume = Deferred()
+        defer_resume.addErrback(self._on_resume_err)
+
+        self.deferreds_resume.append(defer_resume)
+
+        return defer_resume
+
     def set_moreinfo_stats(self, enable):
         """ Called by any thread """
 
@@ -911,15 +962,14 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
             pstate = self.network_get_persistent_state()
             if self.handle is not None:
                 self._logger.debug("LibtorrentDownloadImpl: network_stop: engineresumedata from torrent handle")
+                self.pstate_for_restart = pstate
                 if removestate:
                     self.ltmgr.remove_torrent(self, removecontent)
                     self.handle = None
                 else:
                     self.set_vod_mode(False)
                     self.handle.pause()
-                    pstate.set('state', 'engineresumedata', self.handle.save_resume_data()
-                               if isinstance(self.tdef, TorrentDef) else None)
-                self.pstate_for_restart = pstate
+                    self.save_resume_data()
             else:
                 # This method is also called at Session shutdown, where one may
                 # choose to checkpoint its Download. If the Download was
@@ -1008,7 +1058,8 @@ class LibtorrentDownloadImpl(DownloadConfigInterface):
                 if self.pstate_for_restart is not None:
                     resdata = self.pstate_for_restart.get('state', 'engineresumedata')
             elif isinstance(self.tdef, TorrentDef):
-                resdata = self.handle.save_resume_data()
+                self.save_resume_data()
+
             pstate.set('state', 'engineresumedata', resdata)
             return (self.tdef.get_infohash(), pstate)
 
--- a/Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
+++ b/Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
@@ -1,3 +1,4 @@
+import binascii
 import os
 
 from Tribler.Core.DownloadConfig import DownloadStartupConfig
@@ -29,6 +30,21 @@ class TestLibtorrentDownloadImpl(TestAsServer):
         self.config.set_videoplayer(False)
         self.config.set_torrent_collecting_dir(os.path.join(self.session_base_dir, 'torrent_collecting_dir'))
 
+    def create_tdef(self):
+        """
+        create and save torrent definition used in this test file
+        """
+        tdef = TorrentDef()
+        sourcefn = os.path.join(TESTS_DATA_DIR, 'video.avi')
+        tdef.add_content(sourcefn)
+        tdef.set_tracker("http://localhost/announce")
+        tdef.finalize()
+
+        torrentfn = os.path.join(self.session.get_state_dir(), "gen.torrent")
+        tdef.save(torrentfn)
+
+        return tdef
+
     @deferred(timeout=10)
     def test_can_create_engine_wrapper(self):
         impl = LibtorrentDownloadImpl(self.session, None)
@@ -50,28 +66,14 @@ class TestLibtorrentDownloadImpl(TestAsServer):
         return impl.can_create_engine_wrapper()
 
     def test_get_magnet_link_none(self):
-        tdef = TorrentDef()
-        sourcefn = os.path.join(TESTS_DATA_DIR, 'video.avi')
-        tdef.add_content(sourcefn)
-        tdef.set_tracker("http://localhost/announce")
-        tdef.finalize()
-
-        torrentfn = os.path.join(self.session.get_state_dir(), "gen.torrent")
-        tdef.save(torrentfn)
+        tdef = self.create_tdef()
 
         impl = LibtorrentDownloadImpl(self.session, tdef)
         link = impl.get_magnet_link()
         self.assertEqual(None, link, "Magnet link was not none while it should be!")
 
     def test_get_tdef(self):
-        tdef = TorrentDef()
-        sourcefn = os.path.join(TESTS_DATA_DIR, 'video.avi')
-        tdef.add_content(sourcefn)
-        tdef.set_tracker("http://localhost/announce")
-        tdef.finalize()
-
-        torrentfn = os.path.join(self.session.get_state_dir(), "gen.torrent")
-        tdef.save(torrentfn)
+        tdef = self.create_tdef()
 
         impl = LibtorrentDownloadImpl(self.session, None)
         impl.set_def(tdef)
@@ -79,14 +81,7 @@ class TestLibtorrentDownloadImpl(TestAsServer):
 
     @deferred(timeout=20)
     def test_setup(self):
-        tdef = TorrentDef()
-        sourcefn = os.path.join(TESTS_DATA_DIR, 'video.avi')
-        tdef.add_content(sourcefn)
-        tdef.set_tracker("http://localhost/announce")
-        tdef.finalize()
-
-        torrentfn = os.path.join(self.session.get_state_dir(), "gen.torrent")
-        tdef.save(torrentfn)
+        tdef = self.create_tdef()
 
         impl = LibtorrentDownloadImpl(self.session, tdef)
         def callback((ignored, ignored2)):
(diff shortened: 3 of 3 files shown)


ardhipoetra:
retest this pleas

ardhipoetra:
retest this please

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -759,8 +759,17 @@ def network_shutdown(self):
             self.threadpool.cancel_all_pending_tasks()
             self.threadpool = None
 
+    def save_download_pstate_callback(self, resume_data, infohash, pstate):
+        pstate.set('state', 'engineresumedata', resume_data)
+        self.save_download_pstate(infohash, pstate)
```
lfdversluis: Suppose Tribler is shutting down and a `resume_data` deffered fires, will this cause an error?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -706,6 +724,26 @@ def get_current_speed(self, dir):
         with self.dllock:
             return self.curspeeds[dir]
 
+    def save_resume_data(self):
+        """
+        method to save resume data. It will return either :
+        stored resume data, or deferred if it's not available
+        """
```
lfdversluis: If you want less `instanceof(X, Deferred)` code you could let this function always return a deferred by using `defer.succeed` or use `maybeDeferred(save_resume_data)` on the caller.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -442,13 +446,27 @@ def process_alert(self, alert, alert_type):
...
+        alert to handle save_resume_data_alert
+        it will assign store resume data in a attribute,
+            and called save_download_pstate in LaunchManyCore
+        """
+        self.resume_data = alert.resume_data
+
+        # at this point, Deferred object should call save_download
+        if self.saveresume_called_defer:
+            self.saveresume_called_defer.callback(alert.resume_data)
+            self.saveresume_called_defer = None
```
ardhipoetra: Wait, I thought line 467 means you _call_ the callback. Attaching a callback can be done by `addCallback` isn't it?

There's no severe consequences by not setting it to `None`, though. You just write the checkpoint twice.

Tribler/Core/APIImplementation/LaunchManyCore.py
ardhipoetra: if `session` already killed, maybe yes. Otherwise, I don't think there's a problem with that.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ardhipoetra: Yep, I'm using second approach

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
lfdversluis: I mean you add a callback to the deferred in line 467 then set the deferred to None in 468.

Tribler/Core/APIImplementation/LaunchManyCore.py
lfdversluis: I guess you need a check for that then. Note that LMC extends TaskManager (`class TriblerLaunchMany(TaskManager):`), so if it solves your problem you can let the taskmanager cancel the deferreds for you on shutdown. (note that this may not be the right place to cancel them, you should check this)

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
lfdversluis: Apparently this is not an issue. Made a small example and this works: https://gist.github.com/lfdversluis/68af55e43bfc167f6617f5debd9fbad2

Tribler/Core/APIImplementation/LaunchManyCore.py
ardhipoetra: in dcd07b45333b40804664b5230b5ad0264b2d2d6e, I added the use of taskmanager when create Deferred object. Naturally, it will cleaned when `LaunchManyCore` shutting down.

ardhipoetra:
retest this please

ardhipoetra:
retest this please

ardhipoetra:
I think it's ready. However, I tried retesting several times, but randomly get timeout in either [win32](https://jenkins.tribler.org/job/GH_Tribler_PR_tests_win32/1247/) or [win64](https://jenkins.tribler.org/job/GH_Tribler_PR_tests_win64/1252/) in test. Any clue?

ardhipoetra:
retest this please

ardhipoetra:
@whirm ready for review

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
whirm: Please, always return a deferred as @lfdversluis said. Having to deal with a function returning different data types randomly doesn't make sense.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -706,6 +724,31 @@ def get_current_speed(self, dir):
...
+    def save_resume_data(self):
+        """
+        method to save resume data. It will return either :
+        stored resume data, or deferred if it's not available
+        """
+        ret_val = None
+
+        # prepare deferred entry
+        defer_resume = Deferred()
+        defer_resume.addErrback(lambda _: self._logger.info("Defered resume_data killed before launch"))
```
whirm: This is the equivalent of a catch all try block. Anything that goes wrong, will be registered as the resume data deferred to be killed before launch.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -917,8 +960,7 @@ def network_stop(self, removestate, removecontent):
                 else:
                     self.set_vod_mode(False)
                     self.handle.pause()
-                    pstate.set('state', 'engineresumedata', self.handle.save_resume_data()
-                               if isinstance(self.tdef, TorrentDef) else None)
+                    pstate.set('state', 'engineresumedata', defer.maybeDeferred(self.save_resume_data))
```
whirm: you are setting engineresumedata to a deferred?

whirm:
What tests do we have relative to this modifications?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ardhipoetra: when it's not available, it will stored as deferred. 
But when actually saving it, it will get data first, then save it.

ardhipoetra:
I don't recall we have test that covers this

ardhipoetra:
retest this please

whirm:
retest this please

whirm:
(I had to restart jenkins yesterday)

ardhipoetra:
@whirm all your feedback have been processed. Test error in [Allchannel](https://jenkins.tribler.org/job/Tribler/job/GH/job/PR/job/Allchannel+ChannelCommunity_short/235/) which I don't know why.

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
```diff
@@ -140,3 +140,60 @@ def test_multifile_torrent(self):
         test_dict["a"] = "b"
         pstate.set("state", "engineresumedata", test_dict)
         return impl.network_create_engine_wrapper(pstate)
+
+    @deferred(timeout=20)
+    def test_save_resume(self):
+        """
+        testing call resume data alert
+        """
+        tdef = TorrentDef()
```
devos50: Can you please refactor this creation of a torrent into a separate method? You are duplicating this in `test_save_resume` and `test_save_resume_exists`.

ardhipoetra: I end up refactor those which not in my test. It should be in https://github.com/Tribler/tribler/pull/2265/commits/fb2d2698abbb778700f54e399700e5b58dff9187

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -759,8 +759,17 @@ def network_shutdown(self):
...
+
     def save_download_pstate(self, infohash, pstate):
         """ Called by network thread """
+        resume_data = pstate.get('state', 'engineresumedata')
+        if isinstance(resume_data, Deferred):
```
whirm: Don't do this. `resume_data` has to be the actual value when `save_download_pstate` is called.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -960,7 +964,7 @@ def network_stop(self, removestate, removecontent):
                 else:
                     self.set_vod_mode(False)
                     self.handle.pause()
-                    pstate.set('state', 'engineresumedata', defer.maybeDeferred(self.save_resume_data))
+                    pstate.set('state', 'engineresumedata', self.save_resume_data)
```
whirm: You are assigning the function itself here!

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -1050,7 +1054,7 @@ def network_checkpoint(self):
                 if self.pstate_for_restart is not None:
                     resdata = self.pstate_for_restart.get('state', 'engineresumedata')
             elif isinstance(self.tdef, TorrentDef):
-                resdata = defer.maybeDeferred(self.save_resume_data)
+                resdata = self.save_resume_data
```
whirm: here too

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -917,8 +964,7 @@ def network_stop(self, removestate, removecontent):
                 else:
                     self.set_vod_mode(False)
                     self.handle.pause()
-                    pstate.set('state', 'engineresumedata', self.handle.save_resume_data()
-                               if isinstance(self.tdef, TorrentDef) else None)
+                    pstate.set('state', 'engineresumedata', self.save_resume_data)
```
whirm: You are storing the function itself!

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -1008,7 +1054,7 @@ def network_checkpoint(self):
                 if self.pstate_for_restart is not None:
                     resdata = self.pstate_for_restart.get('state', 'engineresumedata')
             elif isinstance(self.tdef, TorrentDef):
-                resdata = self.handle.save_resume_data()
+                resdata = self.save_resume_data
```
whirm: here too

Tribler/Core/APIImplementation/LaunchManyCore.py
ardhipoetra: What if `resume_data` is not ready yet? In case we are waiting for alert. Can I returning `None` or should it block the call until ready?

whirm: you don't do anything until you've got all you need. If you need more help, please join the IRC channel

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ardhipoetra: my bad, why I missed this? It's fixed now

ardhipoetra:
retest this please

ardhipoetra:
retest this please

whirm:
@ardhipoetra If you retest, please comment on why that particular test run failed. Otherwise we can't see what needs to be fixed.

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -761,10 +761,18 @@ def network_shutdown(self):
 
     def save_download_pstate(self, infohash, pstate):
         """ Called by network thread """
+
+        # if pstate can't be found
+        if not pstate.get('state', 'engineresumedata'):
```
whirm: Shouldn't this be refreshed every time the method is called? 
Otherwise we will be using stale information from that point on.

If we did this then everything on this function could be moved to `on_save_resume_data_alert` which would make it a bit nicer.

ardhipoetra: you mean the `pstate`? It is stored in `pstate_for_restart` in `LibtorrentDownloadImpl`. And it is updated when we get an alert.

whirm: Yes, but you only call `save_resume_data` if `'engineresumedata'` is empty so you will keep using the same data. (If libtorrent doesn't generate resume data alerts by its own initiative, but even if this is the case the proposal still stands)

ardhipoetra: All who called `save_download_pstate` also called `network_checkpoint` beforehand. `network_checkpoint` empties `engineresumedata`, so whenever we called `save_download_pstate` (in this stage), it will wait an alert before actually write it to file.

ardhipoetra: So, how about `save_download_pstate` always call `save_resume_data` and then set the `pstate_for_restart`. And then when receiving an alert, we actually write that one.

whirm: Sounds good

ardhipoetra:
@whirm, ready to review again

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -1008,7 +1041,10 @@ def network_checkpoint(self):
                 if self.pstate_for_restart is not None:
                     resdata = self.pstate_for_restart.get('state', 'engineresumedata')
             elif isinstance(self.tdef, TorrentDef):
-                resdata = self.handle.save_resume_data()
+                self.save_resume_data()
+
+                # whether it's already called or not
+                resdata = self.resume_data
```
whirm: calling save_resume_data() doesn't make self.resume_data contain the data right away. Also,  the pstate dict is being updated on the alert handler, so what follows doesn't make much sense anymore?

whirm: At first glance, I think we can get rid of `self.resume_data` altogether?

ardhipoetra: Yes you're right. I missed that part. It's fixed in be4370154ba96d229f3287bafd3fa3a176c5ac83 now

ardhipoetra:
retest this please

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -761,11 +761,9 @@ def network_shutdown(self):
 
     def save_download_pstate(self, infohash, pstate):
         """ Called by network thread """
-        basename = binascii.hexlify(infohash) + '.state'
-        filename = os.path.join(self.session.get_downloads_pstate_dir(), basename)
 
-        self._logger.debug("tlm: network checkpointing: to file %s", filename)
-        pstate.write_file(filename)
+        self.downloads[infohash].pstate_for_restart = pstate
```
whirm: on_save_for_resume_data is fetching the pstate itself if it's not there yet. Is there a point on making save_download_pstate() receive and set it?

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -911,15 +942,14 @@ def network_stop(self, removestate, removecontent):
             pstate = self.network_get_persistent_state()
             if self.handle is not None:
                 self._logger.debug("LibtorrentDownloadImpl: network_stop: engineresumedata from torrent handle")
+                self.pstate_for_restart = pstate
```
whirm: same here, if `on_save_resume_data()` can fetch it. Why are we doing the same in two other places?

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
```diff
@@ -140,3 +129,47 @@ def test_multifile_torrent(self):
...
+                engine_data = CallbackConfigParser()
+                engine_data.read_file(filename)
+
+                self.assertEqual(tdef.get_infohash(), engine_data.get('state', 'engineresumedata').get('info-hash'))
+
+                d.callback(None)
+
+            return d
+
+        def callback((dummy_1, dummy_2)):
```
whirm: use underscores like in the rest of the Tribler codebase

ardhipoetra: In my IDE, I can't use two underscore in different parameters. Pylint, btw, ignore parameter with prefix "dummy"

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ardhipoetra: Because `pstate` may be used in the `else` statement (below it)

Tribler/Core/APIImplementation/LaunchManyCore.py
ardhipoetra: Because `pstate` is provided by different method and I'm not sure those will return the same value, so I keep the parameter `pstate` there and use it as a basis to save the state

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
```diff
@@ -140,3 +129,47 @@ def test_multifile_torrent(self):
...
+
+        def resume_ready(d=None):
+            """
+            check if resume data is ready
+            """
+            if not d:
+                d = defer.Deferred()
+
+            if not impl.pstate_for_restart:
+                reactor.callLater(2, resume_ready, d)
```
whirm: I think it would make a whole lot more of sense to make save_resume_data() return a deferred that fires once the resume data hits the disk. This way it could not only be used to have nicer testability but also to wait for all the resume data to hit the disk during shutdown.

ardhipoetra: Doesn't that complicate things and back to the idea of using deferred-taskmanager? 

What if the deferred is not fired yet, but we call `save_resume_data()` again? Should it ignore previous call or just overwrite it?

whirm: You can just fire both

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
whirm: if you can't use an underscore for an ignored argument then your IDE is broken and you should use something else. I know using a dummy prefix is also allowed but we want to keep consistency across the codebase.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
whirm: That should probably be done differently, but oh well. I guess it will have to be done later.

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
devos50: @whirm I have the same issue: I can also only use one underscore at most in parameters. And I'm using PyCharm (which is I think one of the most-used IDEs out there).

ardhipoetra: Isn't that `SyntaxError` instead of just pylint warning?

whirm: @devos50 oh, for sure, you cannot have two arguments named the same. we only need one here.
Sorry @ardhipoetra I didn't catch you mentioning the _two_ underscores. Just replace the whole thing with a single underscore. You are ignoring a tuple of two items instead of just ignoring the tuple altogether.

Tribler/Test/Core/Libtorrent/test_libtorrent_download_impl.py
ardhipoetra: @whirm, I think it's ready. I put those deferreds in an array to make sure all call is not overwritten by another call

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -761,11 +761,11 @@ def network_shutdown(self):
...
+
+        self.register_task("save_pstate %d" %len(self.downloads[infohash].deferreds_resume),
```
devos50: Please insert a space after the `%`.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -442,13 +443,38 @@ def process_alert(self, alert, alert_type):
...
 
         if alert_type in alert_types:
             getattr(self, 'on_' + alert_type)(alert)
         else:
             self.update_lt_stats()
 
+    def on_save_resume_data_alert(self, alert):
+        """
+        alert to handle save_resume_data_alert
+        it will assign store resume data in a attribute,
```
devos50: it will assign store**d** resume data in a**n** attribute

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -442,13 +443,38 @@ def process_alert(self, alert, alert_type):
...
+            and write it to file
```
devos50: Write it to **a** file

ardhipoetra:
retest this please

ardhipoetra:
retest this please

ardhipoetra:
retest this please

ardhipoetra:
retest this please

ardhipoetra:
retest this please

whirm:
retest this please

ardhipoetra:
retest this please

PR #2308, opened 2016-06-16
Converted all Tribler documentation to rst format and Sphinx

devos50:
To build the documentation, checkout this PR, cd to doc and execute `make html`.

If you don't want to build the documentation yourself, I've created a zip file containing the docs: 
[tribler_docs.zip](https://github.com/Tribler/tribler/files/323712/tribler_docs.zip)


devos50:
@lfdversluis can you review?

lfdversluis:
I think it's nice to have the version number of Tribler with the doc so you know which version this doc belongs to.

lfdversluis:
The Tribler REST API doesn't specify that tribler must be running to be able to curl to localhost etc.

lfdversluis:
Under building tribler > windows it reads "``__init__.py``" these quotes should form a box

lfdversluis:
Under Architectural Structure, the table is a mess.

lfdversluis:
Review done. Didn't read most of the text as it's copy paste.

devos50:
I've dived into the usage of markdown with Sphinx. In short: it is possible using [recommonmark](https://github.com/rtfd/recommonmark). I tried it out and it seems to work for simple pages.

Now the downside: the fact that `README.md` is a directory above the `doc` directory is a problem since I cannot import it directly in the table of contents tree. Instead, I need to create another file and use the `.. include` directive to import that file. This does not work with markdown which means that the `README` needs to be in RST format in order to be displayed on the docs.

Moreover, `autodoc` (the tool I'm using to extract docstrings from source code and display it in the REST API documentation) in combination with [httpdomain](https://pythonhosted.org/sphinxcontrib-httpdomain/) does not work with markdown.

There are some files which could be imported in markdown format, however, my opinion is that we should be consistent across our repository with file format. I think mixing md and rst files makes it more confusing.

We can use `mkdocs` however but there is no support for httpdomain (we can use `autodoc` however but apparently, markdown inside the docstrings are not rendered correctly).

@whirm what's your opinion? Should I find another tool or shall we continue using RST?

whirm:
I spotted this: https://github.com/devos50/tribler/blob/8eb8eaf0f11d27a8dabb8b400455b55f039add26/Tribler/Core/Modules/restapi/search_endpoint.py#L37

While having a quick glance at the generated docs.

They look great!

whirm:
the events list in events.html is all mangled up

devos50:
@whirm good find, I've fixed them in my last fixup commit.

whirm:
Squash it!

devos50:
@whirm done, ready for merge

PR #2317, opened 2016-06-22
Refactored general excepts in downloadconfig

lfdversluis:


lfdversluis:
```
Failed

Tribler.Test.Core.Libtorrent.test_libtorrent_download_impl.TestLibtorrentDownloadImpl.test_can_create_engine_wrapper (from nosetests)
Failing for the past 1 build (Since Failed#1670 )
Took 2 sec.
Error Message

Couldn't listen on any:32481: [Errno 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted.
```

lfdversluis:
retest this please

lfdversluis:
@devos50 Can you review?

Tribler/Core/DownloadConfig.py
```diff
@@ -193,8 +194,8 @@ def load(filename):
         dlconfig = CallbackConfigParser()
         try:
             dlconfig.read_file(filename)
-        except:
-            raise IOError, "Failed to open download config file"
+        except (ParsingError, IOError):
+            raise IOError, "Failed to open download config file: %s" % filename
```
devos50: Why not raise the exception that is being thrown here? It might be confusing to debug if a `ParsingError` is raised but an `IOError` is displayed to the dev instead.

devos50: Also, if I'm not mistaken, the exact content of the exception (the cause) is not printed? So, if the `ParsingError` happens, I only see `Failed to open download config file: ...` and not what went wrong during parsing.

devos50:
@lfdversluis Comments given

lfdversluis:
@devos50 You think the current code is ok?

devos50:
@lfdversluis yeah it looks fine to me.

lfdversluis:
mac flipped something with
```
mkdir: /Users/tribler/Documents/workspace/workspace/GH_Tribler_PR_tests_mac/output: File exists
Build step 'Execute shell' marked build as failure
Archiving artifacts
```

lfdversluis:
@devos50 I changed some things, could you take a look once more?

Tribler/Core/DownloadConfig.py
```diff
@@ -16,13 +16,16 @@
...
 import os
+from ConfigParser import ParsingError, MissingSectionHeaderError
 from types import StringType
 
 from Tribler.Core.Utilities.configparser import CallbackConfigParser
 from Tribler.Core.defaults import dldefaults
 from Tribler.Core.osutils import get_home_dir
 from Tribler.Core.simpledefs import DLMODE_VOD, STATEDIR_DLCONFIG, UPLOAD
 
+logger = logging.getLogger(__name__)
```
devos50: Maybe it's a bit cleaner to make this a class variable of `DownloadConfigInterface`?

devos50:
@lfdversluis done

Tribler/Core/DownloadConfig.py
lfdversluis: You cannot reference it then from load. It doesn't have self as a param.

devos50: Ah I didn't notice that it's a class method, ignore my comment then.

lfdversluis:
@whirm Ready for your review.

PR #2322, opened 2016-06-22
Removed 'channel' term from filter terms

devos50:
This was messing up my tests lol...

### DIFF
--- a/Tribler/Category/filter_terms.filter
+++ b/Tribler/Category/filter_terms.filter
@@ -619,7 +619,6 @@ centerfolds
 chachi
 chandigarhdicke
 changingroom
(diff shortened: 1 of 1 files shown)


devos50:
@lfdversluis can you review?

lfdversluis:
Well not much to review here, besides to comment that many terms can arguably be removed from this list. This approach is old-fashioned anyway :)

devos50:
Yes I agree, there are more terms to be removed. Still, we should use a better algorithm for the family filter.

devos50:
@whirm ready for your review.

whirm:
lol

PR #2325, opened 2016-06-22
READY: Initialized threadpool size in twistd plugin

devos50:
If we don't do this, we have no thread pool when running the twistd plugin 👎 . This gave me an error if I tried to start the twistd plugin before the upgrader in Tribler starts.

The reason why it probably worked before, is that the thread pool size was initialized by https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Utilities/twisted_thread.py#L54.

### DIFF
--- a/twisted/plugins/tribler_plugin.py
+++ b/twisted/plugins/tribler_plugin.py
@@ -115,6 +115,7 @@ class TriblerServiceMaker(object):
             })
             tribler_service.addService(manhole)
 
+        reactor.suggestThreadPoolSize(1)
         reactor.callWhenRunning(self.start_tribler, options)
 
         return tribler_service


devos50:
retest this please

devos50:
@lfdversluis can you review?

lfdversluis:
Interesting that twisted doesn't create a threadpool on start. I thought it did. 1 Seems small to be, but is apparently what we used before. AFAIK it's not a hard limit but merely a suggestion to Twisted, so it should be ok.

devos50:
Yeah it's weird indeed, maybe they disable it by default for performance or something like that?

devos50:
@whirm ready for review

whirm:
Closing it as talked face to face.

PR #2336, opened 2016-06-25
Fixed bugs when removing torrents from playlists

### LINKED ISSUE #2275: Torrents cannot be removed from a playlist
It appears that torrents cannot be removed from a playlist from the GUI. Trying to remove a torrent has no effect. In the command line, it says the following:

```
ERROR   1465474120.77       community:3322  you are attempting to undo the same message twice. trying to return the previous undo message
```

So probably, it is creating an undo message but it is ignored somehow.

devos50:
When removing torrents from a playlist, there were three errors. First, the GUI was not immediately notified of the removal. Second, there would be Dispersy warnings if we tried to remove the torrent from a playlist a second time (while the first removal request was still ongoing). Third, there was a error in the sql query that caused that the right torrent to be removed was not always returned.

Fixes #2275


devos50:
@lfdversluis can you review?

lfdversluis:
Looks good to me. Have you created an issue for the bug btw?

devos50:
@lfdversluis yes there is, I just forgot to link it. I've added it now.

devos50:
@whirm ready for review

whirm:
I don't see `Tribler/Test/data/bak_new_tribler.sdb.tar.gz ` mentioned on the whole PR?

devos50:
@whirm ah, I've added some playlists in the database that are used during the tests.

whirm:
Oh, so it's a replacement? I couldn't tell that from the GH gui.

PR #2339, opened 2016-06-25
Implemented export torrent endpoint

devos50:


devos50:
@lfdversluis can you review?

Tribler/Core/Modules/restapi/downloads_endpoint.py
```diff
@@ -208,3 +209,37 @@ def render_PATCH(self, request):
...
+
+            **Example response**:
+
+            The contents of the .torrent file.
+        """
+        download = self.session.get_download(self.infohash)
+        if not download:
+            return DownloadExportTorrentEndpoint.return_404(request)
+
+        request.setHeader(b'content-type', 'application/x-bittorrent')
```
lfdversluis: is it necessary to make 'content-type' and the other bytes?

lfdversluis:
If my comment is invalid, I am happy :+1:

Tribler/Core/Modules/restapi/downloads_endpoint.py
devos50: It's necessary to make web browsers happy (so they download the file immediately) :)

devos50:
@whirm ready for review

PR #2355, opened 2016-06-29
WIP: Removal of the wx GUI

### LINKED ISSUE #2174: Call to unexisting methods in SearchGridManager
There are some method calls in SearchGridManager to methods that are not available in Tribler, see  https://github.com/Tribler/tribler/blob/devel/Tribler/Main/vwxGUI/SearchGridManager.py#L1516. Note that `subscribe`, `spam` and `unsubscribe` calls. They do not exist in the `VoteCastDBHandler` class.

devos50:
In this PR, I remove the old wx GUI.

Some notes:
- I'm also removing the tribler profiler in `tribler_profiler.py`. I'm moving this one to the twistd plugin in #2389.
- I'm currently working on refactoring the video server to use Twisted instead, see #2391. This PR is not ready (yet).
- I will remove the blocking calls on reactor thread decorators in Tribler in another PR.

Fixes #2174.


brussee:
Removing 16k loc and still the code coverage goes down XD

devos50:
@brussee much of the code I removed was already covered :(

whirm:
It goes down by one line, don't start crying now...

devos50:
@lfdversluis can you review? I finished this PR. Please also see the notes in my first comment.

The windows test on 32-bit succeeded but somehow, it's not reflected in the checks here. Also, the pylint error is because the nose reactor has to be imported before twisted stuff, causing an import violation. Finally, I'm missing one line in `tribler.py` which starts the twistd plugin. I think that one is hard to cover.

Tribler/Test/test_as_server.py
```diff
@@ -15,37 +15,24 @@
 from Tribler.Core.TorrentDef import TorrentDef
 from Tribler.Core.simpledefs import dlstatus_strings, DLSTATUS_SEEDING, UPLOAD
 
-from Tribler.Main.Utility.utility import initialize_x11_threads
-initialize_x11_threads()
```
whirm: So happy to lose this...

Tribler/Test/test_tunnel_community_negative.py
```diff
@@ -1,133 +0,0 @@
-# Written by Niels Zeilemaker
```
whirm: The tunnel tests shouldn't be removed

Tribler/Core/Video/VideoUtility.py
```diff
@@ -1,181 +0,0 @@
-# Written by Egbert Bouman
```
whirm: are you sure all this should be deleted?

Tribler/Core/Video/utils.py
```diff
@@ -1,192 +0,0 @@
-# Written by Arno Bakker
```
whirm: and this?

Tribler/Core/Video/VideoUtility.py
devos50: We are not using any of this and if we use it, it should be implemented in the new GUI.

devos50:
Will reopen later.

PR #2360, opened 2016-06-29
Workaround a bug were settings are written the wrong way on Windows

pimveldhuisen:


pimotte:
I have carefully codereviewed this hack, and have come to the conclusion it is the most appropriate one for a problem likely no longer exists in devel.

whirm:
For future ref: This is a quick hack to work around the fact that the WX gui is horribly broken. Changing the mintray value and restarting would make the gui stop refreshing its contents while still reacting to user input.

As the WX gui is being deleted within the next few days, it doesn't make sense to pursue a decent fix for this.

whirm:
@pimotte , @pimveldhuisen: remember to mark the PRs as READY

whirm:
also: lol @ the branch name

PR #2374, opened 2016-06-30
Refactored general excepts in LibtorrentDownloadImpl

lfdversluis:


lfdversluis:
I do not see an *easy* way to cover these excepts, you got an idea @devos50 ?

devos50:
No, that seems to be really hard.

PR #2379, opened 2016-06-30
Refactored general except in VideoPlayer

lfdversluis:


lfdversluis:
retest this please

lfdversluis:
@devos50 Capturing an OS error seems to be rather hard. Can you review?

devos50:
Looks good!

Indeed hard to cover without mocking `os.popen2` (which we should not do I guess).

PR #2389, opened 2016-06-30
Integrated yappi in Tribler twistd plugin

devos50:
I moved the yappi profiler to the Tribler twistd plugin since the `tribler_profiler.py` class will soon be removed. The yappi in the twistd plugin has the same functionality as in the original `tribler_profiler.py`.


brussee:
Have you added this to the requirements list?

brussee:
> CPython standard distribution is coming with three profilers. cProfile, Profile and hotshot. cProfile module is implemented as a C module based on lsprof, Profile is in pure Python and the hotshot can be seen as a small subset of a cProfile

Which one are you using and which one will work on Android?

devos50:
@brussee I'm not sure whether we should add it to `debian/control` since it's not necessary for users to have in order to run Tribler. @whirm maybe you can say something about this?

Also, if I'm not mistaken, yappi is another profiler, besides the profiles enumerated in your quote. As far as I know, yappi is a profiler that is written in C and I'm not sure whether it works on Android.

devos50:
@lfdversluis can you review?

lfdversluis:
I am looking at a combination of my created tunnel arguments and the code I refactored in tribler_main.py. Naturally I can only accept!

devos50:
@lfdversluis nice! @whirm ready for your review

whirm:
twistd already has support for profiling:

```
  -p, --profile=       Run in profile mode, dumping results to specified file.
      --profiler=      Name of the profiler to use (profile, cprofile).
                       [default: cprofile]
```
Maybe you could check if extending the list to support yappi is feasible so we can use it in all plugins existing and future.

devos50:
@whirm done :D. You can now use Yappi using:

```
PYTHONPATH=twisted twistd --profile=prof.txt --profiler=yappi -n tribler
```

Can you review?

whirm:
I think this should be made a separate plugin

devos50:
@whirm why do we need a separate plugin for something we can toggle with a flag in our current plugin?

whirm:
Because that feature is usable across all twistd commands

devos50:
@whirm ah I see, you want to reuse the code. That makes sense.

What about moving the profiler code to a separate file in the `twisted` directory and importing that file wherever we want to use the yappi profiler? Or we can make a ServiceMaker base class that does this? Maybe you have some other suggestions?

whirm:
I don't think you need to do anything else than to move it to a yappy_plugin.py module, twistd should pick it up right away.

devos50:
@whirm I tried that (see my current code) but it only works one time. After that, I first have to remove the `dropin.cache` file. Suggestions on what could be causing this?

whirm:
Oh, that's probably because if there's no twisted serviceMaker declared inside  of the module, so it doesn't get into the cache and the next time you run twistd it's not loaded.

Have a look at the twisted documentation to see if there's an alternative way to extend twistd or to mark the module to be loaded every time. Otherwise we will have to put it somewhere else and import it from all plugins.

devos50:
After discussion with @whirm, I moved this script to Dispersy, see https://github.com/Tribler/dispersy/pull/503. I will update this PR to adjust the Dispersy pointer and integrate the yappi profiler in Tribler when that PR is merged.

devos50:
After https://github.com/Tribler/dispersy/pull/504 is merged, I will retest this PR to see whether everything is still stable after increasing the dispersy pointer.

devos50:
@whirm tests passed with the newest dispersy! Ready for your review/merge.

PR #2392, opened 2016-07-01
WIP: Refactored the download mechanism to make it more simple

devos50:
My initial idea with this PR was to only refactor the checkpoint mechanism, however, it seems that I'm ending up with a complete overhaul of the download mechanism.

The current checkpoint mechanism has a grace time parameter in there, probably because we had to wait for the GUI to be stopped. This leads to checkpoints being scheduled on the tread pool around 1.5 seconds after shutdown! Moreover, the checkpointing was combined with stopping the downloads on shutdown which is a complicated design and prone to errors.

I refactored the checkpoint mechanism so it is better streamlined in our current code. This fixes a bug where the thread pool is `None`. Also, it speeds up our test execution time tremendously: a small test reduced the average runtime of REST API tests (where a complete session is started) **from 2.3 seconds to 1.3 seconds**.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -14,10 +14,12 @@ from traceback import print_exc
 from twisted.internet import reactor
 from twisted.internet.defer import Deferred
 from twisted.internet.defer import inlineCallbacks
+from twisted.internet.task import LoopingCall
 
 from Tribler.Core.APIImplementation.threadpoolmanager import ThreadPoolManager
 from Tribler.Core.CacheDB.sqlitecachedb import forceDBThread
 from Tribler.Core.DownloadConfig import DownloadStartupConfig, DefaultDownloadStartupConfig
+from Tribler.Core.Libtorrent.LibtorrentDownloadImpl import LibtorrentDownloadImpl
 from Tribler.Core.Modules.search_manager import SearchManager
 from Tribler.Core.Modules.versioncheck_manager import VersionCheckManager
 from Tribler.Core.Modules.watch_folder import WatchFolder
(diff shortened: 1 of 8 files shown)


Tribler/Core/Session.py
```diff
@@ -472,21 +472,25 @@ def start(self):
...
-            Callback that gets called when the early shutdown has been compelted.
+            Callback that gets called when the early shutdown has been completed.
             Continues the shutdown procedure that is dependant on the early shutdown.
             :param _: ignored parameter of the Deferred
             """
-            self.checkpoint_shutdown(stop=True, checkpoint=checkpoint,
-                                 gracetime=gracetime, hacksessconfcheckpoint=hacksessconfcheckpoint)
+            try:
+                self.save_session_config()
+            except Exception as e:
```
lfdversluis: General exception, please catch the ones that can be thrown

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -595,54 +602,31 @@ def resume_download(self, filename, initialdlstatus=None, initialdlstatus_dict={
...
-                    return
-
-            self.network_shutdown()
+    def network_checkpoint_callback(self, dllist):
+        for d in dllist:
+            try:
+                infohash, pstate = d.network_checkpoint()
+                self._logger.debug("Checkpointing download: %s %s", d.get_def().get_name(), pstate)
+                self.save_download_pstate(infohash, pstate)
+            except Exception as e:
```
lfdversluis: idem as below

Tribler/Core/Session.py
devos50: Yeah I already saw this one in the pylint errors, will take a look. Currently, there's another bug that I have to fix first (see the tests on linux).

brussee:
Why did you close this?

brussee:
> Why did you close this?

^ Idem

synctext:
it is scheduled for a bigger refactor I assume

PR #2415, opened 2016-07-05
Implemented new search relevance ranking algorithm

### LINKED ISSUE #2250: The relevance ranking algorithm when searching in Tribler needs to be redesigned
When searching in Tribler, search results (which can consist of torrents and channels) are ranked according to several criteria. Channels and torrents are sorted separately from each other. The associated channels of the torrents are also determined and in case a torrent is present in more than one channel, the channel with the highest number of votes is taken.

Channels are sorted based on the number of torrents in the channel. The algorithm to sort torrents is more complex. First, an array of five scores are constructed. These values are determined as follows:
1. The number of matching keywords in the name of the torrent
2. The number of matching keywords in the name of the files the torrent contains
3. The position of the lowest matching keyword in the torrent name. For instance, when searching for ‘Away From Keyboard’ and ‘of’ is a matching keyword, this position will be 2.
4. The number of matching keywords in the extensions of the file names of the torrent.
5. A score that is based on several (normalised) attributes of the torrent. This score is determined after the set of local search results are constructed. The following formula is used: _score = 0.8 \* normalized_num_seeder - 0.1 \* normalized_num_negative_votes + 0.1 \* normalized_num_subscriptions_. Note that the number of negatives votes and number of subscriptions do not exist anymore and will always be 0. 

After the torrents and channels are sorted, the results are put in a list. Every five torrent results, a channel result is inserted. This is to make sure that channels are also showing up in the search results (and more important, are visible to the user).

Also, I think it’s better to not be dependent on other search results to make streaming torrents from the database possible (like we do now when normalising). In other words, the final relevance score in the new algorithm should only be dependent on the data that this torrent object has.

I will start thinking about a new (more simple) algorithm that we could use for this and add the proposal in this issue.

**Possible improvements:**
- Improve the relevance algorithm by taking the number of matches inside the torrent name/torrent file names in consideration. For instance, when searching for 'pioneer one', a torrent named 'pioneer one collection' probably has a higher relevance than a torrent named 'pioneer one - episode 3, season 4'. This will prevent situations like the following, when I expect 'test_tribler.txt' to show up higher than 'test.tribler.upc.iso'.
  9167a65ab13479c028a35e1a9625f2c6
- I think the lowest matching term criteria should be removed or get lower priority since I doubt that it's always true that terms that are matching earlier in torrent A than in torrent B makes torrent A more relevant. For instance, torrent A = "game of thrones" and torrent B = "thrones in someplace". If a user searches for 'thrones', torrent B will be marked as more relevant which feels weird.
- Users are making mistakes when typing. For instance, if I type "triler", I might mean "Tribler" or "triller" so I would like to see results that are matching these queries. This can be achieved by calculating the [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) with know keywords. One issue here is performance. Fortunately, sqlite contains support for levenshtein collation.
  4c590b44f5fabf0762f4fc29c01f7fa8
- The sorting of the channels is currently only dependent on the number of torrents in the channel. If I'm not mistaken, it is not even sorted on the matching in the name! See the following example for instance, where I expect the channel with 'pirate' in the name, to show up higher:
  f29b7edfd88b07d83b8d69d10746db87
- When building the FTS3 index in the sqlite database, duplicate words are removed. This means that when searching for 'years', a torrent named 'best years' will be ranked equal to a torrent named 'years and years'. However, 'years and years' should have a higher relevance since the word 'years' occurs twice. Another example is when searching for 'iso'. A torrent file that contains 100 iso files is ranked equivalent to a torrent file that only has one iso file.
- By removing the number of seeders criteria when sorting, less popular torrent have a better chance of coming up higher in the lists, increasing the chance the user will download these and seed the torrent afterwards.
- Our current algorithm only returns results that matches all given keywords. So when searching for 'pirate audio', only torrents returned that have both terms in it, are returned. It might be better to show the user also torrents matching 'pirate' and matching 'audio' (while still giving more relevance to torrents that matches 'pirate audio').

**Proposal for a new algorithm:**

The new algorithm should take the matching with keywords better into account and be less dependent on other (confusing) variables such as number of seeders and number of torrents in the channel.
- The first step should be to adjust our FTS3 index to not filter out duplicate keywords in the torrent name, file names and file extensions. This will have consequences on the amount of storage required and requires rebuilding the FTS3 index in the upgrader.
- Slightly change the FTS matching query to account for partial term results.
- Use the [BM25F](https://en.wikipedia.org/wiki/Okapi_BM25) algorithm to determine the relevance of the query in the torrent name, file names and file extensions. Note that we have to calculate the BM25 score three times. After that, we can take a weighted average of these scores. The score of the torrent name matching should have the highest priority. The relevance scores are sent to the GUI over the REST API.
- Channel name matching is done in the same way, however, the only difference is that we are not using FTS for the channels. We also calculate the BM25 score of the channel name matching. These relevance scores are sent to the GUI over the REST API.
- If after several seconds there are no incoming results, fetch the keyword(s) with the closest Levenshtein distance from the database and display them to the user with a message similar to the following: "No results found so far. Did you mean to search for ..."?

The above algorithm takes all of the improvements into account I outlined above.

devos50:
Fixes #2250


devos50:
@lfdversluis can you review? I will fix the import order error in the pylint after your review. Also, I think it's hard to reduce the number of local variables in the `search_in_local_torrents_db` methods (I'm not even sure how pylint counts them since my count results in a lower amount).

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -14,6 +15,8 @@
 from traceback import print_exc
 from collections import OrderedDict, defaultdict
 from libtorrent import bencode
+
+import math
```
lfdversluis: Import order, was noted but just so you don't forget

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -862,6 +869,58 @@ def freeSpace(self, torrents2del):
...
+        The algorithm is based on BM25. The document length factor is regarded since our "documents" are very small
+        (often a few keywords).
+        See https://en.wikipedia.org/wiki/Okapi_BM25 for more information about BM25.
+        """
+        search_results = []
+        keys_str = ", ".join(keys)
+        keywords = split_into_keywords(query, to_filter_stopwords=True)
+        infohash_index = keys.index('infohash')
+
+        results = self._db.fetchall("SELECT DISTINCT %s, Matchinfo(FullTextIndex, 'pcnalx') "
```
lfdversluis: Did you write this query?

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -862,6 +869,58 @@ def freeSpace(self, torrents2del):
...
+                                    "FROM Torrent T, FullTextIndex "
+                                    "LEFT OUTER JOIN _ChannelTorrents C ON T.torrent_id = C.torrent_id "
+                                    "WHERE t.name IS NOT NULL AND t.torrent_id = FullTextIndex.rowid "
+                                    "AND C.deleted_at IS NULL AND FullTextIndex MATCH ?"
+                                    % keys_str, (" OR ".join(keywords),))
+
+        for result in results:
+            result = list(result)  # We convert the result to a mutable list since we have to decode the infohash
+            result[infohash_index] = str2bin(result[infohash_index])
+            matchinfo = result[len(keys)]  # The matchinfo is the last element in the results tuple
```
lfdversluis: The len(keys) can be moved outside the loop. Saves a O(n) run every time.

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -862,6 +869,58 @@ def freeSpace(self, torrents2del):
...
+
+            unpack_str = 'I' * (3 * num_cols * num_phrases)
+            matchinfo = unpack_from('I' * 9 + unpack_str, matchinfo)[9:]
+
+            scores = []
+
+            for col_ind in xrange(num_cols):
+                score = 0
+                for phrase_ind in xrange(num_phrases):
+                    rows_with_term = matchinfo[3 * (col_ind + phrase_ind * num_cols) + 2]
```
lfdversluis: You may want to explain this expression.

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -862,6 +869,58 @@ def freeSpace(self, torrents2del):
...
+                    term_freq = matchinfo[3 * (col_ind + phrase_ind * num_cols)]
```
lfdversluis: So this is exactly the statement above but - 2?

Tribler/Core/Modules/restapi/search_endpoint.py
```diff
@@ -66,14 +66,16 @@ def render_GET(self, request):
...
+        query = unicode(request.args['q'][0])
+        keywords = split_into_keywords(query)
+        results_local_channels = self.channel_db_handler.search_in_local_channels_db(query)
         results_dict = {"keywords": keywords, "result_list": results_local_channels}
         self.session.notifier.notify(SIGNAL_CHANNEL, SIGNAL_ON_SEARCH_RESULTS, None, results_dict)
 
         torrent_db_columns = ['T.torrent_id', 'infohash', 'T.name', 'length', 'category',
                               'num_seeders', 'num_leechers', 'last_tracker_check']
-        results_local_torrents = self.torrent_db_handler.searchNames(keywords, keys=torrent_db_columns, doSort=False)
+        results_local_torrents = self.torrent_db_handler.search_in_local_torrents_db(request.args['q'][0],
```
lfdversluis: `request.args['q'][0]` can be replaced by `query`

Tribler/Core/Modules/restapi/util.py
```diff
@@ -23,42 +26,84 @@ def return_handled_exception(request, exception):
...
     torrent_name = torrent['name'] if torrent['name'] is not None else "Unnamed torrent"
+    relevance_score = relevance_score_remote_torrent(torrent_name)
 
     return {'id': torrent['torrent_id'], "infohash": torrent['infohash'].encode('hex'), "name": torrent_name,
             'size': torrent['length'], 'category': torrent['category'], 'num_seeders': torrent['num_seeders'],
-            'num_leechers': torrent['num_leechers'], 'last_tracker_check': 0}
+            'num_leechers': torrent['num_leechers'], 'last_tracker_check': 0,
+            'relevance_score': relevance_score}
+
+
```
lfdversluis: remove one new line

Tribler/Core/Modules/restapi/util.py
```diff
@@ -23,42 +26,84 @@ def return_handled_exception(request, exception):
...
+def relevance_score_remote_torrent(torrent_name):
+    """
+    Calculate the relevance score of a remote torrent, based on the name and the matchinfo object
+    of the last torrent from the database.
+    The algorithm used is the same one as in search_in_local_torrents_db in SqliteCacheDBHandler.py.
+    """
+    from Tribler.Core.Session import Session
+    torrent_db = Session.get_instance().open_dbhandler(NTFY_TORRENTS)
+    if torrent_db.latest_matchinfo_torrent is None:
+        return 0.0
```
lfdversluis: inconsistent return. This is a float, below you initialize the score to 0 which is an int.

Tribler/Test/Core/Modules/RestApi/test_events_endpoint.py
```diff
@@ -95,7 +95,7 @@ def test_events(self):
 
         def send_notifications(_):
             self.session.lm.api_manager.root_endpoint.events_endpoint.start_new_query()
-            results_dict = {"keywords": ["test"], "result_list": [('a',) * 9]}
+            results_dict = {"keywords": ["test"], "result_list": [('a',) * 10]}
```
lfdversluis: why did you change all these from 9 to 10?

Tribler/Test/Core/Modules/RestApi/test_util.py
```diff
@@ -1,35 +1,62 @@
...
+        mocked_db = MockObject()
+        mocked_db.latest_matchinfo_torrent = None
+        Session.get_instance().open_dbhandler = lambda _: mocked_db
+
         input = {'torrent_id': 42, 'infohash': 'a', 'name': 'test torrent', 'length': 43,
                  'category': 'other', 'num_seeders': 1, 'num_leechers': 2}
         output = {'id': 42, 'infohash': 'a'.encode('hex'), 'name': 'test torrent', 'size': 43, 'category': 'other',
-                  'num_seeders': 1, 'num_leechers': 2, 'last_tracker_check': 0}
-        self.assertEqual(convert_torrent_to_json(input), output)
+                  'num_seeders': 1, 'num_leechers': 2, 'last_tracker_check': 0, 'relevance_score': 0.0}
```
lfdversluis: Here the inconsistency would show if you added a test where the score would be set to 0 but the list would be empty, returning 0 instead of 0.0

lfdversluis:
Comments given, Windows 64 bit is failing too.

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
devos50: As far as I know, the `len` function is `O(1)` on common datasets (i.e. lists)?

Tribler/Core/Modules/restapi/search_endpoint.py
devos50: Good catch!

Tribler/Core/Modules/restapi/util.py
devos50: PEP8 specifies that there should be two newlines between methods outside a class ;)

Tribler/Core/Modules/restapi/util.py
devos50: True, I'm not sure whether this can cause errors but to be sure and consistent, I've changed the initialization of `score` to `0.0`.

Tribler/Test/Core/Modules/RestApi/test_events_endpoint.py
devos50: Because we have an additional item in the result list now: the relevance score.

devos50:
```
  File "C:\Python27\lib\threading.py", line 763, in run
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Core\Utilities\twisted_thread.py", line 55, in _reactor_runner
  File "C:\Python27\lib\site-packages\twisted\internet\base.py", line 1194, in run
    self.mainLoop()
  File "C:\Python27\lib\site-packages\twisted\internet\base.py", line 1203, in mainLoop
    self.runUntilCurrent()
  File "C:\Python27\lib\site-packages\twisted\internet\base.py", line 825, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Core\APIImplementation\threadpoolmanager.py", line 50, in caller
    fun(*args, **kwargs)
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Core\Libtorrent\LibtorrentMgr.py", line 482, in do_dht_check
    lt_session = self.get_session()
  File "C:\workspace\jenkins\workspace\GH_Tribler_PR_tests_win64\tribler\Tribler\Core\Libtorrent\LibtorrentMgr.py", line 186, in get_session
    if hops not in self.ltsessions:
exceptions.TypeError: argument of type 'NoneType' is not iterable
```

devos50:
@lfdversluis thanks for the review. I don't think the algorithm is perfect right now and it might need some minor tweaking but for now, I think it is ok for a first version of the GUI.

devos50:
@whirm ready for your review

whirm:
@devos50 is @synctext happy with the new algo?

devos50:
@whirm I will discuss it with him tomorrow.

whirm:
Ok, let me know about it

whirm:
Needs a rebase

devos50:
@synctext forgot to ask, can this be merged?

Tribler/Core/CacheDB/SqliteCacheDBHandler.py
lfdversluis: No idea why I thought len wouldn't be a property of the datastructure. Probably a brain derp.

devos50:
Let's see how this algorithm works out. Any improvements to the algorithm can be made later 👍

PR #2416, opened 2016-07-05
WIP: Handle dispersy returning Deferreds

lfdversluis:
Soon™ Dispersy will return deferreds on many calls. This PR makes sure Tribler can handle those.

The following will happen because of this PR:

The **wx GUI** will no longer function, I did not push any changes (by accident) to wx, so it will call the functions but completely ignore the deferreds returned.

**All** tests will run on the nose reactor and uses nose's `deferred` decorator.

Some tests had blocking behavior which blocked the **reactor thread** if they ran on the reactor, I have refactored this to use deferreds, which should be only a good thing.


synctext:
very nice cleaned up history :-)

lfdversluis:
All commits need to be squashed and sorted by file. Moreover, this PR modifies wx which was wrong to begin with. That combined with more than 1.5 weeks of debugging why the allchannel experiment does not perform well (reading logs for several hours a day) makes your mind go dull.

lfdversluis:
@whirm Maybe a silly question, but I have specified my own dispersy branch in the pointer (it's needed for the PR to be checked out because a lot depends on it), yet the tests fail that they cannot find stormdbmanager. Any things I need to do extra?

lfdversluis:
Darn, it checks out the wrong commit of my Dispersy branch. @whirm @devos50 do you know  how to specify the current HEAD of my branch? A log statement in Dispersy is faulty and tests are failing because of that, but this log is already removed in my branch HEAD.

devos50:
@lfdversluis do you mean something like this? http://stackoverflow.com/questions/10443627/force-git-submodule-to-always-stay-current

lfdversluis:
It's solved now, had to update my pointer

lfdversluis:
retest this please

Linux and mac passed, added storm to windows 32 and 64 bit, let's see how that goes!

lfdversluis:
All tests passed :+1:

lfdversluis:
And now a few fail :-1:

devos50:
I will close this issue in favor of IPv8.

PR #2440, opened 2016-07-07
Fix 2395 kill threadpool shim

whirm:


devos50:
I will continue working on this in #2573.

PR #2443, opened 2016-07-08
Remove session-registered notification callbacks

whirm:
when shutting down to avoid getting them triggered half way during
shutdown and crashing because the GUI or any other component has been
removed.

### DIFF
--- a/Tribler/Main/tribler_main.py
+++ b/Tribler/Main/tribler_main.py
@@ -830,6 +830,25 @@ class ABCApp(object):
             # Niels: lets add a max waiting time for this session shutdown.
             session_shutdown_start = time()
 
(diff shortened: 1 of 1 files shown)


devos50:
@whirm please also remove this observer if you can: https://github.com/Tribler/tribler/blob/next/Tribler/Main/vwxGUI/home.py#L878. This is causing `PyDeadObjectErrors` in the tests.

whirm:
I'm not sure where this should be removed. I've already pushed a workaround so no more dead object errors happen. 
But as the gui is being removed I don't think it's worth making a proper fix for it.

whirm:
Once this and @brussee 's fixes are merged to `next` I'll update my next to devel PR and get it merged.

PR #2446, opened 2016-07-08
Fix unhandled ex catcher

whirm:


whirm:
retest this please (clone timeout)

whirm:
retest this please (Now that I want it to fail, it doesnt)

PR #2448, opened 2016-07-08
Removed my channel GUI tests

devos50:
The stuff I removed here is already tested by the API endpoints tests/unit tests

The code coverage might drop a bit, I will write some tests to counter this.


devos50:
The drop in coverage was minimal. I've written some small additional tests to stabilize the drop in coverage.

whirm:
retest this please

devos50:
@lfdversluis can you review?

lfdversluis:
Looks ok. Does these test really cover all that you just deleted?

devos50:
@lfdversluis there are probably a few lines of code that are now missing but these are hard to track. This is due to the fact that these big session tests are hard to control. For instance, in a specific run of the tests, it might be that some error handling code is triggered while this is not the case in some other (more successful) run.

I've tried my best to identify what's different between these old tests and the new refactored tests (which is also not easy since the coverage metric settings are slightly different between the `test_tribler_devel` and `GH_Tribler_PR` jobs) and I've added some additional tests to hopefully stabilize the code coverage.

Anyway, if you can identify some missing lines of code, I will gladly create some unit tests for these.

lfdversluis:
I would flag it as READY. It is more important to get rid of wx at this stage so devel can move forward. If we find uncovered lines with nose and coverage, we can always add tests.

lfdversluis:
(Of course critical code should be covered, we always must have a basis to build on)

devos50:
@lfdversluis I agree, other people are dependent on the wx removal at the moment and writing unit tests can always be done (but should be done in the long run lol).

devos50:
@whirm ready for you review

PR #2468, opened 2016-07-14
ON HOLD: Channel boosting twisted plugin

ardhipoetra:
In this PR, I will work on credit mining twisted plugin, especially for channels.


devos50:
@ardhipoetra good work but I think it might be better to integrate this credit mining plugin in the "main" tribler plugin and just make channel boosting an option you can pass to the script. Otherwise, we keep repeating the same code several times throughout scripts. @whirm what do you think about this?

ardhipoetra:
I agree, credit mining need almost full-fledged session anyway. Can I perhaps extend the class or do it in separate module (that will be called in tribler main plugin)?

devos50:
I'm thinking about a solution where we create classes that are reusable across several plugins since we have some other 'supporter' plugins that are not converted yet (see the files in `Tribler/Main`). I'm afraid that if we are going to put everything in `tribler_plugin.py`, it will become too large. Maybe you can come up with a good solution for this?

synctext:
no opinion on this. general input: features of Tribler will be almost frozen for coming years. We want mature and perfect things.

No need for future extension of core. just refactoring.

ardhipoetra:
This PR is one of the pending stuff in the previous big credit mining PR. I will/have to do it sooner or later. Considering what you said, I will put this into hold right now.

devos50:
No need to create a new plugin for credit mining; please try to integrate it in our current Tribler twistd plugin script.

ardhipoetra:
@devos50 I will close this PR for now. I will do that in another PR instead.

PR #2471, opened 2016-07-15
Removed check when subscribing to channel

devos50:
Since it is not necessary for the channel to be discovered in order to subscribe to this channel.

### DIFF
--- a/Tribler/Core/Modules/restapi/channels/channels_subscription_endpoint.py
+++ b/Tribler/Core/Modules/restapi/channels/channels_subscription_endpoint.py
@@ -84,10 +84,8 @@ class ChannelsModifySubscriptionEndpoint(BaseChannelsEndpoint):
         """
         request.setHeader('Content-Type', 'text/json')
         channel_info = self.get_channel_from_db(self.cid)
-        if channel_info is None:
-            return ChannelsModifySubscriptionEndpoint.return_404(request)
 
-        if channel_info[7] == VOTE_SUBSCRIBE:
+        if channel_info is not None and channel_info[7] == VOTE_SUBSCRIBE:
             request.setResponseCode(http.CONFLICT)
             return json.dumps({"error": ALREADY_SUBSCRIBED_RESPONSE_MSG})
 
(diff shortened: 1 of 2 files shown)


devos50:
@lfdversluis can you review?

devos50:
Also, unit tests appear to be quite stable now 👍

Tribler/Core/Modules/restapi/channels/channels_subscription_endpoint.py
```diff
@@ -84,10 +84,8 @@ def render_PUT(self, request):
         """
         request.setHeader('Content-Type', 'text/json')
         channel_info = self.get_channel_from_db(self.cid)
-        if channel_info is None:
-            return ChannelsModifySubscriptionEndpoint.return_404(request)
 
-        if channel_info[7] == VOTE_SUBSCRIBE:
+        if channel_info is not None and channel_info[7] == VOTE_SUBSCRIBE:
```
lfdversluis: So if the channel cannot be found (is `None`) then you have already subscribed to it?

devos50: No, if the `channel_info` is `None`, the channel is not discovered and this if-statement will never be entered.

I've added the first part of the if-statment to make the 2nd part after the `and` not crash if the channel is not discovered.

lfdversluis: It wouldn't crash previously with that (now removed) if statement?

devos50: Nope because I was returning an error 404 then if `channel_info` is `None` so it would never reach line 90 in the old code.

lfdversluis: My point exactly, so why wouldn't you return a 404 anymore?

devos50: Ah, because it's not necessary for a channel to be discovered if you want to subscribe to it. So if someone passes you a dispersy id of the channel, you can subscribe to it, whether you have discovered that channel already or not. Dispersy takes care of that. By returning the 404 error, I make this possible (again).

lfdversluis: Ok then, all is good

devos50:
@whirm can you review?

PR #2481, opened 2016-07-19
READY: absolute path for android

brussee:
find_executable relies on the executable to be on the path, but on android it is in the current folder and not on the path


brussee:
@whirm Ready.

PR #2485, opened 2016-07-21
Next to devel

whirm:


ardhipoetra:
@whirm, commit https://github.com/Tribler/tribler/pull/2485/commits/e3f15ea314bee11abe36346a8d9be411ccfa4949 should not be merged if all of credit mining fix is not merged yet. Can you revert the merged branch so I can fix this in PR 2461?

PR #2488, opened 2016-07-22
WIP: Excluded pyqt5 from build

devos50:
This will reduce the size of Tribler.app from around 231mb to 100.5mb 👍 

Py2app somehow manages to include pyqt5 in the app bundle. The problem is that I cannot remove pyqt5 from the site packages, otherwise my new GUI won't work anymore on the builder.

Hopefully we can switch to PyInstaller soon!


devos50:
@whirm ready

PR #2509, opened 2016-08-04
Multichain statistics REST endpoint

Captain-Coder:


Captain-Coder:
@devos50 would you like to review? REST seems to be your forte.

devos50:
```
55: W0221: (arguments-differ), MultiChainCommunity.initialize: Arguments number differs from overridden 'initialize' method
```

Can you fix this one?

Tribler/Core/Modules/restapi/multichain_endpoint.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+
+        child_handler_dict = {"stats": MultichainStatsEndpoint}
+
+        for path, child_cls in child_handler_dict.iteritems():
+            self.putChild(path, child_cls(session))
+
+
+class MultichainStatsEndpoint(resource.Resource):
```
devos50: I already created a `DebugEndpoint`. I think it makes more sense if we make this resource a sub-node of the `DebugEndpoint` (i.e. so it can be accessed by `/debug/multichain`).

Tribler/Core/Modules/restapi/multichain_endpoint.py
```diff
@@ -0,0 +1,77 @@
...
+        Search for the multichain community in the dispersy communities.
+        """
+        for community in self.session.get_dispersy_instance().get_communities():
+            if isinstance(community, MultiChainCommunity):
+                return community
+        return None
+
+    def render_GET(self, request):
+        """
+        .. http:get:: /multichain/stats
```
devos50: Minor detail: I would spell out `stats` to `statistics`, both in the URL and JSON response.

Tribler/Test/Community/Multichain/test_multichain_database.py
```diff
@@ -166,8 +166,8 @@ def test_get_total_not_existing(self):
         # Act
         (result_up, result_down) = self.db.get_total(self.block2.public_key_requester)
         # Assert
-        self.assertEqual(-1, result_up)
-        self.assertEqual(-1, result_down)
+        self.assertEqual(0, result_up)
```
devos50: This doesn't seem to be related to the API?

devos50:
@Captain-Coder comments given. Before I merge this PR, I would like someone else to review it. Maybe @pimveldhuisen ?

Tribler/Test/Community/Multichain/test_multichain_database.py
Captain-Coder: I know, it seems like a wierd one. But... it is.. uhmmm.. inconvenient if the api reports the totals as -1 and -1. I suddenly had -2 for total contribution in the QT-gui. And there is no semantic reason to not default to (0, 0). Blocks with both zero up and zero down are invalid, and up and down cannot be negative, so after the first block we cannot have (0, 0). There are also changes to community.py _get_total and database.py get_total that result in 0,0 as the default. The test had to change too to reflect this change.

Similarly, I wonder if the last sequence number should read 0 instead of -1.

Also this change is inline with the upcomming true-halves for multichain.

Tribler/Core/Modules/restapi/multichain_endpoint.py
Captain-Coder: Hummm, interresting point. I'll admit I used your DebugEndpoint as a template for this resource. Mainly I thought it verly likely that there would be further resources in the future. For example /multichain/block/<id>. Or validation results. So I decided to make the /multichain/stats (or /statistics) a separately named resource as opposed to having just a /multichain resource that always serves the stats.

Then the placement under /debug. What is your reasoning about placing the multichain statistics under the /debug resource? I'm not convinced this resource is purely diagnostic. I already have a GUI usecase for it.

Captain-Coder:
@devos50, I'd like to fix the "W0221: (arguments-differ)" item. I looked at it before but the big problem is that I need a tribler session to get to the notifier. I can get that either as an argument to the initialize method (a solution I copied from the tunnel community), or I can do Session.get_instance (which we discussed in IRC as being undesirable). Or did I miss another way of getting to it?

It is perfectly fine to have a None session and/or notifier (and that is the default of this karg), so I expect adding this argument will not upset any code that assumes this community follows the .initialize() convention from dispersy.

Tribler/Core/Modules/restapi/multichain_endpoint.py
pimveldhuisen: I agree this should not be categorized as debug. We currently use it as such, but this should eventually be used to give the user an overview of his contribution to the community.

devos50:
retest this please

devos50:
jenkins had a hiccup

Tribler/Core/Modules/restapi/multichain_endpoint.py
devos50: Ok, let's keep it this way and create a separate root endpoint for `multichain` then ;)

Tribler/Test/Community/Multichain/test_multichain_database.py
devos50: Ok great!

Captain-Coder:
test_remote_channel_search had a timeout, retest this please

devos50:
@pimveldhuisen do you have additional comments on this PR? If you agree, I will merge it.

pimveldhuisen:
I must admit that I don't fully grasp the 'arguments-differ' issue. Otherwise the PR looks fine to me.

PR #2514, opened 2016-08-10
Moved Category package inside Core

devos50:
Since the family filter is a component of libtribler, I propose the move the code inside the Core package. This also fixes the cyclic dependency between the `Category` and `Core` packages.


devos50:
@lfdversluis can you review?

lfdversluis:
looks ok

devos50:
@whirm or @brussee can you also review this?

PR #2523, opened 2016-08-14
Create torrent REST API endpoint changed to POST

### LINKED ISSUE #2522: Create torrent REST API endpoint should follow HTTP spec
GET requests do not have a body https://tools.ietf.org/html/rfc2616#section-9.3

brussee:
Fixes #2522


devos50:
@brussee I just checked, `POST` is indeed the preferred operation for creating a torrent according to https://en.wikipedia.org/wiki/Representational_state_transfer 👍

PR #2532, opened 2016-08-31
Moved multichain crawler service to systemd directory

Captain-Coder:
The multchain-crawler.service was in the twisted plugins directory initially, however we have since added (or more likely I didn't notice) the systemd directory in the root. So it should probably move there.


devos50:
Great, thanks!

PR #2550, opened 2016-09-21
stopVideoIfEqual: don't raise if video player is not initialized

vi:


synctext:
OK to test

devos50:
ok to test

devos50:
@vi thank you for your contribution, however, we don't make changes to the GUI on the `devel` branch anymore. Please reopen this PR to make it point to the `next` branch.

Also, please be a bit more specific about which issue this exactly solves and if possible, how to reproduce it (create a new issue if it does not exist yet).

vi:
Is it OK to open pull request to "next" branch without perosnally checking if it works or needed on the branch?

I though it's easier to just fix the thing than to describe those things in issue.

devos50:
@vi what problem is this PR fixing exactly? Is it an issue that you are experiencing or someone else?

vi:
I've updated `devel` and discovered that deleting, stopping, starting, etc. downloads no longer worked. It wants to check status of embedded video player, but the video player is not available (probably should be separate issue), so the check crashes.

So the simple fix: if video player is not available (`None`), then obviously no need to stop any video in it.

lfdversluis:
@vi yup. The devel branch contains experimental code which may break functionality. This includes the new GUI. As we abandon wx real soon, all (minor) changes will be applied to next. Next is the next (stable) release which features bug fixes mostly for 6.5.x. 

I would suggest you open your PR against next and if all checks pass, it can be merged.

vi:
Can't go back to "next": #2551 .

I think old and new GUI should try to coexist for some time (at least in devel), at least while fixing the old GUI is easy enough.

PR #2564, opened 2016-10-06
WIP: replace session configuration system by ConfigObj

### LINKED ISSUE #1822: We should have a single config file/object
That's why we have sections, it doesn't make sense to have 3 different ways to access/set config options with different features (one upgrades across config versions, the other doesn't even check for that, etc.) which leads to silly bugs and general confusion.

I would also discuss if it's worth switching to ConfigObj, it makes some things easier.

devos50:
In this PR, I refactored the complete configuration system of Tribler. This involves refactoring such that the configuration object is a field in the `Session` object. This distinction makes it easier to merge the `LaunchManyCore` and `Session` classes later.

`ConfigObj` is easier to use than the system we are using now. Moreover, the two configuration files we currently have to deal with, can be merged together to have one single Tribler configuration file. Finally, `ConfigObj` makes it very easy to return a dictionary with all the settings available. We can also pass a config specification that allows us to verify the validity of the configuration file and to set some default values.

Fixes #1822


devos50:
Only tests on Mac are not passing due to a known bug (that apparently surfaces now we are using `ConfigObj`). I will hunt that one down now.

devos50:
All tests are passing! Time to clean up this PR, test it well using the AllChannel experiment and modify Gumby to be compatible with this 👍

devos50:
In favour of #2841

PR #2569, opened 2016-10-12
WIP: Refactored various tests and methods in the base test class

### LINKED ISSUE #2437: Refactor setup file server in tests
Duplicate code is showing up to setup a file server in tests, like here:

https://github.com/brussee/tribler/blob/1ca1ab7e1712911186a97ebfaccb395929298fc9/Tribler/Test/Core/test_torrent_def.py#L38

Should be refactored.

devos50:
In this PR, I refactored various tests to make them use deferreds instead of hard-coded sleep statements. Moreover, I moved tests to other packages that are more suitable.

I also removed some methods that are now unused in the base test class. These methods were mostly used by the wx GUI and were replaced by Twisted primitives if possible.

Fixes #2437.


devos50:
Most of this has been addressed in #2626

PR #2578, opened 2016-10-19
READY: Added save_resume_data None case

### LINKED ISSUE #2577: [devel] save_download_pstate AssertionError

- [X] I have read CONTRIBUTING.md

- [X] I have tried with the latest pre-release version and I still can reproduce the issue.

3aa6bb204bbcc0cf7be3d6e8a355baf544ef6cb6

##### Behavior: #####
When you close Tribler when it is still initializing downloads it will throw an uncaught AssertionError.

```
ERROR   1476869025.44  LaunchManyCore:703   Exception while checkpointing: MyTestTorrent
  File "/home/quinten/tribler/tribler/Tribler/Core/APIImplementation/LaunchManyCore.py", line 700, in network_checkpoint_callback
    self.save_download_pstate(infohash, pstate)
  File "/home/quinten/tribler/tribler/Tribler/Core/APIImplementation/LaunchManyCore.py", line 867, in save_download_pstate
    self.downloads[infohash].save_resume_data())
  File "/home/quinten/tribler/tribler/Tribler/dispersy/taskmanager.py", line 38, in register_task
    assert isinstance(task, (Deferred, DelayedCall, LoopingCall)), (task, type(task) == type(Deferred))
AssertionError: (None, False)

```
##### Cause: #####
`LibtorrentDownloadImpl.save_resume_data()` is annotated as `@waitForHandleAndSynchronize()`: this means the default return value is `None` if it cannot synchronize/is shutting down.

##### Fix: #####
Simply check if `self.downloads[infohash].save_resume_data()` is not `None` before registering the task in `LaunchManyCore.save_download_pstate`.

qstokkink:
This makes sure that when closing Tribler, uninitialized pstates do not cause AssertionErrors.
This fixes #2577 (see for more info).

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -784,9 +784,11 @@ class TriblerLaunchMany(TaskManager):
         """ Called by network thread """
 
         self.downloads[infohash].pstate_for_restart = pstate
+        save_resume_data = self.downloads[infohash].save_resume_data()
 
-        self.register_task("save_pstate %f" % timemod.clock(),
-                           self.downloads[infohash].save_resume_data())
+        if save_resume_data:
+            self.register_task("save_pstate %f" % timemod.clock(),
+                               save_resume_data)
 
     def load_download_pstate(self, filename):
         """ Called by any thread """


qstokkink:
@devos50 ready for review.

devos50:
@qstokkink a better solution would be to have to use deferreds instead where the deferred fires if the handle is available. This avoids the need to check for None and is more in line with our usage of Twisted. However, this requires a major refactoring of `LibtorrentDownloadImpl` as many methods are dependent on the torrent handle so I guess this is the best solution until we are moving to deferreds.

brussee:
@devos50 What is your policy on merging fixes on next to devel?

devos50:
@brussee as soon as we have some fixes available on `next`, I will perform the next to devel merge (excluding GUI-related fixes). Why?

brussee:
@devos50 This PR is one of those fixes on next, and since it concerns the download feature, which I need to work perfectly, I would like to have this fix on devel as well :)

Issue #1882, opened 2016-01-20
slow anonymous downloads: Crypto CPU bottleneck


synctext:
The CPU seams to be the reason for slow <1 MByte/sec anonymous downloads.

_possible problem_
Running crypto on twisted thread blocks all other Tribler activity. Unclear if we needs 256bit GCM mode. Anything that checks a signature, decrypt a message, etc. needs to be traced.

_possible long-term solution_
Separate thread for the tunnel community or even an isolated thread for each community.
Low hanging fruit: parallel execution of relaying, make it multi-threaded. Real threads: Twisted reactor versus import multiprocessing...

_goal_
1) benchmark raw openssl C code GCM MBps.

2) Create a minimal benchmark comparing current situation in Tribler with alternatives. Not re-using Tribler code, but a performance test processing 10.000 UDP 
- Only generating packets and decrypting one layer, then another layer. Current approach.
- Doing the same on a threadpool within Twisted.
- Separate Python OS process, communicate using a ringbuffer. Twisted has build-in support for this, treating it like just another thread.

EDIT: use 10k UDP packets through exit node as benchmark.

whirm:
@synctext I don't think this is related to 6.6

synctext:
Within the roadmap for _high-performance privacy_ it is important to identify bottlenecks. We do not know if the CPU, IO, buffering, or networking is the bottleneck.

The following experiment determines the crypto performance of the HiddenTunnelCommunity at the downloading side.

Step-by-step we build and measure the raw networking throughput expressed in MByte/second in various setups. First is the minimal Dispersy NULL receiver, dropping packets immediately coming from several senders. Next is receiving the packet on the network thread and creating a Twisted event to handle this packet; again by simply dropping it. Third is using the on-data approach of the hidden community. Fourth experiment is benchmarking a single layer of Tor decryption at the receiver. Finally, the full tunnel community.
![image](https://cloud.githubusercontent.com/assets/325224/14645652/cca06f02-0656-11e6-90d0-70250f569b88.png)

synctext:
Connected to: https://github.com/Tribler/tribler/issues/2137

synctext:
Approach: our exit node helpers are maxed out in CPU. It will be extremely valuable to profile them, identify hotspots, test improvements, and benchmark various changes in the field.
This could move our donated network capacity from 9 TByte/day to perhaps even 15TByte/day.

lfdversluis:
[...]

PR #2607, opened 2016-11-01
WIP: TunnelCommunity multiprocessing

qstokkink:
Related to #1882 .

This PR adds a multiprocessing option for the TunnelCommunity by using Twisted's reactor.spawnProcess.

### Architecture
The architecture of this implementation consists of two halves:
 1. The main process's view
 2. The subprocess's view

The following subsections will cover these two views. After that the generic process interface will be discussed.

#### Main process view
classtree_full_parent
Using the `pooled` option in the settings the main process can decide to substitute its TunnelCommunity (/HiddenCommunity) for a PooledTunnelCommunity. This PooledTunnelCommunity offers the same interface as the normal TunnelCommunity. The big difference of course, is that it offloads _(1)_ circuit creation (`create_circuit`), _(2)_ sending LibTorrent data (`send_data`) and _(3)_ retrieving download info/peers (`monitor_downloads`) to other (child) processes.

This offloading to child processes is handled through the ProcessManager class, leaving the PooledTunnelCommunity solely in charge of decorating the TunnelCommunity interface. This keeps the code clean and focused. 

The wrapper for the child processes which the ProcessManager uses, is the TunnelProcess class (`processes/tunnel_childprocess.py`). The TunnelProcess takes care of the high-level interfacing with the TunnelSubprocess class running on the actual child process (more on that in the next subsection).
The low-level twisted Protocol interfacing is handled by the TunnelProcess's parent, the ChildProcess.

#### Subprocess view
classtree_subprocess_fuller
The TunnelSubprocess is in charge of running a TunnelCommunity which supports RPC calls from the main process. This is achieved (in line with the existing `tunnel_helper` code) by running a very light-weight Session. The low-level interfacing with the main process is handled by the TunnelSubprocess's parent class, the Subprocess class.

Note that the TunnelSubprocess is in charge of all encryption/decryption and management of circuit and hop constructs. The TunnelSubprocess will share its circuits and hops with the main process, but it is the only process allowed to edit these.

#### The IProcess
To minimize duplicate code, both the TunnelProcess and TunnelSubprocess inherit from the IProcess interface. This interface is implemented on the ChildProcess by means of the twisted ProcessProtocol.transport and on the Subprocess using `os` file descriptors and raw threads. These raw threads on the Subprocess are delegated to the twisted threadpool a.s.a.p. to avoid threading issues. Building on the IProcess is the RPCProcess which provides generic serialization for RPC calls.

The generic RPC interface, in turn, allows for object synchronization using the SyncDict class. This class is a subclass of the `dict` class, which checks for updates to its entries (RemoteObjects) and sends out minimized update messages/diffs to synchronize the main process's view of the shared objects.

### Data flows
There are three data flows between the main process and a child process. These consist of:
 1. Control data
 2. Raw data
 3. Exit messages

These data flows are implemented using separate process file descriptors (so, next to stdin, stdout and stderr). Note that this leaves the stdout and stderr open for debugging purposes.
The rationale for each of these streams will be discussed in the following subsections; each of these different flows has radically different data characteristics, which would otherwise interfere.

#### Control flow
The control flow handles messages with **high diversity** and **low volume**. This comes down to the main process requesting community setup, circuit creation and download monitoring and the child process forwarding notifications, SyncDict changes and circuit deaths.

#### Raw data flow
The raw data flow deals with **no diversity** and **high volume** messages. Its only purpose is to serialize and deserialize LibTorrent data as fast as possible.

#### Exit messages
The exit messages have a **low latency** requirement. If these messages were to have to wait for both data and control messages this would add unnecessary delay to Tribler shutdowns. This turned out to be especially important in a previous Proof of Concept, in which a bug clogged the pipe and the process could not exit (non-forcefully).

### Reviewing this PR
The commit history of this PR has been fabricated in such a way that:
 1. The first commit handles TunnelCommunity and HiddenCommunity refactoring, such that they utilize SyncDicts and allow for control message propagation to an unknown process entity, should it exist (which it doesn't at this point).
 2. The second commit adds the process views as discussed in the architecture section
 3. The third commit adds the intergration into the Tribler options/defaults
 4. The fourth commit adds all of the unit tests


qstokkink:
Ok.. within the same timestamp a 1.9MB download goes from 0.0 to 1.0 after checkpointing. @devos50 is this the infamous checkpointing bug? (Or is this my error?)

```
DEBUG   1478006948 LibtorrentDownloadImpl:LibtorrentDownloadImpl:490  852ffeabfd7c56d102721fe4ded93ada2bb714ea get resume data {'active_time': 0, 'num_incomplete': 16777215, 'announce_to_lsd': 1, 'seed_mode': 0, 'pieces': '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'paused': 0, 'seeding_time': 0, 'last_scrape': 0, 'info-hash': '\x85/\xfe\xab\xfd|V\xd1\x02r\x1f\xe4\xde\xd9:\xda+\xb7\x14\xea', 'upload_rate_limit': 0, 'max_uploads': 16777215, 'max_connections': 16777215, 'unfinished': [], 'num_downloaded': 16777215, 'total_downloaded': 0, 'priority': 0, 'file-format': 'libtorrent resume file', 'peers6': '', 'added_time': 1478006947, 'banned_peers6': '', 'file_priority': [1], 'last_seen_complete': 0, 'total_uploaded': 0, 'piece_priority': '\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01', 'file-version': 1, 'announce_to_dht': 1, 'auto_managed': 0, 'save_path': '/home/jenkins/workspace/GH_Tribler_PR_tests_linux/tribler/Tribler/Test/data', 'peers': '', 'completed_time': 0, 'blocks per piece': 2, 'download_rate_limit': 0, 'libtorrent-version': '1.0.7.0', 'banned_peers': '', 'sequential_download': 0, 'announce_to_trackers': 1, 'num_complete': 16777215, 'finished_time': 0, 'last_upload': 0, 'trackers': [['http://localhost/announce']], 'super_seeding': 0, 'last_download': 0}
DEBUG   1478006948 LibtorrentDownloadImpl:LibtorrentDownloadImpl:496  tlm: network checkpointing: to file /home/jenkins/workspace/GH_Tribler_PR_tests_linux/tmp/tmpnCpnNV_tribler_test_session/dot.Tribler2/dlcheckpoints/852ffeabfd7c56d102721fe4ded93ada2bb714ea.state
DEBUG   1478006948 LibtorrentDownloadImpl:LibtorrentDownloadImpl:822  Torrent video.avi PROGRESS 1.0 DLSTATE 4 SEEDTIME 1
```

devos50:
@qstokkink no I can't explain that error unfortunately.

qstokkink:
Ok, I see the issue now.
The torrent being downloaded has the save dir: `Tribler/Test/data`
But, it should have the save dir: `TriblerDownloads` as defined by `test_as_server.AbstractServer.dest_dir`

qstokkink:
This is an actual bug in my code. Confirmed to fail on localhost reliably for this branch and pass reliably on devel.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
Alright, the bug is fixed by replacing a line of code I though was duplicate.

The cause is that the `LaunchManyCore` adds its own `tunnel_community` to itself (as it should), but the tests disable this and inject their own community and have the `tunnel_community` fix the reference in the `Session` for them.

We should probably look at refactoring the following line of code into the tests, where it belongs:
[https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/tunnel_community.py#L321](https://github.com/Tribler/tribler/blob/6d3c99501251499577c9b399748c28976121560e/Tribler/community/tunnel/tunnel_community.py#L321)

qstokkink:
@devos50 ready for whenever you feel masochistic enough to review

devos50:
@qstokkink there is much code coverage to gain. When will you write the unit tests for these uncovered lines?

devos50:
@qstokkink I also see many pylint errors that are easy to fix

qstokkink:
@devos50 As far as code coverage goes, I can write unit tests for the following classes somewhat easily:
- Tribler/community/tunnel/processes/iprocess.py
- Tribler/community/tunnel/processes/line_util.py
- Tribler/community/tunnel/processes/processmanager.py
- Tribler/community/tunnel/processes/rpc_defs.py
- Tribler/community/tunnel/processes/rpcprocess.py
- Tribler/community/tunnel/processes/tunnel_childprocess.py
- Tribler/community/tunnel/remotes/circuit.py
- Tribler/community/tunnel/remotes/hop.py
- Tribler/community/tunnel/remotes/relayroute.py
- Tribler/community/tunnel/remotes/remote_object.py
- Tribler/community/tunnel/remotes/sync_dict.py

And I really should, you are correct: I'll mark this as WIP until I have implemented these. The rest would really make more sense being covered by an integration test.

As far as the pylint errors go, I believe fixing them all would be very much possible and very much treating the metric: I am prepared to defend each of these violations.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
Alright, unit tests are now covering all classes which are **not**:
 1. A data class
 2. In charge of spawning processes
 3. Non-functional without an actual process

I think these would be better suited in a Gumby integration test, which will follow later.

Note that inclusion of category 2 would mean having to come in with root access once in a while to clean up orphaned processes. Category 3's would be possible using a really fat testing framework, but then the unit tests _(a)_ become small integration tests, _(b)_ would also require the classes themselves to be refactored to allow for testing and _(c)_ only serve to bring coverage up to 100% for trivial lines, while advanced functionality is already covered by the other unit tests.

For now I will mark this as READY again, feel free to challenge me on any pylint violation or class assignment to the integration tests.

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -275,9 +278,31 @@ def load_communities():
                                                        load=True,
                                                        kargs=multichain_kwargs)
 
-                        from Tribler.community.tunnel.hidden_community_multichain import HiddenTunnelCommunityMultichain
-                        self.tunnel_community = self.dispersy.define_auto_load(
-                            HiddenTunnelCommunityMultichain, dispersy_member, load=True, kargs=tunnel_kwargs)[0]
+                        if self.session.get_tunnel_community_pooled():
+                            if self.session.get_tunnel_community_test_pooled():
```
devos50: I don't think we should add test-dependent code to our Core code base. Instead, you can load this test community in the test(s) where you need this.

devos50:
@qstokkink I started reviewing, however, it might take a few days because I'm also busy with wrapping up the first alpha release of Tribler 7.

Regarding the tests: the coverage of you code is quite low and we want more coverage by unit tests. While at first glance, it seems that writing tests for various classes leads to integration tests, it is often not very hard to create small self-contained unit tests using mocking. Take for instance the `ProcessManager` class (https://github.com/qstokkink/tribler/blob/tunnelprocess_rc2/Tribler/community/tunnel/processes/processmanager.py). Using mocking, it's not very hard to verify the correct behaviour of methods in this class.

I don't expect that you cover all possible lines (since some are almost impossible to cover), however, the coverage of your code by the means of unit tests must be higher. The lack of unit tests is something that has plagued Tribler for many years and has lead to many regression bugs. Please try to keep these unit tests as small as possible and make sure they are only testing the thing that you actually want to test.

You can take a look at the test suite in Tribler, especially the Core tests. There are various classes which are tested using mocking. If you have any questions, please let me know!

qstokkink:
@devos50 I hate having to implement a factory pattern/adding complexity just for testing purposes. Your point about regression bugs is very valid though, so I will (attempt to) squeeze out some more tests.

devos50:
@qstokkink I've tried various torrents on MacOS (Ubuntu, Pioneer, Big Buck Bunny). Some of them are starting the download and others are not. When the download is starting, the subprocesses seems to work correctly 👍

22317dd4b03576499e88ac291271d71b

devos50:
Also, review is coming!

qstokkink:
@devos50 Thanks for the efforts! This last week-and-a-half I have been doing thesis writing, so no real progress on the additional tests: [although I have started working on the Gumby experiment.](https://jenkins.tribler.org/job/pers/job/tunnel_pooled_qstokkink/)

Also, I guess it's at least good to see that the CPU usage is under control, even though the download speed hasn't improved.

Tribler/community/tunnel/processes/subprocess.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @staticmethod
+    def close_all_streams():
+        """Close all registered file descriptors
+
+        :returns: None
+        """
+        # We use the fact that they are assigned
+        # to the range [3, 8].
+        for fno in range(3, 9, 1):
```
devos50: Minor detail: `xrange` requires less memory and lazy-evaluates the `fno` variable in the loop. I prefer to use that over `range`.

There are more occurrences where are you using `range` in the code.

Tribler/community/tunnel/processes/subprocess.py
```diff
@@ -0,0 +1,192 @@
...
+        def helper():
+            lock.acquire(True)
+            try:
+                f.write(packed)
+                f.flush()
+            except IOError:
+                pass
+            finally:
+                lock.release()
+        threading.Thread(target=helper).start()
```
devos50: Why are you creating a thread every time you want to write something to the parent process? What about the overhead due to context switches?

Tribler/community/tunnel/processes/tunnel_subprocess.py
```diff
@@ -0,0 +1,277 @@
...
+class TunnelSubprocess(RPCProcess, Subprocess):
+
+    """The child process's view of the parent process
+
+    In other words, this controls all of the constructs
+    required by the parent process. Most importantly it
+    manages a TunnelCommunity.
+
+    TODO/Future work in this file:
+     - Forward notificiations
```
devos50: Typo in notifications

Tribler/community/tunnel/processes/tunnel_subprocess.py
```diff
@@ -0,0 +1,277 @@
...
+        :type subject: str
+        :param changetype: the change type
+        :type changetype: str
+        :param objectID: the change object
+        :type objectID: object
+        :param args: the optional arguments
+        :type args: [object]
+        :returns: None
+        """
+        reactor.callFromThread(self.setup_community)
```
devos50: This notification is now called from the reactor thread in a very recent commit by me.

Tribler/community/tunnel/processes/tunnel_subprocess.py
```diff
@@ -0,0 +1,277 @@
...
+        config = SessionStartupConfig()
+        working_dir = os.path.join(config.get_state_dir(),
+                                   "SUBPROCESS" + str(os.getpid()))
+        if not os.path.exists(working_dir):
+            os.makedirs(working_dir)
+
+        # Configure MiniSession
+        config.set_state_dir(working_dir)
+        config.set_torrent_checking(False)
+        if not test_mode:
```
devos50: Test-related code.

Tribler/community/tunnel/processes/tunnel_subprocess.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def on_rpc_monitor_infohashes(self, infohashes):
+        """Call monitor_infohashes
+
+        :param infohashes: the infohash tuples to monitor
+        :type infohashes: [(str, int, int)]
+        :return: RPC response code
+        :rtype: str
+        """
+        if not self.session_started:
```
devos50: Shouldn't these kind of events be logged?

Tribler/community/tunnel/processes/tunnel_subprocess.py
```diff
@@ -0,0 +1,277 @@
...
+            return False
+        return blockingCallFromThread(reactor,
+                                      self.community.create_circuit,
+                                      goal_hops,
+                                      ctype,
+                                      None,
+                                      required_endpoint,
+                                      info_hash)
+
+    def on_data(self, s):
```
devos50: Please rename this parameter to be more meaningful

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -1401,3 +1427,22 @@ def increase_bytes_received(self, obj, num_bytes):
         else:
             raise TypeError("Increase_bytes_received() was called with an object that is not a Circuit, " +
                             "RelayRoute or TunnelExitSocket")
+
+    def set_process(self, process):
+        """Set the TUnnelSubprocess we belong to
```
devos50: Extra capital letter in the word TunnelSubprocess

Tribler/community/tunnel/tunnel_community.py
```diff
@@ -895,11 +913,11 @@ def _ours_on_created_extended(self, circuit, message):
             self.remove_circuit(circuit.circuit_id, "error while verifying shared secret, bailing out.")
             return
 
-        circuit.add_hop(hop)
+        circuit.add_hop(hop_id)
         circuit.unverified_hop = None
 
         if circuit.state == CIRCUIT_STATE_EXTENDING:
-            ignore_candidates = [self.crypto.key_to_bin(hop.public_key) for hop in circuit.hops] + \
+            ignore_candidates = [self.hops[h].node_public_key for h in circuit.hops] + \
```
devos50: Please name the `h` variable better.

qstokkink:
@devos50 Thanks for the review!

I have one rebuttal, for several of the comments on the names in the pooled_tunnel_community: the names in this class have been adopted to provide the same interface as the TunnelCommunity. If I were to change these particular method names, I would have to refactor the TunnelCommunity and the rest of Tribler.
The "do_circuits" LoopingCallback name is an exception, but this has also been adopted from the TunnelCommunity for clarity.

This also ties in to the TODO's for duplicate code: I was planning to give the TunnelCommunity and the PooledTunnelCommunity a shared interface/abstract parent class for structural clarity. However, I deemed this too complicated for this PR and left it as a TODO for the future.

Tribler/community/tunnel/pooled_tunnel_community.py
```diff
@@ -0,0 +1,345 @@
...
+    def init_circuit(self, circuit_id, circuit):
+        """Callback for when a synced circuit is localized
+
+        :param circuit_id: the circuit's id
+        :type circuit_id: long
+        :param circuit: the circuit
+        :type circuit: Tribler.community.tunnel.remotes.circuit.Circuit
+        :returns: None
+        """
+        setattr(circuit, 'proxy', self)
```
qstokkink: This is an initializer of an object created with `__new__()` (bypassing the `__init__()` to provide a snapshot of the remote object), the `.proxy` field does not exist (yet), but needs to be localized on the current process.

Tribler/community/tunnel/remotes/sync_dict.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.cls = cls
+        self.sync_interval = sync_interval
+        self.callback = callback
+        self.init_callback = init_callback
+
+        if task_manager and sync_interval > 0:
+            self.register_task(task_manager)
+
+    def register_task(self, task_manager):
```
qstokkink: I opted to attach this to the community class: a SyncDict is a field of a Community, so all events should be attached to and destroyed with the Community it belongs to.

Tribler/community/tunnel/processes/processmanager.py
```diff
@@ -0,0 +1,186 @@
...
+    def get_worker_count(self):
+        """Return the active amount of workers
+
+        :return: the current worker count
+        :rtype: int
+        """
+        return len(self.pool.keys())
+
+    def _create_worker(self, callback):
+        """Create a single worker, without adding it to the pool
```
qstokkink: Otherwise external classes might select a worker from the pool which has not properly been initialized yet.

Tribler/community/tunnel/processes/processmanager.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        if circuit_id in self.circuit_map:
+            worker = self.pool[self.circuit_map[circuit_id]]
+            worker.send_data([cd.sock_addr for cd in candidates],
+                             circuit_id,
+                             dest_address,
+                             source_address,
+                             data)
+        else:
+            logging.warning("Could not find worker registered for " +
```
qstokkink: It can happen that `create_circuit()` returns a `circuit_id` which is not yet known for any circuit shared by the _PooledTunnelCommunity.circuits_ `SyncDict`, even though it exists. This is a brute-force 'fix' for a situation that will resolve itself momentarily. LibTorrent does not respond very well to any fancy fixes for this (like delaying or dropping the data).

Tribler/community/tunnel/processes/subprocess.py
qstokkink: This is to make the write() method re-entrant for different file descriptors. I actually don't know how this weighs up to the overhead, I'll have to test it without the Thread construct.

qstokkink:
I just merged in the new GUI and it was kind of catastrophic.
A plethora of things is going wrong, including the subprocesses not exiting anymore.
For some reason downloading still works though:
newgui
At any rate, I probably wont have this done tomorrow.

devos50:
@qstokkink first try to get it working using the twistd plugin. You can run it with `twistd -n tribler`.

You can add anonymous downloads using the REST API (http://tribler.readthedocs.io/en/latest/restapi/downloads.html)

qstokkink:
@devos50 the twistd plugin (without my code) errors out for me on the first request:
```
~/tribler/tribler$ twistd -n tribler
Unhandled Error
  File "/usr/lib/python2.7/dist-packages/twisted/application/app.py", line 619, in parseOptions
    usage.Options.parseOptions(self, options)
  File "/usr/lib/python2.7/dist-packages/twisted/python/usage.py", line 262, in parseOptions
    for (cmd, short, parser, doc) in self.subCommands:
  File "/usr/lib/python2.7/dist-packages/twisted/application/app.py", line 636, in subCommands
    for plug in sorted(plugins, key=attrgetter('tapname')):
  File "/usr/lib/python2.7/dist-packages/twisted/plugin.py", line 209, in getPlugins
    allDropins = getCache(package)
  File "/usr/lib/python2.7/dist-packages/twisted/plugin.py", line 167, in getCache
    provider = pluginModule.load()
  File "/usr/lib/python2.7/dist-packages/twisted/python/modules.py", line 392, in load
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/usr/lib/python2.7/dist-packages/twisted/python/reflect.py", line 303, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/usr/lib/python2.7/dist-packages/twisted/python/reflect.py", line 250, in _importAndCheckStack
    reraise(excValue, excTraceback)
  File "/home/quinten/tribler/tribler/twisted/plugins/tribler_plugin.py", line 24, in <module>
    from Tribler.dispersy.utils import twistd_yappi
exceptions.ImportError: No module named utils
```
Also, my logging level seems to be set at WARNING, regardless of the logger.conf: ideas?

devos50:
@qstokkink you are probably using an older checkout of Dispersy, run `git submodule update`.

Regarding the logging level, I believe twistd is using a separate configuration for logging but I'm not sure about that.

qstokkink:
@devos50 Woops I had local changes in my dispersy, failing the submodule update 😊, good catch.
Also, the warnings only appear on the GUI.

qstokkink:
Unloading the community is a sore spot:
Using:

Produces:
```
```

devos50:
@qstokkink this method is being executed on a thread other than the reactor:

```
File "/home/quinten/tribler/tribler/Tribler/community/tunnel/pooled_tunnel_community.py", line 99, in unload_community
    yield super(PooledTunnelCommunity, self).unload_community()
```

You should make sure that you unload it on the reactor thread :)

devos50:
@qstokkink can you verify that `dispersy.stop()` returns a `Deferred`?

qstokkink:
@devos50 this is what dispersy.stop() returns:
```
<Deferred at 0x7fa3c80b5998 current result: <twisted.python.failure.Failure <type 'exceptions.AttributeError'>>>
```
Which has the not-so-descriptive error:
```
AttributeError: 'NoneType' object has no attribute 'addCallbacks'
```

qstokkink:
@devos50 Alright, the rest seems to have been stabilized. While I try to figure out the `unload_communities` stuff:
What do I need to change and where do I change it, to get the experimental setting in the new GUI?
I found `TriblerGUI/widgets/settingspage.py`, but I doubt that's all there is to it.

devos50:
@qstokkink I guess that the `@inlineCallbacks` decorator adds the callbacks to the deferred that is being yielded. It also seems that the stop method of Dispersy is returning a Failure, indicating that we didn't catch an error in Dispersy somewhere. Can you print the exception of the failure (http://twistedmatrix.com/documents/8.2.0/api/twisted.python.failure.Failure.html) so we can see where the exception is being thrown?

Regarding the GUI setting: you should edit `qt_resources/mainwindow.ui` with the visualiser and add a new tab to the settings page (named `experimental` or something). There you can add a checkbox and a label to toggle your settings. Next, in `settingspage.py`, you can manage the logic of this checkbox.

Let me know if you have any more questions!

qstokkink:
@devos50 Thanks for the GUI crash course!

As far as the unloading goes: I think I found the magic commands to get it all working correctly, so I wont pursue this any further. If you want (for some reason) to break it again I found adding `@inlineCallbacks` to `Tribler.dispersy.Community.unload_community()` produces the same behavior.

qstokkink:
Alright changelist since last review:
 - Review points have been handled (fixed/discussed) https://github.com/Tribler/tribler/pull/2607/commits/6da7b75fcf33357f023bd791af8e684482782a85
 - Merge has been performed and stabilized https://github.com/Tribler/tribler/pull/2607/commits/811ee204f0cb9d29129d35fd4bf317daf994de1f
 - Button has been added to the GUI https://github.com/Tribler/tribler/pull/2607/commits/68c321a9525759ac743a9f8673c463a5f044c607

Marking this as READY for (final?!) review.

devos50:
@qstokkink please also test your work on Windows 👍

qstokkink:
@devos50 Is there a way to have Jenkins build this for Windows? (Compiling is kind of kicking my butt right now)

devos50:
@qstokkink yes but that might make debugging a bit harder (since you cannot easily add debugging statements to the code and run it again). You can clone the windows build job and point it towards your branch in the Jenkins configuration.

qstokkink:
windowsfds
Well.. that would've been nice to include in the twisted docs.
Marking this as WIP again, until I fix this.

devos50:
Hmm that's annoying. Indeed, this should be added to the docs. The fact that you cannot pass custom file descriptors to child processes on Windows is pretty important.

Do you have an idea on how to fix this issue?

qstokkink:
@devos50 All communication will have to go through the std I/O, leaving stderr for forwarding debug info.

qstokkink:
@devos50 are the GUI tests supposed to pass if I changed the GUI?

devos50:
They should pass if you don't change API behaviour ;)

Now that you've added the experimental tab, would you mind to extend the GUI tests so a screenshot of that tab is taken too? Shouldn't be too hard to do since I already take screenshots of other tabs.

qstokkink:
Alright, the file descriptor std bypass has been implemented (and is working correctly on Windows now). The TODO/bug list is now as follows:

 - [ ] [Windows] Processes do not get force killed
 - [ ] [Windows] All peers are/seem to be unconnectable for a download
 - [ ] [Win/Unix] Take test screenshot

Also, the win64 builder is operational: https://jenkins.tribler.org/job/pers/job/build_tunnel_processes_win64_qstokkink/

qstokkink:
@devos50 The `GUI tests` check is testing Tribler/devel instead of the pull request GUI test code.

devos50:
@qstokkink I've updated the pointer of the GUI tests jobs to the latest PR, thanks for noticing!

qstokkink:
@devos50 am I correct in assuming [the FakeTriblerAPI](https://jenkins.tribler.org/job/Clone_tribler_fakeapi/ws/fakeapi/FakeTriblerAPI/) is not part of the source code? If so, the additional setting will have to be added to the FakeTriblerAPI later to have the GUI tests pass.

devos50:
@qstokkink I've created the fake API provider in order to not be dependent on a running Tribler session. I still have to clean this up but it is lower priority since the fake API is only used for GUI tests and to quickly generate screenshots. If you want, you can create a PR on the fake Tribler repo: https://github.com/devos50/fake-tribler-api. Otherwise, let me know what needs to be added to the API 👍

qstokkink:
Alright, the new screenshot is working:

Now I just need to figure out why the Socks5Server can't keep its Socks5Connection running on Windows and this PR is done.

devos50:
@qstokkink lol at that text, we need to discuss that :P 

I can give you access to one of the windows proxmox machines so you can easily run Tribler on Windows machines. If you come to my office next week, we can arrange that if you wish.

qstokkink:
@devos50 do we have a legal team or someone with any knowledge of what to write on the experimental tab?

Also, if you could make my life easier with a proxmox machine, that would be great.

qstokkink:
retest this please

devos50:
@qstokkink legal-related questions should be redirected to @synctext :)

Also, the mac builder is currently down and I'm unable to access it now so I will fix it monday morning.

devos50:
@qstokkink what's the progress with your Windows bug?

qstokkink:
@devos50 I can confirm it is a bug in my code: the normal TunnelCommunity does work on Windows (with the same common changes: using SyncDicts etc.).

As to why the bug is happening, I still don't know.

Even more annoyingly: I'm trying to get the normal TunnelCommunity Gumby experiment working and it is showing the same behavior on Linux (finds LibTorrent peers and then throws them all away). I am also half-hoping this is the same issue.

devos50:
Hmm that's weird. I would focus on the Gumby experiment; not supporting Windows (yet) for your PR code is not that bad and can be fixed later. Is the issue easily reproducible on your local machine running Gumby (with a very simple scenario)?

qstokkink:
@devos50 Yes it is 100% reproducible.

You can even check it out if you want: https://github.com/qstokkink/gumby/tree/tunnel_experiment : run the `hiddenservices.conf` locally (you may want to decrease the `experiment_time` to something more sensible first though). This has been modified to not have hidden seeding.

The experiment has no issue finding Dispersy _candidates_, but as soon as soon as LibTorrent _peers_ come in, they get dropped again (see the `Seed` progress output).

synctext:
'experimental' tab..
'enable multi-core Tor-like tunnels (beta)'.
Seems like a sufficient warning and deeply burried feature.

synctext:
btw 'multi-core anonymous downloading', that sounds better

qstokkink:
@devos50 I am running from source on Windows now, but I'm getting some weird dll errors. Have you ever seen this before?
pydllerror

devos50:
@qstokkink can you check whether the openssl directory is included in the system environment variable `PATH`?

qstokkink:
@devos50 I added `openssl` and the `openssl/dir` folders to the path, with a higher priority than `system32`, now it gives me the same error three times in a row for each of the dlls with the same name in these folders: none of the dlls seem to have ordinal 3906.

Did the openssl version, which Tribler uses, update?

devos50:
What if you backup and remove the openssl files from the system32 directory?

qstokkink:
@devos50 It fires the same error but for the `ssleay32.dll` in the `openssl` folder.

qstokkink:
@devos50 Ok, I found an inconsistency (I think this is unrelated to my work).
On Linux machines the LibTorrent proxy setting is being written (even though it is not specified in the settings):
settings_lin
On Windows machines it is not (which is intended behavior I guess):
settings_win

qstokkink:
Alright, I still don't know why Windows is dropping the incoming packets, but I did hit a new high-score during testing:
windows_speed

synctext:
6.3 Mbit per second.. Guess best to leave Windows for now and focus on Linux peak performance for thesis figures.

btw We can spend 1 or 2 Bitcoins to rent a few 32 or 64 cores machines if needed...

qstokkink:
Okay. Apparently this is some sort of GUI issue on Windows. Left it running for a while on `0%`, restarted Tribler and suddenly it was seeding. No wonder all of the messages seemed to be coming through correctly: they were coming through correctly.
suddenly_done

qstokkink:
Welp. I finally got it working on Windows.

Turns out Windows has special binary flags for the stdio streams.

Fix incoming in a little bit.

qstokkink:
@devos50 This is confirmed working (again) on Windows and Linux. What do you want to do with this PR? (rebasing this is going to be a little bit of a bitch, so I'll hold off on that until this gets greenlighted for merging)

synctext:
congrats on the Windows victory!

devos50:
@qstokkink good to hear! Why not rebase now? It has to be done anyways.

@synctext do we want to ship this as experimental feature in the upcoming Tribler 7 release?

synctext:
If this is stable and mature, then it's great stuff to merge. Quality over Tribler feature set..

qstokkink:
Alright, history has been rewritten, pylint errors have been minimized (I can't get rid of the remaining ones or everything breaks 😃 ): as far as I can see this is ready for review again.

qstokkink:
TWIMC in the future:
I am keeping this PR in a running state (branch checkout is always runnable).
Even though this code is functionally dormant, you may see the following updates here:

 1. Rebases to devel
 2. Additional unit tests
 3. Temporary fixes for Tribler/devel*

\* If/when this is merged, these will have to be removed before merging

qstokkink:
With the advent of IPv8 and separating the GUI from the core, 90% of this PR is no longer required. We can filter out the useful stuff later, if/when we need it.

PR #2612, opened 2016-11-08
Added torrent info endpoint

devos50:
Also added the ability to select specific files when starting a torrent download.


brussee:
@devos50 How does this affect #2612?

Issue #2676, opened 2016-12-05
Credit mining is incompatible with older versions of libtorrent


devos50:
I'm getting this error when running the credit mining tests on a computer with libtorrent 0.16.18:

```
test_logging (Tribler.Test.Core.CreditMining.test_creditmining.TestBoostingManagerUtilities) ... ERROR
ERROR

======================================================================
ERROR: test_logging (Tribler.Test.Core.CreditMining.test_creditmining.TestBoostingManagerUtilities)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/martijn/tribler/Tribler/Test/test_as_server.py", line 64, in check
    result = fun(*argv, **kwargs)
  File "/home/martijn/tribler/Tribler/Test/Core/CreditMining/test_creditmining.py", line 358, in test_logging
    boost_man.log_statistics()
  File "/home/martijn/tribler/Tribler/Policies/BoostingManager.py", line 458, in log_statistics
    if lt.__version__ == '1.0.9.0':
AttributeError: 'module' object has no attribute '__version__'
-------------------- >> begin captured logging << --------------------
BoostingManager: DEBUG: Status for 6161616161616161616161616161616161616161 : 0 0 | ul_lim : 12, max_ul 13, maxcon 14
--------------------- >> end captured logging << ---------------------
```

Note that `__version__` is not available in older versions of libtorrent.

ardhipoetra:
I tried to compile libtorrent  0.16.18, but it gives me error `undefined symbol: _ZTIN10libtorrent7torrentE` upon launching.

As it's only used for logging, I can wrap it in try-except Attribute error. If the version is unknown, we proceed with not logs anything. Does this approach acceptable?

devos50:
Older versions of libtorrent require `libtorrent.version` to get the version. We should create a method in `LibtorrentMgr` that returns the version string, independent of which version of libtorrent is being used.

I'm sure we have another instance of this somewhere in the code where I had to put a conditional check for the libtorrent version.

ardhipoetra:
I found the conditional code that depends on libtorrent version, but didn't found anything that checks the actual version. It checks the attribute instead. Examples : 
- https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentMgr.py#L302
- https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L742
[...]

PR #2677, opened 2016-12-06
Added function to check libtorrent version

ardhipoetra:
This PR will add a function to check libtorrent version. On old version of libtorrent, we can call this by `libtorrent.version`. However, starting from version RC_1_0, that will be deprecated and replaced by `libtorrent.__version__` (which more python-like).

Relevant : https://github.com/arvidn/libtorrent/blob/master/bindings/python/src/version.cpp

This should close issue #2676 if merged.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -590,6 +590,18 @@ class LibtorrentMgr(TaskManager):
 
         return result
 
+    def get_libtorrent_version(self):
+        """
+        This method returns the version of the used libtorrent
+        library and is required for compatibility purposes
+        """
+        if hasattr(lt, '__version__'):
+            return lt.__version__
+        else:
+            # libtorrent.version is deprecated starting from 1.0
+            return lt.version
+
+
 def encode_atp(atp):
     for k, v in atp.iteritems():
         if isinstance(v, unicode):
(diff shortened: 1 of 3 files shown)


brussee:
_"# piece_priorities will fail in libtorrent 1.0.9"_
Is this a known bug? Because I use this version on Android and some tests are failing.

ardhipoetra:
I found that bug long ago. Sometimes it works, but often I got error. I tried to run the same code with version 1.0.8 and 1.0.10, both of them just okay.

Never find any useful documentation for that error, though.

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -590,6 +590,17 @@ def start_download(self, torrentfilename=None, infohash=None, tdef=None, dconfig
 
         return result
 
+    def get_libtorrent_version(self):
+        """
+        A function to check libtorrent version
```
devos50: 'this method returns the version of the used libtorrent library and is required for compatibility purposes'

Tribler/Test/Core/CreditMining/mock_creditmining.py
```diff
@@ -147,6 +147,12 @@ def __init__(self):
 
         self.add_observer = lambda *_: None
 
+    def get_libtorrent_version(self):
+        """
+        A mock function to check libtorrent version
```
devos50: This comment is not really necessary I guess

ardhipoetra:
@devos50 fixed

devos50:
@ardhipoetra great, please squash and I will merge!

ardhipoetra:
@devos50, ready to merge

PR #2718, opened 2017-01-14
Remove tunnel always notifies with a valid Candidate

### LINKED ISSUE #2717: MULTICHAIN: Tunnel candidate not found
I have been seeing some TunnelCommunity entries in the 1h experiment log:
```
WARNING:TunnelLogger:MULTICHAIN: Tunnel candidate not found
```
This is only printed when previously established (Candidate + Member object) community members can no longer be found. Since the socket address is known, there is no need for this condition to occur (this is all the information needed to construct a Candidate object - see the recently implemented [copy_shallow_candidate()](https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/tunnel_community.py#L505)).

qstokkink:
Fixes #2717

### DIFF
--- a/Tribler/Test/Community/Tunnel/test_tunnelcommunity.py
+++ b/Tribler/Test/Community/Tunnel/test_tunnelcommunity.py
@@ -189,3 +189,94 @@ class TestTunnelCommunity(AbstractTestTunnelCommunity):
         self.assertTrue(self.tunnel_community.notifier.called)
         self.assertNotEqual(self.tunnel_community.notifier.candidate, None)
         self.assertNotEqual(self.tunnel_community.notifier.candidate.get_member(), None)
+
+    @blocking_call_on_reactor_thread
+    def test_reconstruct_candidate_on_tunnel_remove(self):
+        """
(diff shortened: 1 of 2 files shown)


qstokkink:
@devos50 Ready for review:

- Pylint warnings are due to a callback function (which is legal, but pylint doesn't detect it as a callback)
- Coverage decrease is due to moving uncovered code

devos50:
@qstokkink actually, I made it a habit to cover (most) uncovered code in my PRs, at least the code that's not hard to cover with a simple unit test. If we don't maintain a policy on this, the code coverage won't improve. So I would like that you cover these lines with a basic unit test before I review/merge.

Also, it helps to avoid the errors we got in the last PRs in the `copy_shallow_candidate` method.

qstokkink:
retest this please
The coverage and windows jobs crapped out

qstokkink:
@devos50 Fair point, I covered it with a test: ready for review again.

devos50:
@qstokkink thanks!

Issue #2732, opened 2017-01-17
Arch Linux dependencies are out of date


vandenheuvel:
#### Please, fill all relevant items: ####

- [x] I have read CONTRIBUTING.rst

- [x] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
devel, 19ecae714f037916c8ae8a05c962d7e2914331d2

##### Operating system and version: #####
Arch Linux, kernel version 4.8.13-1

The dependencies listed for Arch Linux [here](https://github.com/Tribler/tribler/blob/devel/doc/development/development_on_linux.rst#L67) are no longer up to date. It appears that dependencies python2-chardet, python2-pyqt5, qt5-svg and phonon-qt5-vlc and their dependencies were required for me to run Tribler with GUI. Also, dependencies python2-gmpy2, python2-pillow and wxpython2.8 seem no longer used.

vandenheuvel:
With the dependencies described above, running Tribler on Wayland with Sway works seamlessly. This includes the video player.

devos50:
Thanks!

vandenheuvel:
Updated this in pr https://github.com/Tribler/tribler/pull/2733.

vandenheuvel:
Second try in pr https://github.com/Tribler/tribler/pull/2736.

PR #2736, opened 2017-01-18
Update documentation on dependencies for Arch Linux

vandenheuvel:
Second attempt, this time for `devel`. I removed the python2-gmpy2, python2-pillow and wxpython2.8 packages and added python2-chardet, python2-pyqt5, qt5-svg and phonon-qt5-vlc.

This fixes https://github.com/Tribler/tribler/issues/2732.

### DIFF
--- a/doc/development/development_on_linux.rst
+++ b/doc/development/development_on_linux.rst
@@ -64,4 +64,4 @@ Execute the following command in your terminal:
 
 .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


devos50:
ok to test

devos50:
Hmm we should add some common words to the ignore list of our spelling checker. @qstokkink comments on this?

devos50:
And thanks for the contribution!

qstokkink:
@devos50 the spelling checker only fails on __new__ (non-dictionary) words introduced in a file in the PR, so once you verify the vernacular is correct, it will never go off on the same words again.
The only "flaw" is that these words are not checked cross-file: for this you could (somewhat easily) add a custom dictionary to aspell, if this really starts bothering you.

devos50:
@qstokkink thanks, that's good to know!

PR #2752, opened 2017-01-24
WIP: various fixes for Tribler 7

### LINKED ISSUE #2746: Stopping torrent during "Building circuits" will leave it on that state in the UI, not "Stopped".

Tribler_7.0.0-alpha2-34-g67f495c_x64.exe (from Jenkins)

Windows 10 Home,
Version 1607
OS-Build: 14393.693
 
Selected torrent will change to state "Stopped".

##### Actual behavior: #####
Torrent remains on "Building Circuits" even though every other torrent with same amount of hops (i had them set to 1) will work as normal. Restarting Tribler, the stopped torrents seem to remain in an active state, so if it would have seeded after establishing circuits, then it will do that when you start, same if it would download. I'm no aware if the torrent actually downloads when it's locked in "Building circuits". Either the state is not saved, or it's not changed when you do try to stop it. 

Side note, sometimes you get the option to stop said torrent when you right click, stopping, and then starting will make it work like normal again. Don't think you need to stop if first, haven't investigated enough. 

Open Tribler, go to downloads, and stop a torrent(that will be active once circuits are built) while in "Building circuits" state.

No log output.

### LINKED ISSUE #2749: File sorting based on numbers works incorrectly in the UI.

Tribler_7.0.0-alpha2-34-g67f495c_x64.exe (from Jenkins)

Windows 10 Home,
Version 1607
OS-Build: 14393.693

Say a torrent contains 5 files, and i got 2.5, 10, 25, 60, and 100 percent downloaded respectively. 
Sorting them by percentage, from high to low, should results in this pattern:

* 100 %
* 60 %
* 25 %
* 10 %
* 2.5 %

##### Actual behavior: #####
How it actually sorts:
* 60 %
* 25 %
* 2.5 %
* 100 %
* 10 %

The same applies to sorting by size. This happens when inspecting a selected torrent's files. Which means a 50 MB file will be over a 400 MB file. Or something downloaded 6% will appear above a 57% downloaded file for example.

(Note: Sorting torrents themselves by size seems to work, but i haven't the right sized ones atm. to test it out properly. But a torrent on 1.5GB will be above a 500 MB one. So that seems to work as expected.)

View a torrent with multiple files, and some of those have a size that the leftmost digit is higher than another files leftmost digit, while still being low in size. (Like 6 MB will appear above 570 MB) Also, 200 B will be above 180 MB... 

No log file.

### LINKED ISSUE #2750: Tribler crashing when receiving invalid packet in Dispersy
This is something we already observed before: Tribler crashes when we receive a packet that is invalid in Dispersy by throwing an `AssertionError`. However, an assertion is in my opinion not the way to handle the invalid packets since we cannot guarantee that packets we receive, are always correct. See the following stack trace for instance (IP address removed though):

```
File "TriblerGUI\event_request_manager.py", line 91, in on_read_data
RuntimeError: Unhandled Error
File "lib\multiprocessing\process.py", line 114, in run

File "TriblerGUI\core_manager.py", line 61, in start_tribler_core

File "lib\site-packages\twisted\internet\base.py", line 1195, in run

File "lib\site-packages\twisted\internet\base.py", line 1204, in mainLoop

File "lib\site-packages\twisted\internet\base.py", line 798, in runUntilCurrent

File "Tribler\dispersy\endpoint.py", line 252, in dispersythread_data_came_in

File "Tribler\dispersy\dispersy.py", line 1466, in on_incoming_packets

exceptions.AssertionError: [('<Candidate xxx.xxx.xxx.xxx:1897>', "'\\x00\\x1d'")]
```

I propose to make a method in `dispersy.py` that returns `True` when a packet is valid and passes all checks. If it does not pass the check, we do not process it.

@qstokkink do you agree with this?

### LINKED ISSUE #2751: "Running" multiple downloads before tribler starts causes problems.

(Sorry for vague title, was not sure how to avoid making the title a paragraph to explain well.)
Tribler_7.0.0-alpha2-34-g67f495c_x64.exe (from Jenkins)

Windows 10 Home,
Version 1607
OS-Build: 14393.693

Running two(or more) .torrent files rapidly, will open Tribler normally and let you select specifics for each download.

##### Actual behavior: #####
The opening and download selection works well, however, the issue is that when you press download, you get taken from the start screen to the downloads page, and this removes every download selection windows except the first one (that you pressed download on.) So, the download pop-up disappears, and won't appear again, even if you run open new torrents the download selection will not appear. 

Double click on multiple torrents, while Tribler is closed. Press download on that first torrent, and you'll be taken out of the download selection screen to the downloads list. Since i use a torrent download and run program, they will all start Tribler roughly at the same time, causing the issue. (Unless i have Tribler open already.) Only times this didn't happen was when Tribler didn't move to the downloads screen after pressing download on the first torrent. 

Workaround: Select any torrent file with triblers add torrent function, this will cause all attempted torrents to appear (i think, many of them at least. Didn't count) and make the function work as expected again. (Did not try magnet and directory adding.)

None.

devos50:
Added an endpoint to fetch all trackers of a specific torrent.

Fixes #2746, fixes #2749, fixes #2750, fixes #2751

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -1067,6 +1067,10 @@ class LibtorrentDownloadImpl(DownloadConfigInterface, TaskManager):
                 # stopped before, pstate_for_restart contains its resumedata.
                 # and that should be written into the checkpoint.
                 #
+                self.cancel_pending_task("check_create_wrapper")
+                if self.dlstate == DLSTATUS_CIRCUITS:
+                    self.dlstate = DLSTATUS_STOPPED
+
                 if self.pstate_for_restart is not None:
                     self._logger.debug(
                         "LibtorrentDownloadImpl: network_stop: Reusing previously saved engineresume data for checkpoint")
--- a/Tribler/Core/Modules/restapi/torrents_endpoint.py
+++ b/Tribler/Core/Modules/restapi/torrents_endpoint.py
@@ -96,6 +96,53 @@ class SpecificTorrentEndpoint(resource.Resource):
         self.infohash = infohash
 
         self.putChild("health", TorrentHealthEndpoint(self.session, self.infohash))
+        self.putChild("trackers", TorrentTrackersEndpoint(self.session, self.infohash))
+
+
+class TorrentTrackersEndpoint(resource.Resource):
+    """
+    This class is responsible for fetching all trackers of a specific torrent.
+    """
+
+    def __init__(self, session, infohash):
+        resource.Resource.__init__(self)
+        self.session = session
+        self.infohash = infohash
+        self.torrent_db = self.session.open_dbhandler(NTFY_TORRENTS)
+
+    def render_GET(self, request):
+        """
+        .. http:get:: /torrents/(string: torrent infohash)/tracker
+
+        Fetch all trackers of a specific torrent.
+
+            **Example request**:
+
+            .. sourcecode:: none
+
+                curl http://localhost:8085/torrents/97d2d8f5d37e56cfaeaae151d55f05b077074779/trackers
+
+            **Example response**:
+
+            .. sourcecode:: javascript
+
+                {
+                    "trackers": [
+                        "http://mytracker.com:80/announce",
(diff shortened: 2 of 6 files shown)


devos50:
retest this please

PR #2780, opened 2017-02-02
Twistd failed to reflect the tunnel helper plugin

Captain-Coder:
If there is no logger.conf in the working directory, then invoking twistd will produce an uncaught exception on the stderr. This is because the tunnel_helper unconditionally uses the logger.conf file. This made gumby spew several errors that have nothing to do with my experiments.


devos50:
Thanks!

PR #2788, opened 2017-02-12
READY: Allow UDP associate without circuits

### LINKED ISSUE #2787: SOCKS5 Server should never deny associate requests
Currently the SOCKS5 server denies libtorrent's proxy connection if the tunnel community does not have enough circuits, see https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/Socks5/server.py#L223 .

Libtorrent's response to this, is to not use the proxy and start a naked download.

Usually this is symptomatic of [the tunnel community not being registered correctly in tests](https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L265). However, this may also occur in the wild if circuits are dropped in between [can_create_engine_wrapper](https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L253) and [network_create_engine_wrapper](https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L288) calls (__bad!__).

qstokkink:
Fixes #2787, this switches the SOCKS5 behavior from old:

1. Deny connection if no circuits available
2. Fallback to naked downloading

To new:
1. Always allow connection
2. [Drop packets while no circuits are available](https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/Socks5/server.py#L64)

This also includes a base for future unittests for the Socks5Connection.

Pylint violation in unit test:
I had to use a protected member since [SocksUDPConnection does not yield on the listen_port.stopListening() deferred](https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/Socks5/server.py#L79) causing a dirty reactor.

### DIFF
--- a/Tribler/community/tunnel/Socks5/server.py
+++ b/Tribler/community/tunnel/Socks5/server.py
@@ -220,24 +220,18 @@ class Socks5Connection(Protocol):
         self._logger.error("DENYING SOCKS5 request, reason: %s" % reason)
 
     def on_udp_associate_request(self, connection, request):
-        if self.selection_strategy.has_options(self.hops):
-            # The DST.ADDR and DST.PORT fields contain the address and port that the client expects
-            # to use to send UDP datagrams on for the association.  The server MAY use this information
-            # to limit access to the association.
-            self._udp_socket = SocksUDPConnection(self, request.destination)
-            ip = self.transport.getHost().host
-            port = self._udp_socket.get_listen_port()
+        # The DST.ADDR and DST.PORT fields contain the address and port that the client expects
+        # to use to send UDP datagrams on for the association.  The server MAY use this information
+        # to limit access to the association.
+        self._udp_socket = SocksUDPConnection(self, request.destination)
+        ip = self.transport.getHost().host
+        port = self._udp_socket.get_listen_port()
 
-            self._logger.info("Accepting UDP ASSOCIATE request to %s:%d", ip, port)
+        self._logger.info("Accepting UDP ASSOCIATE request to %s:%d", ip, port)
 
-            response = conversion.encode_reply(
-                0x05, conversion.REP_SUCCEEDED, 0x00, conversion.ADDRESS_TYPE_IPV4, ip, port)
-            self.transport.write(response)
-
-        else:
-            reason = "not enough circuits"
-            self.deny_request(request, reason)
-            self.close(reason)
+        response = conversion.encode_reply(
+            0x05, conversion.REP_SUCCEEDED, 0x00, conversion.ADDRESS_TYPE_IPV4, ip, port)
+        self.transport.write(response)
 
     def select(self, destination):
         if destination not in self.destinations:
--- a/Tribler/community/tunnel/Socks5/server.py
+++ b/Tribler/community/tunnel/Socks5/server.py
@@ -76,8 +76,10 @@ class SocksUDPConnection(DatagramProtocol):
 
     def close(self):
         if self.listen_port:
-            self.listen_port.stopListening()
+            exit_value = self.listen_port.stopListening()
             self.listen_port = None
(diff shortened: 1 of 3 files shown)


devos50:
@qstokkink the Mac builder is offline apparently, I will fix it tomorrow 👍

qstokkink:
@devos50 ready for review (even though Mac GUI+unittests and coverage diff are broken), for pylint violation see OP.

Tribler/community/tunnel/Socks5/server.py
```diff
@@ -219,25 +219,19 @@ def deny_request(self, request, reason):
...
-
-            response = conversion.encode_reply(
-                0x05, conversion.REP_SUCCEEDED, 0x00, conversion.ADDRESS_TYPE_IPV4, ip, port)
-            self.transport.write(response)
-
-        else:
-            reason = "not enough circuits"
-            self.deny_request(request, reason)
-            self.close(reason)
+    def on_udp_associate_request(self, _, request):
```
devos50: For clarity, I would prefer to not use the `_` in non-test code. If I ever want to extend/modify this method, I have to find out what this parameter is.

devos50:
retest this please

Tribler/community/tunnel/Socks5/server.py
qstokkink: Alright, modifying and rebasing this now.

qstokkink:
@devos50 requested change fixed, ready for double check and merge. Mac still down.

egbertbouman:
I just noticed that force_proxy only exists in the Python binding since version 1.0.2. To protect our users as best as we can, maybe we should only support libtorrent >= 1.0.2 (right now we support [libtorrent >= 0.16.18](https://github.com/Tribler/tribler/blob/devel/debian/control#L31)).

qstokkink:
@egbertbouman by always accepting the SOCKS5 connection, it will __never__ go into naked fallback mode and the setting shouldn't even be needed. This should make it safe for older versions too.

devos50:
@qstokkink I guess we cannot easily make the `close` method asynchronous (so we can yield the `stopListening` call)?

devos50:
retest this please

qstokkink:
@devos50 that would be a little bit of work as you need to forward a deferred up two levels, but not difficult and also in line with the [way twisted ports are closed in the SOCKS5 server](https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/Socks5/server.py#L322).

egbertbouman:
@qstokkink OK, sound good.

devos50:
@qstokkink could you please twistify these calls then? This should hopefully prevent dirty reactor errors in the future and it's in line with our twistifaction of Tribler 👍

qstokkink:
Uh oh, internally Port.stopListening() schedules a deferLater call for its own deferred. However, to make sure the connection is closed, the test tearDown blocks on this deferred (blocking_call_on_reactor_thread). So it waits eternally until something finishes which has been scheduled to occur after waiting.

@devos50 Any ideas on how to work around this?

qstokkink:
Hmm.. @devos50 `git rebase --autosquash` removes you from the history. Is this a problem?

devos50:
@qstokkink I don't mind. Can you try to cover line 82 in that code coverage report? When that's done, I will merge.

qstokkink:
@devos50 covered now. Also added an additional test for the same behavior in the Socks5Connection class.

qstokkink:
retest this please

devos50:
retest this please

devos50:
@qstokkink thanks! :)

PR #2792, opened 2017-02-13
WIP: Fix hidden tunnel test

egbertbouman:


devos50:
ok to test

qstokkink:
`self.find_download(info_hash).add_peer(('1.1.1.1', 1024))`

egbertbouman:
@qstokkink Still WIP.

qstokkink:
@egbertbouman You'll need some other changes to the test too, here is the complete commit:
https://github.com/qstokkink/tribler/commit/f86f3e5757d1f26bd6528cd9d1b5cbc73022581a

Tribler/Test/Community/Tunnel/FullSession/test_tunnel_base.py
```diff
@@ -160,26 +161,31 @@ def setup_tunnel_seeder(self, hops):
...
 
-        if hops > 0:  # Safe seeding enabled
-            self.tunnel_community_seeder = self.load_tunnel_community_in_session(self.session2)
-
     def seeder_state_callback(self, ds):
         """
         The callback of the seeder download. For now, this only logs the state of the download that's seeder and is
         useful for debugging purposes.
         """
+        self.tunnel_community_seeder.monitor_downloads([ds])
```
qstokkink: Woops, this is also used by other tests, this needs to be:

```python
if self.tunnel_community_seeder:
    self.tunnel_community_seeder.monitor_downloads([ds])
```

qstokkink:
😃 
```
nosetests -x Tribler/Test/Community/Tunnel/
..................
Ran 18 tests in 106.187s

OK
```

devos50:
Nice :D

qstokkink:
@devos50 some minor cleanup required for the code still (some unused code according to the coverage report, a pylint violation and a duplicate port in use apparently?!), but this should be able to be merged very soon.

devos50:
Yes indeed but that are quick fixes hopefully. Glad it's finally working!

qstokkink:
@devos50 @egbertbouman's magic `.add_peer(('1.1.1.1', 1024))` fix was changed in the HiddenTunnelCommunity itself, this might have been what was holding back in-the-wild deployment.

qstokkink:
retest this please

qstokkink:
Oh wow there are two things going wrong at once here:
https://jenkins.tribler.org/job/GH_Tribler_PR_tests_linux/4150/testReport/junit/Tribler.Test.Community.Search.FullSession.test_search_community/TestSearchCommunity/test_torrent_search/

Neither of which are related to this PR.

devos50:
Hmm there seems to be an incoming packet that's malformed. I suppose it's from the outside world as indicated by the IP address in the packet. @qstokkink what about removing the `packet[1]` length check and just ignore this message (and log it)?

qstokkink:
@devos50 You may want to generalise the check [here in endpoint.py](https://github.com/Tribler/dispersy/blob/ddf0e4b119ff0bde30ae7a1238f4e1d62c5f344b/endpoint.py#L249) to filter on packet size as well. Then you can just keep the check in `dispersy.py` in place too.
As to what exactly is being sent, I cannot distinguish these packets from random garbage. I tried matching the packets to community cid's, conversion versions and such and found nothing. Only the dispersy version seems to match. Furthermore, I can tell you the outside IP did correctly puncture and respond to the missing id request. I did note one packet ending with "\n\r", possibly this is misdirected output?

On the second subject, I personally believe the test runner should never crash, whatever the test produces. But, since there are more important issues at hand, I begrudgingly accept your explanation.

egbertbouman:
retest this please

devos50:
This should actually have been merged into the `next` branch since we want to release this in Tribler 7.

PR #2809, opened 2017-02-18
READY: Various fixes for Tribler 7

### LINKED ISSUE #2790: The python-pip dependency could be removed
I've just (devel, 31841c768bbf4b100974ad48d2e2a9e8877bd5a1) ran Tribler successfully on Ubuntu 16.10 by using the `python-decorator` and `python-libnacl` packages. By doing this, `python-pip` doesn't need to be used anymore and all deps can be installed in a single command.

### LINKED ISSUE #2806: Tribler 7 alpha3's web api binds to WAN IP
Right now if you're connectable on port 8085, others can remote control your Tribler instance. IPs of potential targets can be easily found by crawling a recently added community. Not very anonymous, we should fix this asap.

### LINKED ISSUE #2807: API port used by the GUI should be configurable by the user
While finding https://github.com/Tribler/tribler/issues/2806, I noticed that the port for the web api seems to be hard-coded in 3 different places:
* core_manager.py
* tribler_request_manager.py
* tribler_window.py

Also, the port does not respect the setting in libtribler.conf

devos50:
Fixes #2807, fixes #2790, fixes #2806

### DIFF
--- a/Tribler/Core/Modules/restapi/rest_manager.py
+++ b/Tribler/Core/Modules/restapi/rest_manager.py
@@ -30,7 +30,7 @@ class RESTManager(TaskManager):
         self.root_endpoint = RootEndpoint(self.session)
         site = server.Site(resource=self.root_endpoint)
         site.requestFactory = RESTRequest
-        self.site = reactor.listenTCP(self.session.get_http_api_port(), site)
+        self.site = reactor.listenTCP(self.session.get_http_api_port(), site, interface="127.0.0.1")
 
     def stop(self):
         """
--- a/Tribler/Test/GUI/test_gui.py
+++ b/Tribler/Test/GUI/test_gui.py
@@ -11,16 +11,14 @@ from PyQt5.QtWidgets import QApplication, QListWidget, QTreeWidget
 
 from Tribler.Core.Utilities.network_utils import get_random_port
 import TriblerGUI.core_manager as core_manager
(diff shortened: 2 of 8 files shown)


Tribler/Core/Modules/restapi/rest_manager.py
```diff
@@ -30,7 +30,7 @@ def start(self):
         self.root_endpoint = RootEndpoint(self.session)
         site = server.Site(resource=self.root_endpoint)
         site.requestFactory = RESTRequest
-        self.site = reactor.listenTCP(self.session.get_http_api_port(), site)
+        self.site = reactor.listenTCP(self.session.get_http_api_port(), site, interface="127.0.0.1")
```
vandenheuvel: Could this be implemented in a more flexible way? A command line option (`twistd -n tribler --bind-address=0.0.0.0`) to bind to (a specific) WAN IP would be nice.

devos50: We could make it a configuration option if we need more flexibility (which we might need in the end since controlling/monitoring a remote Tribler instances is quite powerful). What about making it a config option which defaults to localhost? @egbertbouman do you agree with that?

egbertbouman: I think right now we should hold off on that, at least until the web api has some sort of security mechanism. Users may not understand that the web api is unprotected..

Issue #2765, opened 2017-01-29
Many KeyErrors reported in Tribler 7 alpha 3


devos50:
I'm receiving many, many reports about a `KeyError` in the (hidden) tunnel community. I think this issue should be prioritised.

Examples:

```
Traceback (most recent call last):
File "TriblerGUI\event_request_manager.py", line 91, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.KeyError'>: 'LibNaCLPK:\x9bg\x84\xfa\xb64;\xf8\x85\xf9\x1c\xfa\x8fb\x81\xc1P\t\xfeW\xde\xc3\x03\x1ct\xf9\xf55\x1e&\x16\x00Id\x9f\x01L\xcc\xc1\x11;\xc4\xcfa\xa2\x8fa>\x00%\xc4\r%x$\xd0)\xe5(i\xb42eR'
lib\site-packages\twisted\internet\base.py:1195:run
lib\site-packages\twisted\internet\base.py:1204:mainLoop
lib\site-packages\twisted\internet\base.py:825:runUntilCurrent
lib\site-packages\twisted\internet\task.py:239:__call__
--- <exception caught here> ---
lib\site-packages\twisted\internet\defer.py:149:maybeDeferred
Tribler\dispersy\util.py:37:helper
Tribler\community\tunnel\tunnel_community.py:441:do_circuits
Tribler\community\tunnel\tunnel_community.py:538:create_circuit
]
```

```
Traceback (most recent call last):
File "TriblerGUI\event_request_manager.py", line 91, in on_read_data
RuntimeError: Unhandled Error
Traceback (most recent call last):
File "lib\multiprocessing\process.py", line 114, in run

File "TriblerGUI\core_manager.py", line 61, in start_tribler_core

File "lib\site-packages\twisted\internet\base.py", line 1195, in run

File "lib\site-packages\twisted\internet\base.py", line 1204, in mainLoop

--- <exception caught here> ---
File "lib\site-packages\twisted\internet\base.py", line 798, in runUntilCurrent

File "Tribler\community\tunnel\hidden_community.py", line 250, in monitor_downloads

File "Tribler\community\tunnel\hidden_community.py", line 640, in create_introduction_point

File "Tribler\community\tunnel\tunnel_community.py", line 538, in create_circuit

[...]

PR #2811, opened 2017-02-20
Tunnel refactoring

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/2765, https://github.com/Tribler/tribler/issues/2824.


vandenheuvel:
retest this please

devos50:
This PR should go to devel ;)

qstokkink:
Why is the master member being changed? Are these changes no longer compatible with the old TunnelCommunity?

egbertbouman:
@qstokkink Don't worry, everthing is still compatible. I changed the master member because we had the private key of our master member in the comments :astonished:

qstokkink:
@egbertbouman ah. Yes, I would agreee that is a very valid reason to change it 😛 .

devos50:
@egbertbouman can you cover the missing line with a unit test and fix the pylint errors? :)

devos50:
retest this please

devos50:
@egbertbouman I think we should defer the changing of the community ID and do this just before we release Tribler 7 so I can change all the tunnel helpers at once.

synctext:
would prefer we dont do a single big bang change of our PetaByte helper servers. 1 or more now. Rest later.

devos50:
Ok then, I will change the tunnel helper restart script to account for our new community ID.

PR #2848, opened 2017-03-11
Consider torrents without ETA taking the longest

### LINKED ISSUE #1591: Consider torrents without ETA taking the longest
When sorting torrent by ETA, could you please change it so that completed torrents and torrents that don't download are sorted below torrents that have an estimated time. This way, it is an easy way to see what torrent will be complete the soonest, as it is on top of the list, which torrent will take a while, since it will be at the bottom of the still downloading torrents, and which torrent are complete, since they will be on the bottom of the list.

Griffon26:
This fixes #1591 by considering torrents with an ETA of 0 to have an
ETA of infinity.

It does not deteriorate pep8, pylint or the unit tests.


synctext:
ok to test

Issue #2844, opened 2017-03-05
Tribler should have a LICENSE file in the root directory of the project


vandenheuvel:
The current license file seems to apply only to Tribler, not to TriblerGUI. Moreover, it is good to follow [GitHub standards](https://developer.github.com/v3/licenses/#get-a-repositorys-license). And, as a bonus, [this icon](https://help.github.com/assets/images/help/repository/repo-license-indicator.png).

PR #2853, opened 2017-03-13
Update license files and move them into the root directory of the project

vandenheuvel:
Fixes https://github.com/Tribler/tribler/issues/2844

I moved the copy of the LGPL license from `/Tribler/LICENSE.txt` to `/LICENSE`. It is now clear that the license applies not only to `/Tribler/*` but also to the `/TriblerGUI/*` files. We now follow [GitHub conventions](https://developer.github.com/v3/licenses/#get-a-repositorys-license).

The `/LICENSE` file is the GPL license, the LGPL license is in the `/COPYING.LESSER` file as described on [choosealicense.com](https://choosealicense.com/licenses/lgpl-3.0/), a website by GitHub. This is the third version.

- From the `/Tribler/binary-LICENSE-postfix.txt` file I removed the section on wxPython as this software is no longer used in Tribler 7.
- Note that the `/LICENSE` and `/COPYING.LESSER` files no longer start with a [section](https://github.com/Tribler/tribler/blob/1d3ddb8aa3ba2f33b0d8ce81fbea6bda97007183/Tribler/LICENSE.txt#L3) on the history of Tribler and research grants by which Tribler development was supported. I think it is better not to include such a statement as it can get outdated as it is now.
- `pymdht` and `dispersy` also are no longer mentioned as they have their own licenses.
- There is now also no mention of [the files under the MIT license](https://github.com/Tribler/tribler/blob/1d3ddb8aa3ba2f33b0d8ce81fbea6bda97007183/Tribler/LICENSE.txt#L129) as the files mentioned either aren't in Tribler 7 or are changed greatly since their original creation. The latter only holds for `defaults.py`, all other files are no longer present.

In a second commit I removed mentions of the license from the top of all files as this wasn't done consistently and is unnecessary. It is however recommended by the Free Software Foundation, but I personally don't think the hassle is worth it. I used this opportunity to make the start of the files PEP 8 and PEP 257 compliant, that is removal of author names and leaving only a short description relevant to the module. I only left encoding declarations when necessary.


synctext:
good cleanup work!
please sync with Martijn and Egbert which part of the code most desperately needs some love.

Tribler/LICENSE.txt
```diff
@@ -1,679 +0,0 @@
...
-     Tribler/Core/Utilities/bencode.py
-     Tribler/Core/Utilities/convert.py
-     Tribler/Core/Utilities/bitfield.py
-     Tribler/Core/APIImplementation/maketorrent.py
-     Tribler/Core/DecentralizedTracking/ut_pex.py
-     Tribler/Core/RawServer/SocketHandler.py
-     Tribler/Core/RawServer/selectpoll.py
-     Tribler/Core/RawServer/RawServer.py
-     Tribler/Core/defaults.py
-
```
synctext: So it has been checked that all these files are no longer present. Tribler project has replaced them all, thus holds full copyright on these original works.

Tribler/LICENSE.txt
```diff
@@ -1,679 +0,0 @@
...
-       under contract no 231200.
-     - European Community's Seventh Framework Programme in the P2P-NEXT project
-       under contract no 216217 (see ns-LICENSE.txt for details).
-     - BSIK Freeband Communication I-Share project (Dutch Ministry of Economic 
-       Affairs)
-     - Netherlands Organisation for Scientific Research (NWO) grant 612.060.215
-     - Dutch Technology Foundation STW: Veni project DTC.7299
-     - European Community's Sixth Framework Programme in the P2P-FUSION project
-       under contract no 035249.
-
```
synctext: We used to list our grants. This is more of a pointer and courtesy. Can be removed.

Tribler/Test/Community/Bartercast/test_bartercast4_community.py
```diff
@@ -1,5 +1,5 @@
 # Written by Cor-Paul Bezemer
```
devos50: Please also remove this line.

synctext:
On second though, please do not remove all credit from this source file.

The license cleanup was a really good thing. But removing names is on 2nd though the only thing giving the people who worked hard on Tribler 5+ years ago the credit they deserve. @vandenheuvel What is the easiest/best way to fix this? 
https://github.com/Tribler/tribler/pull/2853/files#diff-ab7f68de5beec3adf85e22b2c6f6787eL3

devos50:
@vandenheuvel you can revert this commit and remove the irrelevant parts.

PR #2857, opened 2017-03-14
Remove irrelevant comment in Main.hacks

vandenheuvel:
The comment I removed is about a previous workaround which is already removed.


vandenheuvel:
@devos50 I still cannot request a review.

devos50:
@vandenheuvel Apparently, this is not possible, see https://help.github.com/articles/requesting-a-pull-request-review/

So you should just mention me when the PR is ready I guess.

PR #2865, opened 2017-03-16
TunnelSettings respect socks port settings in session

Captain-Coder:
During gumby refactoring I stubled across this situation. Apparently all tunnel settings needed to have the socks port set explicitly after setting it in the session. This PR makes the tunnel_community_socks5_listen_ports behave the same as the other tunnel_community settings.

### DIFF
--- a/Tribler/community/tunnel/tunnel_community.py
+++ b/Tribler/community/tunnel/tunnel_community.py
@@ -215,7 +215,6 @@ class TunnelSettings(object):
         self.tunnel_logger = logging.getLogger('TunnelLogger')
 
         self.crypto = TunnelCrypto()
-        self.socks_listen_ports = range(1080, 1085)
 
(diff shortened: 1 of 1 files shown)


Tribler/community/tunnel/tunnel_community.py
```diff
@@ -215,7 +215,6 @@ def __init__(self, tribler_session=None):
         self.tunnel_logger = logging.getLogger('TunnelLogger')
 
         self.crypto = TunnelCrypto()
-        self.socks_listen_ports = range(1080, 1085)
```
devos50: Please initialize this value (to something like `[]`).

devos50: Oh whoops, this code is all in the init method, didn't see that 👍

Issue #2860, opened 2017-03-15
get_uniformed_tracker_url: ValueError, invalid literal for int() with base 10: '80 udp'


synctext:
While running 7 .0 Beta on Ubuntu (part of the 26MByte log file).

```
Traceback (most recent call last):
  File "/usr/share/tribler/Tribler/Core/CacheDB/SqliteCacheDBHandler.py", line 345, in addExternalTorrentNoDef
    torrent_id = self._addTorrentToDB(torrentdef, extra_info)
  File "/usr/share/tribler/Tribler/Core/CacheDB/SqliteCacheDBHandler.py", line 431, in _addTorrentToDB
    self._addTorrentTracker(torrent_id, torrentdef, extra_info)
  File "/usr/share/tribler/Tribler/Core/CacheDB/SqliteCacheDBHandler.py", line 485, in _addTorrentTracker
    tracker_url = get_uniformed_tracker_url(announce)
  File "/usr/share/tribler/Tribler/Core/Utilities/tracker_utils.py", line 47, in get_uniformed_tracker_url
    if not url.port:
  File "/usr/lib/python2.7/urlparse.py", line 113, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: '80 udp'
```

vandenheuvel:
Please assign this to me.

PR #2869, opened 2017-03-17
Catch all exceptions of urlparse in tracker_utils

vandenheuvel:
Fixes https://github.com/Tribler/tribler/issues/2860.

I catch all exceptions in that come from parsing the given tracker url with `urlparse`. I believe violating pylint rules is justified.


vandenheuvel:
@devos50 Ready.

PR #2878, opened 2017-03-21
Remove demers community

vandenheuvel:
In this PR I remove the demers community as it is no longer being used.


devos50:
Thanks!

Issue #2842, opened 2017-03-04
Tribler 7 GUI freezes after a couple of days


pimlie:
##### Tribler version/branch+revision: #####
```
$ git log
commit 4d4ef144dc0d94110a350716cfe059d4c08da373
Merge: 1905399 038efde
Author: Martijn de Vos <mdmartijndevos@gmail.com>
Date:   Mon Feb 27 16:13:54 2017 +0100

    Merge pull request #2818 from vandenheuvel/fix_rewrite_tracker_utils
    
    Rewrite tracker_utils to use urlparse
$ git status
On branch devel
Your branch is up-to-date with 'origin/devel'.
```
(this is a pull from last week as it takes a couple of days for this issue to trigger)

##### Operating system and version: #####
Linux Mint 18.1 with Python 2.7.12 in a vm

##### Expected behavior: #####
GUI always stays responsive

##### Actual behavior: #####
After having Tribler 7 open for a couple of days, the GUI becomes unresponsive. The GUI is not updated anymore and (if the GUI was still open at first) when you switch windows the whole GUI window disappears / becomes transparent. The tray icon is still visible but also unresponsive. All other threads seem to be running still, eg when you forcibly close Tribler and restart it downloads have progressed and/or finished. This behaviour has occurred repeatedly the last couple of months.

##### Steps to reproduce the behavior: #####
Because it takes some time for this issue to trigger, it's difficult to test why/when this happens exactly. Not sure if just keeping the window open will reproduce it or that the vm is an issue here.

##### Relevant log file output: #####
Not sure how to debug this, no errors are showing in the STDERR output (but scrollback is limited) and no log file exists in ~/.Tribler or /var/log/. I start tribler as follows:
```
$ cd ~/bin/tribler
$ git pull
$ ./tribler.sh
```
Please let me know how I can help to debug this issue :)

mrspits4ever:
just a suggestion but maybe the bug will trigger agian if you change your date/time a few days forward/back?

qstokkink:
I have had an issue in the past (on Windows) where the GUI would lock up until input was received in the terminal window. It's a long shot, but you could try it.

pimlie:
[...]

PR #2883, opened 2017-03-26
WIP: Added heartbeat mechanism to check aliveness of the core

devos50:


devos50:
Didn't see comment of @synctext in #2842, I will postpone this PR until we need it for better debugging.

synctext:
Logging any missing heartbeat and slow response time of the core would be great debug material in future.

devos50:
Actually, I was thinking about creating a visual plot of request latency in the debug panel for this endpoint but a bit low priority for now.

synctext:
We want more deeper awareness then just responsiveness I fear. e.g. What is the core doing for this 42% CPU usage?

PR #2889, opened 2017-03-27
Move "offset_to_piece" to maketorrent

vandenheuvel:
I moved the `offset_to_piece` method to maketorrent as it is the only place where it is used.


devos50:
@vandenheuvel please rebase this PR 👍

devos50:
Thanks!

Issue #7, opened 2013-03-07
Systemtap: create a Python, SQLite and C++ probe (#6)


synctext:
**#6 SUBTICKET**
Firste step to enable creation of Tribler-specific probes
#6

PR #2890, opened 2017-03-28
Remove prctl import from LaunchManyCore

vandenheuvel:
I removed the import of `prctl` as it isn't used.


vandenheuvel:
@devos50 You can have a look at [this](https://github.com/seveas/python-prctl/blob/master/prctl.py). I believe some names are being set, but I still don't think this should be in production code. Is it a good idea to add a comment to the docs and remove this code?

devos50:
@vandenheuvel see https://github.com/Tribler/tribler/issues/7. Apparently, it was used to set thread names. I'm pretty sure that it can be removed. If we run into any problems, we revert this commit but I don't think we will.

synctext:
Please ensure threads have names. Quite essential for debugging.

synctext:
never mind. thnx for clearing the junk..

PR #2900, opened 2017-04-15
Adds option to use monochrome tray icon.

### LINKED ISSUE #2899: Monochrome icon for macOS
Hi guys, 

I would like to make a small contribution: create monochrome icon for Tribler for macOS: 

tribler-macos

Why? 

screen shot 2017-04-11 at 13 29 03

Would a PR be accepted? I guess `devel` branch?

hovancik:
fixes #2899


devos50:
Ok to test

Tribler/Core/defaults.py
```diff
@@ -251,6 +252,7 @@
 tribler_defaults['Tribler']['window_height'] = 670
 tribler_defaults['Tribler']['sash_position'] = -185
 tribler_defaults['Tribler']['family_filter'] = 1
+tribler_defaults['Tribler']['monochrome_tray_icon'] = False
```
devos50: Please don't add this setting to the core, it's a GUI setting.

Take a look at how we toggle the debug mode in the GUI; you should do the same with this setting.

TriblerGUI/tribler_window.py
```diff
@@ -137,8 +137,12 @@ def __init__(self):
 
         # Create the system tray icon
         if QSystemTrayIcon.isSystemTrayAvailable():
+            print(self.tribler_settings)
```
devos50: Don't forget to remove this print if your PR is almost done.

Tribler/Core/defaults.py
hovancik: Thanks for feedback. I made the changes. Not sure whether it's fine, but changes on icon are made after Tribler restart.

devos50: Looks good! One more thing: maybe you could refactor the code that sets the tray icon in a separate method (named like `update_tray_icon`) and call it when the newly added checkbox state changes. This way, the update of the tray icon happens immediately and no restart is needed. Can you implement this? 👍 

When done, I will merge your PR.

hovancik: I did the change. Maybe I should be checking inside the function whether tray exists? 

Some test is failing because of timeout. Not sure whether it's my false or the test can be restarted.

devos50:
retest this please

PR #2912, opened 2017-04-25
Fixed QIODevice error

### LINKED ISSUE #2902: "Error: no Json object could be decoded from data:" When checking health on torrents (searching)

Tribler_7.0.0-beta-53-g4803abb_x64.exe	

Windows 10, creators update

Using Tribler search feature, you can search and then click on torrents to get a health state.  

##### Actual behavior: #####
It does seem to work as intended, but console is riddled with, 
"QIODevice::read: device not open"
"ERROR:root:No json object could be decoded from data:"

Search for torrent, click on any of them (might not happen on non-standard torrents that appear in the first few lines that are auto-checked.) Should happen with some of them. This could potentially lead to search results not working/appearing. Had that happened, but i can't really be sure if they are related or not. 

Also used a VPN at the time, if that's related. (I notice Trilber uses a lot of extra time to start up with a VPN active. )

```
QIODevice::read: device not open
ERROR:root:No json object could be decoded from data:
```

devos50:
Fixes #2902

### DIFF
--- a/TriblerGUI/core_manager.py
+++ b/TriblerGUI/core_manager.py
@@ -88,6 +88,8 @@ class CoreManager(QObject):
         self.stop_timer = QTimer()
         self.stop_timer.timeout.connect(self.check_stopped)
 
+        self.check_state_timer = QTimer()
+
     def check_stopped(self):
         if not self.core_process.is_alive():
             self.stop_timer.stop()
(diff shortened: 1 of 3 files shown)


qstokkink:
Should [this](https://github.com/Tribler/tribler/blob/c591926b3fddd3a5cc975bc2db1b1534f46467e4/TriblerGUI/event_request_manager.py#L49) be set as `setSingleShot(True)`? Right now the MAC GUI test produces way more than 40 "INFO: Will connect to events endpoint" messages.

qstokkink:
All of the debug tracebacks of `connect()` flow through `core_manager.on_received_state():` with a state of `STARTED`: https://github.com/devos50/tribler/blob/071b39ef24492a2cc4f05b33a21133e5b6579b00/TriblerGUI/core_manager.py#L130

Why this happens is still unknown.

PR #2921, opened 2017-05-01
[WIP] Python 2.6 Backports

hbiyik:
##  Reason for this PR:
This pr contains backport for python 2.6. Python 2.6 is widely used in Kodi / XBMC android ports and android devices is more than half of Kodi users. Python 2.7 is only supported in Kodi 17 Krypton on android , rest of android builds is shipped with embedded py2.6

## Changes in this PR:

**1) Dict comprehensions[DONE]**
Dict Comprehensions are removed. Dict comprehensions came up with python 2.7

**2) Set literals: [DONE]**
Set Literals Removed, this functionality came up with py27

**3) With statements: [DONE]**
With statements can not have multiple contexes in py26, modifed multiple contexted with statements to nested with statements.

**4) os.kill method: [DONE]**
os.kill method is not supported on windows python2.6. os.kill methods are ported to Tribler.Compat.oskill method. Py26 support is added with ctypes

**5) Ordered dict & defaultdict (collections module): [DONE]**
These 2 classes are absent in py26 collections, all imports are ported to Tribler.Compat.collection module. In python2.6 backport_collections module is need for interfacing to work. This creates a new dependency.
https://pypi.python.org/pypi/backport_collections/0.1

This issue also impacts dispersy, slightly simalar but less complicated is done at my fork at
https://github.com/hbiyik/dispersy/commit/2969de4532dde150407d49636e5d1d272e778b04

**6) Argparse module: [DONE]**
Argparse does not exist in py26, however there is a backport with same module name, there is no need for an interface under Tribler.Compat, only creates another dependency.
https://pypi.python.org/pypi/argparse/1.4.0

**7) Tempfile module: [Interfacing DONE, porting WIP]**
There actually is nothing wrong with tempfile in py26, however tempfile under android kodi py26 & 27 buth broken. For this reason all tempfile import are interfaced to Tribler.Compat module.

**8) Import Order Fixes. [DONE]**
Python interpreter under kodi/xbmc is not case sensitive for android. This causes Libtorrent, and configparser modules to be imported instead of libtorrent, Configparser from site-packages. Tribler.Core.Libtorrent renamed to Tribler.Core.LibtorrentInterface, Tribler.Core.Utilities.configparser ranmed to Tribler.Core.Utilities.cfgparser. All the imports in the code are fixed including tests.
This issue also impacts dispersy, see my modification for libnacl in 
https://github.com/hbiyik/dispersy/commit/03a7dfb510ba1d28763c84dd6984993798147f43

**9) Unit testing [py27 DONE, py26 DONE needs testing]**
A test + unittest script is implemented under Tribler/Test/Python. This script checks all .py files recursively under /Tribler folder which includes dispersy and pymdht. Script is a modifed and improved version pyqver
https://github.com/ghewgill/pyqver
It detects grammar problems very good, so far all module + method problems has been adressed by it. If in future some other problem is detected, script can be easyly improved.

Also I have passed all Tribler tests under py27 except credit mining, i assumed that creditmining is heavily under construction. Also dispersy tests all pass. Gonna do testing with py26 under android and windows next.

## Questions about the PR:
**1) Compat module:**
I thought it is the best practise to keep the code clean from tricky backporting problems and generate an interface, however if you have any suggestions i can change it easly. I could not decide which place is the best for it so added in the root folder, may be it can be moved to Trbiler.Core.Modules but it is more than a core module.

**2) New names**
open for suggestions for new names of Libtorrent & configparser modules, or any other approach.

**3) Dispersy changes**
Ordereddict problem and import order problem impacts dispersy as well, i have done work for it but according to comments on this PR i will create another PR for dispersy.

**4) Unit tests**
Is it normal to fail for unittest of Credit mining, is there anything i should know beforehand.

I have passed all unittests for this patch should be ok, but a review is also much appreciated.

Thanks

### DIFF
--- a/doc/development/development_on_windows.rst
+++ b/doc/development/development_on_windows.rst
@@ -3,7 +3,7 @@ This section contains information about setting up a Tribler development environ
 Introduction
 ------------
 
-In this guide, all required dependencies of Tribler will be explained. It presents how to install these dependencies. Some dependencies have to be built from source whereas other dependencies can be installed using a .msi installer. The guide targets Windows 10, 64-bit systems, however, it is probably not very hard to install 32-bit packages.
+In this guide, all required dependencies of Tribler will be explained. It presents how to install these dependencies. Some dependencies have to be built from source whereas other dependencies can be installed using a .msi or .exe installer. The guide targets Windows 7 or higher, 64-bit systems, however, it is probably not very hard to install 32-bit packages.
 
 First, Python 2.7 should be installed. If you already have a Python version installed, please check whether this version is 64 bit before proceeding.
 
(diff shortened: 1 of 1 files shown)


hbiyik:
@synctext @EinNarr 
For notification purposes just in case, here is the PR almost finished for Py26 backport.

qstokkink:
ok to test

hbiyik:
Cosmetics and lint checks are failing because did not do a lint check. Sorry my bad.

However, builds are failing because libnacl is no longer a sub module, a package dependency, the reason for that is

1) all binary dependant packages are also modified in pythonically to achieve a successful cross platform binary add on loading under xbmc / Kodi through script.module.abi

2) libnacl sub module is being loaded directly in disperse, however under Kodi the modified libnacl must be imported from site packages, but it is overridden by submodule.

3) as far as I am aware libnacl is not under development realm of tribler team, so there actually is no need for it to be a submodule, instead should be a package dependency like m2crypto on netifaces etc

To fix build issues I am afraid build scripts inside Jenkins server must be modified to do a libnacl clone from GitHub first. Sorry can't think of any alternative way of it.

synctext:
getting a fresh libnacl from github release url sounds like a good addition.

synctext:
ok to test

devos50:
The PR got closed so no need to retest ;)

PR #2927, opened 2017-05-05
Remove old architecture documentation and structure diagram

vandenheuvel:
I removed old documentation on architecture design and the accompanying structure diagram as it is incomplete, out of date and incorrect.


qstokkink:
Could you merge #2926 into this PR's commit: the other PR also removes the same document (albeit in German).
Furthermore, please change the commit message to `documentation` instead of `documentaiton`.

qstokkink:
@devos50 Do you have any sentimental attachment to the old architecture docs? (If not, you get the honor of pressing the merge button)

devos50:
@vandenheuvel thanks! I will fix the Mac GUI tests on Monday. And indeed, please create a new issue that we need some way to add an architecture diagram to the documentation (preferably auto-generated).

@qstokkink I don't have any sentimental feelings for outdated documentation :)

PR #2930, opened 2017-05-07
Small addition to Arch dependencies

sum01:
I found that Tribler requires `python2-matplotlib` on Arch when I was trying to install without it. I ran into no other dependency issues, and the list seems complete barring that.


qstokkink:
ok to test

devos50:
Thanks!

Issue #2872, opened 2017-03-18
Sign Tribler.app final bundle and Windows .exe


devos50:
Before releasing Tribler 7, we should check the signing process of the .exe on Windows and the .app on macOS.

devos50:
The .exe on Windows is now signed:

![0a77ae6696599084947e5ec454ee42ae](https://cloud.githubusercontent.com/assets/1707075/26752659/63c24f0e-4855-11e7-84b6-78132c825cf4.png)

qstokkink:
Signing of the Mac .app will be deferred to the 7.1 release.

devos50:
We successfully signed `Tribler.app` with my developer credentials, using the `codesign` tool. Since I do not want to use my own private key to do this, we should check whether the TU Delft has an Apple Developer account, or create one for the Tribler team.

devos50:
I stil haven't heard anything back from our ICT department. I will move this issue to backlog since I do not expect that we will get a valid signing certificate anytime soon.

synctext:
just to document our journey somewhat.
From 11 May 2015 until 2018 we used DigiCert. Digicert certificate for Build signing.

We could use the Surfnet facilities, they offer support.
https://dracc.commonsconservancy.org/

qstokkink:
Thanks to @xoriole the Apple release is now signed. The only remaining issue is now Windows, for which there is a separate issue here: #5648

PR #2968, opened 2017-06-03
Fixed Windows signing

devos50:
Related to #2872


devos50:
The Mac GUI tests are failing again, I need physical access to fix this (will do next week). Fortunately, it isn't a blocker for merging.

PR #2971, opened 2017-06-03
Remove unused dependency from Ubuntu documentation

vandenheuvel:
The `python-crypto` is not necessary for Ubuntu. I tested this on 17.04. Moreover, for the Arch dependencies, the package is not listed either.


vandenheuvel:
@devos50

devos50:
I've searched through the code and this package is indeed not required to run Tribler.

PR #2982, opened 2017-06-11
Generic TrustChain transaction blocks

### LINKED ISSUE #2980: Generic transaction blocks for the MultiChain community
I would like to propose generic transaction blocks for the MultiChain community. I need this community for the market to store transactions between traders and the current format of the Multichain is not suitable for this purpose. This means I have to copy the whole community code and make my modifications, leading to a cluttered PR and unnecessary code in the repository. I'm positive that this will happen more often in the future when people use the community.

I propose to slightly modify the structure of multichain blocks to allow generic `tx` blocks. First, I think we should get rid of the hard-coded public key/signature/struct lengths (see the first few lines of `block.py`) and use the `encode()` and `decode()` functions so we can have fields of dynamic lengths. Next, I think we should combine the `up`, `down`, `total_up` and `total_down` fields into a generic string type. This requires some modifications to the database and breaks backwards compatibility (which is not a huge deal since we recently merged true halves, breaking backwards compatibility anyway). Note that these changes also require us to split the validator.

The final design allows developers to create a subclass of `MultiChainCommunity` and design their own transactions using method overriding. We should think a bit further on how to design this in particular. For the Tribler MultiChain (used to account up- and downloads) we can make a `TriblerMultiChainCommunity` subclass.

@qstokkink could you give some comments/pointers on my proposal? I would like to work on this during the weekend (unless you want to do it :) since it has quite some impact on the outstanding market PR. Thanks!

### LINKED ISSUE #2981: The MultiChain community performs a naked reactor call
See: https://github.com/Tribler/tribler/blob/devel/Tribler/community/multichain/community.py#L329

This should be hooked up to a TaskManager task, like for example:
https://github.com/qstokkink/tribler/blob/ce0157f57d9c7de1e903724d566d1d98907dff0c/Tribler/community/multichain/community.py#L345

Otherwise, the task will survive the community `unload()` and keep running for an unloaded community. This can cause dirty reactors for tests etc.

devos50:
This PR implements generic transaction blocks in the TrustChain. Fixes #2980, fixes #2981.

Work done in this PR:
-renamed MultiChain to TrustChain
-refactored the TrustChain implementation to remove the hard-coded `up`, `down`, `total_up` and `total_down` fields. Instead, we have a single `transaction` field now.
-split the TrustChain in two parts: the `TrustChainCommunity` that is responsible for the generic message passing/persistence of blocks and the `TriblerChainCommunity`. For now, I've called it `TriblerChain` but we should discuss the name I guess. It's now very easy for developers to make a subclass of the `TrustChainCommunity` and use this community for their own purposes. Maybe we should extend the documentation with a small guide on how to do this. Also the tests are more separated now (tests for the generic data structure and tests for Tribler-specific constructs).
-created a separated method `validate_transaction` that validates a transaction. This is in line with the existing `validate` method in a block object.

I planned to keep this PR small but it turned out that implementing generic transaction blocks was harder than I thought and required more code changes than I expected :(.


devos50:
@qstokkink I'm not done with this PR yet (needs pylint fixes and more coverage improvements) but could you give me a review already so I can take this into account as early as possible?

Tribler/community/trustchain/block.py
```diff
@@ -13,21 +14,20 @@
...
-block_pack_format = "! Q Q Q Q {0}s I {0}s I {1}s {2}s".format(PK_LENGTH, HASH_LENGTH, SIG_LENGTH)
+block_pack_format = "! {0}s I {0}s I {1}s {2}s".format(PK_LENGTH, HASH_LENGTH, SIG_LENGTH)
 block_pack_size = calcsize(block_pack_format)
 
 
-class MultiChainBlock(object):
+class TrustChainBlock(object):
     """
-    Container for MultiChain block information
+    Container for TrustChain block information
```
qstokkink: Please add some documentation on the usage of this class. In particular I would like to know how to add my own transaction data.

Tribler/community/triblerchain/block.py
```diff
@@ -0,0 +1,117 @@
+from Tribler.community.trustchain.block import TrustChainBlock, ValidationResult, GENESIS_SEQ, GENESIS_HASH, EMPTY_SIG
+
+
+class TriblerChainBlock(TrustChainBlock):
+    """
+    Container for TriblerChain block information
```
qstokkink: Please add documentation on the fields supported by a `transaction` dictionary.

Tribler/community/trustchain/block.py
```diff
@@ -277,67 +248,69 @@ def sign(self, key):
...
-        pack_into(block_pack_format, buff, offset, self.up, self.down, self.total_up, self.total_down, self.public_key,
-                  self.sequence_number, self.link_public_key, self.link_sequence_number, self.previous_hash,
-                  self.signature if signature else EMPTY_SIG)
-        return str(buff)
+        encoded_tx = encode(self.transaction)
+
+        buff = bytearray(block_pack_size)
+        pack_into(block_pack_format, buff, 0, self.public_key, self.sequence_number, self.link_public_key,
+                  self.link_sequence_number, self.previous_hash, self.signature if signature else EMPTY_SIG)
+        return str(encoded_tx) + str(buff)
```
qstokkink: For validation purposes please structure this as `str(buff) + struct.pack("I", str(encoded_tx)) + str(encoded_tx)` (or something like it).

Tribler/community/trustchain/block.py
devos50: I've written some basic documentation on how to use the TrustChain :)

Tribler/community/triblerchain/block.py
devos50: See my comment above.

Tribler/community/trustchain/block.py
```diff
@@ -277,67 +248,69 @@ def sign(self, key):
...
             ret.link_sequence_number = link.sequence_number
         if blk:
-            ret.total_up = blk.total_up + ret.up
-            ret.total_down = blk.total_down + ret.down
             ret.sequence_number = blk.sequence_number + 1
             ret.previous_hash = blk.hash
-        else:
-            ret.total_up = ret.up
-            ret.total_down = ret.down
+        if link is None:
```
devos50: Good find, thanks

Tribler/community/triblerchain/database.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        assert isinstance(database_version, unicode)
+        assert database_version.isdigit()
+        assert int(database_version) >= 0
+        database_version = int(database_version)
+
+        if database_version < self.LATEST_DB_VERSION:
+            # Remove all previous data, since we have only been testing so far, and previous blocks might not be
+            # reliable. In the future, we should implement an actual upgrade procedure
+            self.executescript(self.get_upgrade_script(current_version=2))
```
devos50: I've made the upgrade script slightly more fancy 👍

PR #2995, opened 2017-06-14
Fix decoding of time data from HEX

### LINKED ISSUE #2992: Trust Page requiring wrong dataformat

- [X] I have read CONTRIBUTING.rst

- [X] I have tried with the latest pre-release version and I still can reproduce the issue.

devel b5ac1be98340c1f2df85dd2bba188cf0e3f1032e
Ubuntu 17.04
When I open the TrustPage after I have collected some blocks, it shows me a graph with the correct information according to the blocks.
##### Actual behavior: #####
When I open the TrustPage, Tribler crashes with `ValueError: time data '323031372d30362d31332031303a34303a3030' does not match format '%Y-%m-%d %H:%M:%S'`.

1. Get a Block inside the database.
2. Open the Trust Page.

`ERROR:root:Traceback (most recent call last):
  File "/home/max/git/tribler/TriblerGUI/widgets/trustpagegraph.py", line 109, in received_multichain_blocks
    self.plot_absolute_values()
  File "/home/max/git/tribler/TriblerGUI/widgets/trustpagegraph.py", line 119, in plot_absolute_values
    plot_data[1].append(datetime.datetime.strptime(block["insert_time"], "%Y-%m-%d %H:%M:%S"))
  File "/usr/lib/python2.7/_strptime.py", line 332, in _strptime
    (data_string, format))
ValueError: time data '323031372d30362d31332031303a34303a3030' does not match format '%Y-%m-%d %H:%M:%S'`

Please note that this most probably has something to do with the timestamp of SQLite3, for which DEFAULT CURRENT_TIMESTAMP is used. As far as I have read, this is [dependent on the sqlite3_step()](https://sqlite.org/lang_datefunc.html) call, which in it's turn is [VSF dependent](https://sqlite.org/c3ref/step.html)

MaxVanDeursen:
Fixes #2992

### DIFF
--- a/Tribler/community/trustchain/block.py
+++ b/Tribler/community/trustchain/block.py
@@ -326,7 +326,7 @@ class TrustChainBlock(object):
         for key, value in self.__dict__.iteritems():
             if key == 'key':
                 continue
-            if isinstance(value, basestring):
+            if isinstance(value, basestring) and key != "insert_time":
                 yield key, value.encode("hex")
             else:
                 yield key, value


TriblerGUI/widgets/trustpage.py
```diff
@@ -116,7 +116,11 @@ def plot_absolute_values(self):
 
         # Convert all dates to a datetime object
         for block in self.blocks:
-            plot_data[1].append(datetime.datetime.strptime(block["insert_time"], "%Y-%m-%d %H:%M:%S"))
+            try:
+                date = datetime.datetime.strptime(block["insert_time"], "%Y-%m-%d %H:%M:%S")
+            except ValueError:
+                date = datetime.datetime.strptime(block["insert_time"].decode("HEX"), "%Y-%m-%d %H:%M:%S")
```
devos50: This is not something that should be fixed in the GUI: it seems that the core passes incorrect data. After a quick investigation, it seems to be caused by the `__iter__` method in a block (https://github.com/Tribler/tribler/blob/devel/Tribler/community/trustchain/block.py#L326). Simply ignoring the `insert_time` field here should work.

Issue #3008, opened 2017-06-21
IOError on launch. Tribler RC2-84


Thomasedv:
- [x] I have read CONTRIBUTING.rst

- [x] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
Tribler rc2-84 from Jenkins, Windows 64 bit version, build 404. 

##### Operating system and version: #####
Windows 10

##### Expected behavior: #####
Installation of Tribler and successful startup. 

##### Actual behavior: #####
Crash on startup, GUI stuck on loading gears. 

##### Steps to reproduce the behavior: #####
Install build, (previous build was rc2-13 ) Not a fresh install. 

##### Relevant log file output: #####
![image](https://user-images.githubusercontent.com/1584461/27407421-7b117478-56d8-11e7-84e5-9abcb0842fb6.png)

That file location to the folder Tribler is the install folder. However! Tribler is looking for a folder named Tribler, in my install directory, and there is none.

Thomasedv:
Copied the folder named Tribler from the tribler_source folder, and got it to start up. However met with lots of extra error warnings in the GUIsee image below. Though not exactly sure if it should be considered an issue as this might be because of my attempted workaround. 

![image](https://user-images.githubusercontent.com/1584461/27407998-f996c544-56da-11e7-8143-6bd06e2e9e2c.png)

qstokkink:
Thanks for the report, we will be stabilizing the `devel` (7.1) branch after the release of 7.0 (`next`).

devos50:
@vandenheuvel @MaxVanDeursen do you have time to investigate and fix this issue?

devos50:
Partially fixed by #3009

vandenheuvel:
Probably, [this](https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Config/tribler_config.py#L20) line isn't doing what I would expect it to on that platform. Besides that, I currently have no time to investigate further.

PR #3009, opened 2017-06-22
Set safeseeding value in watch folder with result of function

vandenheuvel:
This (partially) fixes https://github.com/Tribler/tribler/issues/3008.


vandenheuvel:
retest this please

devos50:
It seems that GUI tests on win64 are failing since the deployment tests failed because of issue #3008. This one should be fixed as soon as possible.

PR #3083, opened 2017-09-07
WIP: Fix trust chain transaction sql usability

Captain-Coder:
The current implementation of trust chain packs the transaction details in a blob in the sql database. For triblerchain this is a problem when constructing the subjective work graph (or asking anything aggregating over many transactions really). It is much more efficient to be able to do this in SQL. So this PR changes the way trust chain handles the transaction fields such that they can be used in SQL.


Captain-Coder:
Since we are going for a new tribler chain structure, this will require a rewrite.

PR #2976, opened 2017-06-08
Changes to LibTorrentMgr to enable gumby DHT isolation

Captain-Coder:
Companion to [Gumby PR #330](https://github.com/Tribler/gumby/pull/330)
This is mainly an edit of LibTorrentMgr's session creation code.

PR #3084, opened 2017-09-07
FUTURE: Be ready for current Libtorrent versions

Captain-Coder:
Because some linux distros are inexplicably stuck on very old libtorrent-rasterbar versions (0.16) instead of 1.0 (or even 1.1.4), I've had to do some hacking to get Tribler-core working on Arch linux on a raspberry pi.

So when the new versions (1.1.1) of libtorrent finally ship with Debian Strech and Ubuntu Zesty, this PR will come in handy.

Builds on the work in #2976. So merge that first.

One thing to note, libtorrent no longer does the country lookup for IP, as they deemd it out of scope for a torrent implementation (seems resonable). So I've disabled this, but not fixed this in the GUI, since my tribler-core on pi doesn't run a gui.


devos50:
Closed in favour of #3376

PR #3088, opened 2017-09-11
Added ability to perform a memory dump to the debug pane

devos50:
This helps us to debug memory issues faster.

### DIFF
--- a/Tribler/Core/Modules/restapi/debug_endpoint.py
+++ b/Tribler/Core/Modules/restapi/debug_endpoint.py
@@ -16,7 +16,7 @@ class DebugEndpoint(resource.Resource):
 
         child_handler_dict = {"circuits": DebugCircuitsEndpoint, "open_files": DebugOpenFilesEndpoint,
                               "open_sockets": DebugOpenSocketsEndpoint, "threads": DebugThreadsEndpoint,
-                              "cpu_history": DebugCPUHistoryEndpoint, "memory_history": DebugMemoryHistoryEndpoint}
+                              "cpu": DebugCPUEndpoint, "memory": DebugMemoryEndpoint}
 
         for path, child_cls in child_handler_dict.iteritems():
             self.putChild(path, child_cls(session))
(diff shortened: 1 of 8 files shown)


devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

qstokkink:
retest this please

devos50:
retest this please

qstokkink:
retest this please

devos50:
retest this please

qstokkink:
I think what's failing in the Mac test is a packet coming in after the community has been unloaded, but before the endpoint is closed. This is due to the closing order in Dispersy (which I also ran into with IPv8) at:

https://github.com/Tribler/dispersy/blob/a68fbe7eba4a6542c8d64975ac0504a0cff5e5b6/dispersy.py#L1705-L1719

devos50:
Hmm can we safely swap the closing order (first stop the endpoint and unload the communities afterwards)?

Also, I need to put the thread printing before the assertion error. If the assert triggers, the stacks do not get printed.

devos50:
retest this please

qstokkink:
@devos50 no this is not an easy fix, we can't just switch the order. Some communities send messages when unloaded (for instance the TunnelCommunity). We would need to implement a two phase endpoint shutdown: first disable receiving on the endpoint, then unload the communities and lastly disable sending on the endpoint.

devos50:
@qstokkink I don't think it's related to Dispersy since in the test that went wrong in the Mac build (https://jenkins.tribler.org/job/GH_Tribler_PR_tests_mac/4914/artifact/output/test_runners_output/0.err/*view*/), I don't even start Dispersy.

There are some more instances where we schedule stuff on the thread pool, for instance, our download callback, which might cause this issue. Still, it never gave any problems like this. I will create a separate issue for this and merge this PR.

qstokkink:
@devos50 yes that's a different issue (I was looking at the failed run before that)

PR #3109, opened 2017-09-24
WIP: needs package python-psutil/python-meliae or it won't run

themusicgod1:


devos50:
Also, it seems that we forgot to update these dependencies. We need a better way to keep track of all dependencies (like a script that automatically generates all dependencies we need).

qstokkink:
@devos50 We can iterate all `.py` files in the source tree and run:

```python
import ast
node = ast.parse(file_contents)
for subnode in ast.walk(node):
    if isinstance(subnode, ast.Import):
        print "Found import", [alias.name for alias in subnode.names]
```

This will list all normally defined imports (so no `__import__()` or `imp.load_module()`).

qstokkink:
Ok to test

devos50:
This will be part of the next -> devel backport so I close this one in favor of the backport.

PR #3118, opened 2017-09-27
Error logging to log files inside .Tribler directory

xoriole:
Error & info logs  are now written to respective error.log & info.log files inside .Tribler directory for each platform

### DIFF
--- a/run_tribler.py
+++ b/run_tribler.py
@@ -1,8 +1,42 @@
 import os
 import sys
 import multiprocessing
+import logging.config
 
 from check_os import check_environment
+from check_os import error_and_exit
+
+
+def setup_logging():
+    """
+    Setup logging to write logs to files inside \
+    .Tribler directory in each platforms
+    """
+    # First check if logger.conf is present or not
+    if getattr(sys, 'frozen', False):
+        # If the application is run as a bundle, the pyInstaller bootloader
+        # extends the sys module by a flag frozen=True and sets the app
+        # path into variable _MEIPASS'.
+        application_path = os.environ.get("_MEIPASS", os.path.abspath("."))
+    else:
+        application_path = os.path.dirname(os.path.abspath(__file__))
+
+    log_config = os.path.join(application_path, "logger.conf")
+
+    if not os.path.exists(log_config):
+        print "Log configuration file not found at location '%s'" % log_config
+        return
+
+    log_directory = os.path.abspath(os.environ.get('APPDATA', os.path.expanduser('~')))
+    log_directory = os.path.join(log_directory, '.Tribler', 'logs')
+
+    if not os.path.exists(log_directory):
+        os.makedirs(log_directory)
+
+    logging.info_log_file = '%s/tribler-info.log' % log_directory
+    logging.error_log_file = '%s/tribler-error.log' % log_directory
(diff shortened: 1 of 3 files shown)


qstokkink:
Could you look at [MemoryHandler](https://docs.python.org/2/library/logging.handlers.html#memoryhandler) which flushes to the INFO log on ERROR to reduce disk IO?

run_tribler.py
```diff
@@ -1,30 +1,75 @@
...
+    else:
+        application_path = os.path.dirname(os.path.abspath(__file__))
+
+    log_config = os.path.join(application_path, "logger.conf")
+
+    if not os.path.exists(log_config):
+        print "Log configuration file not found at location '%s'" % log_config
+        return
+
+    if sys.platform == 'linux2' or sys.platform == 'darwin':
```
qstokkink: I think it would be better to use [platform.system](https://docs.python.org/2/library/platform.html#platform.system) here, for generality.

qstokkink: Actually, you could even do:
```python
log_directory = os.path.abspath(os.environ.get('APPDATA', os.path.expanduser('.')))
```
and skip this entire statement.

run_tribler.py
```diff
@@ -1,30 +1,75 @@
...
-    app.parse_sys_args(sys.argv)
-    sys.exit(app.exec_())
+        window = TriblerWindow()
+        window.setWindowTitle("Tribler")
+        app.set_activation_window(window)
+        app.parse_sys_args(sys.argv)
+        sys.exit(app.exec_())
+    except ImportError as ie:
+        print "Import error: {0}".format(ie)
+        logging.exception(ie)
```
qstokkink: Also, related to #3120, could you call `check_os.error_and_exit()` here with the import message.

run_tribler.py
xoriole: Yes, that is neat. I'll make the changes.

`log_directory = os.path.abspath(os.environ.get('APPDATA', os.path.expanduser('.')))`
Instead of `os.path.expanduser('.')` it should be `os.path.expanduser('~')` as we create the directory inside user home.

run_tribler.py
xoriole: yes, that would be nice

run_tribler.py
qstokkink: @xoriole hmm I just checked, I guess neither is really needed:
```python
>>> os.path.abspath(os.path.expanduser('.'))
'/home/quinten'
>>> os.path.abspath(os.path.expanduser('~'))
'/home/quinten'
>>> os.path.abspath(os.path.expanduser(''))
'/home/quinten'
```

xoriole: @qstokkink It might probably be that you are running the python console from your home directory. This is my output on a console initialized at a different directory
```
>>> os.path.abspath(os.path.expanduser('.'))
'/home/sandip/personal/tribler'
>>> os.path.abspath(os.path.expanduser('~'))
'/home/sandip'
>>> os.path.abspath(os.path.expanduser(''))
'/home/sandip/personal/tribler'

```

qstokkink: @xoriole good catch 👍

PR #3141, opened 2017-10-06
Fix memory usage psutil bug

xoriole:
Old version of psutil (<4.0.0) does not have memory_full_info() method, so checked for the existence of the method and added proper fallback for it.

### DIFF
--- a/Tribler/Core/Modules/resource_monitor.py
+++ b/Tribler/Core/Modules/resource_monitor.py
@@ -1,7 +1,7 @@
 import logging
-import psutil
 import time
 
+import psutil
 from twisted.internet.task import LoopingCall
 
 from Tribler.dispersy.taskmanager import TaskManager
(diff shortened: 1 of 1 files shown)


qstokkink:
Has this been tested manually?

Tribler/Core/Modules/resource_monitor.py
```diff
@@ -45,7 +46,17 @@ def check_resources(self):
...
         self.cpu_data.append((time_seconds, self.process.cpu_percent(interval=None)))
-        self.memory_data.append((time_seconds, self.process.memory_full_info().uss))
+
+        # Get the memory usage of the process
+        # psutil package 4.0.0 introduced memory_full_info() method which among other info also returns uss.
+        # uss (Unique Set Size) is probably the most representative metric for determining how much memory is
+        # actually being used by a process.
+        # However, on psutil version < 4.0.0, we fallback to use rss (Resident Set Size) which is the non-swapped
+        # physical memory a process has used
+        if hasattr(self.process, "memory_full_info") and inspect.ismethod(getattr(self.process, "memory_full_info")):
```
devos50: `hasattr` should be enough right?

xoriole: Yes, it is tested manually.
`hasattr` is sufficient for checking the existence of attribute, rest is to check if it is a method.

qstokkink: Why not use the built-in [callable()](https://docs.python.org/2/library/functions.html#callable)?

Issue #3066, opened 2017-08-26
Stuck on (turning gears) loading screen


pimlie:
- [x ] I have read CONTRIBUTING.rst

- [x ] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
/opt/tribler# git status
On branch next
Your branch is up-to-date with 'origin/next'.

##### Operating system and version: #####
both alpine linux as ubuntu 16.04

This is happening quite often the last couple of months. As mentioned before I am running Tribler in a docker container and connect with vnc. Sometimes when I restart Tribler (or the docker container itself) Tribler will not start anymore and hangs on the first loading screen with the double turning gears (so before the menu on the left is loaded).
This happened yesterday evening and I thought lets see if it will do anything in the morning but its still stuck at the gears. It seems no output has been sent to a logfile (yet).

(I thought I had seen an issue about this before but cant seem to find it at the moment)

An strace of the main process is as follows:
```
 time strace -c -p 3569
strace: Process 3569 attached
^Cstrace: Process 3569 detached
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 94.73    0.072573           1     77617           poll
  2.47    0.001892           0     22045      3904 futex
  1.68    0.001289           0     70166           write
  0.42    0.000325           0     34982           writev
  0.41    0.000315           0     31507           read
  0.15    0.000112           0      4462           clone
  0.11    0.000082           0      4462           mprotect
  0.03    0.000020           0      4462           mmap
  0.00    0.000000           0         1           restart_syscall
------ ----------- ----------- --------- --------- ----------------
100.00    0.076608                249704      3904 total

real	3m43.813s
user	0m0.264s
sys	0m5.160s
```

--edit--
to clarify, restarting tribler again does not help. Also switching from the alpine to the ubuntu container and back doesnt help. But both containers are linked to the same `$HOME/.Tribler` volume. 

I looked in the `.Tribler` folder and noticed a `triblerd.lock` file which was persistent also when I shutdown Tribler. After shutting down Tribler, removing the lock file manually then starting Tribler again it proceeds immediately to the normal 'home' screen.

devos50:
Hmm I completely agree, we need more insight in the boot process of Tribler, something like a line in the GUI indicating the status. This makes it easier for us to debug. I will work on this 👍

qstokkink:
[...]

PR #3217, opened 2017-11-10
Added status to startup screen

devos50:
Addresses #3066, however, does not fix it.

I've also added a ctrl+d shortcut to open the debug panel, even when developer mode is not enabled.


qstokkink:
retest this please

qstokkink:
retest this please

PR #3253, opened 2017-11-22
Fix for download path write permission issue

xoriole:
Checked if there is sufficient write permission on the download path before attempting to download the torrent.

### DIFF
--- a/TriblerGUI/dialogs/startdownloaddialog.py
+++ b/TriblerGUI/dialogs/startdownloaddialog.py
@@ -8,7 +8,7 @@ from Tribler.Core.Utilities.utilities import quote_plus_unicode
 from TriblerGUI.dialogs.confirmationdialog import ConfirmationDialog
 from TriblerGUI.dialogs.dialogcontainer import DialogContainer
 from TriblerGUI.tribler_request_manager import TriblerRequestManager
-from TriblerGUI.utilities import get_ui_file_path, format_size, get_gui_setting, get_image_path
+from TriblerGUI.utilities import get_ui_file_path, format_size, get_gui_setting, get_image_path, is_dir_writable
 
 
 class DownloadFileTreeWidgetItem(QTreeWidgetItem):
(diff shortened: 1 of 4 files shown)


xoriole:
retest this please

TriblerGUI/dialogs/startdownloaddialog.py
```diff
@@ -147,12 +147,19 @@ def on_received_metainfo(self, metainfo):
...
+                                                                     "download", "", QFileDialog.ShowDirsOnly)
 
         if len(chosen_dir) != 0:
             self.dialog_widget.destination_input.setCurrentText(chosen_dir)
 
+        if not is_dir_writable(chosen_dir):
+            ConfirmationDialog.show_message(self.dialog_widget, "Insufficient Permissions",
+                                            "Tribler cannot download to <i>%s</i> directory. "
+                                            "Please add proper write permissions to the directory "
+                                            "or choose another download directory" % chosen_dir,
```
devos50: Please add a period after the last sentence.

TriblerGUI/dialogs/startdownloaddialog.py
```diff
@@ -163,7 +170,15 @@ def on_download_clicked(self):
             ConfirmationDialog.show_error(self.window(), "No files selected",
                                           "Please select at least one file to download.")
         else:
-            self.button_clicked.emit(1)
+            download_dir = self.dialog_widget.destination_input.currentText()
+            if not is_dir_writable(download_dir):
+                ConfirmationDialog.show_message(self.dialog_widget, "Insufficient Permissions",
```
devos50: Please extract this string into an attribute so we can reuse it.

TriblerGUI/tribler_window.py
```diff
@@ -444,6 +449,7 @@ def start_download_from_uri(self, uri):
         self.download_uri = uri
 
         if get_gui_setting(self.gui_settings, "ask_download_settings", True, is_bool=True):
+            print "asking download settings"
```
devos50: Don't forget to remove this print statement.

TriblerGUI/utilities.py
```diff
@@ -204,3 +204,13 @@ def get_gui_setting(gui_settings, value, default, is_bool=False):
     if is_bool:
         val = val == True or val == 'true'
     return val
+
+
+def is_dir_writable(path):
+    try:
+        open(os.path.join(path, 'temp'), 'w')
```
devos50: Why don't you use `os.access`? https://docs.python.org/2/library/os.html#files-and-directories

PR #3262, opened 2017-11-29
Fix hide torrent details tab

### LINKED ISSUE #3251: Download details still present even after removing all downloads
Download details of the torrent is still visible even if the last of the torrents in downloads is removed.
Screenshot below:
capture

xoriole:
Issue Fix #3251: When all the downloads are removed, torrent details tab is no longer shown.
Alert type 'torrent_removed_alert' is properly handled.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -294,6 +294,10 @@ class LibtorrentMgr(TaskManager):
             return torrent_handle
 
     def remove_torrent(self, torrentdl, removecontent=False):
+        """
+        Start removing a torrent, the process is completed when a 'torrent_removed_alert'
+        is received in 'process_alert'.
+        """
         handle = torrentdl.handle
         if handle and handle.is_valid():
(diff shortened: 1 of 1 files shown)


Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -301,6 +301,8 @@ def remove_torrent(self, torrentdl, removecontent=False):
                 self.torrents[infohash][1].remove_torrent(handle, int(removecontent))
                 out = self.torrents[infohash][0].deferred_removed
                 self._logger.debug("remove torrent %s", infohash)
+                # del self.torrents[infohash]; Don't remove the torrent here yet, let it be deleted
```
qstokkink: The code in this line comment confused me. Please remove it here and add it as general info to the docstring of the function.
Something like:

```python
def remove_torrent(self, torrentdl, removecontent=False):
"""
Start removing a torrent, the process is completed when a 'torrent_removed_alert' is received in 'process_alert'.
"""
```

qstokkink:
retest this please

Issue #3263, opened 2017-11-30
Torrent already exists in session exception on trying to resolve metadata


xoriole:
On trying to add a new torrent through magnet link, if it fails to resolve metadata (show list of files on the dialog) for the first time, then on trying to add the same magnet link again occasionally results into this error.
![metadata_issue](https://user-images.githubusercontent.com/1442867/33435168-bcc2194e-d5e1-11e7-9988-8b39aad8b5cb.png)

PR #3264, opened 2017-11-30
READY: Fix torrent already exists in session exception

xoriole:
If torrent already exists in the session, it is now removed from the session and added again. This triggers the metadata_received_alert handled by 'process_alert'. This removal and addition of torrent to the session is done for the new torrents only so it does not affect the existing torrents in the downloads.

Issue #3263


qstokkink:
Could you solve it like in `LibtorrentMgr`, instead of catching a RuntimeError and parsing the error message?

https://github.com/Tribler/tribler/blob/6e5e21cbf1f81bcab4caf61443d18d932d03443e/Tribler/Core/Libtorrent/LibtorrentMgr.py#L281-L284

xoriole:
retest this please

xoriole:
retest this please

Issue #3239, opened 2017-11-21
Build your own storagecoin - Group11


PauluzzNL:
Project H: Build your own Storagecoin

You will make your own storagecoin within this project. Filecoin is the largest ICO in the history of cybercurrency, see our analysis. A total of $257 million is given to a project without any running code. The big idea of your project and Filecoin is to turn any storage device into a cash generation machine. Key target is expanding an existing system with mobility of mined storage coins. Your code will allow send credits to another Android wallet. First task is to understand the Trustchain fabric by TUDelft, our 12 years of ledger prototyping and Android implementation. See project A for starting point with coding. Second task is to enable Android-to-Android transfer of coins. Finally, implement private key importing within your Android code and exporting using QR codes from the Tribler credit mining system. 

@devbridie
@PauluzzNL
@mitchellolsthoorn 
@roelofsol
@DavidGSB

RoelofSol:
Our first milestone is: 

Extend Tribler with the feature to generate a QR code from a key pair ( [pybip38](https://pypi.python.org/pypi/pybip38/0.9) )
This QR code can be picked up by a mobile phone using a version of the [TrustChain Android app ](http://trustchain-android.readthedocs.io/en/latest/) . 
The Android app can transfer credits to a second phone without using wifi. ( bluetooth/NFC/QR-codes )


Our first week objective: 

Have local builds of Tribler and Android and get acquainted with their code base.

RoelofSol:
Meeting - Friday - 24 Nov
We heard that there are problems with the available Android app. 
We hope to get a POC QR exchange between PC and Android working. 

Tasks: 

- Activate the Tribler market? and extract the key pair ( David & Paul & Mitchell ) 
- Use Python to generate QR codes and display in Qt5  ( Roelof ) 
- POC Android app to scan and display QR Codes ( Dereck )

RoelofSol:
QR generation examples
![screenshot from 2017-11-28 13-39-17](https://user-images.githubusercontent.com/668529/33320243-dbfa73f2-d441-11e7-8725-6db47fb269a8.png)
![screenshot from 2017-11-28 13-39-58](https://user-images.githubusercontent.com/668529/33320253-e42874a2-d441-11e7-8fba-e2cff28327f8.png)

RoelofSol:
Update existing android app with QR scan to import the key to android
Export QR code button in Tribler. 
Contact Overlay app group & student assistants

mitchellolsthoorn:
[...]

PR #3267, opened 2017-12-02
WIP: Added ability to export chain keys

mitchellolsthoorn:
Needed for #3239, where it is used to export the keys to an android client

Dependencies used:
python-qrcode

### DIFF
--- a/TriblerGUI/widgets/settingspage.py
+++ b/TriblerGUI/widgets/settingspage.py
@@ -1,4 +1,7 @@
-from PyQt5.QtWidgets import QWidget, QFileDialog
+from PyQt5 import QtGui
+from PyQt5.QtWidgets import QWidget, QLabel, QFileDialog
+from PIL.ImageQt import ImageQt
+import qrcode
 
 import Tribler.Core.Utilities.json_util as json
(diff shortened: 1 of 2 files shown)


TriblerGUI/widgets/settingspage.py
```diff
@@ -1,4 +1,7 @@
-from PyQt5.QtWidgets import QWidget, QFileDialog
+from PyQt5 import QtGui
+from PyQt5.QtWidgets import QWidget, QLabel, QFileDialog
+from PIL.ImageQt import ImageQt
+import qrcode
```
devos50: Just an idea, we can make this `qrcode` import conditional, something like:

```
try:
    import qrcode
    has_qr = True
except ImportError:
    has_qr = False
```

And show an error if has_qr is `False`?

mitchellolsthoorn: Yeah, that is probably better, so it causes less dependency errors. Do you have an example how I can call an error widget?

devos50: Yes, see https://github.com/Tribler/tribler/blob/devel/TriblerGUI/dialogs/confirmationdialog.py#L46

mitchellolsthoorn:
@devos50, I updated the code to test if the module is present and also included an error for when the keys are missing or are still being created

mitchellolsthoorn:
This PR has become obsolete and is replaced by PR #3358

PR #3277, opened 2017-12-08
Fix crash on remote host closing connection

### LINKED ISSUE #3274: Tribler can be crashed externally
Sending a `/torrentinfo` request to the core externally and closing the connection prematurely results in  Tribler to crash.
```python
  File "/home/sandip/personal/tribler/TriblerGUI/event_request_manager.py", line 99, in on_read_data
RuntimeError: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionDone'>: Connection was closed cleanly.
]
```

xoriole:
Before any data is written by request.write() it now checks if the request has not already finished and the channel is not None. One case where the channel is None is when a remote host closes the connection before the response is served. In such a case just checking 'request.finished is not True' is not sufficient and channel should also be checked.

Fixes #3274

### DIFF
--- a/Tribler/Core/Modules/restapi/rest_manager.py
+++ b/Tribler/Core/Modules/restapi/rest_manager.py
@@ -41,7 +41,8 @@ class RESTManager(TaskManager):
 
 class RESTRequest(server.Request):
     """
-    This class gracefully takes care of unhandled exceptions raised during the processing of any request.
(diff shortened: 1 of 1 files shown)


xoriole:
retest this please

xoriole:
retest this please

PR #3290, opened 2017-12-13
Custom log directory with separate core and GUI logs

xoriole:
Separated core and GUI logs and added a custom log directory option in settings


Tribler/Core/Modules/restapi/debug_endpoint.py
```diff
@@ -386,7 +387,8 @@ def render_GET(self, request):
             handler.flush()
 
         # Get the location of log file
-        log_file = os.path.join(self.session.get_state_dir(), 'logs', 'tribler-info.log')
+        param_process = request.args['process'][0] if request.args['process'] else 'core'
```
devos50: We can crash Tribler now if we provide anything else than `core` or `gui`. Could you add a check for this?

PR #3308, opened 2017-12-21
Startup checks for LevelDB and SQLite database errors

xoriole:
Checked LevelDB and SQLite databases for common startup errors.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -117,10 +117,14 @@ class TriblerLaunchMany(TaskManager):
             if self.session.get_torrent_store():
                 from Tribler.Core.leveldbstore import LevelDbStore
                 self.torrent_store = LevelDbStore(self.session.get_torrent_store_dir())
+                if not self.torrent_store.get_db():
+                    raise RuntimeError("Torrent store (leveldb) is None which should not normally happen")
 
             if self.session.get_enable_metadata():
                 from Tribler.Core.leveldbstore import LevelDbStore
                 self.metadata_store = LevelDbStore(self.session.get_metadata_store_dir())
+                if not self.metadata_store.get_db():
+                    raise RuntimeError("Metadata store (leveldb) is None which should not normally happen")
 
             # torrent collecting: RemoteTorrentHandler
             if self.session.get_torrent_collecting():
--- a/Tribler/Core/CacheDB/sqlitecachedb.py
+++ b/Tribler/Core/CacheDB/sqlitecachedb.py
@@ -128,6 +128,13 @@ class SQLiteCacheDB(TaskManager):
 
         cursor = self.get_cursor()
 
+        # Check integrity of the database
+        check_response, = self.execute(u"PRAGMA quick_check").next()
+        if check_response != 'ok':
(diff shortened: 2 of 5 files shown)


devos50:
Nice, this is really helpful!

run_tribler.py
```diff
@@ -18,6 +18,9 @@
     # Set up logging
     setup_gui_logging()
 
+    # Check sqlite databases and LevelDb stores
+    check_datastores()
```
devos50: I argue that this should be moved to the Tribler core since this won't work if the GUI and Core are running on separate machines (I always make that assumption for myself).

xoriole: At this point in line 22, neither the GUI nor the core has started yet so I thought it would be a good place to do some preliminary database checks here. Also, we can have more extensive checks when the core is starting. This will surely increase the startup time so some benchmarking would be necessary.

devos50: We can still run these checks before the session has started when moving the code to the core.

Creating a benchmark for booting times is a good idea. We should discuss this at the beginning of the next year 👍

check_os.py
```diff
@@ -117,3 +125,62 @@ def setup_logging(gui=False):
...
+        from leveldb import LevelDB, LevelDBError
+        use_leveldb = True
+    except ImportError:
+        from Tribler.Core.plyveladapter import LevelDB
+        use_leveldb = False
+    except:
+        error_and_exit("Tribler Error", "LevelDB not found")
+
+    try:
+        db_path = store_dir.decode('windows-1252') if sys.platform == "win32" else store_dir
```
devos50: Here we have quite some code duplication. We should extract this into a more generic method.

xoriole: You are right. I'll refactor the code.

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -117,10 +117,12 @@ def register(self, session, sesslock):
             if self.session.get_torrent_store():
                 from Tribler.Core.leveldbstore import LevelDbStore
                 self.torrent_store = LevelDbStore(self.session.get_torrent_store_dir())
+                assert self.torrent_store.get_db() is not None
```
devos50: I think we shouldn't have assert statements in production code. We have unit tests for this (while I understand your reasoning here and why it might be helpful in this particular case). Now that we are handling all cases of the except when creating a leveldb store, I don't think we need these assert statements (if the database cannot be created, we always throw an exception now).

xoriole: Hmm, alright, will make the changes. Looking at the reporter log [#11858](http://reporter.tribler.org/stacktrace/11858), it seems Tribler was already running for 50 seconds so somehow it might have got through the constructor exception handling.

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

PR #3331, opened 2018-01-04
Fix DebugEndpoint file leak

### LINKED ISSUE #3330: debug_endpoint does not close opened files
[`DebugEndpoint`](https://github.com/Tribler/tribler/blob/next/Tribler/Core/Modules/restapi/debug_endpoint.py) does not close its opened files:

https://github.com/Tribler/tribler/blob/486797b090c637e8c1b35059a8ce3962589b79b4/Tribler/Core/Modules/restapi/debug_endpoint.py#L347

https://github.com/Tribler/tribler/blob/486797b090c637e8c1b35059a8ce3962589b79b4/Tribler/Core/Modules/restapi/debug_endpoint.py#L400

https://github.com/Tribler/tribler/blob/486797b090c637e8c1b35059a8ce3962589b79b4/Tribler/Core/Modules/restapi/debug_endpoint.py#L403

Either `with` or `close` should be used. This is leaking file handles.

qstokkink:
Fixes #3330 

This PR closes the file handlers opened by the DebugEndpoint.

This PR also fixes the Windows unittest segfault.

### DIFF
--- a/Tribler/Core/Modules/restapi/debug_endpoint.py
+++ b/Tribler/Core/Modules/restapi/debug_endpoint.py
@@ -344,7 +344,10 @@ class DebugMemoryDumpEndpoint(resource.Resource):
         date_str = datetime.datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
         request.setHeader(b'content-type', 'application/json')
         request.setHeader(b'Content-Disposition', 'attachment; filename=tribler_memory_dump_%s.json' % date_str)
-        return open(dump_file_path).read()
+        content = ""
+        with open(dump_file_path, 'r') as dump_file:
+            content = dump_file.read()
+        return content
 
 
 class DebugLogEndpoint(resource.Resource):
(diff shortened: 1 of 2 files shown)


qstokkink:
@devos50 ready for review! Also, our Windows machine appears to have an old version of `meliae`. Nevertheless, this PR allows for a partial dump even using an old `meliae`.

Tribler/Core/Modules/restapi/debug_endpoint.py
```diff
@@ -12,6 +14,12 @@
 import Tribler.Core.Utilities.json_util as json
 
 
+class MemoryDumpBuffer(StringIO):
```
devos50: Why do we need this `MemoryDumpBuffer` exactly?

qstokkink: `meliae` expects its file handle to support `write()`, `flush()` and `__call__()`, the `StringIO` class does not support the latter.

devos50: Could you add a comment in the code for that? 👍

qstokkink: Done 👍

devos50:
@qstokkink thanks, I just went to EWI for a quick visit and fixed the Mac deployment tester. The fans of this machine were blowing at full speed so I assume some process was stuck internally. After restarting it, the fans went back to normal levels. I also checked the other mac whether there's a Tribler running in the background but I couldn't find it so there's something else that might be causing the failing GUI tests on Mac.

qstokkink:
Hmm.. weird.

PR #3340, opened 2018-01-11
Fixed bugs in torrent checker

devos50:
They shouldn't be set to None since individual torrent checker session still require these attributes in their callback/errbacks.


devos50:
@qstokkink @xoriole @egbertbouman for some very weird reason, I'm unable to reproduce the 'endless shutdown' bug from my home network :/, either when clicking a lot in the GUI or when using the twistd plugin. It always closes within 20 seconds, regardless of how many torrents I click.

xoriole:
@devos50 I'm still able to reproduce the bug with this fix.

qstokkink:
@devos50 I can still reproduce it on Ubuntu 16.10 using the code of this PR.

devos50:
retest this please

PR #3349, opened 2018-01-12
Force restart Tribler

### LINKED ISSUE #3348: Check for zombie Tribler instance at start up
When Tribler is force killed or exits due to some unexpected error, it sometimes leaves behind a zombie core process and subsequent relaunch of Tribler fails.

One solution to this issue could be to check for such processes and kill it before starting the fresh instance of Tribler. But the issue with this approach is that it could lead to database or data corruption if the previously running instance was writing to disk.

For now, I suggest using a popup dialog to inform the user if there is already running instance of Tribler and ask the user if he wants to force restart. A caution message could also be included in the pop up to indicate the possibility of data corruption.

tribler-warning

xoriole:
Asks user whether to force restart Tribler if there is more than one instance running. This will help user to kill any zombie instances which might have been left behind by previous force kill command or some other unexpected exceptions and relaunch Tribler again.

Fixes #3348

### DIFF
--- a/check_os.py
+++ b/check_os.py
@@ -1,9 +1,14 @@
 import logging
 import os
+import subprocess
 import sys
 import tempfile
+
+from Tribler.Core.Modules.process_checker import ProcessChecker
 from Tribler.Core.SessionConfig import SessionStartupConfig
 
+FORCE_RESTART_MESSAGE = "A Tribler instance is already running. Do you want to force restart? " \
+                        "\n\nCaution: force restart could result in data corruption."
 
 def error_and_exit(title, main_text):
     """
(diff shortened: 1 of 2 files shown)


check_os.py
```diff
@@ -1,9 +1,18 @@
+import Tkinter
```
qstokkink: Could you import `Tkinter` in the function itself instead of globally? Otherwise we are stuck with the `Tkinter` module loaded the rest of the application's lifetime, even if we will never need it.

xoriole:
retest this please

qstokkink:
This stuff is always really iffy. Have you tested both Windows from source (`python run_tribler.py`) and Windows from compiled (`tribler.exe`)?

xoriole:
@qstokkink only tested for the frozen environment. I can add the support for running from the source in windows if you think it is necessary. But it does work in Linux from source :D

qstokkink:
@xoriole that would be good for completeness (I believe it's `python.exe` for Windows from source).

run_tribler.py
```diff
@@ -4,7 +4,7 @@
 import logging.config
 
 from check_os import check_environment, check_free_space, setup_gui_logging
-from check_os import error_and_exit
+from check_os import error_and_exit, should_kill_other_tribler_instances
```
qstokkink: Could you put these in the same import statement?

xoriole:
retest this please

PR #3355, opened 2018-01-14
Bug fix in downloadpage

EinNarr:
This bug causes mess when deleting multiple downloads within 1 second. It may delete nothing or wrong download.

### DIFF
--- a/TriblerGUI/widgets/downloadspage.py
+++ b/TriblerGUI/widgets/downloadspage.py
@@ -128,15 +128,11 @@ class DownloadsPage(QWidget):
                 self.window().download_details_widget.update_pages()
 
         # Check whether there are download that should be removed
-        toremove = set()
-        for infohash, item in self.download_widgets.iteritems():
+        for infohash, item in self.download_widgets.copy().iteritems():
             if infohash not in download_infohashes:
                 index = self.window().downloads_list.indexOfTopLevelItem(item)
-                toremove.add((infohash, index))
-
-        for infohash, index in toremove:
-            self.window().downloads_list.takeTopLevelItem(index)
-            del self.download_widgets[infohash]
+                self.window().downloads_list.takeTopLevelItem(index)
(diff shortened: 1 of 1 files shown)


TriblerGUI/widgets/downloadspage.py
```diff
@@ -128,15 +128,11 @@ def on_received_downloads(self, downloads):
                 self.window().download_details_widget.update_pages()
 
         # Check whether there are download that should be removed
-        toremove = set()
-        for infohash, item in self.download_widgets.iteritems():
+        for infohash, item in self.download_widgets.copy().iteritems():
```
qstokkink: You can just use `.items()` instead of `.copy().iteritems()`.

EinNarr: Would you prefer to use .items() instead of .iteritems()? Since other part of this file is in python 2, I picked .iteritems()

.copy() is necessary. We delete items during the iteration, this is not allow in a for-loop.

qstokkink: Yes I would prefer `.items()`, which is perfectly valid Python 2: https://docs.python.org/2/library/stdtypes.html#dict.items

Excerpt from the documentation:
> Return a copy of the dictionary’s list of (key, value) pairs.

EinNarr: In python 3 it returns a new view instead of a copy of list, would that be safe to remove itemd from it?

qstokkink: If we (ever) switch to Python 3, we can use `frozenset()` 👍 .

qstokkink:
Ok to test

qstokkink:
I just noticed this patch is for the `devel` branch, please cherry-pick it onto `next` so we can merge it.

devos50:
Closed in favor of #3356

PR #3364, opened 2018-01-18
Next to devel

### LINKED ISSUE #3367: CI for latest devel is broken
See https://github.com/Tribler/tribler/pull/3366.

devos50:
Fixes #3367


devos50:
@qstokkink @xoriole ready for your review. I've fixed the upgrade issue in https://github.com/Tribler/tribler/pull/3364/commits/46a814974e1bce72c3e9834d62a583c6a28531b4 and I would like to know whether you agree with this solution.

Simply said, I made sure that the current `TriblerConfig` object is reused and not initialized again. I also improved the upgrade script so it only merges the `libtribler.conf` and `tribler.conf` files once (they are removed when the upgrade is done). Also, the trustchain option is only enabled when we are actually upgrading from 7.0 to 7.1.

Also, the mac GUI test fails again, probably due to the popup. I will fix it tomorrow.

xoriole:
retest this please

xoriole:
@devos50 I like the idea of reusing the current ```TriblerConfig``` object but I don't think it's a good idea to remove ```libtribler.conf``` and ```tribler.conf``` after merge. This can lead to inconsistent state if Tribler crashes.

PR #3375, opened 2018-01-20
Remove redundant logic from LibtorrentDownloadImpl.network_stop

vandenheuvel:
This method appears to contain useless statements.


vandenheuvel:
@devos50 Do you think this is a good change, or am I missing something?

devos50:
I'm not 100% sure whether this code can be removed. The VCS annotations suggest that @egbertbouman created this code. I will ask about it on monday.

egbertbouman:
It doesn't look like this is being used anymore. The pstate.set gets called, but nothing is being done with pstate. Feel free to remove!

vandenheuvel:
@egbertbouman thanks, I'll update this branch and put it on `READY` for  @devos50 to review.

devos50:
Thanks!

PR #3383, opened 2018-01-22
Check dialog widget is None

xoriole:
Removed previous dialog if exists when opening a new dialog and also checked if the dialog is none when handling torrent from URL.


TriblerGUI/tribler_window.py
```diff
@@ -456,7 +456,12 @@ def start_download_from_uri(self, uri):
...
 
         if get_gui_setting(self.gui_settings, "ask_download_settings", True, is_bool=True):
-            self.dialog = StartDownloadDialog(self.window().stackedWidget, self.download_uri)
+            # Clear any previous dialog if exists
+            if self.dialog:
+                self.dialog.setParent(None)
+                self.dialog = None
+
+            self.dialog = StartDownloadDialog(self, self.download_uri)
             self.dialog.button_clicked.connect(self.on_start_download_action)
```
qstokkink: Shouldn't you also disconnect this?

xoriole: yes, thanks for reminding

PR #3410, opened 2018-01-29
Hooked IPv8 into MIMEndpoint

qstokkink:
Added the IPv8 stack (without loaded communities) onto the existing Dispersy endpoint.


devos50:
@qstokkink make sure to update the builders so ipv8 is copied into the frozen environment.

PR #3426, opened 2018-01-31
Monitor free disk space

### LINKED ISSUE #3423: IOError: No space left on device
From reporter (Tribler 7.0.0)
```python
File "TriblerGUI\event_request_manager.py", line 99, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.IOError'>: [Errno 28] No space left on device
lib\site-packages\twisted\internet\base.py:1243:run
lib\site-packages\twisted\internet\base.py:1252:mainLoop
lib\site-packages\twisted\internet\base.py:878:runUntilCurrent
lib\site-packages\twisted\internet\task.py:239:__call__
lib\site-packages\twisted\internet\defer.py:150:maybeDeferred
Tribler\dispersy\discovery\community.py:809:clean_and_save
]
```

In the report, we see that Tribler was already running for over 10 hours before it crashed. We do already have a disk space check at the startup but it seems it is not sufficient. We need to continuously monitor the available disk space and notify the user when disk space is below a threshold.

In case the disk space is critically low, we should stop the downloads and also the core, and ask the user to free up the disk and restart Tribler.

xoriole:
Monitor free disk space.
Fixes #3423 `No space left on device`


Tribler/Core/Modules/resource_monitor.py
```diff
@@ -62,6 +66,19 @@ def check_resources(self):
         elif hasattr(self.process, "memory_info") and callable(getattr(self.process, "memory_info")):
             self.memory_data.append((time_seconds, self.process.memory_info().rss))
 
+        # Check for available disk space
+        disk_usage = psutil.disk_usage(".")
```
qstokkink: Is this the same as the disk being downloaded to?

devos50: Interesting question. Insufficient disk space when downloading torrents should be handled by libtorrent already IIRC. We're writing quite some stuff to our Tribler state dir (`.Tribler`) and theoretically, this could be on another disk. This code line fetches the available space of the disk where the _Python source files_ are located, which might be a different disk than the state directory.

TLDR: I would change it to something like `psutil.disk_usage(self.session.config.get_state_dir())`.

Tribler/Core/Modules/resource_monitor.py
```diff
@@ -62,6 +66,19 @@ def check_resources(self):
...
+        self.disk_usage_data.append({"time": time_seconds,
+                                     "total": disk_usage.total,
+                                     "used": disk_usage.used,
+                                     "free": disk_usage.free,
+                                     "percent": disk_usage.percent})
+
+        # Notify session if less than 100MB of disk space is available
+        if disk_usage.free < 100 * (1024 * 1024.0):
```
devos50: either use 1024.0 or 1024, not both please 👍

PR #3437, opened 2018-02-05
FUTURE: Remove deprecated features of LibTorrent

Captain-Coder:
My libtorrent (built from Arch User Repository) does not contain the deprecated features by default. This PR documents the current state of affairs of what is needed to make tribler work with lib torrent without the deprecated features.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -41,7 +41,6 @@ DEFAULT_DHT_ROUTERS = [
     ("router.utorrent.com", 6881)
 ]
 DEFAULT_LT_EXTENSIONS = [
-    lt.create_metadata_plugin,
     lt.create_ut_metadata_plugin,
     lt.create_ut_pex_plugin,
     lt.create_smart_ban_plugin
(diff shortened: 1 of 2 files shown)


Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -204,16 +212,19 @@ def create_session(self, hops=0, store_listen_port=True):
...
             ltsession_settings['upload_rate_limit'] = self.tribler_session.config.get_libtorrent_max_upload_rate()
             ltsession_settings['download_rate_limit'] = self.tribler_session.config.get_libtorrent_max_download_rate()
-            ltsession.set_settings(ltsession_settings)
+            if is_1_1_version:
+                ltsession.apply_settings(ltsession_settings)
+            else:
+                ltsession.set_settings(ltsession_settings)
 
         for router in DEFAULT_DHT_ROUTERS:
             ltsession.add_dht_router(*router)
```
arvidn: it's much better to add these as part of the settings when the session is constructed. (I'll make a note there too)

Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -129,31 +128,30 @@ def shutdown(self):
...
 
+        is_1_1_version = LooseVersion(self.get_libtorrent_version()) >= LooseVersion("1.1.0")
+
         # Due to a bug in Libtorrent 0.16.18, the outgoing_port and num_outgoing_ports value should be set in
         # the settings dictionary
         settings['outgoing_port'] = 0
         settings['num_outgoing_ports'] = 1
 
+        if is_1_1_version:
+            settings["dht_bootstrap_nodes"] = ""
```
arvidn: I imagine you would want to do something like: ``','.join(DEFAULT_DHT_ROUTERS)``

Tribler/Core/Libtorrent/LibtorrentMgr.py
Captain-Coder: Yup, I completely agree! I previously submitted a PR that did just that (aswell as join the DHT nodes). https://github.com/Tribler/tribler/pull/3084/files#diff-63cb62bb807795516a444dcad4ba3097 but it got rejected. Mostly because it broke older versions. This formulation/structure is the minimal that @devos50 came up with that results in the least extra code to support the various versions.

qstokkink:
I'd like someone else to verify, but I think Debian is now pushing `1.1` and we can pick up this PR again.

devos50:
I think this issue is superseded by #4368. Note that I took this PR as a reference for my work in #4368 so it has been helpful. 👍

Issue #3183, opened 2017-11-03
The future of our DHTs plan


synctext:
What is our goal with hidden seeder discovery?

We now have the Libtorrent DHT and the Python based DHT using different ports.
- Do we create a Libtorrent patch to expose their DHT?
- Do we create a DHT community?

Related issues. DHT self-DDos: #3065 Latency influence on Libtorrent: #2620

qstokkink:
Related to #109

synctext:
In introduction point we might require to do "DHT address rewriting" and keep track of DHT ID of the seeders for parsing incoming DHT responses. (DHT NATting) :-1:

egbertbouman:
We could solve this in a number of ways, for instance:

* Run a single (pymdht) DHT instance for each exit node and use this to do all our DHT requests. This would either require detecting and parsing DHT traffic coming from libtorrent, or somehow send the DHT request to the exit node manually (e.g., using a Dispersy message).
* Create our own DHT community. Since pymdht does not support BEP-44, the main benefit is that we could store additional information in the DHT, avoiding the use of the key-request/response stuff.
* As a secondary mechanism maybe we could use the top-n most frequently used public UDP trackers (from our database).

synctext:
Attack nearby keyspace:
```BEP44 allows keys to be the hash of an ed25519 public key with arbitrary 1k signed content. However, this protocol extension introduces new problems. An attacker can identify a hash they want to disrupt and generate nearby hashes so that responsibility for a hash moves onto nodes the attacker controls. This attack is addressed in dht_sec, but there is no analogous protection for BEP44 dht_store keys.```
https://gist.github.com/substack/eadd13302d785dc13aac#file-results-markdown

egbertbouman:
Status:
- Kademlia implementation: fast and designed for integrated NAT traversal
- routing table functional, designed as isolated DHT community https://github.com/egbertbouman/tribler/blob/dht/Tribler/community/dht/routing.py#L119
- the current protocol is iterative, which means that due to connectablity issues we will not always get a response
- besides smaller unit tests, we also have a larger test with a network of 25 nodes https://github.com/egbertbouman/tribler/blob/dht/Tribler/Test/Community/DHT/test_community.py#L109
- no node balancing test yet or any other deployments tests

synctext:
ToDo in Design 3: caching mechanism for popular torrents or Tribler channels. (e.g. no trackers needed)

qstokkink:
[...]

PR #3438, opened 2018-02-05
WIP: Refactor DHT usage

Captain-Coder:
This implements https://github.com/Tribler/tribler/issues/3183#issuecomment-344624844 option 1.

- [X] Send DHT requests to the exit's dht implementation & add to download
- [X] Disable libtorrent dht when hops > 0

Optional feature:
- [ ] Run a single dht on an exit. Currently every tunnel helper gets one. If an exit runs multiple of those, the exit will have multiple dht's running

Also a minor fix, when DLSTATUS == Metadata, we do want to find introduction points, to get metadata that way.


devos50:
I assume this PR is superseded by the implementation and deployment of EgbertDHT?

PR #3445, opened 2018-02-07
Ported the MarketCommunity to IPv8

devos50:
This PR ports the market community to IPv8 structure, which leads to a significant cleanup of the code.

Due to inner workings of the Python import mechanism, I had to add the IPv8 base files for testing to the Tribler repository. While this leads to code duplication, we can't think of any easy way of preventing this.

Depends on #3462


devos50:
We should update the submodule pointer of ipv8 before merging this 👍

qstokkink:
You could also do that in this PR.

qstokkink:
Some pop-up documentation:

Be sure to add your overlay to `<LaunchManyCore instance>.ipv8.overlays` and add a walking strategy to `<LaunchManyCore instance>.ipv8.strategies` (probably best to do this from within `LaunchManyCore.load_communities()`).

We should also enable the ipv8 `DiscoveryCommunity` in the `IPv8Module` class.

devos50:
This PR requires that the tunnel community is refactored properly first since the `TrustchainWallet` in this PR expects a IPv8 compatible trustchain community. I will work towards that first.

devos50:
retest this please

devos50:
retest this please

PR #3449, opened 2018-02-08
Ignore invalid infohash

xoriole:
If the info hash is found invalid when trying to add a torrent, it is simply logged and ignored.


xoriole:
retest this please

Issue #1410, opened 2015-05-08
Remove some session locks in LaunchManyCore and Libtorrent


LipuFei:
@NielsZeilemaker Since you have changed the RawServer to a twisted thing, I guess all those NoDispersyLocks do not make sense any more, and they can be removed, right?

LipuFei:
I was trying to remove the policies and I found there is a lot of functions in LaunchManyCore and Libtorrent that have crazy chains of callbacks. It's already giving me a big headache... If it is ok to remove them, I will try to do it as well.

NielsZeilemaker:
Sure, I would keep some of the locks around the download list
Op 8 mei 2015 12:36 PM schreef "Lipu Fei" <notifications@github.com>:

vandenheuvel:
@devos50 Could you assign this to me? This is close to https://github.com/Tribler/tribler/issues/2286.

synctext:
Closing this issue, please re-open if still relevant after ASYNC refactor work.

PR #3451, opened 2018-02-08
WIP: Remove locks

vandenheuvel:
Solves #1410.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -430,7 +430,6 @@ class TriblerLaunchMany(TaskManager):
     def add(self, tdef, dscfg, pstate=None, setupDelay=0, hidden=False,
             share_mode=False, checkpoint_disabled=False):
         """ Called by any thread """
-        d = None
         with self.session_lock:
             if not isinstance(tdef, TorrentDefNoMetainfo) and not tdef.is_finalized():
                 raise ValueError("TorrentDef not finalized")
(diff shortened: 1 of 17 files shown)


synctext:
Big change! This addresses the 'crazy chain of callbacks'?

devos50:
@vandenheuvel first of all, thank you for making contributions. While it may be tempting to remove the (at first sight, useless) locks from the session/downloads, they actually need to be there if I remember correctly. I just had some discussion about it with others but we couldn't reach a consent. I suspect that the main reason is that we still call some methods from the thread pool. Please correct me if I'm wrong.

Also, before you decide to refactor some critical code or remove some functionalities (like the locks), please discuss it with us first so we can give our opinion. This might save you some time in the end.

vandenheuvel:
@devos50 thanks for your response. Don't worry; this cost me very little time.

Could you point me to code managing the thread pool?

vandenheuvel:
@qstokkink do you know which code @devos50 is referencing?

qstokkink:
@vandenheuvel that is sadly not my area of expertise within Tribler (also I believe that part of the code is +- 10 years old)

devos50:
@vandenheuvel https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L1053

qstokkink:
Most of the changes in this PR are outdated and will naturally be fixed as we keep refactoring each component in Tribler. As such, I will close this PR.

PR #3452, opened 2018-02-09
Refactored SOCKS5 server

devos50:
This PR significantly refactors the SOCKS 5 server. The main improvement is that the server is cleaned from any tunnel community-related logic and is really a proxy right now.

I've added a so-called `TunnelDispatcher`, which can be seen as a "secondary" proxy between SOCKS5 and the tunnel community. The dispatcher makes sure that the right circuits are chosen for data coming out of the SOCKS5 and that the right SOCKS server is selected when there's incoming data from the tunnel community. We create a SOCKS5 server for each number of hops. This design improves separation between the tunnels and SOCKS5 and is a requirement for future engineering (which includes porting the tunnels to IPv8).


devos50:
@qstokkink can you also review this?

PR #3469, opened 2018-02-15
Added Tribler version to title bar in GUI

### LINKED ISSUE #3457: Version information of important dependencies
It would be useful in debugging if we can get easily see information about the versions of all the important dependencies in Tribler. To implement this, we can have a debug endpoint for versions and add a panel to show the results in Debug pane. 

We can start with the following dependencies to begin with:
- Libtorrent
- OpenSSL
- Qt, PyQt

devos50:
Fixes #3457

### DIFF
--- a/TriblerGUI/debug_window.py
+++ b/TriblerGUI/debug_window.py
@@ -108,7 +108,7 @@ class DebugWindow(QMainWindow):
     community information.
     """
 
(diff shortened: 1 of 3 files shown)


devos50:
retest this please

qstokkink:
retest this please

PR #3480, opened 2018-02-26
READY: better support for a large number of downloads

egbertbouman:


egbertbouman:
@devos50 Done. The uncovered lines in `test_libttorrent_mgr.py` were due to a bug in the unit test itself, but that's fixed now.

devos50:
Thanks!

PR #3483, opened 2018-02-27
Fixed mainline DHT when used for hidden seeding

devos50:
In this PR, I load the mainline DHT before loading IPv8 overlays. I also disabled thread pool processing of messages in the market and tunnel overlays.


devos50:
retest this please

PR #3491, opened 2018-03-04
Fixed reactor conflict in tests

### LINKED ISSUE #3490: ipv8_util.py fighting with twisted_thread.py
The `devel` tests are currently spewing a `ReactorAlreadyRunning` exception.

This is because both `ipv8_util.py` and `twisted_thread.py` are trying to set the Twisted reactor to their own threaded reactor.

The solution is to have `ipv8_util.py` use the `deferred()` function from `twisted_thread.py`. The contents of the new `ipv8_util.py` should be:

```python
from twisted.internet.defer import inlineCallbacks

from Tribler.Test.twisted_thread import deferred


def twisted_wrapper(arg):
    """
    Wrap a twisted test. Optionally supply a test timeout.
    Note that arg might either be a func or the timeout.
    """
    if isinstance(arg, (int, long)):
        return lambda x: deferred(arg)(inlineCallbacks(x))
    return deferred(timeout=10000)(inlineCallbacks(arg))
```

qstokkink:
Fixes #3490


qstokkink:
@devos50 this fixes the segfault on the market tests.

devos50:
@qstokkink hmm weird, the core dump suggested that the segfault was caused by sqlite. But it might ultimately be caused by some weird reactor magic.

devos50:
@qstokkink want me to merge it?

qstokkink:
@devos50 please 👍

Issue #3150, opened 2017-10-09
Tribler 7 RC3 hangs on startup


vandenheuvel:
On tag `v7.0.0-rc3`, https://github.com/Tribler/tribler/commit/e8ae41e33890a144d5d1d62f6a58ce8c60ae2f5f. Arch Linux based on 4.13.4-1. I have installed the `python2-meliae` package from the AUR.

When starting Tribler without a state directory, I receive the following error:
```
INFO    1507545687.25 single_application:23    __init__()
INFO    1507545687.32 single_application:49      self._outSocket.error() = '0'
INFO    1507545687.32 single_application:52    received QLocalSocket.ConnectionRefusedError; removing server.
INFO    1507545687.32 single_application:64    close()
INFO    1507545687.32 single_application:71    close(): returning
INFO    1507545687.32 single_application:60    __init__(): returning
INFO    1507545687.64 event_request_manager:117   Will connect to events endpoint
INFO    1507545687.66 event_request_manager:42    Got Tribler core error: 1
INFO    1507545687.68   sqlitecachedb:136   begin page_size upgrade...
INFO    1507545687.69   sqlitecachedb:140   ...end page_size upgrade
INFO    1507545687.69   sqlitecachedb:160   Initializing new database...
INFO    1507545687.70   sqlitecachedb:181   Reading database version...
INFO    1507545687.70   sqlitecachedb:185   Current database version is 28
INFO    1507545687.70   sqlitecachedb:203   Beginning the first transaction...
INFO    1507545687.71 torrent_upgrade65:25    torrent collecting directory not found, skip: None
INFO    1507545687.71   sqlitecachedb:225   Start committing...
INFO    1507545687.71   sqlitecachedb:238   Beginning another transaction...

```
The GUI loading screen does show up (the grinding gears), but I am not yet being asked for my keyring password.

hovancik:
I'm seeing similar from time to time on debian: 

the screen with gears shows up but nothing after... What helps sometimes is to restart app, but mostly I need to reinstall.

xoriole:
Encountered a similar problem. Tribler gets stuck on the loading screen.
Looking at the log, this was the last logged message:
```
Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!
```
[...]

PR #3498, opened 2018-03-08
Refactored Process Checker

xoriole:
Refactored Process Checker so that it does not create a lock file immediately. This fixes the problem with running Tribler with twistd possibly caused by the fix for zombie process.

Also, added a restart timer which restarts the Core if it does not restart within 30 seconds. This fixes one of the problems causing Tribler to get stuck at gear window.
> Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!

### DIFF
--- a/Tribler/Core/Modules/process_checker.py
+++ b/Tribler/Core/Modules/process_checker.py
@@ -1,4 +1,5 @@
 import os
+import psutil
 
 from Tribler.Core.SessionConfig import SessionStartupConfig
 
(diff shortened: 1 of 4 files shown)


devos50:
@xoriole nice work! Improving the process checker is something that clearly needed to be done.

However, I have a few concerns about restarting Tribler if the core does not start correctly. First of all, the timeout of 30 seconds is in my opinion quite random and not based on any performance metrics. I suggest that we should first work towards a (very basic!) performance suite for the startup times of Tribler. What if the user has a database filled with 100.000 torrents or has 100 torrent downloads running? How does this impact the time Tribler needs to load? If the startup duration of Tribler can become higher than 30 seconds within the realm of regular Tribler usage, I'm not sure whether it's a good idea to restart Tribler after a fixed time interval. This might lead to additional problems.

Second, I feel like we're not solving the problem here. Could you elaborate a bit more on the error you report (the inconsistency detected)? I would rather attempt to solve that specific issue instead of trying again to start Tribler. Maybe after that we might conclude that we don't need to restart at all 👍

xoriole:
@devos50 Yes, 30 seconds is an arbitrary number and not based on performance metrics. Personally, I don't think Tribler should take more than half a minute to just load and number of torrents in the database or number of downloads should not affect the loading time. But yes, I fully agree with developing a performance suite for the startup times so we have a better idea of it.

The problem I mentioned above about Tribler being stuck because of the inconsistency detected by the linker is caused by a [race condition](https://sourceware.org/ml/libc-alpha/2016-01/msg00220.html) between tls allocation at thread creation and dlopen. It occurs occasionally and I've also encountered it a few [times.](https://github.com/Tribler/tribler/issues/3150#issuecomment-336059704) Tribler does work after a restart in this case.

synctext:
Can Tribler optionally report a startup or shutdown error details if it takes over 30 seconds? Not crash report, but slow report. Plus kill button obviously..

qstokkink:
We could look into a Tribler `debug mode`, which toggles on `DelayedCall.debug = True` before running. That way we get verbose output of the `DelayedCall`s in the reactor and do something like this (copied from IPv8 call inspector):

```python
running_threads = []
for threadId, stack in sys._current_frames().items():
	raw_tb = traceback.extract_stack(stack)
	running_threads.append((threadId, ''.join(traceback.format_list(raw_tb))))

delayed_calls = []
for call in reactor.getDelayedCalls():
	delayed_calls.append((str(call), ''.join(call.creator)))
```

xoriole:
retest this please

xoriole:
retest this please

xoriole:
- Process Checker also checks for stale lock file now. This addresses the issue of stuck on gear window because of existing invalid lock file (#3066).
- Removed the code for restarting the core if it fails to load under 30 seconds in favor of creating startup  performance suites.

qstokkink:
http://psutil.readthedocs.io/en/latest/index.html#psutil.boot_time

PR #3512, opened 2018-03-17
Various improvements for the market

devos50:
Again, a PR with various improvements for the market.

The most important change in this PR is that I've added an order book synchronization mechanism back. It uses bloom filters to minimize the amount of traffic exchanged between market participants. I've also added a ping mechanism to discover offline matchmakers (and they are now automatically removed from the `self.matchmakers` set).

Finally, I've fixed a few bugs regarding record signing in TriblerChain.


devos50:
retest this please

devos50:
retest this please

PR #3515, opened 2018-03-20
Fixed issue with sending crash reports

xoriole:
Fixed issue with base URL on sending crash reports using TriblerRequestManager.
Added a GUI test to ensure crash reports are sent properly.


devos50:
Does this solve the issue with reports that cannot be sent that some users experience?

xoriole:
Yes, it should fix that issue

PR #3539, opened 2018-03-28
Fixed bugs in tests, updated IPv8 pointer

devos50:


Tribler/Core/CacheDB/sqlitecachedb.py
```diff
@@ -6,19 +6,16 @@
 import logging
 import os
 
-from Tribler.Core.Utilities.install_dir import get_lib_path
+import apsw
```
qstokkink: This should be at the top of the file.

Tribler/Core/Utilities/network_utils.py
```diff
@@ -30,6 +32,8 @@ def get_random_port(socket_type="all", min_port=5000, max_port=60000):
             break
         try_port += 1
 
+    CLAIMED_PORTS.append(try_port)
```
xoriole: `working_port` instead of `try_port` with a check if not None

PR #3552, opened 2018-04-02
Refactored VoD logic

### LINKED ISSUE #3546: VOD mode is broken in latest next
Video on demand doesn't work anymore in 7.1. It gives me the following error:

```
ERROR   1522665876.61  tribler_window:92    Traceback (most recent call last):
  File "/Users/martijndevos/Documents/tribler/TriblerGUI/widgets/downloadspage.py", line 165, in on_received_downloads
    self.window().video_player_page.update_with_download_info(download)
  File "/Users/martijndevos/Documents/tribler/TriblerGUI/widgets/videoplayerpage.py", line 121, in update_with_download_info
    if len(download["files"]) > 0 and not self.window().left_menu_playlist.loaded_list:
KeyError: 'files'
```

The reason for this is that we made the `files` parameter optional when performing a request to `/downloads`.

devos50:
In this PR, I refactored the VoD logic. Currently, the code that manages the VoD, is fragmented into several classes. I tried to make the logic a bit better by giving the `VideoPage` class more responsibilities. Note that this is only a minor step and more work should be done eventually.

I also added an endpoint to fetch files and the status of files in a specific download. Also, I've modified the GUI tests so the files are fetched from the fake core when starting a VoD session.

Finally, I fixed the annoying issue where a crash during the GUI tests would lead to a segmentation fault (or abort trap on MacOS). When an error occurs during the GUI tests, the test will now fail and provide us with debug output 👍 

Fixes #3546

### DIFF
--- a/Tribler/Core/Modules/restapi/downloads_endpoint.py
+++ b/Tribler/Core/Modules/restapi/downloads_endpoint.py
@@ -85,6 +85,25 @@ class DownloadBaseEndpoint(resource.Resource):
 
         return download_config, None
 
+    def get_files_info_json(self, download):
+        """
+        Return file information as JSON from a specified download.
+        """
+        files_json = []
+        files_completion = dict((name, progress) for name, progress in download.get_state().get_files_completion())
+        selected_files = download.get_selected_files()
+        file_index = 0
+        for fn, size in download.get_def().get_files_with_length():
+            files_json.append({
+                "index": file_index,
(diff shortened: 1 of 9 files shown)


devos50:
There's a zombie fake core running on the Mac Mini, which prevents the GUI tests from completing successfully. I'm unable to finish this PR since I need physical access to the machine.

Issue #3547, opened 2018-04-02
Files in download always report a progress of 0 for multi-file torrents


devos50:
When browsing through the files in a download, Tribler always reports that they are 0.0% done, even when the download itself has completed. It seems that this behavior only occurs when browsing through the progress of torrents with multiple files.

devos50:
The issue is that the file name returned by `tdef.get_files_with_length()` is different from the one returned by `state.get_files_completion()`. This causes that the progress is always 0.0 (due to this line: https://github.com/Tribler/tribler/blob/next/Tribler/Core/Modules/restapi/downloads_endpoint.py#L243).

I'm not sure how to fix this easily. @egbertbouman could you take a look at this?

PR #3557, opened 2018-04-04
Fix file progress for multi file downloads

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/3547.


devos50:
Thanks for fixing!

PR #3560, opened 2018-04-04
Added the Yappi profiler to the core and GUI

devos50:
I've added the Yappi profiler to the core and the GUI. The profiler itself can be started from the HTTP API with requests to `/debug/profiler`. I've also added an additional tab in the debug panel where the profiler can be enabled and disabled.

When the profiler is stopped, the data file is written to the `logs` directory in the state dir of Tribler. The filename contains the timestamp when the profiler was started. When shutting down Tribler while the profiler is still running, this file is also written to the disk.

### DIFF
--- a/Tribler/Core/Modules/resource_monitor.py
+++ b/Tribler/Core/Modules/resource_monitor.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
 
 import psutil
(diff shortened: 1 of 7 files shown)


devos50:
This PR is ready. The missing lines in the code coverage report are hard to cover without a system where Yappi is not installed.

Tribler/Core/Modules/resource_monitor.py
```diff
@@ -33,8 +44,52 @@ def start(self):
...
+    def stop_profiler(self):
+        """
+        Stop yappi and write the stats to the output directory.
+        Return the path of the yappi statistics file.
+        """
+        if not self.profiler_running:
+            raise RuntimeError("Profiler is not running")
+
+        if not HAS_YAPPI:
+            raise RuntimeError("Yappi is not installed")
```
xoriole: It would be nice to add more comments on this runtime error. Considering user is seeing this error, it might be a good idea to point how to install yappi and that Tribler needs to be restarted after yappi installation to enable it.

PR #3562, opened 2018-04-05
Improved stability by spawning new Tribler subprocess instead of forking the main process

devos50:
In this PR, the way the core is initialized is changed significantly. Instead of using `multiprocessing` to create a new subprocess, we now use the `subprocess` module where a new process is created instead. This already solved the annoying keyring segmentation faults we got. I suspect that this also results in more stable code execution in the core process. The fake Tribler core is now started on a separate thread during the GUI tests.

In addition, I fixed a bug where the error reporter in the GUI didn't work correctly. This was caused by incompatibilities between Tkinter and Qt which were competing on the main thread (my suspicion). I changed to code so we only use Qt now.

TODO:
- [x] make sure GUI tests still work
- [x] test changes in frozen environments
  - [x] MacOS
  - [x] Win32
  - [x] Win64
  - [x] Ubuntu

Proof that all deployment tests have passed: https://jenkins.tribler.org/job/Build_Tribler_next/238/


run_tribler.py
```diff
@@ -1,51 +1,99 @@
...
 import logging.config
 
+import signal
 from check_os import check_environment, check_free_space, error_and_exit, setup_gui_logging, \
     should_kill_other_tribler_instances
 
 
+def start_tribler_core(base_path, api_port):
+    """
+    This method is invoked by multiprocessing when the Tribler core is started and will start a Tribler session.
```
xoriole: Please update the doc string here.

PR #3575, opened 2018-04-09
Implemented basic strategy for freerider prevention

### LINKED ISSUE #3573: Free-ride policy: give users with lower token balances less priority at the exit nodes
Now that the payouts and anonymous downloading is functional, we should actually use the token balance to do something useful. The first iteration of such a policy will focus on the exit nodes. In particular, we create a _basic_ mechanism that works in the case of over- and undersupply of bandwidth where freeriders get less priority.

After a small discussion with @synctext, we came up with a design for this. We create a fixed amount of _slots_ on each exit node process. Each slot is capable of managing exactly one circuit. The current implementation restricts the maximum number of open circuits on the exit nodes to 100 but I think this is rather high. Instead, we could use 20 slots for a first version or even less?

We assume two types of slots: _competing slots_ and _equal opportunity slots_. Competing slots are reserved for users with the highest token amount. This also means that when all slots are filled and a circuit request comes in from a user with a high token balance, an existing circuit in one of the competing slots might be destroyed. Equal opportunity slots are comparable with the optimistic unchoke slots in BitTorrent and accept all incoming circuit requests.

Note that the total number of slots and the number of competing/equal opportunity slots can be easily changed.

To implement this, we need to do the following code changes:
- allow exit nodes to reject circuit requests. The initiator of a circuit should pick a new exit node when a circuit extension fails.
- next, an exit node needs to keep track of the allocated slots and circuits.
- finally, we implement the competing and equal opportunity slots.

devos50:
In this PR, I've implemented a basic freeride policy, based on the token balance of users. See issue #3573 for more information.

Depends on https://github.com/Tribler/py-ipv8/pull/99

Fixes #3573

### DIFF
--- a/Tribler/Test/Community/Market/test_community.py
+++ b/Tribler/Test/Community/Market/test_community.py
@@ -232,6 +232,7 @@ class TestMarketCommunity(TestMarketCommunityBase):
         self.add_node_to_experiment(self.create_node())
         self.nodes[3].discovery.take_step()
         yield self.deliver_messages(timeout=.5)
+        yield self.sleep(0.2)  # For processing the tick blocks
 
         self.assertTrue(self.nodes[3].overlay.order_book.get_tick(ask_order.order_id))
         self.assertTrue(self.nodes[3].overlay.order_book.get_tick(bid_order.order_id))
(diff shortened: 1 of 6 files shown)


qstokkink:
retest this please (testing the external import checker)

qstokkink:
On the failing test: it seems that `PriceLevel.remove_tick` is being called with a tick which is not part of the linked list.

PR #3588, opened 2018-04-19
WIP: Benchmarking decorators

xoriole:


qstokkink:
retest this please

qstokkink:
We probably don't want this inside of Tribler, but measured by Yappi in Gumby instead.

PR #3614, opened 2018-05-04
Integrity check is only done if there is WAL or SHM file present

xoriole:
Integrity check at the startup is done only when necessary because it is quite expensive and slows down the startup process. It seems it is sufficient to do integrity check only when there are leftover rollback files (WAL and/or SHM) present.

### DIFF
--- a/Tribler/Test/Core/test_sqlitecachedb.py
+++ b/Tribler/Test/Core/test_sqlitecachedb.py
@@ -127,6 +127,28 @@ class TestSqliteCacheDB(TriblerCoreTest):
         sqlite_test_2.execute = execute
         sqlite_test_2.initialize()
 
+    @blocking_call_on_reactor_thread
+    def test_integrity_check_triggered(self):
+        """ Tests if integrity check is triggered if temporary rollback files are present."""
+        def do_integrity_check():
+            do_integrity_check.called = True
+
+        db_path = os.path.join(self.session_base_dir, "test_db.db")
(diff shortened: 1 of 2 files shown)


qstokkink:
As there is no unit test: did you check this with a naturally created wal and shm file?

xoriole:
Yes, I confirmed with a manual test for it.

xoriole:
retest this please

Tribler/Test/Core/test_sqlitecachedb.py
```diff
@@ -127,6 +127,23 @@ def execute(sql):
         sqlite_test_2.execute = execute
         sqlite_test_2.initialize()
 
+    @blocking_call_on_reactor_thread
+    def test_integrity_check_triggerd(self):
```
qstokkink: Please change `triggerd` to `triggered`.

Tribler/Test/Core/test_sqlitecachedb.py
```diff
@@ -127,6 +127,23 @@ def execute(sql):
...
+        def do_integrity_check(expected=False):
+            self.assertTrue(expected, "Integrity check was not expected")
+
```
qstokkink: Please set `do_integrity_check.called = False` here and then set `do_integrity_check.called = True` inside `def do_integrity_check()`. Then you can do an assert on whether it was called or not externally.

PR #3630, opened 2018-05-15
Use IPv8 endpoint, then packets data to Dispersy

ichorid:
The long-awaited pull request, that enables the new Twisted-based IPv8 UDP endpoint. Dispersy is now merely an `EndpointListener` being served by IPv8.
This required changing initialization order to IPv8 first, Dispersy second, and  `_use_main_thread=False` compatibility workaround for unit tests using `deliver_messages` method.
In addition, the patch removes explicit teardown command for IPv8 endpoint `self.ipv8.endpoint.close()`, since it is now handled by IPv8 itself.


qstokkink:
@xoriole do you want to review this too?

qstokkink:
For formality I will also mention that you can use the `StandaloneEndpoint` instead of the `MIMEndpoint` as it is slightly more efficient. I don't really care that much though.

PR #3633, opened 2018-05-18
Add bootstrap override option

ichorid:
This pull request removes `use_testnet` option and replaces it with more general `bootstrap_override` option, that allows to, well, override bootstrap servers list. It is very useful to build your own isolated testnets without hacking the code.
The corresponding new option is added to `tunnel_helper` and Tribler Twisted plugins (it seems not to have any effect for the latter, probably because of Dispersy endpoint interference).
Anyways, support for this option is required to proceed with building Jenkins performance tests.

p.s.
as a small bonus, there is a small change to `.gitignore` to ignore log file generated by `nosetests`.

p.p.s
monstrous regexp for checking IP addresses is a shameless copypaste from a popular Stackexchange answer.

### DIFF
--- a/twisted/plugins/tunnel_helper_plugin.py
+++ b/twisted/plugins/tunnel_helper_plugin.py
@@ -6,6 +6,8 @@ import os
 import random
 import signal
 import threading
+import re
+
 
 from twisted.application.service import MultiService, IServiceMaker
 from twisted.conch import manhole_tap
@@ -17,12 +19,12 @@ from twisted.protocols.basic import LineReceiver
 from twisted.python import usage
 from twisted.python.log import msg
 from zope.interface import implements
+from socket import inet_aton
 
 from Tribler.Core.Config.tribler_config import TriblerConfig
 from Tribler.Core.Session import Session
 from Tribler.Core.simpledefs import NTFY_TUNNEL, NTFY_REMOVE
 from Tribler.dispersy.tool.clean_observers import clean_twisted_observers
-
 
 # Register yappi profiler
 from Tribler.dispersy.utils import twistd_yappi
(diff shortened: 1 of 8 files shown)


qstokkink:
I'm guessing this may screw with @devos50 's experiments.

twisted/plugins/tunnel_helper_plugin.py
```diff
@@ -23,10 +24,48 @@
 from Tribler.Core.simpledefs import NTFY_TUNNEL, NTFY_REMOVE
 from Tribler.dispersy.tool.clean_observers import clean_twisted_observers
 
-
 # Register yappi profiler
 from Tribler.dispersy.utils import twistd_yappi
 
+def is_valid_ipv4(ip):
```
qstokkink: You can steal a shorter version from Dispersy:

https://github.com/Tribler/dispersy/blob/7a2893bb4110da79ec7300652aa7baa4afb40a91/util.py#L247

ichorid: I don't want to add another dependency on Dispersy. And I am not even sure that this check should be there. We got no policy on this kind of checks.

ichorid: Ok, changed that to use "inet_aton".

ichorid:
BTW, @devos50  approved this change.

ichorid:
There was an error in check for windows64. I guess it's a bug in the check, some kind of race condition. I filed it under #3634

qstokkink:
retest this please

devos50:
@ichorid can this be merged?

ichorid:
Oops, put "Thanks everyone" message into pull request description...
Could someone with admin rights fix the message to "Add bootstrap override option" ?

qstokkink:
@ichorid it's fine: no one reads the additional message anyway.

ichorid:
@qstokkink, it's in the commit message on Git!
Sure, it demonstrates the nice atmosphere inside our team, but it's not very useful for those who read the Git log...

PR #3658, opened 2018-06-01
Re-adding plain download when safe seeding is enabled

### LINKED ISSUE #3644: Encrypted anonymous seeding using proxies option checked, does not work.
Everry version since first 7.1 build
WIndows 10 build  17672
In settings, general, check this option. Download a Torrent not anonymous and with 7.0 it would start seeding anonymous, with 7.1 it does not.
When finished downloading it would check and seed anonymous automatically.
##### Actual behavior: #####
When download is done, Torrent starts seeding none anonymous.
!!! When right click on Torrent in Library, just selection menu should come up, it should NOT act as if a left click was also done!!!! Test on left or right mousebutton click and do not act on mousebutton down event.

devos50:
Fixed #3644

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -671,6 +671,10 @@ class TriblerLaunchMany(TaskManager):
                 if safename in self.previous_active_downloads:
                     self.session.notifier.notify(NTFY_TORRENT, NTFY_FINISHED, tdef.get_infohash(), safename)
                     do_checkpoint = True
+                elif download.get_hops() == 0 and download.get_safe_seeding():
+                    # Re-add the download with anonymity enabled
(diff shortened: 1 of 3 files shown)


Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -671,6 +671,10 @@ def sesscb_states_callback(self, states_list):
                 if safename in self.previous_active_downloads:
                     self.session.notifier.notify(NTFY_TORRENT, NTFY_FINISHED, tdef.get_infohash(), safename)
                     do_checkpoint = True
+                elif download.get_hops() == 0 and download.get_safe_seeding():
+                    # Re-add the download with anonymity enabled
+                    hops = self.session.config.get_default_number_hops()
```
qstokkink: What if get_default_number_hops is `0`?

PR #3661, opened 2018-06-04
Codereview

ichorid:
This pull request starts the "codereview" branch that will contain (non-intrusive) review of Tribler codebase.

For each file (package) reviewed, `CODE REVIEW header is added with general description of what the file (package) does, and on the general condition of code.
`OBJECTION` headers mark most obvious problems that should be acted upon.

In addition, as an example (and kind of mischief), I deleted some unused functions in Tribler.

No code-changing work should be done on this branch. Its only for comments.


qstokkink:
It would be great to have the findings of this PR collected in one or multiple issues. Based on discussion and apparent use of this PR and the corresponding branch, it does not make sense for us to maintain a code review branch. Therefore I will close this PR.

PR #3671, opened 2018-06-12
Sorted channel torrents and added health check button

xoriole:
Channel torrents are sorted based on the liveness score. 
The score is computed (@qstokkink's suggestion) as:
```python 
liveness_score = num_seeders + 0.5 * num_leechers 
```

If the health info for a torrent is unknown the then the seeder and leecher score is both set to 0.5 keeping the overall score in the range of 0 to 1 such that score will always be higher than that of dead torrents and just lower than any barely alive torrent. With this good torrents appear at top then unknown torrents and dead torrents at the end.

Additionally added a button in torrent details tab to force health check.

### DIFF
--- a/TriblerGUI/widgets/channelpage.py
+++ b/TriblerGUI/widgets/channelpage.py
@@ -72,7 +72,20 @@ class ChannelPage(QWidget):
             self.window().channel_torrents_list.set_data_items(self.playlists + self.torrents)
 
     def received_torrents_in_channel(self, results):
-        for result in results['torrents']:
+        def sort_key(torrent):
+            """ Scoring algorithm for sorting the torrent to show liveness. The score is basically the sum of number
+                of seeders and leechers. If swarm info is unknown, we give unknown seeder and leecher as 0.5 & 0.4 so
+                that the sum is less than 1 and higher than zero. This means unknown torrents will have higher score
+                than dead torrent with no seeders and leechers and lower score than any barely alive torrent with a
+                single seeder or leecher.
+            """
+            seeder_score = torrent['num_seeders'] if torrent['num_seeders'] or torrent['last_tracker_check'] > 0\
+                else 0.5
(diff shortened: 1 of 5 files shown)


TriblerGUI/widgets/channelpage.py
```diff
@@ -72,7 +72,20 @@ def update_result_list(self):
...
+                of seeders and leechers. If swarm info is unknown, we give unknown seeder and leecher as 0.5 & 0.4 so
+                that the sum is less than 1 and higher than zero. This means unknown torrents will have higher score
+                than dead torrent with no seeders and leechers and lower score than any barely alive torrent with a
+                single seeder or leecher.
+            """
+            seeder_score = torrent['num_seeders'] if torrent['num_seeders'] or torrent['last_tracker_check'] > 0\
+                else 0.5
+            leecher_score = torrent['num_leechers'] if torrent['num_leechers'] or torrent['last_tracker_check'] > 0\
+                else 0.4
+            return seeder_score + leecher_score
```
qstokkink: Could you turn this into `seeder_score + .5 * leecher_score`: that is more in line with the expected download speed (based on homogenous bandwidth and uniformly distributed leecher file completion)?

TriblerGUI/widgets/torrentdetailstabwidget.py
```diff
@@ -90,3 +104,47 @@ def update_with_torrent(self, torrent_info):
...
         self.request_mgr.perform_request("torrents/%s" % self.torrent_info["infohash"], self.on_torrent_info)
+
+    def on_check_health_clicked(self, timeout=15):
+        if self.is_health_checking and (time.time() - self.last_health_check_ts < timeout):
+            return
+
+        self.is_health_checking = True
+        self.last_health_check_ts = time.time()
+        self.health_request_mgr = TriblerRequestManager()
+        self.health_request_mgr.perform_request("torrents/%s/health?timeout=15&refresh=%d" %
```
devos50: Is it right that this timeout is hardcoded?

xoriole:
restest this please

PR #3697, opened 2018-06-29
debian: return to previous dir after running the tribler script

FFY00:
RIght now, after running tribler via terminal, we stay in `/usr/share/tribler`. Depending on your environment you may not experience this issue. I changed the script to use `pushd` and `popd` instead of `cd`.


qstokkink:
Ok to test

devos50:
@FFY00 thank you for your contribution!

PR #3705, opened 2018-07-01
READY: Modernize Python 2 code to get ready for Python 3

cclauss:
$ __futurize --stage1 -w .__  # http://python-future.org/futurize_cheatsheet.html

Minimal, safe changes required to make this code syntax compatible with both Python 2 and Python 3. More changes will be required to complete the port to Python 3 but this should provide a solid start in that process without breaking backward compatibility with legacy Python.


qstokkink:
Ok to test

qstokkink:
It seems you are trying to do a relative import of `libtorrent` and `plyveladapter`, which are external libraries.

devos50:
Please see the comment in the other PR #3851

PR #3709, opened 2018-07-02
Various improvements for the market community

devos50:
In this PR, I made some improvements to the market community:
- changed the implementation of the matching engine from a recursive, inefficient matching algorithm, to a significantly more efficient, iterative one
- merged order books together (an ask to sell BTC for EUR is essentially the same as a bid to buy EUR for BTC)
- using the `lazy_wrapper` decorator instead of manual parsing an IPv8 packet, this also removes the global time from each payload
- removed the message number from each packet sent in the market (should improve efficiency)
- disabled the periodic reputation calculation
- merged the `Price` and `Quantity` objects together into a base class: `AssetAmount`
- added a version number to each TrustChain block which contains market information
- improved security of the market by adding numerous checks for block validity
- converted from `Price`/`Quantity` pairs to `AssetPair` objects
- improved the GUI tests so they cover more of the market screens (transactions/orders/ticks etc)

Functional experiment that shows correctness of these changes: https://jenkins.tribler.org/job/market_experiment_validation/19/


devos50:
Although this PR is very close to being ready, I discovered a fundamental flaw in the mechanism to convert asset prices/quantities from/to integer values. I need to come up with a solution for this first.

devos50:
I thought a bit about the market and how to represent prices/quantities as integers. In this PR, I changed all floating point representations of the price/quantity of assets to integers. In line with most decentralized exchanges that allow users to create their own assets adopt, each asset has a precision, which defines the number of decimals this asset has. For Euro and most other fiat currency, the precision would be 2 since 0.01 is the minimum unit for this asset. For Bitcoin, the precision would be 8 since 1 Satoshi = 0.00000001 BTC. The market now only works with the smallest unit of assets, which is always an integer (when considering euros, the market only works with amount of cents and for Bitcoin, we only express it in Satoshi).

When creating an order in the market, one has to specify a price (i.e. a price of 2 BTC/MB means that I pay 2 BTC to get 1 MB). However, I didn’t immediately realize that this price can also be a floating point. Since the market does not support floating point prices anymore, this results in an error and in some cases, users are unable to post orders in the market.

So how can we fix this? My solution would be the following, which basically is a mix of ideas implemented by Bitshares and Waves. Instead of storing a price and quantity (and the asset types of the price/quantity) in each order, we store a pair of assets and their amounts. Note that these asset amounts should be integer values and never floating point values. The JSON that represents an order, now looks something like this:

```
{
    "is_ask": True,
    "assets": {
        "asset1_type": BTC,
        "asset1_amount": 100000000,
        "asset2_type": MB,
        "asset2_amount": 10
    }
    "trader_id": ...,
    ...
}
```

This order would sell `asset1` (BTC) for `asset2` (bandwidth tokens). Since a buy order for BTC against MB is the same as a sell order for MB against BTC, we add an additional constraint: `asset1_type` < `asset2_type`. We can now dynamically compute a price from each order.

This design gets rid of all floating point representations. However, it requires some time and major changes to the protocol to convert the market to this design.

devos50:
This PR is ready for review. I will add additional tests tomorrow and do some extensive manual testing to see whether everything is working. If not, I don't expect too big changes to the code base.

Tribler/community/market/community.py
```diff
@@ -1343,7 +1359,7 @@ def received_order_status(self, _, payload):
             "traded_quantity": payload.traded_quantity.amount,
             "timeout": float(payload.timeout),
             "timestamp": float(payload.timestamp),
-            "ip": payload.address.ip,
+            "address": payload.address.ip,
```
qstokkink: IP addresses (+ports) are not stable. Your own address is just an estimation of whatever the interface address of your most connected-to NAT box is at that time. This especially holds for port stability, which may be reassigned as needed.

Even worse would be if you were using this on a mobile device and, god forbid, performing some form of physical movement.

Apparently this was already in the code, but I just noticed 😃

Tribler/Core/Modules/restapi/market/asks_bids_endpoint.py
```diff
@@ -1,5 +1,7 @@
 import json
```
qstokkink: Please `import json_util as json` here so we can catch the nasty `UnicodeDecodeError`s

Tribler/community/market/community.py
devos50: We discussed this 👍

PR #3720, opened 2018-07-09
Make Notifier identify torrents by infohash

ichorid:
Makes Notifier identify torrent objects by the infohash instead of the title. It is much safer this way.


qstokkink:
@ichorid could you also get rid of the `seeding_download_list` in that method?

PR #3733, opened 2018-07-12
WIP: Hide download details on deselect

qstokkink:
This PR hides the download details tab upon deselecting all downloads.


Nubro01:
Going into the 5th year since this request and still nothing has been done!
How difficult is it to create a flag and test that flag and enable or disable this?

Issue #3759, opened 2018-07-23
Lots of Unnamed Torrents in channels


qstokkink:
There are a lot of unnamed torrents appearing in channels.

![afbeelding](https://user-images.githubusercontent.com/3630389/43084377-da3db116-8e98-11e8-812d-a743f3ac666d.png)

qstokkink:
Actually these seem to be corrupt in the `Torrent` database:

![afbeelding](https://user-images.githubusercontent.com/3630389/43187570-0994d656-8ff2-11e8-849d-0110e261821b.png)

qstokkink:
I printed out all the torrent updates:

```python
INFO    1532509355.25                 community:110  (PopularityCommunity)  Got torrent info response from ('XXX.XXX.XXX.XXX', 24773)
CRITICAL 1532509355.25      SqliteCacheDBHandler:501  (TorrentDBHandler)  Updating 7I0Dpi3LSw+W8O7bBcXOAqU6P0Q=: {'comment': '', 'num_files': 0, 'length': 0, 'name': '', 'creation_date': 0}
```

This might be related.

qstokkink:
So, clicking download actually shows files in one of these unnamed torrents:

![afbeelding](https://user-images.githubusercontent.com/3630389/43193026-3fc63712-8fff-11e8-91d9-a159c03371b9.png)

But when it starts downloading, all the info is gone:

![afbeelding](https://user-images.githubusercontent.com/3630389/43193053-4c5918aa-8fff-11e8-894f-692ce0bbefab.png)

![afbeelding](https://user-images.githubusercontent.com/3630389/43193061-52d052fc-8fff-11e8-96d0-4bfa6072a17f.png)

PR #3766, opened 2018-07-30
Update unnamed torrents on metainfo check

xoriole:
Unnamed torrents are now updated on metainfo check.
Related #3759

### DIFF
--- a/Tribler/Core/CacheDB/SqliteCacheDBHandler.py
+++ b/Tribler/Core/CacheDB/SqliteCacheDBHandler.py
@@ -515,6 +515,23 @@ class TorrentDBHandler(BasicDBHandler):
         if notify:
             self.notifier.notify(NTFY_TORRENTS, NTFY_UPDATE, infohash)
 
+    def update_torrent_with_metainfo(self, infohash, metainfo):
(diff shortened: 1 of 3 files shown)


Tribler/Core/CacheDB/SqliteCacheDBHandler.py
```diff
@@ -515,6 +515,20 @@ def updateTorrent(self, infohash, notify=True, **kw):  # watch the schema of dat
         if notify:
             self.notifier.notify(NTFY_TORRENTS, NTFY_UPDATE, infohash)
 
+    def update_torrent_with_metainfo(self, infohash, metainfo):
+        """ Updates name, length and num files from metainfo. """
```
qstokkink: Shouldn't you try to load this from the database (if it exists)? You run the risk of overwriting known data now.

xoriole: Check from the database is done by the caller (torrentinfo_endpoint.py:69). But I agree with your comment.

qstokkink: Yeah, it would be cleaner if the `torrentinfo_endpoint.py` does not concern itself with the existing database state. That should be in this function

PR #3776, opened 2018-08-09
Performing direct payouts to other Tribler clients

### LINKED ISSUE #3718: Direct payouts between Tribler clients when doing plain seeding
Currently, we payout bandwidth tokens to relays and exit nodes, to compensate them for their provided services. This resulted in some confusion since 'normal' seeding, without using proxies, is not being rewarded by tokens. Rewarding plain seeding (to other Tribler clients), will reinforce the token economy and provide more incentive to seed.

We should explore whether it is possible to do so. For instance, we might use the libtorrent user agent to check whether another user is running Tribler. If so, we keep track of the amount of uploaded bytes to this user and periodically construct a block to pay this other user. This of course assumes that we know the public (Dispersy/IPv8) port and public key of our counterparts. This information should be exchanged between clients somehow.

devos50:
This PR implements direct payouts between Tribler instances (zero-hop downloads).

TODO:
- [x] write unit tests
- [x] test the implementation with 100+ Tribler instances and varying download sizes (done: https://jenkins-ci.tribler.org/job/pers/job/direct_download_devos50/96/)

Fixes #3718

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -15,6 +15,7 @@ from traceback import print_exc
 from Tribler.Core.CacheDB.sqlitecachedb import forceDBThread
 from Tribler.Core.DecentralizedTracking.dht_provider import MainlineDHTProvider
 from Tribler.Core.DownloadConfig import DownloadStartupConfig, DefaultDownloadStartupConfig
+from Tribler.Core.Modules.payout_manager import PayoutManager
 from Tribler.Core.Modules.resource_monitor import ResourceMonitor
 from Tribler.Core.Modules.search_manager import SearchManager
 from Tribler.Core.Modules.versioncheck_manager import VersionCheckManager
@@ -109,6 +110,7 @@ class TriblerLaunchMany(TaskManager):
         self.credit_mining_manager = None
         self.market_community = None
         self.dht_community = None
+        self.payout_manager = None
 
     def register(self, session, session_lock):
         assert isInIOThread()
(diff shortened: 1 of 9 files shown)


devos50:
While I haven't finished the big experiment yet, I think this PR is ready for review 👍

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -709,6 +714,13 @@ def sesscb_states_callback(self, states_list):
                     hops = self.session.config.get_default_number_hops()
                     self.update_download_hops(download, hops)
 
+            # Check the peers of this download every five seconds and add them to the payout manager when
+            # this peer runs a Tribler instance
+            if self.state_cb_count % 5 == 0 and download.get_hops() == 0 and self.payout_manager:
+                for peer in download.get_peerlist():
+                    if 'Tribler' in peer["extended_version"]:
```
qstokkink: Can you do a `.startswith()` here? It is more efficient.

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -709,6 +714,13 @@ def sesscb_states_callback(self, states_list):
...
+                        self.payout_manager.update_peer(peer["id"].decode('hex'), infohash, peer["dtotal"])
```
qstokkink: Also, there might be garbage in the peer id: please do some input validation here/handle the exceptions from invalid hex encoding.

devos50: The peer ID is already encoded in hex format: https://github.com/Tribler/tribler/blob/next/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py#L852

qstokkink: Alright 👍

devos50:
Status update: it seems that the direct payouts is working correctly, however, I think there is a bug in the DHTDiscoveryCommunity since peers are removed after around one minute, even though they are pinged.

Functional experiment (where I removed the (buggy?) code that deletes peers from the DHT storage): https://jenkins-ci.tribler.org/job/pers/job/direct_download_devos50/96/

Final interaction graph (not all peers are downloading something):
trustchain_interactions

PR #3813, opened 2018-08-24
Integrated IPv8 REST API in Tribler

devos50:
Requires https://github.com/Tribler/fake-tribler-api/pull/28


devos50:
retest this please

Issue #3805, opened 2018-08-20
Too much checking torrent.


Aldo-f:
#### Please, fill all relevant items: ####
- [x] I have read CONTRIBUTING.rst
- [x] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
7.0.2
7.1.0-beta

##### Operating system and version: #####
Windows 10 home
v 1803

##### Steps to reproduce the behavior: #####
Changing anonymity from none to 1 hop.
Changing anonymity from 1 hop to none.
(on D-disk; for tiny and huge torrents; Tribler installed on D)

##### Expected behavior: #####
After building circuits starting to upload/download.

##### Actual behavior: #####
First checks, (then building circuits,) then seeds.
Doesn't need to check, the torrent doesn't change. 

##### Relevant log file output: #####
Didn't find relevant log files on %APPDATA%\.Tribler\logs

##### Also: ##### 
Can't seem to successfully shut down Tribler, it always asks me to force shutdown.
And even that takes more than five minutes, so i need to use task manager to shut it down.
(not a problem in 7.1.0-beta (on C-disk))

devos50:
Thank you for your report. When changing anonymity, Tribler internally removes the torrent from libtorrent and adds it again, with different parameters.

It should not be hard to preserve the state of a download and inject it when re-adding the download to libtorrent. We will give some priority to this issue. @egbertbouman do you want to take this one?

egbertbouman:
@devos50 Sure, I'll take a look. It should just be a matter of reusing the resume data. This will also prevent the download/upload counters from resetting.

arvidn:
I don't really know what Tribler does with the torrent when removing it and re-adding it, but chances are that it could be changed in-place (unless you're actually changing the torrent itself, causing the info-hash to change).

The way to prevent the re-checking is to save resume data for the torrent. This is an asynchronous operation and will return the resume data in an alert. In the current stable version of libtorrent you get back a bencoded buffer than can then be passed back in via the `add_torrent_params` object.

In master (the next major release of libtorrent), this has been simplified a bit by separating the serialisation of the resume data from returning it. There, the resume data is all stored in a structured form in the `add_torrent_params` object directly, and it can then be serialised/deserialised to its bencoded form independently.

Issue #3199, opened 2017-11-09
Change hops while force-rechecking/changing hops clears ratio


qstokkink:
If you change the anon-hops while performing a torrent recheck, the ratio is cleared.

qstokkink:
From https://www.libtorrent.org/reference-Core.html#force-recheck:

> force_recheck puts the torrent back in a state where it assumes to have no resume data

I guess the resume data gets wiped by design.

Issue #3173, opened 2017-10-22
Changing Anonymous hop counter, triggers a check of the whole file.


Nubro01:
- [ ] I have tried with the latest pre-release version and I still can reproduce the issue.
Yes
##### Tribler version/branch+revision: #####
Tribler 7
##### Operating system and version: #####
Windows 10 Insider 17017
##### Steps to reproduce the behavior: #####
Right click an entry with the Library and select a different number of hops
##### Expected behavior: #####
Hop counter is changed and files are not checked.
##### Actual behavior: #####
Hop counter is changed and file is checked again.

qstokkink:
Dev. note: this can probably be achieved by modifying the LibTorrent checkpoint file directly to share it between LibTorrent sessions.

PR #3825, opened 2018-08-27
Reuse pstate when changing hop count

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/3805, https://github.com/Tribler/tribler/issues/3173,  https://github.com/Tribler/tribler/issues/3199.
Also fixes LibtorrentDownloadImpl.setup not correctly handling the pstate argument.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -593,6 +593,8 @@ class TriblerLaunchMany(TaskManager):
         """
         infohash = binascii.hexlify(download.tdef.get_infohash())
         self._logger.info("Updating the amount of hops of download %s", infohash)
+        pstate = download.get_persistent_download_config()
+        pstate.set('state', 'engineresumedata', (yield download.save_resume_data()))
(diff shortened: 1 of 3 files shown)


Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -593,6 +593,8 @@ def update_download_hops(self, download, new_hops):
         """
         infohash = binascii.hexlify(download.tdef.get_infohash())
         self._logger.info("Updating the amount of hops of download %s", infohash)
+        pstate = download.get_persistent_download_config()
+        pstate.set('state', 'engineresumedata', (yield download.save_resume_data()))
```
devos50: What if the torrent is in the state where it's still fetching metadata? Does the `save_resume_data` fire with `None` then?

egbertbouman: Yep, it returns `None`. In this case libtorrent will give us a `save_resume_data_failed_alert`, which will trigger the errback being called, but since the `Failure` is then trapped in `_on_resume_err`, we end up with `None`.

PR #3830, opened 2018-08-28
Fixed bug on adding torrent from HTTP URL that resolves to a magnet link

### LINKED ISSUE #3828: Twisted unsupported scheme: ''
From reporter - Tribler 7.1.0-beta (Darwin-17.7.0-x86_64-i386-64bit) [#15833](http://reporter.tribler.org/stacktrace/15833)
```python

File "TriblerGUI/event_request_manager.py", line 127, in on_read_data
RuntimeError: [Failure instance: Traceback: <class 'twisted.web.error.SchemeNotSupported'>: Unsupported scheme: ''
twisted/internet/defer.py:651:_runCallbacks
twisted/web/client.py:2044:_handleResponse
twisted/web/client.py:2026:_handleRedirect
twisted/web/client.py:1795:request
twisted/web/client.py:1596:request
twisted/web/client.py:1580:_getEndpoint
twisted/web/client.py:1460:endpointForURI
]
```

xoriole:
Fixes #3828 

With previous PR #3795 Twisted agent now supports HTTPS so we no longer need treq.
Therefore, I'm removing this dependency.

### DIFF
--- a/Tribler/Core/Modules/restapi/torrentinfo_endpoint.py
+++ b/Tribler/Core/Modules/restapi/torrentinfo_endpoint.py
@@ -106,8 +106,14 @@ class TorrentInfoEndpoint(resource.Resource):
                 request.setResponseCode(http.INTERNAL_SERVER_ERROR)
                 return json.dumps({"error": "error while decoding torrent file"})
         elif uri.startswith('http'):
-            def _on_loaded(metadata):
-                metainfo_deferred.callback(bdecode(metadata))
+            def _on_loaded(response):
+                if response.startswith('magnet'):
+                    _, infohash, _ = parse_magnetlink(response)
+                    if infohash:
+                        self.session.lm.ltmgr.get_metainfo(response, callback=metainfo_deferred.callback, timeout=20,
+                                                           timeout_callback=on_metainfo_timeout, notify=True)
+                        return
+                metainfo_deferred.callback(bdecode(response))
             http_get(uri.encode('utf-8')).addCallback(_on_loaded).addErrback(on_lookup_error)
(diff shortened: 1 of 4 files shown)


Tribler/Core/Modules/restapi/torrentinfo_endpoint.py
```diff
@@ -106,8 +106,12 @@ def on_lookup_error(failure):
                 request.setResponseCode(http.INTERNAL_SERVER_ERROR)
                 return json.dumps({"error": "error while decoding torrent file"})
         elif uri.startswith('http'):
-            def _on_loaded(metadata):
-                metainfo_deferred.callback(bdecode(metadata))
+            def _on_loaded(response):
+                if response.startswith('magnet:?xt=urn:btih:'):
```
devos50: You should be careful with this. It could be that the magnet link does not start with the `btih` element but instead the name. I would rather have something like:

```
if response.startswith('magnet:'):
```

Tribler/Core/Utilities/utilities.py
```diff
@@ -356,16 +356,22 @@ def http_get(uri):
     def _on_response(response):
         if response.code == http.OK:
             return readBody(response)
+        if response.code == http.FOUND:
+            # Check if location header contains magnet link
+            location_headers = response.headers.getRawHeaders("location")
+            if not location_headers:
+                return fail()
+            new_uri = location_headers[0]
+            if new_uri.startswith('magnet:?xt=urn:btih:'):
```
devos50: See my comment above

devos50:
@xoriole could you rebase it first?

PR #3836, opened 2018-08-30
Replaced Electrum with bitcoinlib

### LINKED ISSUE #3831: Explore libbitcoin as an alternative to Electrum
Currently, we use Electrum to provide the functionality to trade Bitcoin on our decentralized market. However, this library is suffering from an increasing number of issues when being integrated with Tribler. The main issue is that Electrum has moved to Python 3 and deprecated the 'old' Python 2.7 implementation, which will likely suffer from vulnerabilities which are not fixed. Besides that, the Electrum library adds many dependencies to the Tribler code base.

As an alternative, I found [libbitcoin](https://github.com/1200wd/bitcoinlib). A quick investigation shows that it supports the same features as Electrum (creating wallets, managing balance and fetching fee estimations). It also contains support for Litecoin/Dash wallets, supports Python 2.7 and has a nice code base. It is, however, a less popular alternative compared to Electrum.

Goal: investigate whether libbitcoin can replace Electrum

devos50:
In this PR, I replaced the Electrum library with bitcoinlib. Note that bitcoinlib is optional, however, we should ship it with the Tribler client on Windows and Mac. On Linux, it is optional and users can install the wallet themselves using pip (users will get a warning if the library is not installed when trying to create a Bitcoin wallet).

Another advantage of bitcoinlib is that it also supports Litecoin and Dash. Support for these wallets will be added in a later PR.

Fixes #3831

### DIFF
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -207,13 +207,13 @@ class AbstractServer(BaseTestCase):
                 f = open(filename, 'a')
             else:
                 f = open(filename, 'w')
-                print >> f, "annotation start end"
+                f.write("annotation start end\n")
 
             AbstractServer._annotate_counter += 1
             _annotation = re.sub('[^a-zA-Z0-9_]', '_', annotation)
             _annotation = u"%d_" % AbstractServer._annotate_counter + _annotation
 
-            print >> f, _annotation, self.annotate_dict[annotation], time.time()
+            f.write("%s %s %s\n" % (_annotation, self.annotate_dict[annotation], time.time()))
             f.close()
 
     def get_bucket_range_port(self):


xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

Tribler/Core/Modules/wallet/btc_wallet.py
```diff
@@ -213,31 +64,10 @@ def get_balance(self):
...
-                server = self.get_daemon().get_server(config)
-                result = server.run_cmdline(options)
-
-                if not result[0]:  # Transaction failed
-                    return fail(RuntimeError(result[1]))
-
-                return succeed(str(result[1]))
+                self._logger.info("Creating Bitcoin payment with amount %f to address %s", amount, address)
+                tx = self.wallet.send_to(address, int(amount))
+                return str(tx.hash)
```
xoriole: Should this be ` return succeed(str(tx.hash))`?

devos50:
retest this please

devos50:
Ready for review. The uncovered lines are quite hard to cover with a test :/

PR #3841, opened 2018-08-31
Fixed tray icon already deleted bug

### LINKED ISSUE #3840: RuntimeError: wrapped C/C++ object of type QSystemTrayIcon has been deleted
From reporter - Tribler 7.1.0-beta [#15912](http://reporter.tribler.org/stacktrace/15912)
Linux-4.14.66-1-MANJARO-x86_64-with-glibc2.2.5
Python 2.7.15 (default, Jun 27 2018, 13:05:28) 

```python
File "/usr/share/tribler/TriblerGUI/tribler_window.py", line 306, in on_torrent_finished
"Download of %s has finished." % torrent_info["name"])
RuntimeError: wrapped C/C++ object of type QSystemTrayIcon has been deleted
```

xoriole:
Fixed tray icon already deleted bug.

Fixes #3840

### DIFF
--- a/TriblerGUI/tribler_window.py
+++ b/TriblerGUI/tribler_window.py
@@ -1,6 +1,7 @@
 import glob
 import logging
 import os
+import sip
 import sys
 import traceback
 from urllib import pathname2url, unquote
@@ -301,7 +302,7 @@ class TriblerWindow(QMainWindow):
         close_dialog.show()
 
     def on_torrent_finished(self, torrent_info):
-        if self.tray_icon:
+        if self.tray_icon and not sip.isdeleted(self.tray_icon):
             self.window().tray_icon.showMessage("Download finished",
                                                 "Download of %s has finished." % torrent_info["name"])
 


TriblerGUI/tribler_window.py
```diff
@@ -301,7 +302,7 @@ def on_low_storage(self):
         close_dialog.show()
 
     def on_torrent_finished(self, torrent_info):
-        if self.tray_icon:
+        if self.tray_icon and not sip.isdeleted(self.tray_icon):
```
qstokkink: This won't work. We've had this before. As the deletion of objects in Qt is threaded, in between this check and the showMessage, the tray icon can be removed. Instead, you should catch a `RuntimeError` every time you interface with the tray icon. We do this in other places as well.

qstokkink: And.. it's merged.

devos50: Well, I thought it solved the problem. I will revert it and fix it properly then I guess.

qstokkink: See this:

https://github.com/Tribler/tribler/blob/4cec53cced28608efb5884ab9d02556c7234b6b7/TriblerGUI/widgets/downloadspage.py#L86

qstokkink: I just did a search, the market community is apparently also interfacing with the `tray_icon` directly. This should also be changed.

PR #3844, opened 2018-08-31
Properly fixed tray icon + small market fix

### LINKED ISSUE #3843: tray_icon should not be called directly
The tray icon should never be used directly, its removal is asynchronous (in C, so actually asynchronous). This means that at any point in the code a `RuntimeError` can be thrown.

We should make sure that all occurrences of interfacing with the tray icon are properly wrapped in a `try .. except` block (https://github.com/Tribler/tribler/search?q=tray_icon&unscoped_q=tray_icon).

Related to https://github.com/Tribler/tribler/pull/3841/files#r214415651

devos50:
The calls for showing notifications/removing the tray icon is now wrapped in a try/except block.

Next to fixing the tray icon in the GUI, I'm also running the market tests on the reactor thread now. They consistently pass on Linux now (I've tested and verified this on bbq).

Fixes #3843

### DIFF
--- a/Tribler/Test/Community/Market/test_community.py
+++ b/Tribler/Test/Community/Market/test_community.py
@@ -24,7 +24,7 @@ class TestMarketCommunityBase(TestBase):
         super(TestMarketCommunityBase, self).setUp()
         self.initialize(MarketCommunity, self.NUM_NODES)
         for node in self.nodes:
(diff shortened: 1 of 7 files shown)


devos50:
retest this please

devos50:
retest this please

PR #3851, opened 2018-09-04
print() is a function in Python 3 in check_os.py

cclauss:


devos50:
Thank you for your contribution. Instead of making small changes across the code base, we prefer a more controlled approach when migrating the code to Python 3. As a first step, we want to get the unit tests functional under Python 3, which requires us to make sure everything can be imported correctly first. Next, we have to remove submodules we are not going to port to Python 3 (Dispersy and pymdht). Please also see #3883.

If you want to help us with this effort and come up with a roadmap, please reach out to us on IRC. We always value contributions 👍

cclauss:
This PR followed the __futurize__ plan which recommends whacking the syntax errors first so that linting tools can recognize all the remaining issues which usually surface as changed imports and __undefined names__ like basename, cmp, raw_input, reload, xrange, unicode.  Once those are fixed then clean up what remains with special focus on bytes vs. strings.

PR #3870, opened 2018-09-11
Fixed a few bugs in the market, updated IPv8 pointer

devos50:
In this PR, I fixed three bugs in the market:
- fixed a bug in the matching engine where two orders of the same trader could be matched
- fixed order cancellation test
- changed logic for simultaneous order creation. I made the tests more challenging by removing the `yield` statement before `create_ask`/`create_bid`, which always errors out without the new logic.

I also updated the IPv8 pointer, added a configuration option to toggle the IPv8 statistics and mocked the `balance` method of a BTC wallet during tests (so we don't make requests to external websites).


qstokkink:
afbeelding

PR #3872, opened 2018-09-14
Fix having two directories for Channel tests (C/channel)

ichorid:
The directories had the same name except for the 1st character.
It worked on Windows because filesystem semantics there is case-insensitive.


qstokkink:
Could you make your commit messages a bit more.. descriptive?

PR #3873, opened 2018-09-14
Fix dummy process unable to terminate in test_process_checker

ichorid:
The dummy process does not react to `terminate` and only dies with the interpreter. This patch adds a correct, `Value`-based shared var mechanism to shut it down.

### DIFF
--- a/Tribler/Test/Core/Modules/test_process_checker.py
+++ b/Tribler/Test/Core/Modules/test_process_checker.py
@@ -1,12 +1,14 @@
 import os
-from multiprocessing import Process
+from multiprocessing import Process, Value
+from time import sleep
 
 from Tribler.Core.Modules.process_checker import ProcessChecker, LOCK_FILE_NAME
 from Tribler.Test.test_as_server import AbstractServer
 
 
-def process_dummy_function():
-    while True:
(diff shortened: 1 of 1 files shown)


devos50:
@ichorid please rebase on the latest commit on the next branch; it seems you have some of my work in this PR.

ichorid:
yup, already noticed that...

Tribler/Test/Core/Modules/test_process_checker.py
```diff
@@ -1,12 +1,14 @@
...
 
 from Tribler.Core.Modules.process_checker import ProcessChecker, LOCK_FILE_NAME
 from Tribler.Test.test_as_server import AbstractServer
 
 
-def process_dummy_function():
-    while True:
+def process_dummy_function(stop_flag):
+    while stop_flag.value == 0:
+        sleep(0.001)
```
qstokkink: Granularity of sleep is only in 10's of ms (0.01).

ichorid: ow!

qstokkink:
Please wait until the unit tests are finished before you merge.

PR #3842, opened 2018-08-31
Trial testing system

ichorid:
This PR contains all the changes necessary to support Twisted Trial testing system.

PR #3775, opened 2018-08-07
WIP: MetadataStore module ("Channel 2.0")

ichorid:
This module is the base of our upcoming Channel2.0 subsystem. Currently, it provides 3 features:
* PonyORM-based objects/DB bindings for Torrents and Channels metadata
* Serialization-deserialization of Torrent and Channels metadata (XDRLib)
* Torrent-based dissemination of Channels (new-style Channel can be serialized to disk in the form of a torrent)

At the moment, the module is integrated with _some_ of the REST endpoints. It is done in a dual-stack manner (i.e. search endpoint returns concatenated search results over _both_ new- and old-style channels).
By default, it is possible to load the .mdblob file containing channel metadata entry from disk. Tribler should start to download the corresponding torrent automatically into the correct "channels dir", and add scan the dir for binary metadata blobs when the download finishes. This is the intended usage mode for first experimental versions - just download channels by hand.
To enable the ability to create the new-style channels, the user _must_ tick the corresponding checkbox in the GUI "preferences" dialog box. If the option is enabled, new-style channel will be created from "My channel" dialog.

**TODO and problems**
* **First**, as @xoriole pointed out, we still don't know how to seed unique torrents from behind the NAT. One solution can be using hidden seeding (if it works).
* **Second**, there is a LOT of stuff to be changed in the remaining REST endpoints to support new-style channel addresses (which are basically hex-encoded IPv8 public keys).
* **Third** the saved "personal channel" is not seeded by default. 
* **Fourth** current channel "torrent" form uses "single MD entry per file" scheme. It is inefficient for very large channels, but very convenient for testing purposes.
* And add more unit tests, and state checks for channel downloads, and move stuff around, and refactor unit tests, etc, etc...

So, the draft plan is: 
1. Fix the remaining REST endpoints (and add tests for them).
2. Add tracker field to the TorrentMD object and serialized form.
3. Get the channel torrents seeding working.
4. Start squashing all MD entries from the same channel update into a single file.
5. Work through TODOs and FIXMEs in the code.

**Design notes**
* Some methods are "dangling" and not bound to objects. This is because there is still no fitting objects to attach them to. This objects should emerge eventually.
* We have an option to make serialization procedure completely automatical based on class structure.
* The class structure is designed to be general, but not *too* general.
* `serialize` method **changes input dict by design**. It is discussable, though.

PR #3875, opened 2018-09-14
WIP: Disable Watchdog in tests

ichorid:
Watchdog is a nice concept and is really useful in production code. However, enabling in during tests is counterproductive, since it effectively hides some errors. We really want our tests to be sensitive to this kind of stuff.
In addition, its usage results in "unclean reactor" errors with Trial testing system.
Thus, it is removed from base test class.

### DIFF
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -100,7 +100,6 @@ class AbstractServer(BaseTestCase):
 
         if annotate:
             self.annotate(self._testMethodName, start=True)
-        self.watchdog.start()
 
     def setUpCleanup(self):
         # Change to an existing dir before cleaning up.
@@ -168,12 +167,6 @@ class AbstractServer(BaseTestCase):
 
         process_unhandled_exceptions()
         process_unhandled_twisted_exceptions()
-
-        self.watchdog.join(2)
-        if self.watchdog.is_alive():
-            self._logger.critical("The WatchDog didn't stop!")
-            self.watchdog.print_all_stacks()
-            raise RuntimeError("Couldn't stop the WatchDog")
 
         if self.file_server:
             yield maybeDeferred(self.file_server.stopListening).addCallback(self.checkReactor)


qstokkink:
Hmm.. we should only merge this once we finally switch to trial though. The Watchdog makes sure that our test suite actually exits when there is a deadlock. Otherwise we depend on Jenkins timing out (anyone not running Jenkins would have to guess if the test suite locked up or not).

devos50:
@qstokkink @ichorid does trial have a different mechanism to deal with deadlocks during the tests?

I'm not convinced yet why we should remove watchdog. Can you give an example of which errors it may hide?

qstokkink:
Error from mac failure:

```
WARNING 1536935173        services:bitcoinlib.services.services:152  bitcoind.testnet.gettransactions(([u'mvJZwx7eZM6ffvR5o7MvgeYGshi9G22S7k', u'mrWMZ4Dpdv2U9vqXWBogCJzoyse61Enqk3'],)) Error __init__() got an unexpected keyword argument 'strict'
```

devos50:
@qstokkink see #3876, I will look into it.

ichorid:
I retested https://github.com/Tribler/tribler/pull/3842 and https://github.com/Tribler/tribler/pull/3775 with the watchdog enabled/disabled and with/without the [thread shutdown patch](https://github.com/Tribler/tribler/pull/3873) for both Trial and Nose based runners.
The tests show that the watchdog is **unable to kill the stuck thread** if `unittest` is provided by `twisted.trial`.

Therefore, the watchdog in its current state is useless for Tribler tests. We should either disable it or make it function properly. Or we can just ignore it and create an issue for enhancing it to:
1. be able to *really* kill all kinds of runaway processes;
2. provide a special "unittest-compatible-mode" in which whenever it acts it would raise an error.

qstokkink:
I support fixing it instead of removing it. The only reason I see for removing it, is if we were to switch all Tribler tests to IPv8's TestBase (which also has deadlock detection).

devos50:
Let's discuss this on Monday.

qstokkink:
It does not seem like this will ever get merged. As this PR is also stale, I will remove it for now as it is cluttering our PR list.

PR #3877, opened 2018-09-14
Remove broken test_remote_torrent_handler

ichorid:
These tests are horribly broken under Trial testing system for some reason. We don't want them anyways, since these are for the outdated TFTP-baset rtorrent downloader that was broken for half a year and no one noticed it. And the whole subsystem is going to be replaced soon by GigaChannels, etc, etc...


devos50:
I agree that these tests are not very good. Normally, I would argue not to remove tests until we remove the functionality itself (since this would leave us with uncovered code). However, removing these tests does not seem to impact code coverage so I'm fine with removing them.

Tribler/Test/test_remote_torrent_handler.py
```diff
@@ -1,123 +0,0 @@
...
-    @deferred(timeout=20)
-    def test_metadata_download(self):
-        """
-        Testing whether downloading torrent metadata from another peer is successful
-        """
-        session1_port = self.session.config.get_dispersy_port()
-
-        # Add thumbnails to the store of the second session
-        thumb_file = os.path.join(unicode(TESTS_DATA_DIR), u"41aea20908363a80d44234e8fef07fab506cd3b4",
-                                  u"421px-Pots_10k_100k.jpeg")
```
devos50: I think you can also remove this thumbnail?

qstokkink: Yes, this seems to be the only place this jpeg is used. Please remove it now or we probably won't remember to remove it in the future.

ichorid: Did that

ichorid:
I'm going to merge it myself, if no one minds...

PR #3879, opened 2018-09-15
Libtorrent callbacks on main thread

ichorid:
This is a blocker for Trial testing system migration. It is interesting to note that this is the last call there that is still on pool thread, and not on main thread.


Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -931,7 +931,7 @@ def reschedule_cb():
 
                             reactor.callFromThread(reschedule_cb)
 
-                    reactor.callInThread(session_getstate_usercallback_target)
+                    reactor.callFromThread(session_getstate_usercallback_target)
```
devos50: Judging from the `LibtorrentDownloadImpl.py` file, I think we can be pretty sure that `reschedule_cb` is always called from a reactor thread, so I would simply invoke `session_getstate_usercallback_target` directly (or remove this additional method call to reduce complexity).

ichorid: Just removed all the `callFromThread` stuff from there. Gonna refactor all the Libtorrent wrapper subsystem next month anyway...

PR #3880, opened 2018-09-16
Switched to register_anonymous_task where applicable

qstokkink:
This removes the need for custom random id creation.

Also switched away from `save_pstate` id based on time, which is unsafe.

### DIFF
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -1019,8 +1019,7 @@ class TriblerLaunchMany(TaskManager):
 
         self.downloads[infohash].pstate_for_restart = pstate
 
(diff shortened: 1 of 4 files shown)


qstokkink:
@devos50 https://jenkins-ci.tribler.org/job/GH_Tribler_PR_tests_mac/385/testReport/junit/Tribler.Test.Core.Modules.Wallet.test_trustchain_wallet/TestTrustchainWallet/test_monitor_tx_existing/

devos50:
Hmm there is no error/requests to other services during these tests, although the timeout of 1 second is on the low side since TrustChain blocks are signed/verified during the test (and our Mac tester is so slow). I would increase the timeout of this test to 2 seconds. If you agree, could you squeeze in that change in this PR?

PR #3884, opened 2018-09-17
WIP: Renamed config_parser, moved Tribler/Core/Upgrade package to Python 3

### LINKED ISSUE #3801: tribler/Tribler/Core/Utilities/configparser.py needs to be renamed
This file has a name collision with https://docs.python.org/3/library/configparser.html that it will want to import so PyLint wil rasie a __W0406: (import-self), : Module import itself__ error when we move to Python 3.  This was not a problem in Python 2 because https://docs.python.org/2.7/library/configparser.html uses __ConfigParser.py___ so there is no name collision.

- [ ] I have read CONTRIBUTING.rst

- [ ] I have tried with the latest pre-release version and I still can reproduce the issue.

devos50:
Fixes #3801 

```
MacBook-Pro-van-Martijn-2:tribler martijndevos$ find Tribler/Core/Upgrade/ \( -path 'Tribler/Core/DecentralizedTracking' -o -path 'Tribler/Core/DecentralizedTracking' \) -prune -o -type f -iname '*.py' | xargs pylint --py3k

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```


qstokkink:
You need to import `urllib_future` and call `urllib_future.urlparse`.

Issue #3897, opened 2018-09-21
Absolute vs relative file path convention in Tribler config: make Tribler portable


ichorid:
We really want to be consistent with our path treatment on different platforms. However, we would also want to be flexible for advanced users. Therefore, we should treat any option regarding Tribler files as **either relative to Tribler dir, or absolute system-level path**.
One important use case is, say, putting your metadata folder on SSD, and your download dir on an HDD (without using mount points and other system-level stuff).

This could be easily achieved by doing something like:
``` python
def get_some_filename(self):
   filename = self.config["general"]["some_filename"]
   return filename if os.path.isabs(filename) else os.path.join(self.get_state_dir(), filename)
```

One important side-effect of doing this is that Tribler becomes *portable* by default, which is important for security-concerned advanced users.

ichorid:
This could be done with a simple wrapper, or even be inheriting from something for even shorter notation.

ichorid:
Now, Tribler shows a really broken behavior regarding using `TSTATEDIR` environment variable:
if one copies the existing state directory and then points to it with `TSTATEDIR`, Tribler will *partially* use the old dir, because `triblerd.conf` holds some entries pointing back to the original path **including the state dir itself**.

This is like making a road sign stating "This is a road sign".

PR #3923, opened 2018-10-02
READY: Channel 2.0, collective effort

ichorid:
This pull request is a proper version of #3906 and #3775, suitable for collective editing.


ichorid:
retest this please

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
```diff
@@ -1,5 +1,6 @@
 import os
-from datetime import datetime
+from datetime import datetime, time
+from time import sleep
```
devos50: Please avoid using sleeps in tests (I think this is an unused import anyway) 👍

ichorid:
@devos50, why did you removed directory comparison tests, test blobs generation utilities and tests for handling incorrect blobs/filenames/formats?

Tribler/Core/Config/config.spec
```diff
@@ -41,6 +41,11 @@ live_edges_enabled = boolean(default=True)
 bitcoinlib_enabled = boolean(default=True)
 dummy_wallets_enabled = boolean(default=False)
 
+[chant]
+enabled = boolean(default=True)
+channel_edit = boolean(default=False)
+channels_dir = string(default='channels')
```
devos50: I wonder whether we really need this configuration option. Maybe we could hard-code it for now? Or am I missing an important use-case for this?

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -0,0 +1,259 @@
...
+BLOB_EXTENSION = '.mdblob'
+
+
+def create_torrent_from_dir(directory, torrent_filename):
+    fs = file_storage()
+    add_files(fs, directory)
+    t = create_torrent(fs)
+    # For a torrent created with flags=19 with 200+ small files
+    # libtorrent client_test can't see its files on disk.
+    # optimize_alignment + merkle + mutable_torrent_support = 19
```
devos50: Does this mean that piece alignment is disabled for the first version?

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -0,0 +1,259 @@
...
+            my_channel.sign()
+            return my_channel
+
+        def consolidate_channel_torrent(self):
+            """
+            Delete the channel dir contents and create it anew.
+            Use it to consolidate fragmented channel torrent directories.
+            :param key: The public/private key, used to sign the data
+            """
+
```
devos50: Minor comment: you can remove this enter.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -0,0 +1,259 @@
...
+            """
+            Collect new/uncommitted and marked for deletion metadata entries, commit them to channel torrent and
+            remove the obsolete entries if the commit succeeds.
+            :return The new infohash, should be used to update the downloads
+            """
+            new_infohash = None
+            try:
+                new_infohash = self.update_channel_torrent(self.staged_entries_list)
+            except IOError:
+                print ("Error during channel torrent commit, not going to garbage collect the channel")
```
devos50: Please avoid using print statements. Instead, you should use the `logging` framework.

Tribler/Core/Modules/MetadataStore/OrmBindings/metadata.py
```diff
@@ -0,0 +1,76 @@
...
+        _payload_class = MetadataPayload
+        _my_key = None
+
+        def __init__(self, *args, **kwargs):
+            super(Metadata, self).__init__(*args, **kwargs)
+            # If no key/signature given, sign with our own key.
+            if "public_key" not in kwargs or (kwargs["public_key"]==self._my_key and "signature" not in kwargs):
+                self.sign(self._my_key)
+
+        def _serialized(self, key=None):
```
devos50: Please add some comments here, especially that it returns the serialized data and a signature as a tuple.

Tribler/Core/Modules/MetadataStore/OrmBindings/metadata.py
```diff
@@ -0,0 +1,76 @@
...
+            return self._payload_class(**self.to_dict())._serialized(key)
+
+        def serialized(self, key=None):
+            return ''.join(self._serialized(key))
+
+        def _serialized_delete(self):
+            """
+            Create a special command to delete this metadata and encode it for transfer.
```
devos50: What do you mean by a 'special command'?

Tribler/Core/Modules/restapi/channels/channels_torrents_endpoint.py
```diff
@@ -61,14 +66,35 @@ def render_GET(self, request):
...
-        if channel_info is None:
-            return ChannelsTorrentsEndpoint.return_404(request)
+        chant_dirty = False
+        if self.is_chant_channel:
+            with db_session:
+                channel = self.session.lm.mds.ChannelMetadata.get(public_key=self.cid)
+                if channel:
+                    if channel == self.session.lm.mds.get_my_channel():
+                        # That's our channel, it gets special treatment
+                        uncommitted = map(lambda x: convert_torrent_metadata_to_tuple(x, UNCOMMITTED),
```
devos50: Could you rename `x` to something more descriptive?

Tribler/Core/Modules/restapi/util.py
```diff
@@ -26,6 +28,38 @@ def return_handled_exception(request, exception):
...
+def convert_channel_metadata_to_tuple(metadata):
+    """
+    Convert some given channel metadata to a tuple, similar to returned channels from the database.
+    :param metadata: The metadata to convert.
+    :return: A tuple with information about the torrent.
+    """
+    votes = 1
+    my_vote = 2
+    spam = 0
+    relevance = 0.9
```
devos50: Could you add a todo here since this relevance score is not based on anything?

Tribler/Core/Modules/restapi/util.py
```diff
@@ -62,6 +96,9 @@ def convert_db_torrent_to_json(torrent, include_rel_score=False):
                 "category": torrent[4], "num_seeders": torrent[5] or 0, "num_leechers": torrent[6] or 0,
                 "last_tracker_check": torrent[7] or 0}
 
+    if len(torrent) >= 11:
```
devos50: This whole mechanism with these "magic tuples" is so fragile but it will be removed when we can get rid of Dispersy fortunately 👍

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
```diff
@@ -0,0 +1,222 @@
...
+        self.session.lm.mds.TorrentMetadata.from_dict(
+            dict(self.torrent_template, public_key=channel.public_key))
+        channel.commit_channel_torrent()
+        channel.delete_torrent_from_channel(tdef.get_infohash())
+
+        self.assertEqual(1, len(channel.contents_list))
+        self.assertEqual(2, len(os.listdir(my_dir)))
+        channel.consolidate_channel_torrent()
+        self.assertEqual(1, len(os.listdir(my_dir)))
+
```
devos50: Please remove these superfluous newlines.

Tribler/Test/Core/Modules/MetadataStore/test_store.py
```diff
@@ -0,0 +1,77 @@
...
+        loaded_metadata = self.metadata_store.process_channel_dir_file(metadata_path)
+        # Make sure the original metadata is deleted
+        self.assertIsNone(loaded_metadata)
+        self.assertIsNone(self.metadata_store.TorrentMetadata.get(infohash='1'*20))
+
+        # Test an unknown metadata type, this should raise an exception
+        invalid_metadata = os.path.join(self.session_base_dir,'invalidtype.mdblob')
+        make_wrong_payload(invalid_metadata)
+        self.assertRaises(UnknownBlobTypeException, self.metadata_store.process_channel_dir_file, invalid_metadata)
+
```
devos50: Please remove a newline here.

TriblerGUI/tribler_window.py
```diff
@@ -603,6 +606,15 @@ def on_add_torrent_browse_file(self):
             [self.pending_uri_requests.append(u"file:%s" % filename) for filename in filenames[0]]
             self.process_uri_request()
 
+    def on_add_mdblob_browse_file(self):
+        filenames = QFileDialog.getOpenFileNames(self,
+                                                 "Please select the .mdblob file",
+                                                 QDir.homePath(),
+                                                 "Tribler metadata files (*.mdblob)")
+        if len(filenames[0]) > 0:
+            [self.pending_uri_requests.append(u"file:%s" % filename) for filename in filenames[0]]
```
devos50: I would prefer a more explicit `for` statement here, instead of using list comprehensions without using the resulting list.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -0,0 +1,259 @@
...
+            my_channel = cls(public_key=buffer(cls._my_key.pub().key_to_bin()), title=title,
+                             tags=description, subscribed=True)
+            my_channel.sign()
+            return my_channel
+
+        def consolidate_channel_torrent(self):
+            """
+            Delete the channel dir contents and create it anew.
+            Use it to consolidate fragmented channel torrent directories.
+            :param key: The public/private key, used to sign the data
```
devos50: You can remove this line.

TriblerGUI/widgets/editchannelpage.py
```diff
@@ -191,6 +209,18 @@ def on_edit_channel_save_button_pressed(self):
                                                                   (channel_name, channel_description)).encode('utf-8'),
                                                      method='POST')
 
+    def clicked_edit_channel_commit_button(self):
+        self.editchannel_request_mgr = TriblerRequestManager()
+        self.editchannel_request_mgr.perform_request("mychannel", self.on_channel_committed,
+                                                     data=unicode('commit_changes=True').encode('utf-8'),
```
devos50: Instead of `True`, we use 0 and 1 instead in requests.

Tribler/Core/Config/config.spec
ichorid: I don't understand which of three options do you point at, so let me elaborate on this:
* `enabled` is a standard option to disable the module completely (e.g. for testing purposes);
* `channel_edit` option is necessary for the dual-stack mode. We will remove it as soon as the old Channels implementation is gone;
* `channels_dir` option originally had an ability *to point beyond Tribler home dir*, but you removed that code for some reason. This was meant to provide the user an ability to move their channel directory to a place more suitable for storing rarely-used info, like torrents and downloads. Many users own an SSD and HDD, and use SSD for system apps and games, and HDD for video, music, download, torrents and other 'bulk' data. If we don't provide this option, the channel data is store twice at their primary storage device: once in the SQLite database, and once in the torrent form.

https://github.com/Tribler/tribler/issues/3897 related

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
ichorid: yes. We are append-only bulk storage anyways, so it makes no sense now, especially with that problem of 200+ small files.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
ichorid: I always go through pylint after the PR is accepted, so don't worry about these minor stuff ))

Tribler/Core/Modules/MetadataStore/OrmBindings/metadata.py
ichorid: done

Tribler/Core/Modules/restapi/util.py
ichorid: exactly!

Tribler/Core/Config/config.spec
devos50: My question was about the `channels_dir` option (which is the line the comment is based on). I agree that the amount of required storage might become a problem with the channel functionality. You are also right that many users use different hard-drives/partitions for different types of content. In fact, many Tribler users have their state directory on another directory than their Tribler installation.

While this option provides flexibility, I think the ability to move the state directory to a different directory is enough for a vast majority of our user base. Also, I would like to minimize the number of configuration options. Once we have more insight into the (storage) requirements for channels, we could restore this option.

ichorid: So, you insist on removing it?

devos50: I just discussed this with the other team members, please leave it in for now 👍

ichorid: My point is, we should either remove it completely or allow it to point beyond the Tribler config dir. Otherwise, it is useless.

devos50: I agree. With this option, you can move it outside the Tribler config directory, right? Then just leave it in.

devos50:
retest this please

devos50:
retest this please

devos50:
I fixed the pylint checker for you, it was failing because of the `.mdblob` files in the tests directory:

```
Total: 2826 lines
Violation: 47 lines
% Quality: 98%
```

Please make sure to fix some of these violations 👍

ichorid:
Pony author about SQL indexes:

> 
> Alexander Kozlovsky, [16.10.18 15:20]
> You can specify unique index to column in one of the following way:
> class MyEntity(db.Entity):
>     foo = Required(str, unique=True)
>     bar = Optional(str, unique='index_name')
> 
> 
> In the former attribute, the index name will be defined automatically (for SQLite it will be unq_MyEntity_foo). If index with this name is not exists in the database it will be added automatically if you call generate_mapping with option create_tables=True)
> 
> In the latter attribute index name are given explicitly. This is useful if you already have index with that name in the database. If you just write index=True, then Pony will create a new index with a "canonical" name, and having two duplicate indexes will slow down inserts/updates
> 
> Inside db_session Pony has in-memory index for each unique database index which is known to Pony. So, if you add several objects inside the same db_session with the same value of an unique attribute, Pony will produce error immediately aven the creation of the second object in memory, even before INSERTs were performed.
> 
> If you already have a row inside a database with a specific value of an unquie column, and then create a new object with the same value of this column, then, if previous object was not loaded in current db_session, Pony will not try immediately search a previous object with this unique value, because it will lead to unnecessary SELECT queries. Instead, Pony will try to do INSERT in a usual way on flush/commit, and then the database will throw exception on foreign key violation.
> 
> If you want to know beforehand that such unique value exists, you can check it in the following way:
> if not MyEntity.exists(unique_attr=value):
>     obj2 = MyEntity(unique_attr=value, other_attr=value2)
> else:
>     print(value, 'already exists')
> 
> 
> But it will load previous object into memory, so it may be more efficient to just create a new object and catch error later on commit
> 
> When you do MyEntity.exists(foo=bar), Pony will execute a query
> SELECT * FROM MyEntity WHERE foo = ?
> If foo column have an unique index, this query will be very fast. The resulted object will be cached in db_session and added to in-memory index, so the second call of MyEntity.exists will find object in db_session cache and will not execute the same query one more time
> 
> Alexander Kozlovsky, [16.10.18 15:21]
> In short, if you define an attribute as unique, everything should be good
> 
> Alexander Kozlovsky, [16.10.18 15:28]
> in-memory db_session cache for each unique index is a dict with attribute value as a key and object as a value. For composite indexes the dict key is a tuple of attribute values

ichorid:
So, we really want our `signature` field to be unique. And maybe push `public_key` to a separate table.

ichorid:
retest this please

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -117,6 +117,7 @@ def __init__(self, session, tdef):
         self.vod_index = None
         self.orig_files = None
         self.finished_callback = None
+        self.finished_callback_already_called = False
```
devos50: I wonder if this will have unexpected consequences, since the finished alert raised by libtorrent is not always predictable.

Maybe it's better to handle this error at a higher level (when processing the channel blobs)? Is there a way to not process the same data again, while allowing `finished_callback` to fire multiple times?

ichorid: I've already added the code to skip blobs based on current torrent version, so it is relatively safe to remove this. However, it would still be dangerous to try to process the dir multiple times.

But what do you mean by "unpredictable"? Do you mean that it can trigger prematurely, not just repeatedly?

Tribler/Core/Config/config.spec
ichorid: Did just that

TriblerGUI/tribler_window.py
ichorid: yep

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ichorid: Ok, I'll leave it out. The checks at the `process_dir` level should be enough.

ichorid: The warning is still there, though.

Tribler/Core/Modules/restapi/channels/channels_torrents_endpoint.py
ichorid: `convert_torrent_metadata_to_tuple` already tells the reader that the `x` is `torrent_metadata`. If we change it to something longer the line will break and readability will suffer.
Really, long identifiers are only good for long scopes. At small scopes, short identifiers are much more readable.

Tribler/Core/Modules/restapi/util.py
ichorid: Done

ichorid:
retest this please

ichorid:
All tests pass, but status is red. What the schmuck? :face_with_head_bandage:

qstokkink:
@ichorid Jenkins was updated, the latest version of xUnit dies horribly. We'll attempt to resolve this.

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

ichorid:
Thanks, guys!

devos50:
Feedback:
- please hide the delete button if the state of a torrent is 'to delete' in the GUI.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -0,0 +1,304 @@
...
+
+CHANNEL_DIR_NAME_LENGTH = 60  # Its not 40 so it could be distinguished from infohash
+BLOB_EXTENSION = '.mdblob'
+CHUNK_SIZE_LIMIT = 1 * 1024 * 1024  # We use 1MB chunks as a workaround for Python's lack of string pointers
+
+
+def create_torrent_from_dir(directory, torrent_filename):
+    fs = file_storage()
+    add_files(fs, directory)
+    t = create_torrent(fs)
```
qstokkink: This should use the piece size of `CHUNK_SIZE_LIMIT` and flags of `16 | 1 = 17` for mutable torrent support.

devos50:
Link to channel storage on DAS5: https://jenkins-ci.tribler.org/job/pers/job/channels2_experiment_devos50/62/artifact/output/localhost/node301/.TriblerModule-14364-1/channels/a64981ca91bdc7084f598b0179fe5723f6c869f512fdb0c2be504c4df38a/

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
ichorid: As we discussed, we can optimize it later based on usage statistics.

qstokkink:
By the way, don't forget I exposed the [`ipv8.keyvault.crypto.default_eccrypto`](https://github.com/Tribler/py-ipv8/blob/master/ipv8/keyvault/crypto.py#L142) for you.

ichorid:
retest this please

ichorid:
retest this please

Tribler/Core/Modules/MetadataStore/serialization.py
```diff
@@ -0,0 +1,261 @@
...
+from Tribler.pyipv8.ipv8.messaging.serialization import Serializer
+
+EPOCH = datetime(1970, 1, 1)
+INFOHASH_SIZE = 20  # bytes
+
+SIGNATURE_SIZE = 64
+EMPTY_SIG = '0' * 64
+
+crypto = default_eccrypto
+serializer = Serializer()
```
qstokkink: There is also a `default_serializer` available. Actually, why redefine these here anyway?

ichorid: oops, overlooked that...

PR #3940, opened 2018-10-04
Fixed crash when filtering downloads

devos50:
Fixes #3940


devos50:
@xoriole is there anything we can do regarding sonarqube?

xoriole:
I've already set retry for sonarcloud request. There's not much we can do on our side.

PR #3943, opened 2018-10-05
READY: pip for bitcoinlib workaround

Dmole:
pip's return codes are unreliable and it's messages misleading so users may need extra guidance.


devos50:
ok to test

Dmole:
I considered
```
sudo chown -R $(id -un) ~/.local ; sleep 1 &  pip install bitcoinlib --user
```
but that's an ugly hack.

PR #3953, opened 2018-10-09
Remove unused Edit Channel buttons from everywhere

ichorid:
Remove hidden Edit Channel button from the GUI. The button's purpose was to be able to edit other people's Dispersy channels, but this never worked properly, so the button was hidden long time ago. And no one missed it.
Now it's gone for good.

### DIFF
--- a/TriblerGUI/widgets/editchannelpage.py
+++ b/TriblerGUI/widgets/editchannelpage.py
@@ -21,7 +21,7 @@ from TriblerGUI.utilities import get_image_path
 
 class EditChannelPage(QWidget):
     """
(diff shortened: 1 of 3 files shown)


TriblerGUI/widgets/editchannelpage.py
```diff
@@ -21,7 +21,7 @@
 
 class EditChannelPage(QWidget):
     """
-    This class is responsible for managing lists and data on the your channel page, including torrents, playlists
+    This class is responsible for managing lists and data on jour channel page, including torrents, playlists
```
devos50: Typo, please fix this 👍

ichorid: he-he

PR #3999, opened 2018-11-01
Fix chant-related DHT timeout error handling in torrents endpoint

ichorid:
And subscribe to the downloaded channel by default.


ichorid:
I could not figure out how to properly simulate DHT timeouts for `get_metainfo` to test the `timeout_callback`, so I tested it by hand (by turning off my network), and it worked. One way or another, this code will be gone when we refactor the torrent checker after Dispersy is out.

PR #4008, opened 2018-11-06
GigaChannel Gossip

### LINKED ISSUE #3997: download_channel finish deferred never called
When downloading a channel the finished deferred is never called, even though the channel torrent is already seeding. I used the following code:

```python
download, finished_deferred = self.tribler_session.lm.download_channel(channel)
```

I added a print statement to the libtorrent alert as follows:

```python
    @checkHandleAndSynchronize()
    def on_torrent_finished_alert(self, alert):
        self.update_lt_status(self.handle.status())
        print "I FINISHED!", self.finished_callback # This line was added
         ...
```

And this outputted the following:

```
I FINISHED! None
```

When I print the `download.finished_callback` on the caller I get the following though:

```
<function <lambda> at 0x7f2e085b2ed8>
```

(Note: I'm only running channel downloads and no other torrent downloads)

qstokkink:
Fixes #3997.

For graphs see: https://github.com/Tribler/gumby/pull/383.

This PR implements the most naive gossip protocol for GigaChannels. The selection algorithm is one line:

```python
db.ChannelMetadata.select(lambda g: g.subscribed==True).random(limit)
```

The remaining community code simply gossips these randomly selected subscribed channels (up to 7) to neighbors. Upon receipt of such gossiped channels, they will be queued and each channel will downloaded once the previous download finished or 10 seconds have passed.

This PR also enables libtorrent's LSD (when DHT is enabled) for 0 hop downloads, which is essential for channel downloading between nodes in the same NAT layer.

### DIFF
--- a/Tribler/Test/Community/gigachannel/test_community.py
+++ b/Tribler/Test/Community/gigachannel/test_community.py
@@ -1,0 +1,229 @@
+from twisted.internet.defer import inlineCallbacks
+
+from Tribler.community.gigachannel.community import ChannelDownloadCache, GigaChannelCommunity
+from Tribler.pyipv8.ipv8.peer import Peer
+from Tribler.pyipv8.ipv8.test.base import TestBase
+from Tribler.Test.mocking.channel import MockChannel
+from Tribler.Test.mocking.download import MockDownload
+from Tribler.Test.mocking.session import MockSession
+
+
+class TestGigaChannelUnits(TestBase):
+
+    """
+    Unit tests for the GigaChannel community which do not need a real Session.
+    """
+
+    def setUp(self):
+        super(TestGigaChannelUnits, self).setUp()
+        self.session = MockSession()
+
+        self.initialize(GigaChannelCommunity, 1)
+
+    def create_node(self, *args, **kwargs):
+        kwargs['tribler_session'] = self.session
+        return super(TestGigaChannelUnits, self).create_node(*args, **kwargs)
+
+    def _setup_fetch_next(self):
+        """
+        Setup phase for fetch_next() tests.
+
+        Provides:
+         - Database entry for a mocked Channel.
+         - download_channel() functionality for the mocked channel.
+         - Pending overlay.download_queue for the mocked channel.
+        """
+        channel, download = self._setup_download_completed()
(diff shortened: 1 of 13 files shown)


synctext:
What does .version do?

qstokkink:
@synctext it provides the version information of a channel. Each update to a channel gives you a new version. We currently use an unsigned long to store this, so we are bound to an upper limit of version 18,446,744,073,709,551,615.

qstokkink:
I found myself copy-pasting `deliver_messages()` and `sleep()` from IPv8's `TestBase` again. Once Dispersy is gone, we should really merge that into `TestBase` and get rid of the duplicate code (which is also in the TunnelCommunity tests).

Anyway, as I can defend the remaining coverage and Pylint warnings, ready for review I guess.

Tribler/Test/Community/gigachannel/test_community.py
```diff
@@ -0,0 +1,320 @@
+from twisted.internet.defer import Deferred, inlineCallbacks
+
+from Tribler.community.gigachannel.community import ChannelDownloadCache, GigaChannelCommunity
+from Tribler.pyipv8.ipv8.peer import Peer
+from Tribler.pyipv8.ipv8.test.base import TestBase
+
+
+class MockChannel(object):
```
devos50: Some of these mock classes are helpful in other tests. Could you extract them to a more generic location so we can reuse them across different tests?

Tribler/community/gigachannel/community.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        Callback for when the download states are updated in Tribler.
+        We still need to filter out the channel downloads from this list.
+
+        :param states_list: the list of download states
+        :type states_list: [DownloadState]
+        :returns: None
+        """
+        for ds in states_list:
+            if ds.get_download().dlconfig.get('download_defaults', 'channel_download'):
```
devos50: You can use `dlconfig.get_channel_download()` instead.

qstokkink: I tried, but it's apparently some kind of other runtime configuration copy, which is missing the method.

Issue #4010, opened 2018-11-08
Pony missing in Ubuntu


xoriole:
Pony is missing in the Debian build of experimental release, unfortunately. It is not available in the Debian repository. Currently, it is only available via pip which is an issue. 
We can install pony during installation via post-install script but that is not recommended and is considered a security risk. Another option is to create a PPA with dependencies or deploy Tribler as a snap package.

ichorid:
Duplicate of #4224

PR #4011, opened 2018-11-08
Added pony as post installation with pip

xoriole:
Added pony as post installation with pip


xoriole:
retest this please

xoriole:
retest this please

debian/tribler.postinst
```diff
@@ -0,0 +1,2 @@
+# Install pony orm with pip since it is available in Debian repository
```
qstokkink: *not available

xoriole: right

devos50:
This fixes issue #4010 right?

qstokkink:
What was the reason again that we never used `pip`?

qstokkink:
We need to decide what to do with this.

PR #4013, opened 2018-11-09
Updated ipv8 pointer

qstokkink:
This pointer update also requires references to `DiscoveryCommunity` to be refactored and `TestBase` subclasses to return the value of `super( ... , self).tearDown()`.


devos50:
So I guess we have to update Gumby?

qstokkink:
@devos50 indeed, I included that in my superduper major PR though.

PR #4017, opened 2018-11-13
No new REST requests are processed after shutdown is initiated

xoriole:
No new REST requests are processed after the shutdown sequence is initiated. This prevents any error from GUI requests while the core is shutting down.

### DIFF
--- a/Tribler/Test/Core/Modules/RestApi/test_rest_manager.py
+++ b/Tribler/Test/Core/Modules/RestApi/test_rest_manager.py
@@ -1,3 +1,5 @@
+import os
+
 from Tribler.Core.exceptions import TriblerException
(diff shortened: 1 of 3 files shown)


Tribler/Test/Core/Modules/RestApi/test_rest_manager.py
```diff
@@ -35,3 +37,26 @@ def verify_error_message(body):
         self.should_check_equality = False
         return self.do_request('channels/discovered', expected_code=500, expected_json=None, request_type='PUT',
                                post_data=post_data).addCallback(verify_error_message)
+
+    @trial_timeout(1)
```
devos50: Isn't one second a bit on the low side?

xoriole:
retest this please

PR #4035, opened 2018-11-16
Discard old channel versions

### LINKED ISSUE #4033: Channel upgrades cause the old metadata to be lost
Right now, when we update a channel we bump its version and infohash. This means that all of the old versions of the channel we have in the download list will no longer have a matching infohash in the metadata database.

The terminal becomes flooded with messages like this:

```
ERROR   1542376218.96                 community:124  (GigaChannelCommunity)  We are downloading channel 3ad23eae39c5af1caa80b5b207757969f5f117e8, but no database entry exists!
ERROR   1542376218.96                 community:124  (GigaChannelCommunity)  We are downloading channel 81ebcd1a1c52efd37078f8bf4c0e8e35be95fd9d, but no database entry exists!
```

qstokkink:
Fixes #4033.

### DIFF
--- a/Tribler/community/gigachannel/community.py
+++ b/Tribler/community/gigachannel/community.py
@@ -1,6 +1,9 @@
+from time import time
+
 from pony.orm import db_session
 
 from Tribler.community.gigachannel.payload import TruncatedChannelPayload, TruncatedChannelPlayloadBlob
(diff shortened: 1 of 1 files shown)


qstokkink:
retest this please

PR #4041, opened 2018-11-19
Fixed key error in pubsub subscription status

xoriole:
Fixes https://github.com/Tribler/tribler/issues/4034


xoriole:
retest this please

PR #4052, opened 2018-11-21
Debug build for Tribler (--trace-debug & --trace-exception)

xoriole:
**Debug build for Tribler**
If debug is enabled with runtime argument `--trace-debug`, `--trace-exception`, code tracing is enabled and all the calls and exceptions are logged. It should be easier now to debug the issues where Tribler fails to boot.

### DIFF
--- a/check_os.py
+++ b/check_os.py
@@ -1,9 +1,11 @@
 import logging.config
 import os
-import psutil
-import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


qstokkink:
Hmm.. on my Windows machine I never get `'exception'` events.

PR #4058, opened 2018-11-23
Allow code execution in Tribler GUI (disabled by default)

devos50:
In this PR, I've added a flag to the Tribler GUI to allow for remote code execution. If this flag is set, a TCP server will be started and accepts incoming connections. These connections can be used to execute some code. This mechanism enables us to run the Tribler application tester in the shipped Tribler releases, without having to make a separate build.


TriblerGUI/tribler_app.py
```diff
@@ -28,6 +32,16 @@ def parse_sys_args(self, args):
             elif arg.startswith('magnet'):
                 self.handle_uri(arg)
 
+        if '--allow-code-injection' in sys.argv[1:]:
+            variables = globals().copy()
+            variables.update(locals())
```
qstokkink: For reference: this is slightly dangerous.

PR #4064, opened 2018-11-26
Fixed homepage items running off screen

qstokkink:
This PR fixed the homepage items running off the screen (requiring a scrollbar) for widescreens.

afbeelding


devos50:
That was easy 👍

PR #4088, opened 2018-12-06
Various fixes

### LINKED ISSUE #4087: Tribler shows "0.0 GB Token balance" when not fully initialized

v7.1.3

Linux

1. Cold-start Tribler
2. Wait for about 10 seconds until window appears
3. Wait for a minute until cogs stop spinning and it shows the main window
4. Check token balance at the top

It either shows actual token balance or nothing (or something like "--.- GB Token balance").

##### Actual behavior: #####

It displays zero token balance for some time, before revealing the real balance.

devos50:
Some fixes for bugs that were revealed by the application tester 👍 

I also added some constructs for logging circuit rejections (for scientific purposes). In addition to that, I introduced a flag in the tunnel helpers to log removed exit sockets. This could help us to debug circuits.

Fixes #4087


devos50:
I found a bug in my code so I had to change my last commit.

Issue #3883, opened 2018-09-17
Python 3 compatibility


devos50:
This issue hosts all discussion regarding the upgrade to Python 3. It follows the methodology of https://github.com/Tribler/py-ipv8/issues/265.

Everyone is welcome to help us to move our code base to Python 3. While pull requests are more than welcome, please keep the following things in mind:
- make sure that your (new) code is compatible with both Python 2 and Python 3. To help developers with identifying code that is not backward compatible, we run a few checks on your code. Please address issues raised by these checkers accordingly, or motivate why something is not worth fixing.
- keep your pull requests small and concise. This makes it easier for us to review, or revert some changes in case a (major) error slipped through.
- if you have a question, please leave a comment on this issue.

Python 3 migration guidelines:
- use the `six` library and the `six.moves` library as much as possible.
- for hex encoding/decoding, use `hexlify` and `unhexlify` in the `binascii` package. We found out this is significantly faster than using the `encode` or `decode` functions in `codecs`.

devos50:
Also, we have to remove the pymdht submodule if we are going to do this (in favour of EgbertDHT). Otherwise, we have to port pymdht to Python 3, which requires a significant amount of engineering effort.

qstokkink:
As I discovered the Python 3 static analysis is in no way enough to claim a class is ported, we should also do test driven porting with Python 3 actually running the tests.

We could do two things here:
 1. First fix all of the static analysis issues and then move on to the tests.
 2. Port to Python 3 per module, making sure the tests pass with Python 3 for this specific module.

From my experience I prefer (2) as this gives us more insight into the remaining work.

cclauss:
Create a branch and try[__futurize --stage1 -w .__](https://python-future.org/automatic_conversion.html#stage-1-safe-fixes) to get a sense of the initial changes that will be required.

ichorid:
Some interesting links to note:
Twisted is compatible with asyncio (e.g. it is officially supported mixing them together):
https://meejah.ca/blog/python3-twisted-and-asyncio
this provides us with a clear migration path: de-Dispersification -> Python3 -> twisted+asyncio -> asyncio

Comparison of Twisted concepts and resulting code to Asyncio:
https://asyncio.readthedocs.io/en/latest/twisted.html

cclauss:
I agree with [the approach suggested above](https://github.com/Tribler/tribler/issues/3883#issuecomment-427280018) by @qstokkink 

[...]

PR #4094, opened 2018-12-07
Minimal changes to pass the imports on Python 3

### LINKED ISSUE #4092: Python 3 syntax errors in community/market/community.py

It is a syntax error in Python 3 for __lambda__ parameters to be surrounded by parentheses.  This happens in six places in this file and it can be automatically fixed with the command: __2to3 -f lib2to3.fixes.fix_tuple_params Tribler/community/market/community.py__.  However the resulting code is quite ugly because it needs to be able to deal with the situation where the caller passes a few parameters or a single tuple containing multiple parameters.  A manual approach is more readable than the automated approach.

[flake8](http://flake8.pycqa.org) testing of https://github.com/Tribler/tribler on Python 3.7.1

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./Tribler/community/market/community.py:595:68: E999 SyntaxError: invalid syntax
        return self.create_new_tick_block(tick).addCallback(lambda (blk, _): on_block_created(blk))
                                                                   ^
```

devos50:
In this PR, I made the minimal required changes to at least start executing the nosetests on Python 3. These changes should not break compatibility with the old Python 2 code. If the tests in this PR pass, this will give @cclauss a stable basis to work on.

I made a few changes to prevent us from porting Dispersy/pymdht. In particular, I made the imports that import either Dispersy or pymdht local. There were only a few of these and I think this is acceptable to do so we do not have to wait for these submodules to be removed.

All dependencies except for Meliae are available for Python 3, either via `pip` or `apt`.

Fixes #4092 

Related to #3883

### DIFF
--- a/Tribler/util.py
+++ b/Tribler/util.py
@@ -1,0 +1,11 @@
+"""
+This file contains various utility methods.
+"""
+from __future__ import absolute_import
+
+import sys
+
+if sys.version_info.major > 2:
+    cast_to_unicode_utf8 = lambda x: "".join([chr(c) for c in x]) if isinstance(x, bytes) else str(x)
+else:
+    cast_to_unicode_utf8 = lambda x: x.decode('utf-8')
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -3,6 +3,8 @@ LaunchManyCore
 
 Author(s): Arno Bakker, Niels Zeilemaker
 """
+from __future__ import absolute_import
+
 import binascii
 import logging
 import os
--- a/Tribler/Core/APIImplementation/LaunchManyCore.py
+++ b/Tribler/Core/APIImplementation/LaunchManyCore.py
@@ -23,7 +23,6 @@ from twisted.internet.threads import deferToThread
 from twisted.python.threadable import isInIOThread
 
 from Tribler.Core.CacheDB.sqlitecachedb import forceDBThread
-from Tribler.Core.DecentralizedTracking.dht_provider import MainlineDHTProvider
 from Tribler.Core.DownloadConfig import DownloadStartupConfig, DefaultDownloadStartupConfig
 from Tribler.Core.Modules.MetadataStore.serialization import ChannelMetadataPayload, float2time
 from Tribler.Core.Modules.MetadataStore.store import MetadataStore
@@ -296,6 +295,7 @@ class TriblerLaunchMany(TaskManager):
             competing_slots = self.session.config.get_tunnel_community_competing_slots()
 
             if self.mainline_dht:
+                from Tribler.Core.DecentralizedTracking.dht_provider import MainlineDHTProvider
                 dht_provider = MainlineDHTProvider(self.mainline_dht, self.session.config.get_dispersy_port())
             else:
                 dht_provider = DHTCommunityProvider(self.dht_community, self.session.config.get_dispersy_port())
--- a/Tribler/Core/Category/Category.py
+++ b/Tribler/Core/Category/Category.py
@@ -3,10 +3,12 @@ Category.
 
 Author(s):  Yuan Yuan, Jelle Roozenburg
 """
+from __future__ import absolute_import
+
 import logging
 import os
 import re
-from ConfigParser import MissingSectionHeaderError, ParsingError
+from six.moves.configparser import MissingSectionHeaderError, ParsingError
 
 from Tribler.Core.Category.FamilyFilter import XXXFilter
 from Tribler.Core.Category.init_category import getCategoryInfo
--- a/Tribler/Core/Category/init_category.py
+++ b/Tribler/Core/Category/init_category.py
@@ -3,7 +3,9 @@ Give the initial category information.
 
 Author(s): Yuan Yuan
 """
-import ConfigParser
+from __future__ import absolute_import
+
+from six.moves import configparser
 
 
 def __split_list(string):
@@ -38,7 +40,7 @@ def __get_default():
 
 
 def getCategoryInfo(filename):
-    config = ConfigParser.ConfigParser()
+    config = configparser.ConfigParser()
     config.readfp(open(filename))
 
     cate_list = []
--- a/Tribler/Core/CreditMining/CreditMiningSource.py
+++ b/Tribler/Core/CreditMining/CreditMiningSource.py
@@ -5,8 +5,6 @@ import logging
 from binascii import hexlify, unhexlify
 
 from Tribler.dispersy.exception import CommunityNotFoundException
-from Tribler.community.allchannel.community import AllChannelCommunity
-from Tribler.community.channel.community import ChannelCommunity
 from Tribler.Core.simpledefs import NTFY_DISCOVERED, NTFY_TORRENT, NTFY_CHANNELCAST
 from Tribler.pyipv8.ipv8.taskmanager import TaskManager
 
@@ -65,6 +63,9 @@ class ChannelSource(BaseSource):
         try:
             self.community = dispersy.get_community(unhexlify(self.source), True)
         except CommunityNotFoundException:
+            from Tribler.community.allchannel.community import AllChannelCommunity
+            from Tribler.community.channel.community import ChannelCommunity
+
             allchannelcommunity = None
             for community in dispersy.get_communities():
                 if isinstance(community, AllChannelCommunity):
--- a/Tribler/Core/CreditMining/CreditMiningSource.py
+++ b/Tribler/Core/CreditMining/CreditMiningSource.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 import logging
 
 from binascii import hexlify, unhexlify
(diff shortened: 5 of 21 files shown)


Tribler/util.py
```diff
@@ -0,0 +1,41 @@
+"""
+This file contains various utility methods.
+"""
+import sys
+from collections import namedtuple
+
```
qstokkink: `StringIO`, `is_long_or_int` and part of `urllib_future` are already defined in:
https://github.com/Tribler/py-ipv8/blob/master/ipv8/util.py

Please reuse/extend those to get rid of code duplication.

devos50: Do we actually want to depend on the IPv8 submodule for these variables? Although it has some code duplication, I think it makes sense to also have them in the Tribler code base.

devos50:
I've made the requested changes and added TODO statements above the Dispersy-related imports. Also, I rely on the Python3-specific constructs already provided by IPv8 now. Finally, I made two changes to fix the startup/shutdown procedure of the tests in a Python 3 environment. We are now ready for further Python 3 migration effort.

Tribler/util.py
cclauss: I am struggling with why we use __six__ and then rewrite much of the same functionality in custom code in __util.py__.  Why not leverage __six__ as much as possible?

cclauss: __six.moves__ does so much of this in a more common way.

devos50:
I have setup a dedicated linux tester that runs the unit tests in Python 3, see https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux_py3/. I will attach this one to our PR tester after this PR has been merged (since it will not pass with these changes yet).

devos50:
retest this please

devos50:
I have also setup a job to check for new Python 3 violations. This is simply a copy of our existing Pylint checker, but with the `--py3k` option passed to it. It will fail if one introduces a new Python 3 violation 👍

cclauss:
On Python 2 can you please add a [redefining-builtins-modules](http://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=redefining-builtins-modules) option to the PyLint command to keep it from protesting every time we use __six__ to redefine a Python 2 __builtin__?  That should speed the review process.

devos50:
retest this please

devos50:
retest this please

cclauss:
Did __pip install flake8__ happen?  You can __flake8 —version__ to _force_some output.

devos50:
@cclauss running into this issue: https://github.com/Bachmann1234/diff-cover/issues/54. I will try to debug diff-cover myself to fix this. It seems that flake8 is very helpful!

devos50:
retest this please

cclauss:
It is is easier to get started, do not do the diff trick on flake8...  Just give it the whole repo which is will test in about a minute.  Python 2 should just pass and print a __0__.  Once that is working, you can get fancier.

devos50:
retest this please

devos50:
Fixed a bug in our script that invokes pylint! For some reason, we passed `--option=Tribler` which checks **all** files in the Tribler subdirectory. When upgrading to a newer version of the `diff-cover` tool, we noticed that our pylint checker became very slow. This is the case since for each diff file, diff-cover spawns a process that will check the whole Tribler codebase!

Also, the py3k checker makes more sense now and gives me errors I haven't fixed (yet)! 👍

devos50:
@cclauss yes, that will probably be the easiest option for now. To be continued tomorrow...

devos50:
Nah, just added the flask8 checker, see https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_flake8_check/. It fails when there is a new violation. It is also very fast to check.

Tribler/util.py
devos50: @qstokkink could you comment on the usage of `six`? In IPv8, you only used it for metaclasses right?

devos50:
@qstokkink ready again. I made quite some changes, both to the pylint checkers and the code. The remaining py2k checker error is unavoidable.

cclauss:
cast_to_unicode_utf8 = lambda x: unicode(x, 'utf-8') -->
cast_to_unicode_utf8 = lambda x: x.decode('utf-8') ???

Tribler/util.py
qstokkink: This is related to https://github.com/Tribler/tribler/pull/4096#issuecomment-445440146.

**tl;dr** I agree, we should use `six` as much as possible. As I missed some of `six`'s functionality when porting IPv8, I also sinned by making duplicate functionality (see https://github.com/Tribler/py-ipv8/issues/382).

devos50: Ok, I will change my PR to use `six` instead.

devos50:
@cclauss thanks, that would be better indeed!

qstokkink:
One mental note, somewhat unrelated to this PR specifically, on `.encode('utf-8')`/`.decode('utf-8')`: we use strings for binary data a lot in our codebase. We should take care of that fact that we cannot transform arbitrary strings into unicode, for example:

```python
>>> unicode('\x80', 'utf-8')
UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 0: invalid start byte
```

So, every unicode transform in a PR warrants a double-check on the possible string data (i.e. only UTF-8 code page compatible bytes).

devos50:
retest this please

devos50:
@qstokkink

devos50:
`six` is already the latest version on the server where pylint runs. I guess we have to ignore these warnings, or disable them in the `pylintrc` file.

cclauss:
http://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=redefining-builtins-modules#variables-checker

Tribler/util.py
```diff
@@ -0,0 +1,11 @@
...
+This file contains various utility methods.
+"""
+from __future__ import absolute_import
+
+import sys
+
+if sys.version_info.major > 2:
+    cast_to_unicode_utf8 = lambda x: "".join([chr(c) for c in x]) if isinstance(x, bytes) else str(x)
+else:
+    cast_to_unicode_utf8 = lambda x: x.decode('utf-8')
```
qstokkink: Just to double check, are you sure you want this?

In Python 3: put all bytes in a `bytes` into shorts (/`chr`s) of a unicode string
In Python 2: interpret all bytes in a `str` as utf-8 encoded data in a unicode string

Both will create a unicode string, but the Python 2 version will error out for all values which are not in the utf-8 character map (for example '\x80').

devos50: > Both will create a unicode string, but the Python 2 version will error out for all values which are not in the utf-8 character map (for example '\x80').

I did not change the behavior for the Python 2 side and yes, it can error out. The code where I invoke `cast_to_unicode_utf8` _should_ be able to deal with this error.

Maybe we can make a `cast_to_unicode_safe` method in another PR that ignores characters that are not decodable?

cclauss: Also worthy of study are __six.ensure_binary(), six.ensure_str(), and six.ensure_text()__ which are in __master__ on GitHub but they have never been published in a release.
https://github.com/benjaminp/six/blob/master/six.py#L853-L908

cclauss:
LGTM: Please resolve conflict.

devos50:
This PR still has some work to do, changing everything to use `six` now...

Tribler/util.py
cclauss: A new version of __six__ landed overnight that that officially ships those three functions in it.
* https://github.com/benjaminp/six/blob/master/CHANGES

qstokkink: After some digging, I also hit some paydirt with this one:

```python
>>> u'\x80'.encode('raw_unicode_escape')
'\x80'
>>> '\x80'.decode('raw_unicode_escape')
u'\x80'
```

cclauss: $ python3 -c "'\x80'.decode('raw_unicode_escape')" --> AttributeError :-(

I have found that the easiest conversions are done with:
* __codecs.encode(b'x', 'hex')__
* __codecs.decode('x', 'hex')__
See: https://github.com/Tribler/tribler/pull/4100/files#diff-68ce5114e24c3b06aea5e0411154d52fR38

qstokkink: @cclauss That was Python 2 syntax, here's Python 3 for the same:

```python
>>> codecs.encode(u'\x80', 'raw_unicode_escape')
'\x80'
>>> codecs.decode('\x80', 'raw_unicode_escape')
u'\x80'
```

devos50: @qstokkink want me to change the `cast_to_unicode_utf8` to that?

cclauss: Are __cast_to_unicode_utf8()__ and the new __six.ensure_test()__ equivalent?

qstokkink: @devos50 If this is the behavior you want:

```python
import codecs
a = b'abc'
print(type(a), a)
b = codecs.decode(a, 'raw_unicode_escape')
print(type(b), b)
```

```
<class 'bytes'> b'abc'
<class 'str'> abc
```

Personally I like this. I think I can get rid of some of my custom constructs in IPv8 now as well.

devos50: @cclauss while I would say the ensure methods are helpful, the newest version of `six` is not available in the (stable) Debian repositories and we cannot use these methods safely across all linux distros. At one point, we might move to pip-based dependencies which provide us with newer versions of Linux packages.

qstokkink: That would be good, we still have many files to port.

PR #4097, opened 2018-12-07
Fix xrange in Tribler/Test for Python 3

cclauss:


qstokkink:
Ok to test

qstokkink:
Also, I added you to the whitelist for PR request testing, such that we don't have to manually approve them for testing. Please don't bomb us with PRs :)

devos50:
@cclauss @qstokkink at this point, I would argue to enforce the Py3k check if possible before merging a PR. Do you agree?

Also, should we get #4094 merged first so we can work towards the first functional unit tests?

cclauss:
My vote would be to have the Py3 tests run but off of the critical path.  For now, the important thing is that the **Python 2 tests always pass**.  Here is an example of Travis CI running in __allow_failures__ mode.
allow failures  As long as the Python 2 test passes, the build is green.   We run the Python 3 test every time so we have a good status of where we are.

devos50:
@cclauss that would be a good idea. Running our unit tests sequentially takes almost an hour, which is undesired when we want to quickly test out whether our proposed changes work. For that reason, we are using [another (Python 2.7) library](https://github.com/tribler/gumby) to run all unit tests in parallel, in different "buckets". It would require some changes to Gumby and the configuration files. After #4094 is merged, I will deploy https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux_py3/ and make it to run all tests in the `Tribler/Core` directory sequentially. I predict that takes an acceptable amount of time. More fancy enhancements can be added later.

cclauss:
Understood.  Let's not mess with it then.  We just let the __Python 3 tests run in normal mode__.

devos50:
Thanks!

PR #4101, opened 2018-12-09
Fix buffer in Tribler/Test for Python 3

cclauss:


devos50:
ok to test

devos50:
Hmm it seems I didn't add you to the Tribler/Tribler team yet. The CI permission stuff might be fixed now.

devos50:
@cclauss It seems that sqlite does not accept the `memoryview` object. For this situation, we have written this: https://github.com/Tribler/py-ipv8/blob/master/ipv8/database.py#L42. You could consider using that?

devos50:
Thanks!

PR #4102, opened 2018-12-09
Fixed bug where torrent_info was removed in loop

devos50:
[Caught by the application tester](https://jenkins-ci.tribler.org/job/Deployment-Tests/job/Run_Tribler_Release/job/Run_Win7_32bit/65/console). It seems that `torrent_info` is removed while the interpreter executes a for-loop. To solve this, I've added a check in every loop for the existence of torrent info.


devos50:
retest this please

Issue #4103, opened 2018-12-09
GUI tests occasionally throws an error in Windows


devos50:
Error:

```
======================================================================
ERROR: test_add_download_url (Tribler.Test.GUI.test_gui.TriblerGUITest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\workspace\jenkins2\workspace\GH_Tribler_PR_Tests\PR_GUI_win64\tribler\Tribler\Test\GUI\test_gui.py", line 469, in test_add_download_url
    self.wait_for_signal(window.dialog.received_metainfo)
  File "C:\workspace\jenkins2\workspace\GH_Tribler_PR_Tests\PR_GUI_win64\tribler\Tribler\Test\GUI\test_gui.py", line 197, in wait_for_signal
    raise TimeoutException("Signal %s not raised within 10 seconds" % signal)
TimeoutException: Signal <bound PYQT_SIGNAL received_metainfo of StartDownloadDialog object at 0x000000000B155168> not raised within 10 seconds
-------------------- >> begin captured logging << --------------------
```

See https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_GUI_win64/826/

PR #4128, opened 2018-12-18
from six.moves.configparser import RawConfigParser

cclauss:
For compatibility with both Python 2 and Python 3


qstokkink:
retest this please. Failed on #4103

Issue #4131, opened 2018-12-22
Order price shown incorrectly in the Market UI


MattSkala:
##### Tribler version/branch+revision: #####
7.2.0-exp1

##### Operating system and version: #####
macOS 10.14.1

##### Steps to reproduce the behavior: #####
1. Go to the Market screen
2. Click on the "Buy BTC for MB" button
3. Fill in quantity: 1 BTC, price per unit: 100 MB/BTC

##### Expected behavior: #####
An order with the volume of 1 BTC and the price of 100 MB/BTC should be created.

##### Actual behavior: #####
An order with the volume of 1 BTC and the price of 0.01 MB/BTC is added to the list, the same price is shown in the order detail. There is only an issue with the price displayed in the UI. The bid itself is created with the correct amount, as one can verify via the API:

```
{
    "trader_id": "131dac16fa36cb3c9ff1d8382a4c3a862852da49",
    "timestamp": 1545497508.527097,
    "block_hash": "8a95ce69dfe0c3e1fafa8ae0b0e99325b20c26c6e5bfaff2e141b43d99ee7a11",
    "timeout": 3600,
    "traded": 0,
    "order_number": 11,
    "assets": {
        "second": {
            "amount": 100,
            "type": "MB"
        },
        "first": {
            "amount": 100000000,
            "type": "BTC"
        }
    }
}
```

devos50:
The price in the user interface is displayed in the `MB/BTC` asset pair. It might make more sense to reverse this and make it the same as the user has entered it.

Thanks for reporting!

devos50:
Fixed by #4132.

PR #4132, opened 2018-12-22
Added price details to GUI list with asks/bids

devos50:
For clarity, I now show the details of the price of each market order.

Related to #4131 

schermafbeelding 2018-12-22 om 21 01 55


MattSkala:
Are the price details really correct in this way? In the market for BTC/MB pair, the user enters the price in MB/BTC, but then the price in the list is calculated as BTC/MB. So IMO, the column title should say BTC/MB instead of MB/BTC.

PR #4135, opened 2018-12-27
Record transaction history

andrebreis:
This feature will, if the record_transactions config is set to True (default is False):
- Upon receiving a tx_init/tx_done block, create the transaction in the transactions table in the local database. If it already exists, if the received block has more recent information, it will update the corresponding entry.

This will allow for the creation of transaction history to study the market.


devos50:
ok to test

Tribler/community/market/community.py
```diff
@@ -164,6 +165,7 @@ def __init__(self, *args, **kwargs):
         self.matchmakers = set()
         self.request_cache = RequestCache()
         self.cancelled_orders = set()  # Keep track of cancelled orders so we don't add them again to the orderbook.
+        self.config = TriblerConfig()
```
devos50: Please do not create a new instance of `TriblerConfig`. Instead, use the configuration inside the `tribler_session` object (`tribler_session.config`). **make sure to check whether `tribler_session` exists or not!**

Tribler/community/market/core/transaction.py
```diff
@@ -167,6 +167,52 @@ def from_database(cls, data, payments):
 
         return transaction
 
+    @classmethod
+    def from_block(cls, block_info):
+        """
+        Create a Transaction object based on information in an tx_init/tx_done block.
```
devos50: in an -> in a

Tribler/community/market/database.py
```diff
@@ -243,6 +243,29 @@ def add_transaction(self, transaction):
         for payment in transaction.payments:
             self.add_payment(payment)
 
+    def insert_or_update_transaction(self, transaction):
+        """
+        Inserts or updates a specific transaction in the database, according to the timestamp
```
devos50: Please end this sentence with a period, and add that older transaction are not updated 👍

Tribler/community/market/community.py
andrebreis: Fixed.

Tribler/community/market/core/transaction.py
andrebreis: Fixed.

Tribler/community/market/database.py
andrebreis: Fixed.

PR #4136, opened 2019-01-06
WIP: Gigachannel gui refactor

devos50:


devos50:
retest this please

Tribler/Core/Modules/MetadataStore/OrmBindings/torrent_metadata.py
```diff
@@ -61,4 +62,56 @@ def get_auto_complete_terms(cls, keyword, max_terms, limit=100):
                     all_terms.add(term)
             return list(all_terms)
 
+        @classmethod
+        @db_session
+        def get_random_torrents(cls, limit):
+            """
+            Return some random torrents from the database.
+            """
+            return TorrentMetadata.select().random(limit)
```
qstokkink: Is it intended that there is not filtering on `REGULAR_TORRENT` here?

ichorid: The filtering should be applied, indeed. Otherwise, the thing will return both objects of class `TorrentMetadata` and `ChannelMetadata`, that is inherited from it.
This really depends on what we are trying to achieve with this. I would add an option like `regular_torrent_only=True` by default.

devos50: Thanks! Forgot about that.

qstokkink:
retest this please

qstokkink:
retest this please

devos50:
Flake8 together with diff-cover is not working correctly, in particular, the import order check. Since diff-cover ensures that only changes lines will be checked, it will make decisions based on a limited scope which are often incorrect. I will change the flake8 tool so it checks the full content of edited files in a PR.

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

ichorid:
Some remaining problems with the GUI: the "Apply Changes" button should only appear with notification on "Uncommitted changes", FireFox-notification style. At this moment, the notification does not appear on delete/restore actions in your channel.

ichorid:
oops, wrong one

PR #4155, opened 2019-01-15
Market API return order upon creation

andrebreis:


devos50:
ok to test

devos50:
@andrebreis could you fix the flake8 issues?

PR #4090, opened 2018-12-06
WIP-Collective Effort: Rework GUI to support GigaChannel

ichorid:
This patch replaces QT List-based view of the channels and their contents with a TableView-based widget that uses QT MVC idiom The provided TableModel is capable of loading data lazily from the REST endpoint. The search endpoint is modified to send the data in small pages and do sorts and searches. The TableView serves the model with a custom Delegate. In addition, data from the old Channels implementation can be served through (ugly) dual-stack adapters in the same table.

There is a LOT of bugs to fix and work to do to finish this thing:
* No unit tests provided! Feel free to add some or fix the old ones.
* Piles of dead code there. The new system probably made thousands of lines of code obsolete. Please, help in removing it.
* The new code is horrible. Do refactor as you please.
* The table appearance actually sucks... Whatever you do to make it look better - please, do it.
* Etc, etc...

Having said all that, there are only two positive sides about this PR:
1. It works
2. It is able to instantly search and display channel with millions of entries (tested).

### UPDATE
This PR now includes lots of other changes and fixes related to GigaChannels. At the point of its merging, it will provide everything necessary for GigaChannels to become production-ready.

Things left to do (beside usual requirements like test coverage):

- [x] Remove "Libnaclpk" prefix from public keys in mdblob format.
- [x] Add simple differential compression to squashed mdblobs (omit repeated fields after the first mdblob).
- [ ] Fix hover states for "subscribe" button.
- [x] Add channel size indicator to channel page.
- [x] Add "xxx" column to Pony DB
- [x] Add torrent health table
- [ ] Add ability to edit metadata entries
- [ ] Limit metadata size and add an indicator for it to the GUI
- [x] Add converter for foreign legacy channels
- [x] Add converter for the user's own legacy channel to proper GigaChannel
- [ ] Add error handling/delays/marking of broken channels
- [x] Add cleanup/handling of user's channel directory after db cleanup
- [ ] Add protection from lz4 archive bombs
- [x] Get rid of the `apsw` dependency and rewrite the upgrader so it uses `pony` or `sqlite3`.
- [x] Update the documentation regarding dependencies

PR #4167, opened 2019-01-24
Removed pymdht

devos50:
We are using EgbertDHT instead so there is no need anymore to load pymdht. This will also fix an error we observed in the 7.1.5 release.

When this PR is merged, we should remove it in #4090


devos50:
We could but I'm going for the slightly more conservative approach here and just disable the thing. We can remove all the pymdht-related configs and submodule in #4090.

PR #4169, opened 2019-01-25
Reorder tunnel close events to notify tunnel dispatcher early

Captain-Coder:
The circuit_dead() call is responsible for updating the destinations dict of the dispatcher. By having this call inside the deferred callback, it only runs after the tunnel is fully closed and paid for. During this time the dispatcher will drop any data still being sent over this circuit instead of already selecting another circuit to use.

So the call to circuit_dead() should happen as early as possible. This must be after the call to the super since we need the circuit to be in an unselectable state before we remove it from the dispatcher's destinations (to not risk it immediately reselecting the closing circuit).

### DIFF
--- a/Tribler/community/triblertunnel/community.py
+++ b/Tribler/community/triblertunnel/community.py
@@ -1,23 +1,27 @@
-import os
-import time
-import sys
+from __future__ import absolute_import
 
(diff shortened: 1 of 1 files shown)


devos50:
retest this please

devos50:
It seems that our windows tests are failing. I will check it out.

PR #4182, opened 2019-01-29
Already called twisted error captured in session instead of crashing Tribler

xoriole:
Already called twisted error is captured and logged in session handler instead of crashing Tribler

### DIFF
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -212,6 +212,10 @@ class Session(object):
                 self._logger.error("Internal Twisted response error, consider updating your Twisted version.")
                 return
 
+            if 'twisted.internet.error.AlreadyCalled' in text:
+                self._logger.error("Tried to cancel an already called event\n%s", text)
+                return
+
             # We already have a check for invalid infohash when adding a torrent, but if somehow we get this
             # error then we simply log and ignore it.
             if 'exceptions.RuntimeError: invalid info-hash' in text:


xoriole:
It seems Tunnel tests were failing for quite a while
failed_tests

xoriole:
retest this please

xoriole:
@qstokkink Are you referring to the tunnel full session tests or the AlreadyCalled error?
For the tunnel tests, the issue was due to libtorrent manager not shutting down properly which is because it was waiting for libtorrent session updates to confirm session is ready to be closed. My second commit in this PR fixes these failing tests.

For AlreadyCalled error, I think it is okay to catch in session handler than do a hard crash.

xoriole:
retest this please

qstokkink:
@xoriole I was referring to the AlreadyCalled error. In general I agree that we should never crash when running production. However, the AlreadyCalled error is one that should never happen by design and it may not be safe to recover from it by ignoring it. We probably still want to send an error report.

xoriole:
One of the reason to put this to ignore errors is that the traceback is not so useful to debug the issue.
```python
File "TriblerGUI\event_request_manager.py", line 133, in on_read_data
RuntimeError: Unhandled Error
File "run_tribler.py", line 94, in <module>

File "run_tribler.py", line 86, in start_tribler_core

File "lib\site-packages\twisted\internet\base.py", line 1243, in run

File "lib\site-packages\twisted\internet\base.py", line 1252, in mainLoop

File "lib\site-packages\twisted\internet\base.py", line 878, in runUntilCurrent

File "lib\site-packages\twisted\internet\tcp.py", line 292, in connectionLost

File "lib\site-packages\twisted\web\http.py", line 2190, in connectionLost

File "lib\site-packages\twisted\internet\base.py", line 99, in cancel

twisted.internet.error.AlreadyCalled: Tried to cancel an already-called event.
```
Anyway, lets hold the fix for the next release then. I'll exclude this for 7.2.0-rc1.

qstokkink:
@xoriole that one I would agree to ignore, that's inside Twisted and we have no way of catching it otherwise.

Can't we somehow disable catching this error when testing?

qstokkink:
Actually, when I force an `AlreadyCalled` error this code doesn't catch it.

qstokkink:
After some digging: `AlreadyCalled` is not equal to `AlreadyCalledError`, in which case this is perfectly fine to merge.

PR #4199, opened 2019-02-05
Update tribler_config.py

Dmole:
Save collected_torrents next to credit_mining because data > 1GB should have a user selected destination.


qstokkink:
Ok to test. [Very soon we're actually removing this feature and replacing it with something better](https://github.com/Tribler/tribler/pull/4090/files#diff-78f349d5baa61ccb8ba0a2c3d472ec6bL595).

devos50:
Yup, the torrent store will be removed in Tribler 7.3.

qstokkink:
`get_torrent_store_dir` is now completely gone in the `devel` branch.

PR #4220, opened 2019-02-12
Fixed e2e payout test

devos50:
After the change that randomly picks the first relay node, it is possible to have an e2e circuit that goes like 0 (downloader) -> 2 -> 0 -> 2 (seeder). In this case, node 1 will remain with a balance of zero after payouts.


qstokkink:
retest this please, for good measure

PR #4231, opened 2019-02-19
Removed statistics/overlays endpoint from Tribler

devos50:
In conjunction with https://github.com/Tribler/fake-tribler-api/pull/41


devos50:
The error in the GUI tests is a known one and we should look into it as soon as we can.

PR #4235, opened 2019-02-21
Added exit node ratio strategy

qstokkink:
This strategy attempts to maintain a ratio of:

`exit node` : `non-exit node` = `7` : `9`


qstokkink:
I am going to make the wild claim that the GUI tests failure is not my fault. Ready for review.

PR #4253, opened 2019-02-25
READY: Stop old entries from being pushed back by gossip

ichorid:
This PR fixes the infamous 💩-storm problem, when old versions of channel entries were pushed back by gossip, resulting in duplicate entries. In addition, it corrects the wrong "0 by default" timestamps policy on TorrentMetadata entries.
To prevent old 💩 from getting back, I've added a check into the `process_metadata` method, that forbids the addition of the metadata entry to the database, if its timestamp is lower than the `local_version` of the corresponding channel. An important exception to this rule is the case of `local_version` of 0, which indicates that the channel is in the preview mode.

EDIT:
I've also streamlined and generalized channels contents update procedure.


devos50:
@ichorid tests are failing :(

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -228,44 +226,34 @@ def process_payload(self, payload):
...
+            # Get the corresponding channel from local database to see if we really need to update our
+            # local contents of the channel by comparing the channel's local_version with the payload's timestamp.
+            # This check is necessary to prevent other peers pushing deleted entries into the
+            # channels we are subscribed to.
+            # If local channel version is 0, we are still in preview mode and willing to collect everything.
+            channel = self.ChannelMetadata.get(public_key=payload.public_key, id_=payload.origin_id)
+            if channel and (channel.local_version != 0) and (payload.timestamp <= channel.local_version):
+                return None, NO_ACTION
             elif payload.metadata_type == REGULAR_TORRENT:
+                # self.TorrentMetadata.
```
devos50: This can be removed?

ichorid: oops, totally forgot about this one ))

PR #4254, opened 2019-02-26
Quick fix for negative payouts

### LINKED ISSUE #4234: Losing tokens without downloading
I have no downloads but I am still losing tokens, how can that be?

image

I can reproduce this on version 7.2.2 and the 'next' branch from github.

The screenshot is made while I was using revision https://github.com/Tribler/tribler/commit/407afd318d95a62c6c99c30f074389906da0d043

devos50:
This PR implements my proposed solution 1 in #4234. In essence, we only continue the payout process when (parts of) the chain of the transaction counterparty can successfully be validated. This should prevent circuit peers from making payments to the next hop with funds they have not fully claimed yet.

Fixes #4234


devos50:
retest this please

PR #4260, opened 2019-02-27
Fixed test_subscriptions

devos50:
Using these subscribe/unsubscribe signals is not reliable on Windows. Instead, I use a `qWait` and take a screenshot afterwards.


devos50:
retest this please

PR #4275, opened 2019-03-04
unicode --> six.text_type in market/core payment.py and transaction.py

cclauss:


devos50:
@cclauss This PR is ready?

PR #4285, opened 2019-03-04
Fixed pretty_date method, hidden services test and window variable in TriblerRequestManager

devos50:
Fixes #4284 

This should fix the application testers, see https://jenkins-ci.tribler.org/job/Deployment-Tests/job/Run_Tribler_Custom/5/

PR #4286, opened 2019-03-04
unicode(s) --> six.text_type(s) in TorrentDef.py

cclauss:
Ready for review.

Please ignore the flake8 _undefined name_ on Python 3 only because the follow-up PR will target fixing instances of __unicode(s, encoding)__.

I have no idea what the __Combined coverage diff__ is trying to say.

### DIFF
--- a/Tribler/Core/TorrentDef.py
+++ b/Tribler/Core/TorrentDef.py
@@ -12,6 +12,8 @@ from hashlib import sha1
 
 from libtorrent import bencode, bdecode
 
+from six import text_type
+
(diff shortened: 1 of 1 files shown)


devos50:
@cclauss the combined coverage diff tool helps us to identify which lines you touched that are *not* covered by any unit test. We do not expect you to cover these lines that are not covered, however, we should be a bit more careful when reviewing since there is no unit test that executes the modified line.

devos50:
Also, there currently is a new error in the linux tests which you might occasionally trigger. #4285 will fix this.

PR #4287, opened 2019-03-04
READY: unicode(s, enc) --> ensure_unicode(s, enc)

cclauss:
As discussed in https://github.com/Tribler/tribler/issues/3883

In _Tribler/Core/Utilities/unicode.py_, __ensure_unicode()__ is defined to be like [__six.ensure_text()__](https://six.readthedocs.io/#six.ensure_text) except that encoding parameter is not optional.  See [__six>=1.12.0__](https://github.com/benjaminp/six/blob/master/CHANGES).

### DIFF
--- a/Tribler/Core/Utilities/unicode.py
+++ b/Tribler/Core/Utilities/unicode.py
@@ -7,7 +7,18 @@ from __future__ import absolute_import
 
 import sys
 
-from six import text_type
+from six import binary_type, text_type
+
+
+def ensure_text(s, encoding='utf-8', errors='strict'):
+    """Copied from https://github.com/benjaminp/six/blob/master/six.py
+    """
+    if isinstance(s, binary_type):
+        return s.decode(encoding, errors)
+    elif isinstance(s, text_type):
+        return s
+    else:
+        raise TypeError("not expecting type '%s'" % type(s))
 
 
 def bin2unicode(bin, possible_encoding='utf_8'):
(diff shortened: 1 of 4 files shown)


cclauss:
retest this please

Tribler/Core/Utilities/unicode.py
```diff
@@ -7,7 +7,18 @@
 
 import sys
 
-from six import text_type
+from six import binary_type, text_type
+
+
+def ensure_text(s, encoding='utf-8', errors='strict'):
```
qstokkink: Could you make the encoding a non-optional field here?

My fear is that people start using this as a drop-in replacement for `unicode`, whereas this is only a replacement for the functionality of `unicode(string, encoding)` and not `unicode(obj)`. For instance:

```python
>>> unicode(1)
u'1'
>>> unicode(1, "UTF-8")
TypeError: coercing to Unicode: need string or buffer, int found
```

cclauss: Let's pick a different function name to reduce confusion...

__ensure_unicode()__ is like __six.ensure_text()__ except that encoding parameter is ___not___ optional

qstokkink: Perfect 👍

devos50:
@cclauss rebasing on top of the latest commit on devel probably fixes the print statements issue

cclauss:
I can't really rebase...  Can we just ignore the _Print statements_ errors for this PR?

PR #4305, opened 2019-03-05
Remote search in Giga channel community

xoriole:
Added remote search in giga channel community.
- Sends request to 5 peers to get the response
- Responses are added to the top of the results in the GUI and also added to the database
- Changed community id


Tribler/community/gigachannel/community.py
```diff
@@ -92,13 +111,67 @@ def on_blob(self, peer, blob):
...
+        with db_session:
+            try:
+                metadata_result = self.metadata_store.process_compressed_mdblob(response.raw_blob)
+            except RuntimeError as re:
+                self.logger.error("Could not process mdblob: %s", re)
+                return
+
+            search_results = [(dict(type={REGULAR_TORRENT: 'torrent', CHANNEL_TORRENT: 'channel'}[r.metadata_type],
+                                    **(r.to_simple_dict()))) for (r, n) in metadata_result]
+        self.notifier.notify(SIGNAL_GIGACHANNEL_COMMUNITY, SIGNAL_ON_SEARCH_RESULTS, search_results, '')
```
devos50: `self.notifier` can be `None`, please check for that.

Tribler/Core/Modules/restapi/events_endpoint.py
```diff
@@ -183,6 +185,13 @@ def on_circuit_removed(self, subject, changetype, circuit, *args):
             }
             self.write_data({"type": "circuit_removed", "event": event})
 
+    def on_search_response(self, subject, changetype, circuit, *args):
+        event = {
+            "search_query": args[0],
```
devos50: I would just name this `query`.

Tribler/Core/Modules/restapi/search_endpoint.py
```diff
@@ -100,6 +99,9 @@ def render_GET(self, request):
             search_results = [(dict(type={REGULAR_TORRENT: 'torrent', CHANNEL_TORRENT: 'channel'}[r.metadata_type],
                                     **(r.to_simple_dict()))) for r in pony_query]
 
+        if self.session.lm.gigachannel_community and sanitized["first"] == 1:
```
devos50: A comment might be helpful here 👍

Tribler/community/gigachannel/community.py
```diff
@@ -92,13 +111,67 @@ def on_blob(self, peer, blob):
         if reply_blob:
             self.endpoint.send(peer.address, self.ezr_pack(self.NEWS_PUSH_MESSAGE, RawBlobPayload(reply_blob)))
 
+    # def create_search_request(self, query_filter, metadata_type='', sort_by="HEALTH", sort_asc=0, hide_xxx=True):
```
devos50: Either uncomment this code or remote it from the PR 👍

Tribler/community/gigachannel/community.py
```diff
@@ -27,18 +34,30 @@ class GigaChannelCommunity(Community):
     Community to gossip around gigachannels.
     """
 
-    master_peer = Peer(unhexlify("3081a7301006072a8648ce3d020106052b8104002703819200040448a078b597b62d3761a061872cd86"
-                                 "10f58cb513f1dc21e66dd59f1e01d582f633b182d9ca6e5859a9a34e61eb77b768e5e9202f642fd50c6"
-                                 "0b89d8d8b0bdc355cdf8caac262f6707c80da00b1bcbe7bf91ed5015e5163a76a2b2e630afac96925f5"
-                                 "daa8556605043c6da4db7d26113cba9f9cbe63fddf74625117598317e05cb5b8cbd606d0911683570ad"
-                                 "bb921c91"))
+    master_peer = Peer(unhexlify("3081a7301006072a8648ce3d020106052b8104002703819200040514b5f2a1d2293976a5627ce9f0148"
```
devos50: Please use the short key format. You can generate it with the key gen tool in IPv8.

Tribler/community/gigachannel/community.py
```diff
@@ -92,13 +111,67 @@ def on_blob(self, peer, blob):
...
+        self.search_counter += 1
+        request = SearchRequestPayload(self.search_counter, query_filter, metadata_type, sort_by, sort_asc, hide_xxx)
+        self._logger.debug("Started remote search for query:%s", query_filter)
+
+        for peer in self.get_peers()[:max_search_peers]:
+            self.endpoint.send(peer.address, self.ezr_pack(self.SEARCH_REQUEST, request))
+
+    @lazy_wrapper(SearchRequestPayload)
+    def on_search_request(self, peer, request):
+        unicode_query_filter = cast_to_unicode_utf8(request.query_filter)
```
devos50: Please test the mechanism with UTF-8 characters, i.e. by doing this in the unit test.

TriblerGUI/widgets/searchresultspage.py
```diff
@@ -56,3 +56,6 @@ def clicked_tab_button(self, _):
             self.model.type_filter = 'torrent'
 
         self.perform_search(self.query)
+
+    def received_search_result(self, response):
+        self.controller.load_remote_results(response)
```
devos50: Missing newline.

TriblerGUI/widgets/triblertablecontrollers.py
```diff
@@ -118,8 +118,18 @@ def on_search_results(self, response):
         if self.num_search_results_label:
             self.num_search_results_label.setText("%d results" % response['total'])
 
-        if response['first'] >= self.model.rowCount():
-            self.model.add_items(response['results'])
+        # if response['first'] >= self.model.rowCount():
```
devos50: Commented-out code

Tribler/community/gigachannel/community.py
qstokkink: Seconded, you can use: `python generate_key.py --public`

Tribler/community/gigachannel/community.py
```diff
@@ -4,17 +4,24 @@
...
 from Tribler.pyipv8.ipv8.lazy_community import lazy_wrapper
 from Tribler.pyipv8.ipv8.messaging.lazy_payload import VariablePayload
 from Tribler.pyipv8.ipv8.peer import Peer
+from Tribler.pyipv8.ipv8.requestcache import RequestCache
+from Tribler.util import cast_to_unicode_utf8
 
 minimal_blob_size = 200
 maximum_payload_size = 1024
 max_entries = maximum_payload_size // minimal_blob_size
+max_search_peers = 3
```
qstokkink: Why only 3?

Tribler/community/gigachannel/community.py
```diff
@@ -92,13 +111,63 @@ def on_blob(self, peer, blob):
...
+            "metadata_type": metadata_type,
+            "exclude_legacy": True
+        }
+
+        with db_session:
+            db_results, total = self.metadata_store.TorrentMetadata.get_entries(**request_dict)
+            if total > 0:
+                blob = entries_to_chunk(db_results[:max_entries], maximum_payload_size)[0]
+
+        self.endpoint.send(peer.address, self.ezr_pack(self.SEARCH_RESPONSE, SearchResponsePayload(request.id, blob)))
```
qstokkink: What if total = 0?

Tribler/community/gigachannel/community.py
qstokkink: Also, whoa, being able to SQL inject attack your own database wasn't that big of a deal. But now **you're processing raw queries from third parties**.

xoriole: @devos50 Will update the test for UTF-8 characters.
@qstokkink query filter is a search query from the GUI which is processed by Pony ORM, SQL injection is not possible. It is not a SQL query to be processed as it is.

Tribler/community/gigachannel/community.py
xoriole: done

Tribler/community/gigachannel/community.py
xoriole: Good point, done

Tribler/community/gigachannel/community.py
xoriole: How many would you suggest? 5/10?

Tribler/community/gigachannel/community.py
xoriole: good point, checked

Tribler/Core/Modules/restapi/events_endpoint.py
xoriole: removed query from event, results should be sufficient

Tribler/community/gigachannel/community.py
qstokkink: Had a bit of an in-depth look into how Pony/FTS5 structures queries:

Queries you get from third parties **are parsed** using the FTS5 grammar, but **queries do not interact with the SQL data**. In other words, you can still force someone to parse horrible queries, but you can't actually modify their data. For example, something like this is a perfectly legal query:

```python
u'(((("a"*)OR("b"*))AND(("c"*)OR("d"*))OR(("e"*)OR("f"*))OR(("g"*)OR("h"*))OR(("i"*)OR("j"*))OR(("k"*)OR("l"*))OR(("m"*)OR("n"*))OR(("o"*)OR("p"*))) NOT (((^"q"*)OR(^"r"*))OR((^"s"*)OR(^"t"*))OR((^"u"*)OR(^"v"*))OR((^"w"*)OR(^"x"*))OR((^"y"*)OR(^"z"*))OR((^"a"*)OR(^"b"*))OR((^"c"*)OR(^"d"*))OR((^"e"*)OR(^"f"*)))) OR (((("b"*)OR("c"*))AND(("d"*)OR("e"*))OR(("f"*)OR("g"*))OR(("h"*)OR("i"*))OR(("j"*)OR("k"*))OR(("l"*)OR("m"*))OR(("n"*)OR("o"*))OR(("p"*)OR("q"*))) NOT (((^"r"*)OR(^"s"*))OR((^"t"*)OR(^"u"*))OR((^"v"*)OR(^"w"*))OR((^"x"*)OR(^"y"*))OR((^"z"*)OR(^"a"*))OR((^"b"*)OR(^"c"*))OR((^"d"*)OR(^"e"*))OR((^"f"*)OR(^"g"*))))'
```

**tl;dr** this allows you to be a jerk to others, but it's not a security risk. This is fine for now.

Tribler/community/gigachannel/community.py
```diff
@@ -89,13 +104,64 @@ def on_blob(self, peer, blob):
...
+        if self.search_counter != response.id:
+            return
+        with db_session:
+            try:
+                metadata_result = self.metadata_store.process_compressed_mdblob(response.raw_blob)
+            except (TransactionIntegrityError, CacheIndexError) as err:
+                self._logger.error("DB transaction error when tried to process search payload: %s", str(err))
+                return
+
+            search_results = [(dict(type={REGULAR_TORRENT: 'torrent', CHANNEL_TORRENT: 'channel'}[r.metadata_type],
```
ichorid: `process_compressed_mdblob` can have other types of return values. The interpretation of these values is important. For example, it can respond with a `GOT_NEWER_VERSION` and a corresponding DB object. So, you have to check the resulting action type and act properly:
1. Only notify on `UPDATED_OUR_VERSION` or `UNKNOWN_CHANNEL'/`UNKNOWN_TORRENT`.
2. If there is `GOT_NEWER_VERSION` response, you should send back the packet with the updated version, as it is done in the gossip protocol.

Tribler/community/gigachannel/community.py
```diff
@@ -89,13 +104,64 @@ def on_blob(self, peer, blob):
...
+            db_results, total = self.metadata_store.TorrentMetadata.get_entries(**request_dict)
+            if total > 0:
+                result_blob = entries_to_chunk(db_results[:max_entries], maximum_payload_size)[0]
+        if result_blob:
+            self.endpoint.send(peer.address, self.ezr_pack(self.SEARCH_RESPONSE,
+                                                           SearchResponsePayload(request.id, result_blob)))
+
+    @lazy_wrapper(SearchResponsePayload)
+    def on_search_response(self, peer, response):
+        if self.search_counter != response.id:
```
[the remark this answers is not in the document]
xoriole: Number cache is a good point. I added that.

Tribler/community/gigachannel/community.py
xoriole: Do we want this for this release? I think we could optimize this later.

Tribler/community/gigachannel/request.py
```diff
@@ -0,0 +1,42 @@
...
+
+            # Check if all expected responses are received
+            if all(self.requested_peers.values()):
+                self.remove_request()
+
+            return True
+        return False
+
+    def remove_request(self):
+        if self.request_cache.has(self.prefix, self.number):
```
qstokkink: This happens automatically in the RequestCache on timeout.

Tribler/Test/Core/Modules/RestApi/test_events_endpoint.py
```diff
@@ -102,6 +105,8 @@ def send_notifications(_):
             self.session.notifier.notify(SIGNAL_RESOURCE_CHECK, SIGNAL_LOW_SPACE, None, {})
             self.session.notifier.notify(NTFY_CREDIT_MINING, NTFY_ERROR, None, {"message": "Some credit mining error"})
             self.session.notifier.notify(NTFY_TUNNEL, NTFY_REMOVE, Circuit(1234, None), 'test')
+            self.session.notifier.notify(SIGNAL_GIGACHANNEL_COMMUNITY, SIGNAL_ON_SEARCH_RESULTS, None,
```
devos50: I think you forgot to increase the `self.messages_to_wait_for` variable.

Tribler/community/gigachannel/community.py
```diff
@@ -89,13 +106,84 @@ def on_blob(self, peer, blob):
...
+        cleared before adding a new search request to prevent incorrect results being pushed to the GUI.
+        """
+        search_candidates = self.get_peers()[:max_search_peers]
+        search_request_cache = SearchRequestCache(self.request_cache, search_candidates)
+        self.request_cache.clear()
+        self.request_cache.add(search_request_cache)
+
+        search_request_payload = SearchRequestPayload(search_request_cache.number, query_filter.encode('utf8'),
+                                                      metadata_type, sort_by, sort_asc, hide_xxx)
+        self._logger.debug("Started remote search for query:%s", query_filter)
```
devos50: I think this can be on INFO level. Minor comment: please add a space after the `:`.

Tribler/community/gigachannel/request.py
```diff
@@ -0,0 +1,42 @@
...
+    This request cache keeps track of all outstanding search requests within the GigaChannelCommunity.
+    """
+    def __init__(self, request_cache, peers):
+        super(SearchRequestCache, self).__init__(request_cache, u"remote-search-request")
+        self.request_cache = request_cache
+        self.requested_peers = {hexlify(peer.mid): False for peer in peers}
+
+    @property
+    def timeout_delay(self):
+        return 10.0
```
devos50: We could adopt the old timeout that the `SearchCommunity` used.

Tribler/community/gigachannel/request.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        mid = hexlify(peer.mid)
+        if mid in self.requested_peers and not self.requested_peers[mid]:
+            self.requested_peers[mid] = True
+
+            # Check if all expected responses are received
+            if all(self.requested_peers.values()):
+                self.remove_request()
+
+            return True
```
devos50: Personally, I don't like these side effects of a method that seems to only process stuff to me. The exact meaning of these boolean return values are not immediately clear for developers. Can we avoid returning anything here?

xoriole: This result of this method is used to decide if the response is to be processed in the community. I can break it two methods if you prefer.

Tribler/community/gigachannel/community.py
ichorid: @xoriole , it is not about optimization. It is about other peers not being able to crash our Tribler instance with an unhandled response. And it is really easy to add response packet sent on `GOT_NEWER_VERSION`. Just copy it from gossip message handling in the same community.

Tribler/community/gigachannel/request.py
```diff
@@ -0,0 +1,42 @@
...
+        return False
+
+    def remove_request(self):
+        if self.request_cache.has(self.prefix, self.number):
+            self.request_cache.pop(self.prefix, self.number)
```
qstokkink: You can easily make this thread-safe by simply `try` `except KeyError` around the `pop()`.

Tribler/community/gigachannel/community.py
```diff
@@ -82,20 +98,103 @@ def on_blob(self, peer, blob):
...
+        with db_session:
+            try:
+                metadata_result = self.metadata_store.process_compressed_mdblob(response.raw_blob)
+            except (TransactionIntegrityError, CacheIndexError) as err:
+                self._logger.error("DB transaction error when tried to process search payload: %s", str(err))
+                return
+
+            search_results = [(dict(type={REGULAR_TORRENT: 'torrent', CHANNEL_TORRENT: 'channel'}[r.metadata_type],
+                                    **(r.to_simple_dict()))) for (r, _) in metadata_result
+                              if r and (r.metadata_type == CHANNEL_TORRENT or r.metadata_type == REGULAR_TORRENT)]
```
ichorid: these lines should be changed to: 
```python
                                    **(r.to_simple_dict()))) for (r, action) in metadata_result
                              if r and (a in [UPDATED_OUR_VERSION, UNKNOWN_TORRENT, UNKNOWN_CHANNEL])]
```
otherwise, the peer can send you DELETE entry, and then `process_payload` will return `(None, DELETED_METADATA)` and trying to get `None.metadata_type` will result in a crash.

ichorid: Oh, sorry, I see you already check for `r` existence ))

PR #4308, opened 2019-03-05
map() --> comprehensions

cclauss:


devos50:
@cclauss just a quick inspection reveals that we can remove the whole `VideoUtility.py` file, so no need to make changes in this file. You want to do this in this PR?

PR #4318, opened 2019-03-06
str are already Unicode utf-8 by default in Python 3

cclauss:
```
ERROR: test_get_index
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/internet/utils.py", line 201, in runWithWarningsSuppressed
    reraise(exc_info[1], exc_info[2])
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/python/compat.py", line 464, in reraise
    raise exception.with_traceback(traceback)
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/internet/utils.py", line 197, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Test/test_as_server.py", line 62, in check
    result = fun(*argv, **kwargs)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Test/Core/test_torrent_def.py", line 445, in test_get_index
    self.assertEqual(t.get_index_of_file_in_files('a.txt'), 0)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Core/TorrentDef.py", line 790, in get_index_of_file_in_files
    intorrentpath = maketorrent.pathlist2filename(file_dict['path'])
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Core/Utilities/maketorrent.py", line 258, in pathlist2filename
    return fullpath.decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

### DIFF
--- a/Tribler/Core/Utilities/maketorrent.py
+++ b/Tribler/Core/Utilities/maketorrent.py
@@ -250,10 +250,7 @@ def filename2pathlist(path, skipfirst=False):
 
 def pathlist2filename(pathlist):
     """ Convert a multi-file torrent file 'path' entry to a filename. """
-    fullpath = ''
-    for elem in pathlist:
-        fullpath = os.path.join(fullpath, elem)
-
+    fullpath = os.path.join(*pathlist)
     try:
         return fullpath.decode('utf-8')
     except AttributeError:
--- a/Tribler/Core/Utilities/maketorrent.py
+++ b/Tribler/Core/Utilities/maketorrent.py
@@ -257,7 +257,7 @@ def pathlist2filename(pathlist):
     try:
         return fullpath.decode('utf-8')
     except AttributeError:
-        return fullpath  # Python 3: AttributeError: 'str' object has no attribute 'decode' 
+        return fullpath  # Python 3: AttributeError: 'str' object has no attribute 'decode'
     except UnicodeDecodeError:
         charenc = chardet.detect(fullpath)['encoding']
         if not charenc:
--- a/Tribler/Core/Utilities/maketorrent.py
+++ b/Tribler/Core/Utilities/maketorrent.py
@@ -256,6 +256,8 @@ def pathlist2filename(pathlist):
 
     try:
         return fullpath.decode('utf-8')
+    except AttributeError:
+        return fullpath  # Python 3: AttributeError: 'str' object has no attribute 'decode' 
     except UnicodeDecodeError:
         charenc = chardet.detect(fullpath)['encoding']
         if not charenc:
--- a/Tribler/Core/Utilities/maketorrent.py
+++ b/Tribler/Core/Utilities/maketorrent.py
@@ -5,6 +5,7 @@ Author(s): Arno Bakker, Bram Cohen
 """
 from __future__ import absolute_import
 
+import codecs
 import logging
 import os
 from copy import copy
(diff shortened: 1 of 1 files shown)


cclauss:
please retest this

cclauss:
retest this please

qstokkink:
Hmm.. from the function definition is seems that it is expecting a `bytes`-like object with an -apparently unknown- encoding.

It seems to me like the fix for this is to make all methods accessing `pathlist2filename` feed it an encoded `bytes` instead.

cclauss:
What do we want `pathlist2filename()` to OUTPUT? filename as str or filename as bytes?  The current code makes me thing that we want the output to be filename as str (i.e. Unicode utf-8).

If `pathlist` is ('a', 'b', 'c') then this PR does the right thing but if `pathlist` is (b'a', b'b', b'c') then what should we do?

cclauss:
After further consideration, I stand behind this PR as it is.  If the AttributeError is raised then we are in Python 3 with `filepath` as a str so we can just return it.

If `filepath` is bytes then it has a .encode() attribute which will either run successfully or raise an UnicodeDecodeError which is explicitly handled just like in legacy Python.

cclauss:
Retest this please

cclauss:
We should probably start considering adoption of https://docs.python.org/3/library/typing.html

qstokkink:
It seems the input comes from the `libtorrent` module. Specifically the metainfo dict:

https://github.com/Tribler/tribler/blob/5f60319d0c94335fcf075a71b43b1f70c70b56bd/Tribler/Core/TorrentDef.py#L787-L790

@devos50 since you looked at this last, do you by chance know if the metainfo dict is `bytes` or `str` in Python 3?

devos50:
@qstokkink the metainfo dictionary that is spit out by libtorrent is all `bytes`.

cclauss:
```python
def as_raw_unicode(value=u""):
    if isinstance(value, bytes):
        return codecs.decode(value, 'raw_unicode_escape')
    else:
        return six.text_type(value)
```

qstokkink:
In that case we should replace:

```python
return fullpath.decode('utf-8')
```

...with...

```python
return codecs.decode(fullpath, 'utf-8')
```

...in this PR I think.

cclauss:
Making that change.  It does raise the question for me, for want instances of `s` is:
`s.decode('utf-8') != codecs.decode(s, 'utf-8')` ?

qstokkink:
> Making that change. It does raise the question for me, for want instances of s is:
s.decode('utf-8') != codecs.decode(s, 'utf-8') ?

Woops good point, for some reason I got it in my head that bytes no longer supports any `decode`.

cclauss:
The Python Unicode HOWTO https://docs.python.org/3/howto/unicode.html says:
> The low-level routines for registering and accessing the available encodings are found in the [__codecs__](https://docs.python.org/3/library/codecs.html) module. Implementing new encodings also requires understanding the codecs module. However, the encoding and decoding functions returned by this module are usually more low-level than is comfortable, and writing new encodings is a specialized task, so the module won’t be covered in this HOWTO.

PR #4319, opened 2019-03-06
Fix for DHT not making any requests

egbertbouman:


Tribler/community/triblertunnel/community.py
```diff
@@ -460,6 +461,15 @@ def monitor_downloads(self, dslist):
                 hops[info_hash] = hop_count
                 if download.get_state().get_status() in [DLSTATUS_DOWNLOADING, DLSTATUS_SEEDING, DLSTATUS_METADATA]:
                     active_downloads_per_hop[hop_count] = active_downloads_per_hop.get(hop_count, 0) + 1
+
+                    # Ugly work-around for the libtorrent DHT not making any requests
+                    # after a period of having no circuits
```
arvidn: Is there a fix I could put in libtorrent for this?
Iirc, there is a limit on the number of times a node is tried, and it’s ignored if it fails sufficient number of times. Perhaps there should be a way to reset these counters, or maybe some logic in libtorrent to detect that the problem is with the internet connection, rather than the nodes (and not increment the fail counters). I’m open to suggestions

egbertbouman: I did some more testing, and the download does recover after a while. I think it just waits for the next DHT announce, which takes 15min. During this period, the DHT has no peers and doesn't do anything. Eventually, while re-announcing, the DHT routers are contacted and I see get_peers messages for the torrent I'm trying to download.

I'm not sure what the best way to fix this is. Resetting the counters probably won't help in our case, as the DHT has no peers. Perhaps lowering the dht announce interval if the dht has no peers available? BTW, I saw that there is also a force_dht_announce method in libtorrent, so at least we should be using that instead of restarting the download.

egbertbouman:
retest this please

PR #4324, opened 2019-03-07
Moved torrent validation to libtorrent

devos50:
Instead of having a manual `.torrent` validator (which is extremely strict), we can simply try to convert the meta info dictionary to a `libtorrent.torrent_info` object. The constructor of `torrent_info` contains several checks to see whether the torrent is valid or not. If not, it will raise a `RuntimeError` with the message about what is wrong exactly.


devos50:
retest this please

PR #4327, opened 2019-03-07
Move channel dir processing to a separate thread

ichorid:
This PR moves the processing of GigaChannel data downloaded with libtorrent to a background thread spawned with `deferToThread`. Holding DB lock for a long time **severely** affects Twisted reactor performance. Thus, this PR implements a congestion-control algorithm to dynamically break the workload into smaller batches. The default target time for processing a single batch - **100ms** - was devised by experimentation. If the routine runs in the background thread, 50ms `sleep` intervals are inserted between batches, so that other threads got a chance to acquire the DB lock. Again, this proved to be necessary by hands-on experience.
Community payloads are still processed on the main thread.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -331,7 +331,7 @@ class MetadataStore(object):
                                                    infohash=database_blob(payload.infohash))
         if node and node.timestamp < payload.timestamp:
             node.delete()
-            result.append((node, DELETED_METADATA))
+            result.append((None, DELETED_METADATA))
 
         # Check for the older version of the same node
         node = self.TorrentMetadata.get_for_update(public_key=database_blob(payload.public_key), id_=payload.id_)
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -7,7 +7,6 @@ from datetime import datetime, timedelta
 from time import sleep
 
 import lz4.frame
-
 from pony import orm
 from pony.orm import db_session
 
@@ -281,7 +280,8 @@ class MetadataStore(object):
                 elif target_coeff > 1.0:
                     self.batch_size = int(float(self.batch_size) / target_coeff)
                 self.batch_size += 1  # we want to guarantee that at least something will go through
-            self._logger.debug(("Added payload batch to DB (entries, seconds): %i %f", self.batch_size, batch_end_time))
+            self._logger.debug(("Added payload batch to DB (entries, seconds): %i %f", (self.batch_size,
+                                float(batch_end_time.total_seconds()))))
             start = end
         return result
 
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import logging
 import os
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 from time import sleep
 
 import lz4.frame
+
 from pony import orm
 from pony.orm import db_session
 
@@ -280,8 +281,8 @@ class MetadataStore(object):
                 elif target_coeff > 1.0:
                     self.batch_size = int(float(self.batch_size) / target_coeff)
                 self.batch_size += 1  # we want to guarantee that at least something will go through
-            self._logger.debug(("Added payload batch to DB (entries, seconds): %i %f", (self.batch_size,
-                                float(batch_end_time.total_seconds()))))
+            self._logger.debug(("Added payload batch to DB (entries, seconds): %i %f",
+                                (self.batch_size, float(batch_end_time.total_seconds()))))
             start = end
         return result
 
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -3,7 +3,8 @@ from __future__ import absolute_import
 import logging
 import os
 from binascii import hexlify
-from datetime import datetime
+from datetime import datetime, timedelta
+from time import sleep
 
 import lz4.frame
 
@@ -14,7 +15,7 @@ from Tribler.Core.Modules.MetadataStore.OrmBindings import (
     channel_metadata, channel_node, misc, torrent_metadata, torrent_state, tracker_state)
 from Tribler.Core.Modules.MetadataStore.OrmBindings.channel_metadata import BLOB_EXTENSION
 from Tribler.Core.Modules.MetadataStore.serialization import (
-    CHANNEL_NODE, CHANNEL_TORRENT, DELETED, REGULAR_TORRENT, read_payload_with_offset, time2int)
+    CHANNEL_TORRENT, DELETED, REGULAR_TORRENT, read_payload_with_offset, time2int)
 from Tribler.Core.exceptions import InvalidSignatureException
 from Tribler.pyipv8.ipv8.database import database_blob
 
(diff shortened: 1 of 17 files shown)


ichorid:
@devos50 , could you please look why the GUI test failing?

devos50:
@ichorid it seems that relying on signals in PyQt5 on Windows is unreliable. I will rewrite this test to avoid waiting for signals.

ichorid:
retest this please

devos50:
@ichorid I haven't rewritten the GUI test yet, I will do so tomorrow.

ichorid:
))

devos50:
Hopefully fixed the GUI tests in #4375

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -140,13 +141,34 @@ def render_POST(self, request):
...
                 request.setResponseCode(http.NOT_FOUND)
                 return json.dumps({"error": "this channel cannot be found"})
 
             channel.subscribed = to_subscribe
-            channel.local_version = 0
+
+        def delete_channel():
+            # TODO: this should be eventually moved to a garbage-collector like subprocess in MetadataStore
+            with db_session:
+                c = self.session.lm.mds.ChannelMetadata.get_for_update(public_key=database_blob(self.channel_pk))
```
devos50: Minor comment: I always try to avoid one letter variable names. Could you rename it to `channel` for instance?

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -140,13 +141,34 @@ def render_POST(self, request):
...
+                c.local_version = 0
+            while True:
+                if self.session.lm.mds._shutting_down:
+                    self.session.lm.mds._db.disconnect()
+                    return
+                with db_session:
+                    contents = c.contents[:100]
+                    if len(contents) > 3:  # leave just 3 entries for preview purposes
+                        # FIXME: PONY BUG - can't use bulk delete because of broken FOREIGN KEY constraints
+                        for t in contents[:3]:
```
devos50: See my comment about one letter variable names.

Tribler/Test/Core/Modules/RestApi/test_metadata_endpoint.py
```diff
@@ -123,6 +124,30 @@ def test_subscribe(self):
...
+        with db_session:
+            channel = self.session.lm.mds.ChannelMetadata.select()[:][0]
+            channel.subscribed = True
+
+        self.should_check_equality = False
+        post_params = {'subscribe': '0'}
+        channel_pk = hexlify(channel.public_key)
+
+        def on_response(_):
+            # stupid workaround for polling results of a background process
```
devos50: I don't get why this complicated logic (with a sleep!) is necessary?

Tribler/Core/Modules/restapi/metadata_endpoint.py
ichorid: Long and descriptive variable names should only be used for long-lived variables. If a variable lives only a few lines of code, and every usage fits in a single screen, it is perfectly fine to use very short, even single-character name. It makes lines shorter and more readable. The programmer "loads" the variable meaning in his brain, and it persists there for a few lines of code.

In fact, using single-character names even at long spans of code is fine if it fits one of two special use cases:
1) There is an unspoken arrangement that some letter stands for a variable with a special usage type, e.g. `I` and `j` stand for a loop counter, `g` stands for `lambda` in Pony expressions, etc;
2) It stands for a very often used pointer, like `self` (mostly used in C).

Tribler/Test/Core/Modules/RestApi/test_metadata_endpoint.py
ichorid: https://github.com/Tribler/tribler/blob/0149098456ae1cbba4d44ead3e0c2ccb6ec08c4d/Tribler/Core/Modules/restapi/metadata_endpoint.py#L171

The deletion is handled by a background thread without `Deferred` attached to it. So, I have to poll the database constantly to check if it is finished.

Tribler/Core/Modules/restapi/metadata_endpoint.py
devos50: In this situation, the fact that you are using a one letter variable name induced additional time to understand what's going on. When reviewing the code, I did not understand what the type of `c` was (also complicated by the fact that Python is essentially typeless), so I had to read the rest of the line in more detail. If the variable would be named `channel`, it would be immediately clear for me and it would likely avoid the (minor) overhead to defer the type of `c`.

> It makes lines shorter

While it makes lines shorter, I prefer clearness above shorter and more compact code. Since we do not have type information and annotations, naming variables appropriately in Python is important.

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -143,98 +149,161 @@ def __init__(self, db_filename, channels_dir, my_key):
...
-    def process_channel_dir(self, dirname, channel_id):
+    def process_channel_dir(self, dirname, channel_id, external_thread=True):
         """
         Load all metadata blobs in a given directory.
         :param dirname: The directory containing the metadata blobs.
+        :param external_thread: indicate to lower levels that this is running on a background thread
         :param channel_id: public_key of the channel.
         """
-        # We use multiple separate db_sessions here to limit memory usage when reading big channels
+        # We use multiple separate db_sessions here to limit the memory and reactor time impact,
```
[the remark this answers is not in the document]
ichorid: This comment applies to the whole method, why do you think it should be moved?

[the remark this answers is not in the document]
devos50: I actually meant to respond to the second line of comments (talking about the existence of channels). But on second thought, I guess this is fine since the if statement is only a few lines below the comment.

Tribler/Test/Core/Modules/RestApi/test_metadata_endpoint.py
devos50: Right. Is this the only location where you have this kind of check? Otherwise, it might be a good idea to extract this code and make a generic method out of it.

Also, please use asynchronous sleeps instead of `time.sleep`.

Tribler/Core/Modules/restapi/metadata_endpoint.py
ichorid: You're right.
```
channel c = self.session.lm.mds.ChannelMetadata.get_for_update(public_key=database_blob(self.channel_pk))```
would provide a mental shortcut, however
```
c = self.session.lm.mds.ChannelMetadata.get_for_update(public_key=database_blob(self.channel_pk)) ```
requires the programmer to read the whole line, which is inconvenient.
Unfortunately, Python is typeless.
For me, combination of  `ChannelMedatata` with `get` unambigously point to the fact that `c` is of ChannelMetadata type. But that is because I am familiar with Pony, and I use a consistent notation for one-letter identifiers.
The fact that it took you some additional effort to derive the data type tells us that for an average Python programmer Pony notation does not indicate the type clearly enough.

I will change the identifier.

ichorid:
retest this please

Tribler/Core/Modules/restapi/search_endpoint.py
```diff
@@ -109,14 +106,29 @@ def render_GET(self, request):
...
+            request.write(json.dumps({
+                "results": search_results,
+                "first": sanitized["first"],
+                "last": sanitized["last"],
+                "sort_by": sanitized["sort_by"],
+                "sort_asc": sanitized["sort_asc"],
+                "total": total
+            }))
+            request.finish()
+        deferToThread(search_db).addCallback(on_search_results)
```
devos50: Do you have any hint on how much this speeds up the search?

ichorid:
This is not about speeding up the search. This is about freeing the reactor
to do other stuff, like moving ipv8 packets.

пт, 15 марта 2019 г., 10:37 Martijn de Vos <notifications@github.com>:

PR #4332, opened 2019-03-08
Refactored TorrentDef

devos50:
This is a first attempt to refactor the `TorrentDef` class, which contained a significant amount of old, unused code and outdated comments. In particular, I made the following changes:
- Changed the constructor of `TorrentDef`. It accepts a metainfo and/or some torrent parameters.
- Removed the notion of finalizing a `TorrentDef` object. Instead, calling the `save` method will generate metainfo for you and optionally write away a `.torrent` file if you pass the `torrent_filepath` parameter.
- Removed the methods to create some torrent information in `maketorrent.py`. Instead, we are now using the utility method in `torrent_utils.py` which used libtorrent functionality to generate metainfo. This results in the removal of a significant amount of code.
- Removed the name parameter when making a HTTP request to the `createtorrent` endpoint - libtorrent does not allow us to name torrent files. This should be handled by the gigachannels logic.


ichorid:
Holy balls! Net -1k lines!

PR #4351, opened 2019-03-09
READY: Pass a str, not bytes into TraderId()

cclauss:
```
ERROR: Test whether an order is unreserved when address resolution fails
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Test/Community/Market/test_community.py", line 279, in test_address_resolv_fail
    yield self.introduce_nodes()
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/pyipv8/ipv8/test/base.py", line 160, in introduce_nodes
    node.overlay.walk_to(other.endpoint.wan_address)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/pyipv8/ipv8/community.py", line 323, in walk_to
    packet = self.create_introduction_request(address)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/community/market/community.py", line 279, in create_introduction_request
    extra_payload = InfoPayload(TraderId(self.mid), Timestamp.now(), self.is_matchmaker)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/community/market/core/message.py", line 16, in __init__
    raise ValueError("Trader id must be a string")
ValueError: Trader id must be a string
```

PR #4353, opened 2019-03-09
TraderId() takes bytes, not str

cclauss:
Opposite approach to #4351
```python
class TraderId(object):
    def __init__(self, trader_id):
        # type: (bytes) -> None
```
https://mypy.readthedocs.io/en/latest/cheat_sheet.html#classes

```
ERROR: Test whether an order is unreserved when address resolution fails
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/Test/Community/Market/test_community.py", line 279, in test_address_resolv_fail
    yield self.introduce_nodes()
  File "/home/jenkins/.local/lib/python3.5/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/pyipv8/ipv8/test/base.py", line 160, in introduce_nodes
    node.overlay.walk_to(other.endpoint.wan_address)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/pyipv8/ipv8/community.py", line 323, in walk_to
    packet = self.create_introduction_request(address)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/community/market/community.py", line 279, in create_introduction_request
    extra_payload = InfoPayload(TraderId(self.mid), Timestamp.now(), self.is_matchmaker)
  File "/home/jenkins/workspace/test_tribler_python3/tribler/Tribler/community/market/core/message.py", line 16, in __init__
    raise ValueError("Trader id must be a string")
ValueError: Trader id must be a string
```

### DIFF
--- a/Tribler/community/market/core/tick.py
+++ b/Tribler/community/market/core/tick.py
@@ -56,7 +56,7 @@ class Tick(object):
         is_ask, traded, block_hash = data
 
         tick_cls = Ask if is_ask else Bid
-        order_id = OrderId(TraderId(str(trader_id)), OrderNumber(order_number))
+        order_id = OrderId(TraderId(trader_id), OrderNumber(order_number))
         return tick_cls(order_id, AssetPair(AssetAmount(asset1_amount, str(asset1_type)),
                                             AssetAmount(asset2_amount, str(asset2_type))),
                         Timeout(timeout), Timestamp(timestamp), traded=traded, block_hash=str(block_hash))
--- a/Tribler/community/market/core/tick.py
+++ b/Tribler/community/market/core/tick.py
@@ -62,8 +62,8 @@ class Tick(object):
                         Timeout(timeout), Timestamp(timestamp), traded=traded, block_hash=str(block_hash))
 
     def to_database(self):
-        return (text_type(self.order_id.trader_id), int(self.order_id.order_number), self.assets.first.amount,
-                text_type(self.assets.first.asset_id), self.assets.second.amount,
+        return (database_blob(self.order_id.trader_id.to_string()), int(self.order_id.order_number),
+                self.assets.first.amount, text_type(self.assets.first.asset_id), self.assets.second.amount,
                 text_type(self.assets.second.asset_id), int(self.timeout), float(self.timestamp), self.is_ask(),
                 self.traded, database_blob(self.block_hash))
 
--- a/Tribler/Test/Community/Market/test_block.py
+++ b/Tribler/Test/Community/Market/test_block.py
@@ -23,14 +23,14 @@ class TestMarketBlock(AbstractServer):
     def setUp(self):
         yield super(TestMarketBlock, self).setUp()
 
-        self.ask = Ask(OrderId(TraderId('0' * 40), OrderNumber(1)),
+        self.ask = Ask(OrderId(TraderId(b'0' * 40), OrderNumber(1)),
                        AssetPair(AssetAmount(30, 'BTC'), AssetAmount(30, 'MB')), Timeout(30), Timestamp(0.0), True)
-        self.bid = Ask(OrderId(TraderId('1' * 40), OrderNumber(1)),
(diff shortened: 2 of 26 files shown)


cclauss:
Looks like many calls to__TraderId()__ are passing in __str__, not __bytes__.

cclauss:
It is unclear to me how to fix this Linux test is failing:
* https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux/1672/testReport/Tribler.Test.Community.Market.test_database/TestDatabase/test_add_remove_tick/
* https://github.com/Tribler/tribler/blob/devel/Tribler/Test/Community/Market/test_database.py#L166

cclauss:
The database tables are set up as: __trader_id            TEXT NOT NULL__
* https://github.com/Tribler/tribler/blob/devel/Tribler/community/market/database.py#L25-L109

Overall, I would feel more comfortable if we adopted the _Unicode Sandwich_ approach and we moved _str_ inside our program and wrote _str_ to the database and converted to _bytes_ only just before sending to the wire and just after receiving from the wire.

devos50:
> The database tables are set up as: trader_id TEXT NOT NULL Is that OK for storing bytes in the DB?

Yes, that should be ok 👍 

Actually, I want to refactor the `TraderId` a bit so it does not take a hexlified (byte) string but instead, a 20-bytes member ID. I would like to do so before the release of Tribler 7.3. Do you mind if I append these changes to this PR?

cclauss:
Feel free to append.

devos50:
@cclauss for now, I decided to just fix the database issue.

Please have a look at my commit. I adopted the same code conventions as we use in the `TrustChainBlock` class in the IPv8 library (which is a submodule of Tribler and fully Python 2/3 compatible). The most important changes are that I check for `binary_type` in the `__init__` method of `TraderId` (instead of `bytes`), and that I convert the `TraderId` object to a `buffer`/`bytes` before adding it to the database.

We should do the same with the `AssetAmount` and `AssetPair` classes.

cclauss:
This does not look right to me...

Tribler/community/market/core/message.py:18
```python
        trader_id = trader_id if isinstance(trader_id, bytes) else str(trader_id)

        if not isinstance(trader_id, binary_type):
            raise ValueError("Trader id must be a binary type")
```

devos50:
> In Python 3, why convert to Unicode and then raise if it is not bytes?

As far as I see, we are not converting to Unicode. The reason for this statement is to maintain compatibility with Python 2, in particular, to deal with what the sqlite database returns to us.

In Python 2, a query to the sqlite database returns `buffer` objects for columns with the `TEXT` type. In Python 3, `bytes` objects are returned. In Python 2, `buffer` is not an instance of `binary_type`, so I convert it to a `str`. We should, in general, make sure to pass `str`/`bytes` objects to the `__init__` method of `TraderId`, which is what you did in this PR 👍

cclauss:
Would it work to cast to six.binary_type in the else clause instead of str because str in Python 3 is Unicode?

devos50:
@cclauss yup, that might be safer indeed in Python 3. I will change it.

Tribler/community/market/core/tick.py
```diff
@@ -62,8 +62,8 @@ def from_database(cls, data):
                         Timeout(timeout), Timestamp(timestamp), traded=traded, block_hash=str(block_hash))
 
     def to_database(self):
-        return (text_type(self.order_id.trader_id), int(self.order_id.order_number), self.assets.first.amount,
-                text_type(self.assets.first.asset_id), self.assets.second.amount,
+        return (database_blob(self.order_id.trader_id.to_string()), int(self.order_id.order_number),
```
xoriole: Casting trader_id to `database_blob` might be necessary in `to_database()` of the following files as well.
- community/market/core/payment.py
- community/market/core/order.py
- community/market/core/transaction.py

devos50: Will do this later 👍

PR #4357, opened 2019-03-10
WIP/ON HOLD: Twistified startup upgrader

qstokkink:
afbeelding

@ichorid 

Finally found a way to get the core's upgrader ticks through to the GUI. We probably don't want to merge this PR as-is, but it should be able to serve as a reference for the eventual solution.

Also, the upgrader text looks a bit weird.

### DIFF
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -12,7 +12,7 @@ import sys
 from threading import RLock
 
 from twisted.internet import threads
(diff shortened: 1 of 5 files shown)


qstokkink:
This PR has been taken over by @xoriole

ichorid:
I'm working on a patch that makes the upgrader offline again. The patch is based on this PR (merged it by hand in my branch) and, thus, makes this PR obsolete. 

@qstokkink , thanks for your help.

PR #4368, opened 2019-03-12
Made Tribler compatible with libtorrent 1.2.0

devos50:
Two months ago, libtorrent 1.2.0 has been released with many fixes and improvements. This PR should make Tribler compatible with 1.2.0.

While having overlap with #3437, this PR ensures that Tribler still works with libtorrent versions between 1.0.0 and 1.1.0 (in contrast to #3437).

**Important:** this PR drops support for libtorrent 0.16.19 and older. Since 0.16.19 has been released almost four years ago, I think we can safely drop support for these old versions of libtorrent.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -254,15 +254,15 @@ class LibtorrentDownloadImpl(DownloadConfigInterface, TaskManager):
             atp = {}
             atp["save_path"] = os.path.normpath(os.path.join(get_default_dest_dir(), self.get_dest_dir()))
             atp["storage_mode"] = lt.storage_mode_t.storage_mode_sparse
-            atp["paused"] = True
-            atp["auto_managed"] = False
-            atp["duplicate_is_error"] = False
             atp["hops"] = self.get_hops()
+            atp["flags"] = lt.add_torrent_params_flags_t.flag_paused | \
+                           lt.add_torrent_params_flags_t.flag_duplicate_is_error | \
+                           lt.add_torrent_params_flags_t.flag_update_subscribe
 
             if share_mode:
-                atp["flags"] = lt.add_torrent_params_flags_t.flag_share_mode
+                atp["flags"] = atp["flags"] | lt.add_torrent_params_flags_t.flag_share_mode
             if upload_mode:
-                atp["flags"] = lt.add_torrent_params_flags_t.flag_upload_mode
+                atp["flags"] = atp["flags"] | lt.add_torrent_params_flags_t.flag_upload_mode
 
             self.set_checkpoint_disabled(checkpoint_disabled)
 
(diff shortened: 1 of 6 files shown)


arvidn:
I'm interested in any feedback of (unnecessary) backwards compatibility issues. I tried to keep it mostly compatible (apart from some of the major changes in using `std` types instead of `boost` and the alert pointers). I'd be happy to fix things in the next 1.2 point release!

devos50:
@arvidn thanks! Getting stuff to work was less effort than I initially thought (judging from the changelog). Most of the breaking stuff is in the C++ side which we do not really touch. We have to place a few conditionals here and there in order to support older versions but if there are only a few, that should be ok.

It seems that a few of our unit tests are breaking with libtorrent 1.2.0, even though the testers do not show this (they use libtorrent 1.1.x for testing). The process seems to freeze and I am not entirely sure why. The minimal test that is crashing creates two Tribler sessions, one that start seeding a single file and another Tribler instance that downloads the seeded file. When printing all the libtorrent alerts of both the seeding and downloading libtorrent session always results in the `libtorrent.save_resume_data_alert` being the last alert.

```
<class 'libtorrent.save_resume_data_alert'>  # print type(alert)
video.avi resume data generated  # print(alert)
{'active_time': 5, 'num_incomplete': 16777215, 'seed_mode': 0, 'pieces': '\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01', 'paused': 0, 'seeding_time': 4, 'info-hash': '\x9d[-\xbcR\x80s%\xbf\xc2\x8dh\x8f+\xb0?\x8d\x1evg', 'upload_rate_limit': -1, 'max_uploads': -1, 'max_connections': 16777215, 'num_downloaded': 16777215, 'total_downloaded': 0, 'file-format': 'libtorrent resume file', 'peers6': '', 'added_time': 1552400731, 'file_priority': [1], 'last_seen_complete': 1552400735, 'total_uploaded': 1942100, 'file-version': 1, 'auto_managed': 0, 'save_path': '/home/jenkins/tribler/Tribler/Test/data', 'peers': '\x7f\x00\x00\x01\x99\xee\x82\xa1w\xd2+l', 'completed_time': 1552400733, 'allocation': 'sparse', 'download_rate_limit': -1, 'libtorrent-version': '1.2.0.0', 'sequential_download': 0, 'num_complete': 16777215, 'finished_time': 4, 'last_upload': 1552400735, 'super_seeding': 0, 'last_download': 0}
```

@arvidn were there any breaking changes regarding the resume data?

arvidn:
the main difference regarding the resume data is that the `save_resume_data_alert` now contains an `add_torrent_params` object as the resume data. However, when building with deprecated functions (which is the default), the `entry` is there too, as a copy.

Could the freezing be an issue with waiting for the resume data alerts to come back, and missing one?

PR #4369, opened 2019-03-12
Proper lists URL encoding and UTF-8 fixes

### LINKED ISSUE #4365: Creating a torrent from several files is broken
Latest `devel`:
bug

ichorid:
This PR implements proper, Twisted `http.parse_qs`-compatible encoding of URL arguments and makes our test suite use the same way of encoding URLs that we use in Tribler. Also, it changes Tribler config file encoding to `utf-8`. This change is necessary to conform with the "unicode sandwich" ideology and accommodate the changes in the endpoints data encoding.
Fixes #4365

### DIFF
--- a/Tribler/Core/Config/tribler_config.py
+++ b/Tribler/Core/Config/tribler_config.py
@@ -43,9 +43,9 @@ class TriblerConfig(object):
         if config is None:
             file_name = os.path.join(self.get_default_state_dir(), FILENAME)
             if os.path.exists(file_name):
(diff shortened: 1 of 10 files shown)


qstokkink:
Note that Windows uses `cp65001`, `mbcs` and `ascii` usually. Forcing `utf-8` in the config could kill our Windows support.

xoriole:
I like that default_encoding is UTF-8 but for Windows we should use system encoding. 
Ref: https://github.com/Tribler/tribler/blob/devel/run_tribler.py#L18

PR #4374, opened 2019-03-13
READY: Various GUI channel view fixes

### LINKED ISSUE #4347: Put non-legacy channels first in the GUI
By default, non-legacy channels should always appear on top of the list, no matter what sorting option was selected. This could require some tricky SQL.

### LINKED ISSUE #4352: Channels search result tab problems
1. The channel size column is not shown on the channel tab in search results. This is frustrating.
2. :heart: icon is still shown for legacy channels. Clicking on that has some (small) probability of triggering a crash of the GUI with a message like "no infohash attribute when trying to download".

ichorid:
This PR contains several fixes for GUI problems regarding GigaChannel:
 * Fixes #4347 
 * Fixes #4352 
and some others.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/torrent_metadata.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/torrent_metadata.py
@@ -117,6 +117,7 @@ def define_binding(db):
             pony_query = cls.search_keyword(query_filter, lim=1000) if query_filter else select(g for g in cls)
 
             # Sort the query
+            sort_expression = None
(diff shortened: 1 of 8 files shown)


devos50:
@ichorid the width of the search table seemed off due to the changing visibility of various columns. Hopefully you don't mind that I appended a commit to fix this in this PR 👍 

Also, I think that it is slightly confusing that no heart is shown on your channel, so I changed the text in the category label to 'my channel' if it is your channel. Please also review my changes and if you agree, you can merge this PR.

PR #4377, opened 2019-03-15
Add title and category editing to the GUI

ichorid:
This PR adds the ability to edit a GigaChannel entry through the GUI. The editing is done inline in the MyChannel table. The user can change the torrent's category by selecting one from a combo box. Torrent's title can also be edited by double-clicking on it.
Also, this PR changes Legacy entries status label from 6 to 1000, and removes FTS indexing from `tags` column, so the database format is not fully compatible with the previous versions.

This PR is really important for 7.3 release, as it provides a basis for crowdsourcing the torrents channels.

EDIT:
Also, this PR includes a refactoring of `triblertablecontrollers.py` and **breaking changes to DB schema** (removing unused tables and properties).

### DIFF
--- a/Tribler/Core/Category/category.conf
+++ b/Tribler/Core/Category/category.conf
@@ -1,6 +1,6 @@
 [Video]
 rank = 1
-displayname = Video Files
+displayname = Video
 suffix = asf, asp, avi, flc, fli, flic, mkv, mov, movie, mpeg, mpg, qicktime, ram, rm, rmvb, rpm, vob, wma, wmv
 minfilesize = 50
 maxfilesize = 10000000
@@ -12,7 +12,7 @@ matchpercentage = 0.5
 
 [VideoClips]
 rank = 2
-displayname = Video Clips
+displayname = VideoClips
 suffix = asf, asp, avi, flc, fli, flic, mkv, mov, movie, mpeg, mpg, qicktime, ram, rm, rmvb, rpm, vob, wma, wmv, mp4, flv
 minfilesize = 0
 maxfilesize = 50
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
@@ -13,14 +13,11 @@ from Tribler.pyipv8.ipv8.database import database_blob
 from Tribler.pyipv8.ipv8.keyvault.crypto import default_eccrypto
 
 # Metadata, torrents and channel statuses
-NEW = 0
-TODELETE = 1
-COMMITTED = 2
-JUST_RECEIVED = 3
-UPDATE_AVAILABLE = 4
-PREVIEW_UPDATE_AVAILABLE = 5
-UPDATED = 6
-LEGACY_ENTRY = 1000
+NEW = 0  # The entry is newly created and is not published yet. It will be committed at the next commit.
+TODELETE = 1  # The entry is marked to be removed at the next commit.
+COMMITTED = 2  # The entry is committed and seeded.
+UPDATED = 6  # One of the entry's properties was updated. It will be committed at the next commit.
+LEGACY_ENTRY = 1000  # The entry was converted from the old Tribler DB. It has no signature and should not be shared.
 
 PUBLIC_KEY_LEN = 64
 
@@ -61,9 +58,6 @@ def define_binding(db, logger=None, key=None, clock=None):
         # Local
         added_on = orm.Optional(datetime, default=datetime.utcnow)
         status = orm.Optional(int, default=COMMITTED)
-
-        parents = orm.Set('ChannelNode', reverse='children')
-        children = orm.Set('ChannelNode', reverse='parents')
 
         # Special properties
         _payload_class = ChannelNodePayload
@@ -198,12 +192,4 @@ def define_binding(db, logger=None, key=None, clock=None):
         @classmethod
         def from_dict(cls, dct):
             return cls(**dct)
-
-        def update_properties(self, update_dict):
-            #TODO: check if properties really changed
-            self.status = UPDATED
-            self.set(**update_dict)
-            self.timestamp = self._clock.tick()
-            self.sign(self._my_key)
-
     return ChannelNode
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
@@ -19,7 +19,8 @@ COMMITTED = 2
 JUST_RECEIVED = 3
 UPDATE_AVAILABLE = 4
 PREVIEW_UPDATE_AVAILABLE = 5
-LEGACY_ENTRY = 6
+UPDATED = 6
+LEGACY_ENTRY = 1000
 
 PUBLIC_KEY_LEN = 64
 
@@ -197,5 +198,12 @@ def define_binding(db, logger=None, key=None, clock=None):
         @classmethod
         def from_dict(cls, dct):
             return cls(**dct)
+
+        def update_properties(self, update_dict):
+            #TODO: check if properties really changed
+            self.status = UPDATED
+            self.set(**update_dict)
+            self.timestamp = self._clock.tick()
+            self.sign(self._my_key)
 
     return ChannelNode
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -24,7 +24,6 @@ CLOCK_STATE_FILE = "clock.state"
 NO_ACTION = 0
 UNKNOWN_CHANNEL = 1
 UPDATED_OUR_VERSION = 2
-GOT_SAME_VERSION = 3
 GOT_NEWER_VERSION = 4
 UNKNOWN_TORRENT = 5
 DELETED_METADATA = 6
@@ -327,9 +326,16 @@ class MetadataStore(object):
         result = []
         node = self.TorrentMetadata.get_for_update(public_key=database_blob(payload.public_key),
                                                    infohash=database_blob(payload.infohash))
-        if node and node.timestamp < payload.timestamp:
-            node.delete()
-            result.append((None, DELETED_METADATA))
+        if node:
+            if node.timestamp < payload.timestamp:
+                node.delete()
+                result.append((None, DELETED_METADATA))
+            elif node.timestamp > payload.timestamp:
+                result.append((node, GOT_NEWER_VERSION))
+                return result
+            else:
+                return result
+            # Otherwise, we got the same version locally and do nothing.
 
         # Check for the older version of the same node
         node = self.TorrentMetadata.get_for_update(public_key=database_blob(payload.public_key), id_=payload.id_)
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -34,14 +34,13 @@ DELETED_METADATA = 6
 # maintained by SQL triggers.
 sql_create_fts_table = """
     CREATE VIRTUAL TABLE IF NOT EXISTS FtsIndex USING FTS5
-        (title, tags, content='ChannelNode', prefix = '2 3 4 5',
+        (title, content='ChannelNode', prefix = '2 3 4 5',
          tokenize='porter unicode61 remove_diacritics 1');"""
 
 sql_add_fts_trigger_insert = """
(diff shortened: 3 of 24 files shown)


devos50:
@ichorid note that the tests are failing and that there are several flake8 errors.

ichorid:
> @ichorid note that the tests are failing and that there are several flake8 errors.

That's why it is still "draft" pull request :wink:

devos50:
Ah, I got already requested to do a review so that’s why I asked.

ichorid:
> Ah, I got already requested to do a review so that’s why I asked.

oh, sorry. Next time I will only do this when its really finished.

devos50:
Well, it’s ok if you want a preliminary review round for instance (i.e. to give some high level feedback) ;)

ichorid:
> Well, it’s ok if you want a preliminary review round for instance (i.e. to give some high level feedback) ;)

Yep, that was my intention here. I signalled you by requesting a review "*hey, look at this*", but there was a warning that this thing is still in the works (*draft* PR).

Tribler/Core/Category/category.conf
```diff
@@ -12,7 +12,7 @@ matchpercentage = 0.5
 
 [VideoClips]
 rank = 2
-displayname = Video Clips
+displayname = VideoClips
```
devos50: Why not display it with the proper spelling (which is with a space between the words)?

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
```diff
@@ -19,7 +19,8 @@
 JUST_RECEIVED = 3
 UPDATE_AVAILABLE = 4
 PREVIEW_UPDATE_AVAILABLE = 5
-LEGACY_ENTRY = 6
+UPDATED = 6
```
devos50: Could you please add some comments here to indicate what each status exactly means?

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -34,14 +33,13 @@
 # maintained by SQL triggers.
 sql_create_fts_table = """
     CREATE VIRTUAL TABLE IF NOT EXISTS FtsIndex USING FTS5
-        (title, tags, content='ChannelNode', prefix = '2 3 4 5',
+        (title, content='ChannelNode', prefix = '2 3 4 5',
```
devos50: Do we have to remove our metadata database to run this?

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -101,7 +101,7 @@ def render_GET(self, request):
             channels_list = [channel.to_simple_dict() for channel in channels]
 
         return json.dumps({
-            "channels": channels_list,
+            "results": channels_list,
```
devos50: I am not entirely sure whether I agree with this change. On one hand, it is more consistent. On the other hand, a `results` key is quite generic. What is the motivation behind this change?

Tribler/Core/Modules/restapi/mychannel_endpoint.py
```diff
@@ -329,10 +327,14 @@ def __init__(self, session, infohash):
     @db_session
     def render_PATCH(self, request):
         parameters = http.parse_qs(request.content.read(), 1)
-        if 'status' not in parameters:
+        if 'status' not in parameters and 'tags' not in parameters and 'title' not in parameters:
```
devos50: The error text does not reflect the conditional (since it does not only check for status). Also, could you please document this method and give an example of a request? In particular, we should communicate the type of the `tags` parameter to the developer.

Tribler/Core/Modules/restapi/search_endpoint.py
```diff
@@ -116,6 +116,8 @@ def search_db():
 
         def on_search_results(search_results_tuple):
             search_results, total = search_results_tuple
+            if not search_results:
```
devos50: There's no JSON response here?

Tribler/Test/Core/Modules/MetadataStore/test_store.py
```diff
@@ -229,6 +230,35 @@ def test_process_payload_reject_old(self):
...
+
+        torrent3 = self.mds.TorrentMetadata(title='blabla', timestamp=12, id_=4,
+                                            infohash=payload.infohash)
+        payload3 = torrent._payload_class(**torrent3.to_dict())
+        torrent3.delete()
+
+        self.mds.process_payload(payload2)
+        self.assertEqual(GOT_NEWER_VERSION, self.mds.process_payload(payload)[0][1])
+
+        # In this corner case the newle arrived payload contains a newer node
```
devos50: newle -> newly

Tribler/Test/Core/Modules/RestApi/test_mychannel_endpoint.py
```diff
@@ -222,19 +220,19 @@ def test_update_my_torrents_no_channel(self):
         return self.do_request('mychannel/torrents', request_type='POST', post_data=post_params, expected_code=404)
 
     @trial_timeout(10)
-    def test_update_my_torrents(self):
+    def test_update_my_torrents_POST(self):
         """
-        Test whether we get an error if we update multiple torrents in your uncreated channel
+        Test updating/creating a torrent in a personal channel with POST
```
devos50: with POST -> with a POST request (minor comment)

TriblerGUI/qt_resources/mainwindow.ui
```diff
@@ -3307,7 +3310,7 @@ font-weight: bold;</string>
               </widget>
              </item>
              <item>
-              <widget class="QLabel" name="num_search_results_label">
+              <widget class="QLabel" name="num_results_label">
```
devos50: In Qt, names of `QWidget` objects should be unique across the whole `.ui` file. That's why I name these objects as specific as possible. The name `num_results_label` could result in a name collision.

TriblerGUI/widgets/channelpage.py
```diff
@@ -50,4 +52,4 @@ def initialize_with_channel(self, channel_info):
 
     def load_torrents(self):
         self.controller.model.reset()
-        self.controller.load_torrents(1, 50)  # Load the first 50 torrents
+        self.controller.perform_query(first=1, last=50)  # Load the first 50 torrents
```
devos50: I like using the full keyword argument name here 👍

TriblerGUI/widgets/discoveredpage.py
```diff
@@ -36,4 +36,4 @@ def initialize_discovered_page(self, gui_settings):
 
     def load_discovered_channels(self):
         self.controller.model.reset()
-        self.controller.load_channels(1, 50)  # Load the first 50 discovered channels
+        self.controller.perform_query(start=1, end=50)  # Load the first 50 discovered channels
```
devos50: Probably not related to changes in this PR, but we are using start and end here as variable names and in `channelpage.py` we are using first and last. We should make these names consistent.

TriblerGUI/widgets/lazytableview.py
```diff
@@ -82,7 +82,7 @@ def on_subscribe_control_clicked(self, index):
         item = index.model().data_items[index.row()]
         # skip LEGACY entries, regular torrents and personal channel
         if (u'subscribed' not in item or
-                item[u'status'] == 6 or
+                item[u'status'] == 1000 or
```
devos50: We should re-use the metadata state definitions here.

TriblerGUI/widgets/tablecontentdelegate.py
```diff
@@ -4,13 +4,29 @@
...
-from PyQt5.QtWidgets import QStyle, QStyledItemDelegate
+from PyQt5.QtWidgets import QStyle, QStyledItemDelegate, QComboBox
 
 from TriblerGUI.defs import ACTION_BUTTONS, COMMIT_STATUS_COMMITTED, COMMIT_STATUS_NEW, COMMIT_STATUS_TODELETE, \
     HEALTH_CHECKING, HEALTH_DEAD, HEALTH_ERROR, HEALTH_GOOD, HEALTH_MOOT, HEALTH_UNCHECKED
 from TriblerGUI.utilities import get_health, get_image_path
 from TriblerGUI.widgets.tableiconbuttons import DeleteIconButton, DownloadIconButton, PlayIconButton
 
+#TODO: build this automatically and/or move it somewhere
+CATEGORY_LIST = [
```
devos50: I would move these definitions to the `defs.py` file in the `TriblerGUI` package.

Tribler/Core/Category/category.conf
ichorid: This thing is only used to tag Legacy channels now. However, the tag should be a single word. Hence, this change.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
ichorid: Good point, I'll do that.

Tribler/Core/Modules/MetadataStore/store.py
ichorid: Actually, we don't have to. But it is a very good idea to do so because your FTS index gets much more balanced afterward.

TriblerGUI/widgets/tablecontentdelegate.py
ichorid: Good point!

TriblerGUI/widgets/lazytableview.py
ichorid: But I guess you were against adding dependencies to TrilberGUI from the Core, right?

Tribler/Core/Modules/restapi/metadata_endpoint.py
ichorid: The reason is the big refactoring of `triblertablecontrollers.py`. Having the API consistent across different endpoints really helps with inheritance and code reuse. Both for API writers and users.

TriblerGUI/qt_resources/mainwindow.ui
ichorid: Hmm. There are not collisions now. I guess we're fine with this name now. We can always rename it later.

TriblerGUI/widgets/discoveredpage.py
ichorid: Agree.

Tribler/Core/Modules/restapi/search_endpoint.py
ichorid: Hmm... Maybe you're right, and there should be a zero-results response instead. I used this thing as a temporary workaround for a problem.

Tribler/Core/Modules/restapi/mychannel_endpoint.py
ichorid: Done;

Tribler/Core/Modules/MetadataStore/store.py
ichorid: Ok, now I've broken the schema. So, we have to remove our database.

TriblerGUI/widgets/lazytableview.py
ichorid: Hm. I'll just add them to `defs.py`.

devos50: Yes but at this point I rather have a link to the core (which we can later refactor) instead of having a magic variable which is unclear.

devos50:
@ichorid one more thing, since you are changing the REST API response structure, please also make the appropriate changes to the fake Tribler API (https://github.com/Tribler/fake-tribler-api), in particular [here](https://github.com/Tribler/fake-tribler-api/blob/devel/FakeTriblerAPI/endpoints/metadata_endpoint.py#L78) and [here](https://github.com/Tribler/fake-tribler-api/blob/devel/FakeTriblerAPI/endpoints/metadata_endpoint.py#L153).

devos50:
@ichorid additional comment, please do not try to squeeze too much in PRs. For now you can continue but enabling changing the category and torrent name could be in one PR, and the refactoring of the Tribler table controllers in the GUI could have been in another PR 👍

ichorid:
> @ichorid additional comment, please do not try to squeeze too much in PRs. For now you can continue but enabling changing the category and torrent name could be in one PR, and the refactoring of the Tribler table controllers in the GUI could have been in another PR +1

Yeah, I was thinking about this too... However, these things depend on each other, so I really can't pipeline them. And I'm running out of time before my trip.
Also, would you really prefer 3 different PRs that break DB schema instead of one :wink:

ichorid:
Corresponding Fake Tribler API PR: https://github.com/Tribler/fake-tribler-api/pull/47

ichorid:
retest this please

ichorid:
retest this please

PR #4379, opened 2019-03-16
Update dependencies for Arch Linux

elvetemedve:
Currently Tribler 7.2.2 does not start on Arch Linux, because of few missing Python 2 modules.

This change adds these missing modules to the documentation.


devos50:
ok to test

devos50:
Thank you for your contribution! 👍

PR #4384, opened 2019-03-18
Binary on verbs: agent.request(b'GET', uri, headers, None)

cclauss:
As discussed at https://github.com/Tribler/tribler/pull/4382#issuecomment-473809894


devos50:
Github probably does not like our load:

```
hudson.plugins.git.GitException: Failed to fetch from https://github.com/Tribler/tribler.git
```

PR #4389, opened 2019-03-18
Added BEP33 DHT health check

### LINKED ISSUE #4321: Our DHT-based torrent health check is unreliable
To check the health of a torrent from the DHT, we download the torrent in upload mode. When we receive the metadata, we get the list of peers we are connected to and consider them as leechers (or seeders if they have everything). This mechanism does not accurately represent the real swarm size. To get proper popularity estimations, we need something better.

One solution is to rely on the `get_peers` method (which is exposed in the Python bindings since libtorrent 1.1).

Related to #4256.

devos50:
Fixes #4321 

By default, BEP33 checks are disabled and it is checked whether ones libtorrent core has support for it. If not, the user cannot perform these checks.

Since BEP33 support is not enabled on the testing machines, there are a few uncovered lines.

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -16,20 +16,21 @@ from distutils.version import LooseVersion
 from shutil import rmtree
 
 import libtorrent as lt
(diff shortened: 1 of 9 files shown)


devos50:
retest this please

PR #4393, opened 2019-03-19
block_type must be bytes

cclauss:
As suggested at https://github.com/Tribler/py-ipv8/pull/462


devos50:
@cclauss there seems to be a syntax error

devos50:
Kudos for the readability changes 👍

Issue #4222, opened 2019-02-12
Torrents out of Tribler dir


ichorid:
Tribler supports seeding torrent files from any dir on any path in the system. However, users cannot share their existing media collections downloaded with other bittorrent clients. To do so, they must move their entire collections to the Tribler Downloads folder and then add the corresponding .torrent files to Tribler as downloads. That is very unintuitive.
Besides, many users on Windows use more then one drive to store media (films, music), while their Tribler user directory resides on drive C:. This forces them to copy or reorganize their files, which is unacceptable for many.
In addition, we should discuss creating a procedure to map files to torrents, so users can just scan their entire media collection and immediately join the corresponding swarms (possible student project?).

ichorid:
A user complaining about this problem:
https://forum.tribler.org/t/exact-steps-to-start-seeding-from-torrent-files-i-already-have/5054

zimio:
This sounds interesting. Can I give it a go at solving this issue? Is more discussion needed?

ichorid:
@zimio , thanks for your interest in Tribler project! Your contribution would be very welcome.

At this moment Tribler is capable of seeding files outside of Tribler Downloads folder. It does that by default, i.e. a user does not have to move anything if they just want to share their files. However, when someone wants to share the torrent collection they already have to join the existing swarms, there is a problem:
 * If the user adds a .torrent file, Tribler rightfully recognizes this action as a *new download*. It is correct and expected behavior since Tribler knows nothing about where the user keeps their files for this torrent;
 * If the user adds the files from some torrent, Tribler recognizes this action as a *creation of a new torrent*, which is again, completely right and expected, since Tribler does not know anything about the existence of the corresponding .torrent file.

@zimio , I suggest you download a couple of torrents with some other client and then trying to share them with Tribler, to get in touch with the problem. After that, could you please suggest your thoughts/solution to this?

zimio:
Hello thanks. I've been thinking about this.

So we have two cases. Adding a new torrent file and adding new files.

1-) Adding new torrent file.

Since the problem is that we don't know the location of the files already downloaded by the torrent and that information is not available in the torrent file itself. Then, we should ask the user to give us a directory where we can scan for files. Could be analogous to the share directory in other P2P clients.

So Tribler goes and finds files with the same hashes as the ones in the torrent file added. 

If a full match is found (all files), then we start seeding the torrent.

If a partial match is found (only some files), then we start seeding only those parts we already have and ask user if he wants to download the rest.

[...]

PR #4433, opened 2019-03-25
initial proof of concept approach

zimio:
This pull request is not for merging. Its purpose is to get feedback on the way forward to solve this issue:
https://github.com/Tribler/tribler/issues/4222

This PRQ has a naive approach . 

I think there are several optimizations we can make. First, store the directory structure in a NoSQL key-based database for faster look up, instead of repetitively going to disk searching the same structure. 

The next optimization I think can be made is when checking whether a file is indeed a file for seeding... I don't know if libtorrent has a function for this, but fast checking could be done by checking pieces of the file, perhaps randomly and having threshold before actually trying to check all of the pieces...

I need feedback on this, to make sure I'm doing the right thing.

### DIFF
--- a/Tribler/Core/Modules/find_files.py
+++ b/Tribler/Core/Modules/find_files.py
@@ -1,0 +1,33 @@
+from __future__ import absolute_import
+
+import os
+
+# Finds files that match torrent to seed
+class FindSeedFiles:
+
+    def __init__(self, watch_directory, torrent_defs):
+        self.watch_directory = watch_directory
+        self.torrent_defs = torrent_defs
+
+    def scan(self):
+        # return dictionary with torrent as key and list of files to seed
+        seed_files = {}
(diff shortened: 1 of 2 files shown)


devos50:
ok to test

ichorid:
@zimio , you're doing it the right way. You created a proof-of-concept unoptimized implementation and a corresponding unit test. We can always optimize it later.
However, I must warn you that we are very conservative about adding any new external modules, because we have to install these on all of our build machines, and correctly package the stuff on all 3 platforms. So, there should be a really good argument for us to include a new dependency.

Regarding your "piece checking" idea, please clarify it. Or better yet,  try to implement it as a proof-of-concept.

ichorid:
Closed as it seems dead now...

synctext:
Feel free to reopen or resubmit if anybody wants to pick this up.

PR #4437, opened 2019-03-26
Fixed Tribler crashing other processes in Ubuntu

### LINKED ISSUE #4435: Restarting Tribler kills other system processes in Linux (Ubuntu)
If there is a leftover Trible core running in the background, Tribler shows a popup message to the user informing about the existing core and asking whether to force restart. If the user chooses to force restart, the old core process is terminated and Tribler restarts with a fresh core. Sometimes the restart also crashes other system processes.

xoriole:
Fixes #4435

### DIFF
--- a/check_os.py
+++ b/check_os.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+
 import logging.config
 import os
 import sys
@@ -11,8 +12,9 @@ import psutil
 from Tribler.Core.Config.tribler_config import TriblerConfig
 from Tribler.Core.Modules.process_checker import ProcessChecker
 
-FORCE_RESTART_MESSAGE = "A Tribler instance is already running. Do you want to force restart? " \
-                        "\n\nCaution: force restart could result in data corruption."
+FORCE_RESTART_MESSAGE = "An existing Tribler core process (PID:%s) is already running. \n\n" \
+                        "Do you want to stop the process and do a clean restart instead?"
+
 
 def error_and_exit(title, main_text):
     """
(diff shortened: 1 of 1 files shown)


check_os.py
```diff
@@ -145,41 +147,72 @@ def should_kill_other_tribler_instances():
...
+    """
+    def is_tribler_process(name):
+        """
+        Checks if the given name is of a Tribler processs. It checks a few potential keywords that
+        could be present in a Tribler process name across different platforms.
+        :param name: Process name
+        :return: True if pid is a Tribler process else False
+        """
+        name = name.lower()
+        process_keywords = ['usr/bin/python', 'run_tribler.py', 'tribler.exe',
```
devos50: What about `tribler.sh` ?

check_os.py
```diff
@@ -11,8 +12,9 @@
 from Tribler.Core.Config.tribler_config import TriblerConfig
 from Tribler.Core.Modules.process_checker import ProcessChecker
 
-FORCE_RESTART_MESSAGE = "A Tribler instance is already running. Do you want to force restart? " \
-                        "\n\nCaution: force restart could result in data corruption."
+FORCE_RESTART_MESSAGE = "An existing Tribler core instance (PID:%s) is already running. \n\n" \
+                        "Do you want to do a clean restart (YES) or use the existing core (NO)?"
```
devos50: Maybe leave the second part (use the existing core...) out, I think this message is a bit confusing otherwise.

check_os.py
xoriole: The issue is significant in the frozen environment so maybe we can leave that out.

PR #4446, opened 2019-04-01
Fixed a GUI reference import in Core

xoriole:
Fixed a GUI reference import in Core

### DIFF
--- a/Tribler/Core/Utilities/utilities.py
+++ b/Tribler/Core/Utilities/utilities.py
@@ -27,8 +27,6 @@ from twisted.web.http_headers import Headers
 from Tribler.Core.exceptions import HttpError
 from Tribler.Core.version import version_id
 
-from TriblerGUI.utilities import quote_plus_unicode
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


xoriole:
retest this please

Tribler/Core/Utilities/utilities.py
```diff
@@ -14,7 +14,7 @@
 from libtorrent import bdecode, bencode
 
 import six
-from six.moves.urllib.parse import parse_qsl, urlsplit
+from six.moves.urllib.parse import parse_qsl, quote_plus, urlsplit
```
devos50: Why are you not using `quote_plus_unicode` anymore? If you simply need `quote_plus`, you did not have to move the utilities to the core?

xoriole: There are two instances where `quote_plus_unicode` is used. One in Core in this file `utilities.py` and the other in `startdownloaddialog.py`. If looking for a minimal change, it would be to move the `compose_magnetlink()` of core `utilities.py`  method to the GUI. But I think that function fits to the Core.

PR #4449, opened 2019-04-02
Disabling IPv8 during wallet tests

devos50:
It was making connections to the bootstrap servers and listening for TrustChain blocks.


devos50:
retest this please

PR #4453, opened 2019-04-04
Add bootstrap module 

grimadas:


devos50:
retest this please

grimadas:
Latest gumby build https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_bootstrap/17/

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -76,6 +76,8 @@ def __init__(self):
 
         self.shutdownstarttime = None
 
+        self.bootstrap_session = None
```
devos50: I would suggest to rename this attribute to `bootstrap_download`.

Tribler/Core/Config/config.spec
```diff
@@ -25,6 +25,11 @@ ec_keypair_filename = string(default='')
 testnet_keypair_filename = string(default='')
 live_edges_enabled = boolean(default=True)
 
+[bootstrap]
+enabled = boolean(default=True)
+download_rate = integer(min=1, default=1000000)
```
devos50: Could you rename this to `max_download_rate` (since that is essentially what it is)?

Tribler/Test/API/test_bootstrap.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def start_download(self, dscfg):
+        download = self.session.start_download_from_tdef(self.tdef, dscfg)
+        download.set_state_callback(self.downloader_state_callback)
+        download.add_peer(("127.0.0.1", self.seeder_session.config.get_libtorrent_port()))
+
+    @trial_timeout(20)
+    def test_seeding(self):
+        """
+        Test whether a torrent is correctly seeded
```
devos50: This is not only testing for correct seeding behaviour right? Since the goal of this test is to test the payout mechanism, I would rename this test to `test_bootstrap` and change the docstring.

Tribler/Test/Core/test_launch_many_cores.py
```diff
@@ -176,3 +177,69 @@ def test_load_communities(self):
...
         self.assertTrue(self.get_community(TrustChainCommunity))
+
+
+class TestLaunchManyCoreSeederBootstrapSession(TestAsServer):
+
+    def setUpPreSession(self):
+        TestAsServer.setUpPreSession(self)
+
+        # Enable all communities
+        config_sections = ['trustchain', 'ipv8', 'bootstrap', 'libtorrent']
```
devos50: If I'm not mistaken, this test is connecting to the real TrustChain network and IPv8 bootstrap servers. Since this can lead to unexpected behaviour, we should avoid this. Why do you need trustchain/IPv8 enabled here and can you avoid this?

Tribler/Test/API/test_bootstrap.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @inlineCallbacks
+    def setUp(self):
+        yield super(TestBootSeed, self).setUp()
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.test_deferred = Deferred()
+        self.sourcefn = os.path.join(TESTS_DATA_DIR, 'bootstrap.block')
+        self.tdef = create_dummy_tdef(self.sourcefn, 25)
+        self._logger.debug("Creating file with infohash %s", hexlify(self.tdef.infohash))
+        self.count = 0
```
devos50: Unused variable?

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -343,7 +345,24 @@ def init(self):
...
+                    tdef = TorrentDef()
+                    tdef.add_content(full_path)
+                    tdef.set_piece_length(2**16)
+                    tdef.save()
+                    self._logger.debug("Seeding bootstrap file %s", hexlify(tdef.infohash))
+                    self.bootstrap_session = self.session.start_download_from_tdef(tdef, download_startup_config=dcfg)
+                else:
+                    # Download bootstrap file from current seeders
+                    infohash = self.session.config.get_bootstrap_infohash()
+                    self._logger.debug("Starting bootstrap downloading %s ", infohash)
```
devos50: There's a space after `%s` that can be removed.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -199,7 +202,7 @@ def get_handle(self):
         self.deferreds_handle.append(deferred)
         return deferred
 
-    def setup(self, dcfg=None, pstate=None, wrapperDelay=0, share_mode=False, checkpoint_disabled=False):
+    def setup(self, dcfg=None, pstate=None, wrapperDelay=0, share_mode=False, checkpoint_disabled=False, hidden = False):
```
devos50: Please remove the space before and after the `=` sign.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -125,6 +125,9 @@ def __init__(self, session, tdef):
         self.finished_deferred_already_called = False
         self.is_bootstrap_download = False
 
+        # If the download is hidden in GUI
```
devos50: It's not only hidden in the GUI, it is not returned when requested through the REST API (`GET /downloads`). Can you make the comment match this?

Tribler/Test/Core/test_launch_many_cores.py
grimadas: Bootstrap module will be not reachable if trustchain and ipv8 are not available, since it depends on direct payouts, creating blocks etc.

grimadas: But I agree that we don't need them in this test

devos50: Hmm this is an interesting one. I assume that if you leave out this test, it results in much uncovered code in `LaunchManyCore`. It is a bit hard to test since it's part of the `init` method when starting Tribler.

Still, having IPv8/TrustChain is undesired. Even if you replace IPv8 immediately after startup, it could be enough for IPv8 to do a bootstrap attempt and sending introduction requests to the bootstrap servers.

Tribler/Core/bootstrap.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import absolute_import
+
+import logging
+from binascii import hexlify, unhexlify
+
+from Tribler.Core.DownloadConfig import DownloadStartupConfig
+from Tribler.Core.TorrentDef import TorrentDef, TorrentDefNoMetainfo
+
+
+class Bootstrap(object):
```
devos50: A few lines explaining the purpose of this class would be helpful for new developers 👍

grimadas: Done

Issue #3770, opened 2018-08-06
Update developer dependencies


qstokkink:
Related to #3265. I had to do the following to get Tribler to run from source on Ubuntu 18 (using `pip` as much as possible): 

```bash
# Install python
sudo apt install python2.7 python-minimal python-pip

# Install py-ipv8 python dependencies
cd Tribler/pyipv8
pip install --upgrade -r requirements.txt

# Install sytem dependencies
sudo apt install ffmpeg libssl-dev libx11-6 vlc

# Install Tribler python dependencies
sudo apt install python-apsw python-cherrypy3 python-keyring python-keyrings.alt python-libtorrent python-m2crypto python-meliae python-pil python-pyqt5 python-pyqt5.qtsvg python-scipy
pip install bitcoinlib chardet configobj decorator dnspython ecdsa feedparser jsonrpclib leveldb matplotlib netifaces networkx pbkdf2 pony protobuf psutil pyaes pyasn1 pysocks requests
```

`pip install service_identity` segfaults for me so I have to do:

```bash
git clone --recursive https://github.com/pyca/service_identity
cd service_identity
sudo python setup.py install
```

ichorid:
Still broken?

devos50:
I'm pretty confident that this is sufficiently addressed, given our recent changes to the developer documentation. This issue did also target Python 2.

qstokkink:
Yes, this was fixed by #4454 (but that PR didn't use the GitHub `fixes` auto-close apparently).

PR #4454, opened 2019-04-05
Update development_on_linux.rst

cuileri:
With the help #3770 as well as my own experience during installing Tribler on Ubuntu 18.04, instructions were updated.


devos50:
Thanks!

PR #4474, opened 2019-04-29
Relative paths support in Tribler config

### LINKED ISSUE #3897: Absolute vs relative file path convention in Tribler config: make Tribler portable

ichorid:
This PR solves https://github.com/Tribler/tribler/issues/3897 by adding support for relative paths in Tribler config and `.state` files.

Support for relative paths is done in a bit hackish way for `.state` files, as our Libtorrent wrapper is a mess and this is not the time for a full refactoring. Nonetheless, these changes are important for 7.3, because now we start to keep *technical* torrents in the Tribler state dir (channels and bootstrap). Its contents should be independent of the global OS dir path. Otherwise, users can face unexpected consequencies moving stuff around. In addition, this PR makes it possible to create a truly portable Tribler installation.

Now, the paths in Tribler config work in the following way:
 * `get_` methods always return a file's absolute path
 * `set_` methods accept both relative and absolute paths
 * if a `set_` method is applied to an absolute path that is a sub-path of the current Tribler state dir, it is instead saved as a relative path.

Thus, these methods are completely compatible with the old behavior.

DISCLAIMER: a few lines that remove empty(?) `.state` files in `LaunchManyCore.py` are not covered, and I'm not going to cover them, as I do not fully understand what they do. I just moved them around for clarity. In fact, these were *never* properly covered before and I even added a test for `resume_download` method, which is 8 years old!

Closes #3897


qstokkink:
Relative paths have always been an issue on Windows with multiple disks. Make sure you especially test on Windows thoroughly.

PR #4480, opened 2019-05-02
READY: Foreground DB upgrader

### LINKED ISSUE #4439: Almost no content found by Triber v7.3.0-beta1

release-7.3.0-beta1 branch, 65d974cfa262d55f342fb06d6ca1c7b7d6f82e0d

Linux

Use Tribler v7.2 for a while, ensure some content get found. Then try 7.3.

Content continues to be found. My own downloads are also found using search.

##### Actual behavior: #####

Only test channels and torrents seem to be available. Even my own downloads do not seem to be indexed.

### Note ###

Now I see "Upgrading Tribler database". I assume searching starts working after this.

Is there a transitional period where both old and new distributed storages work together? Or it is a full "network split" between legacy and new Triblers, disincentivising users from switching to new version because of nobody switched yet?

### LINKED ISSUE #4440: 7.3 database migration stuck.

release-7.3.0-beta1 branch, 65d974cfa262d55f342fb06d6ca1c7b7d6f82e0d

Linux

Run Tribler 7.3 after using previous Tribler versions, leave it running overnight.

Database migration completes successfully.

##### Actual behavior: #####

Stuck at "Upgrading Tribler database: 8313/147866", not counting forward.

Restarting Tribler resumed migration process although.

Inaccessible due to log flood flushing away terminal history fast.

### LINKED ISSUE #4441: 7.3 database migration is slow unless I turn off fsync/fdatasync

release-7.3.0-beta1 branch, 65d974cfa262d55f342fb06d6ca1c7b7d6f82e0d

Linux

Use 7.2 and earlier for a while, then run 7.3.

Database migration is fast enough to complete overnight, even on HDD.

##### Actual behavior: #####

It converts only 0.7 batches per second and seems to fully synchonize to disk after every batch.

Disabling filsystem sync globally or using `eatmydata` bumps the speed up to 60 batches per second.

ichorid:
This PR moves 72 -> 73(Pony) upgrader out of the background thread to be run in the foreground at Tribler startup instead.

This greatly speed-ups the process by as we can use larger batches and **disable disk synchronization** during the upgrade process for the newly created DB. This is safe to do, since the old DB is read-only, and the conversion process can always be started from a clean slate if power fails during the upgrade process.
Disabling sync makes the upgrade process **run on an HDD as fast as on an SSD**. On my machine, a 400Mb `tribler.sdb` is converted in less than 1 minute.

To give users some control over the process, a "skip" button along with a text indicator of the state of the upgrade process are added to the "spinning gears" screen. Pressing the button produces a confirmation dialog. Even if a user decides to skip the conversion process, their personal channel will still be converted.
upgrader

Closes #4441 
Closes #4439 
Closes #4440

### DIFF
--- a/Tribler/Core/Modules/restapi/events_endpoint.py
+++ b/Tribler/Core/Modules/restapi/events_endpoint.py
@@ -31,7 +31,6 @@ class EventsEndpoint(resource.Resource):
       includes information about whether Tribler has started already or not and the version of Tribler used.
     - search_result_channel: This event dictionary contains a search result with a channel that has been found.
     - search_result_torrent: This event dictionary contains a search result with a torrent that has been found.
-    - upgrader_started: An indication that the Tribler upgrader has started.
     - upgrader_finished: An indication that the Tribler upgrader has finished.
     - upgrader_tick: An indication that the state of the upgrader has changed. The dictionary contains a human-readable
       string with the new state.
@@ -70,7 +69,6 @@ class EventsEndpoint(resource.Resource):
         self.infohashes_sent = set()
         self.channel_cids_sent = set()
 
-        self.session.add_observer(self.on_upgrader_started, NTFY_UPGRADER, [NTFY_STARTED])
         self.session.add_observer(self.on_upgrader_finished, NTFY_UPGRADER, [NTFY_FINISHED])
         self.session.add_observer(self.on_upgrader_tick, NTFY_UPGRADER_TICK, [NTFY_STARTED])
         self.session.add_observer(self.on_watch_folder_corrupt_torrent,
@@ -110,9 +108,6 @@ class EventsEndpoint(resource.Resource):
             return
         else:
             [request.write(message_str + '\n') for request in self.events_requests]
-
-    def on_upgrader_started(self, subject, changetype, objectID, *args):
-        self.write_data({"type": "upgrader_started"})
 
     def on_upgrader_finished(self, subject, changetype, objectID, *args):
         self.write_data({"type": "upgrader_finished"})
--- a/Tribler/Core/Modules/restapi/upgrader_endpoint.py
+++ b/Tribler/Core/Modules/restapi/upgrader_endpoint.py
@@ -2,9 +2,13 @@ from __future__ import absolute_import
 
 import logging
 
-from twisted.web import resource
+from six import viewitems
+
+from twisted.web import http, resource
 
 import Tribler.Core.Utilities.json_util as json
+
+SKIP_DB_UPGRADE_STR = "skip_db_upgrade"
 
 
 class UpgraderEndpoint(resource.Resource):
@@ -17,28 +21,47 @@ class UpgraderEndpoint(resource.Resource):
         self._logger = logging.getLogger(self.__class__.__name__)
         self.session = session
 
-    def render_DELETE(self, _):
+    def render_POST(self, request):
         """
-        .. http:put:: /upgrader
+        .. http:post:: /upgrader
 
-        A PUT request to this endpoint will skip the DB upgrade process, if it is running.
+        A POST request to this endpoint will skip the DB upgrade process, if it is running.
 
             **Example request**:
 
-            .. sourcecode:: none
+            .. sourcecode:: javascript
 
-                curl -X DELETE http://localhost:8085/upgrader
+                {
+                    "skip_db_upgrade": true
+                }
+
+
+                curl -X POST http://localhost:8085/upgrader
 
             **Example response**:
 
             .. sourcecode:: javascript
 
                 {
-                    "skip": True
+                    "skip_db_upgrade": true
                 }
         """
 
-        if self.session.upgrader:
+        parameters_raw = http.parse_qs(request.content.read(), 1)
+        parameters = {}
+
+        # FIXME: make all endpoints Unicode-compatible in a unified way
+        for param, val in viewitems(parameters_raw):
+            parameters.update({param: [item.decode('utf-8') for item in val]})
+
+        if SKIP_DB_UPGRADE_STR not in parameters:
+            request.setResponseCode(http.BAD_REQUEST)
+            return json.twisted_dumps({"error": "attribute to change is missing"})
+        elif not self.session.upgrader:
+            request.setResponseCode(http.NOT_FOUND)
+            return json.twisted_dumps({"error": "upgrader is not running"})
+
+        if self.session.upgrader and parameters[SKIP_DB_UPGRADE_STR]:
             self.session.upgrader.skip()
 
-        return json.twisted_dumps({"skip": True})
+        return json.twisted_dumps({SKIP_DB_UPGRADE_STR: True})
--- a/Tribler/Core/Modules/restapi/upgrader_endpoint.py
+++ b/Tribler/Core/Modules/restapi/upgrader_endpoint.py
@@ -1,0 +1,44 @@
+from __future__ import absolute_import
+
+import logging
+
+from twisted.web import resource
+
+import Tribler.Core.Utilities.json_util as json
+
+
+class UpgraderEndpoint(resource.Resource):
+    """
+    With this endpoint you can control DB upgrade process of Tribler.
+    """
+
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+    def render_DELETE(self, _):
+        """
+        .. http:put:: /upgrader
+
+        A PUT request to this endpoint will skip the DB upgrade process, if it is running.
+
+            **Example request**:
+
+            .. sourcecode:: none
+
+                curl -X DELETE http://localhost:8085/upgrader
+
+            **Example response**:
+
+            .. sourcecode:: javascript
+
+                {
+                    "skip": True
+                }
+        """
+
+        if self.session.upgrader:
+            self.session.upgrader.skip()
+
+        return json.twisted_dumps({"skip": True})
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -12,7 +12,7 @@ import sys
 from threading import RLock
 
 from twisted.internet import threads
-from twisted.internet.defer import fail, inlineCallbacks
+from twisted.internet.defer import fail, inlineCallbacks, succeed
 from twisted.python.failure import Failure
 from twisted.python.log import addObserver
 from twisted.python.threadable import isInIOThread
@@ -88,7 +88,6 @@ class Session(object):
         self.readable_status = ''  # Human-readable string to indicate the status during startup/shutdown of Tribler
 
         self.autoload_discovery = autoload_discovery
-
 
     def create_state_directory_structure(self):
         """Create directory structure of the state directory."""
@@ -406,9 +405,15 @@ class Session(object):
         if self.upgrader_enabled:
             self.upgrader = TriblerUpgrader(self)
             self.readable_status = STATE_UPGRADING_READABLE
-            self.upgrader.run()
+            upgrader_deferred = self.upgrader.run()
+        else:
+            upgrader_deferred = succeed(None)
 
-        startup_deferred = self.lm.register(self, self.session_lock)
+        def after_upgrade(_):
+            self.upgrader = None
+
+        startup_deferred = upgrader_deferred.addCallbacks(lambda _: self.lm.register(self, self.session_lock),
+                                                          lambda _: None).addCallbacks(after_upgrade, lambda _: None)
 
         def load_checkpoint(_):
             if self.config.get_libtorrent_enabled():
@@ -447,9 +452,6 @@ class Session(object):
             if self.lm.mds:
                 self.notify_shutdown_state("Shutting down Metadata Store...")
                 self.lm.mds.shutdown()
-
-            if self.upgrader:
-                self.upgrader.shutdown()
 
             # We close the API manager as late as possible during shutdown.
             if self.lm.api_manager is not None:
--- a/Tribler/Core/Session.py
+++ b/Tribler/Core/Session.py
@@ -412,8 +412,12 @@ class Session(object):
         def after_upgrade(_):
             self.upgrader = None
 
-        startup_deferred = upgrader_deferred.addCallbacks(lambda _: self.lm.register(self, self.session_lock),
-                                                          lambda _: None).addCallbacks(after_upgrade, lambda _: None)
+        def log_upgrader_error(failure):
+            self._logger.error("Error in Upgrader callback chain: %s", failure)
+
+        startup_deferred = upgrader_deferred.\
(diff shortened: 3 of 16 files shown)


Tribler/Core/Modules/restapi/events_endpoint.py
```diff
@@ -96,20 +97,25 @@ def __init__(self, session):
...
-            # The message contains invalid characters; fix them
-            message_str = json.twisted_dumps(fix_unicode_dict(message))
+        if message:
+            self.unsent_messages.append(message)
 
         if len(self.events_requests) == 0:
             return
-        else:
+
+        for m in self.unsent_messages:
```
devos50: The first design of the events endpoint actually had similar behaviour where messages are queued up if there is no available socket. However, we removed this functionality for the reason that event messages are not important enough to be queued up, and a missing event should not affect the user. In addition, I noticed that the queue size of this implementation is not limited and there is a possibility for memory leaks. @ichorid what is the motivation behind this commit (besides making sure that messages are queued up and flushed to the GUI if there is no socket available)?

Tribler/Core/Modules/restapi/upgrader_endpoint.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    With this endpoint you can control DB upgrade process of Tribler.
+    """
+
+    def __init__(self, session):
+        resource.Resource.__init__(self)
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+    def render_DELETE(self, _):
```
devos50: The semantics of this HTTP verb (`DELETE`) does not match the actual behaviour since it is not actually deleting the upgrader itself. I would construct an endpoint to skip the upgrade process as something like: `POST /upgrader/skip`.

Tribler/Core/Modules/restapi/upgrader_endpoint.py
```diff
@@ -0,0 +1,44 @@
...
+            **Example response**:
+
+            .. sourcecode:: javascript
+
+                {
+                    "skip": True
+                }
+        """
+
+        if self.session.upgrader:
```
devos50: You should return an error if the upgrader is not there (probably error `404` since a resource could not be found).

Tribler/Core/Session.py
```diff
@@ -406,9 +405,15 @@ def start(self):
...
+            upgrader_deferred = self.upgrader.run()
+        else:
+            upgrader_deferred = succeed(None)
 
-        startup_deferred = self.lm.register(self, self.session_lock)
+        def after_upgrade(_):
+            self.upgrader = None
+
+        startup_deferred = upgrader_deferred.addCallbacks(lambda _: self.lm.register(self, self.session_lock),
+                                                          lambda _: None).addCallbacks(after_upgrade, lambda _: None)
```
devos50: We should at least log calls to an `errBack`. I think there is a default `addCallback` we have implemented recently in Tribler. Using that one provides some logging 👍

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -263,6 +262,25 @@ def convert_personal_channel(self):
...
+            self.notifier_callback("%i entries converted in %i seconds (%i e/s)" % (amount, int(elapsed),
+                                                                                    int(amount / elapsed)))
+            yield deferLater(reactor, 0.001, lambda: None)
+
+    @inlineCallbacks
+    def update_convert_progress(self, amount, total, elapsed):
+        if self.notifier_callback:
+            elapsed = 0.0001 if elapsed == 0.0 else elapsed
+            amount = amount or 1
+            est_speed = amount/elapsed
```
devos50: Could you add a space before and after the `/`?

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -263,6 +262,25 @@ def convert_personal_channel(self):
...
+            eta = str(datetime.timedelta(seconds=int((total - amount) / est_speed)))
```
devos50: Please make sure it is Python 3 compatible 👍

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -263,6 +262,25 @@ def convert_personal_channel(self):
 
         v.value = CONVERSION_FINISHED
 
+    @inlineCallbacks
+    def update_convert_total(self, amount, elapsed):
+        if self.notifier_callback:
+            self.notifier_callback("%i entries converted in %i seconds (%i e/s)" % (amount, int(elapsed),
+                                                                                    int(amount / elapsed)))
+            yield deferLater(reactor, 0.001, lambda: None)
```
devos50: I assume that the notifier callback invokes the event endpoint and passes a message to the GUI? If so, I think sending notifications to the GUI at millisecond speed is too fast. We might want to send a message to the GUI e.g. every half a second?

Please correct me if I'm wrong.

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -385,22 +394,14 @@ def mark_conversion_finished(self):
...
-            if not self.shutting_down:
-                self.convert_discovered_channels()
-            if not self.shutting_down:
-                self.update_trackers_info()
-            if not self.shutting_down:
-                self.mark_conversion_finished()
-            self.mds._db.disconnect()
-
-        return deferToThread(background_stuff)
+        self.mds.clock = None  # We should never touch the clock during legacy conversions
```
devos50: Are we sure this is safe to do? There might be a better (potentially safer) solution to prevent the clock from being touched, e.g. by setting a flag in the `mds` object.

Tribler/Core/Upgrade/upgrade.py
```diff
@@ -18,10 +20,13 @@ def __init__(self, session):
         self.notified = False
         self.is_done = False
         self.failed = True
-        self.finished_deferred = None
-        self.pony_upgrader = None
 
         self.current_status = u"Initializing"
+        self._dtp72 = None
```
devos50: I can infer from the name what the variable does, but overall I think it can be improved. Suggestion: `self._dispersy_to_pony_migrator`?

Tribler/Test/Core/Upgrade/test_upgrader.py
```diff
@@ -54,10 +41,40 @@ def test_upgrade_72_to_pony(self):
...
+            os.path.dirname(os.path.realpath(__file__))), '..', 'data', 'upgrade_databases', 'tribler_v29.sdb'))
+        old_database_path = os.path.join(self.session.config.get_state_dir(), 'sqlite', 'tribler.sdb')
+        new_database_path = os.path.join(self.session.config.get_state_dir(), 'sqlite', 'metadata.db')
+        channels_dir = os.path.join(self.session.config.get_chant_channels_dir())
+
+        shutil.copyfile(OLD_DB_SAMPLE, old_database_path)
+
+        from Tribler.Core.Upgrade.db72_to_pony import DispersyToPonyMigration
+
+        original_init = DispersyToPonyMigration.__init__
```
devos50: This is _very_ dangerous to do. If the test fails for some reason after this line, and before you replace the init method with the old one, the `DispersyToPonyMigration` initializer will be stuck with the patched method and might fail in subsequent tests. We had these kind of errors in the past and they are a headache to debug.

If possible, try to avoid monkey patching methods on raw classes.

TriblerGUI/tribler_window.py
```diff
@@ -949,6 +946,26 @@ def clicked_force_shutdown(self):
         # Stop the Qt application
         QApplication.quit()
 
+    def clicked_skip_conversion(self):
+        self.dialog = ConfirmationDialog(self, "Abort the conversion of old channels",
+                                         "Upgrade procedure is now converting channels data "
```
devos50: Please start the sentence with: "The upgrade procedure..."

Channels data -> torrents in channels

Tribler/Core/Modules/restapi/events_endpoint.py
ichorid: When TriblerCore starts, it fires the upgrader, which sends 3 messages to the GUI:
1. Upgrade procedure started
2. Upgrader tick (possibly repeated many times)
3. Upgrader finished
This messages should always be sent in order, for upgrade notification/skip procedure to work properly. However, GUI<->Core messaging at startup gives no guarantees on this. For example, message (1) is never seen by the GUI due to some kind of race condition.

To prevent a potential memory leak we can limit the queue size and make it work FIFO.

Tribler/Core/Modules/restapi/upgrader_endpoint.py
ichorid: Fine point.

Tribler/Core/Modules/restapi/upgrader_endpoint.py
ichorid: Good point, I'll fix that.

Tribler/Core/Session.py
ichorid: I'll see if the default `addCallback` works.

Tribler/Core/Upgrade/db72_to_pony.py
ichorid: OK.

Tribler/Core/Upgrade/db72_to_pony.py
ichorid: How could this be non-compatible? It is just generating strings, there is no character input.

Tribler/Core/Upgrade/db72_to_pony.py
ichorid: This is reactor magic. The only reason why this `deferLater` is here is to give other Deferreds running on the same reactor a chance to do their stuff in between `notifier_callback` calls. This thing only truly happens once per conversion loop iteration (a single batch), that is *about 1 second*.  The whole thing works like this:
"at the end of converting each batch (that takes approx 1 second), give control to the reactor and schedule `notifier_callback` to be run in 1 millisecond."

Tribler/Core/Upgrade/db72_to_pony.py
ichorid: Actually, this *is* the safest solution. The thing is here mostly for the programmer's convenience. Upgrading is and should be a completely deterministic process. We *want* Tribler to crash if the clock is touched during the conversion process, because *this should never, ever happen*. Any other behavior can potentially create nearly-uncatchable inconsistencies in produced metadata entities.

Tribler/Core/Upgrade/upgrade.py
ichorid: ... and then some nice-looking single-liners suddenly become two-liners. One can infer from the name what it is, all of its usages fit on a single screen, it is only used locally and making it longer will result in some line wraps.  I'll leave it as it is.

Tribler/Test/Core/Upgrade/test_upgrader.py
ichorid: I used monkey patching because there are only a couple of tests in the file, basically testing the same thing. The problem with this kind of stuff is that the alternative is to introduce more backreferences, which we try to avoid. I'll see what I can do, but I would leave it as it is, instead of adding backreference to e.g. `session` in `DispersyToPonyMigration`. "Backreference" often means "stuck memory".

TriblerGUI/tribler_window.py
ichorid: OK

Tribler/Core/Modules/restapi/events_endpoint.py
devos50: > This messages should always be sent in order, for upgrade notification/skip procedure to work properly.

If possible, I would try to remove any assumptions/requirements on the order of messages being sent/received. Is this possible in this situation?

Tribler/Core/Upgrade/db72_to_pony.py
devos50: Ah I see, missed that 👍

Tribler/Core/Upgrade/db72_to_pony.py
devos50: I think that the Python exception system is built for detection of this kind of behaviour. If we adopt a defensive programming style, we should have some kind of flag whether it can be touched and if it is touched when this flag is set, we raise a `RuntimeError`. However, this might be a bit too much overengineering for this specific situation so I guess it's fine like this.

Tribler/Core/Modules/restapi/events_endpoint.py
ichorid: Yes, that is possible, at least for the particular case of the current upgrader logic.
However, I've already redone the thing to use, literally, FIFO queue with limited size.
Do you insist on removing the queue solution and going instead for non-order-dependent-upgrader-notifications solution?

devos50: In this case, I would suggest to revert this change and not rely on any order if not too much effort 👍

devos50:
@ichorid don't forget to address this Pylint issue:

```
120: W0612: (unused-variable), MetadataStore.__init__.sqlite_disable_sync: Unused variable 'sqlite_disable_sync'
```

ichorid:
> @ichorid don't forget to address this Pylint issue:
> 
> ```
> 120: W0612: (unused-variable), MetadataStore.__init__.sqlite_disable_sync: Unused variable 'sqlite_disable_sync'
> ```

That's intentional. This is Pony's way of plugging in code on DB startup. I've added a comment on this.

Tribler/Test/Core/Modules/RestApi/test_upgrader_endpoint.py
```diff
@@ -0,0 +1,36 @@
...
+
+from twisted.internet.defer import inlineCallbacks
+
+from Tribler.Core.Modules.restapi.upgrader_endpoint import SKIP_DB_UPGRADE_STR
+from Tribler.Test.Core.Modules.RestApi.base_api_test import AbstractApiTest
+from Tribler.Test.Core.base_test import MockObject
+from Tribler.Test.tools import trial_timeout
+
+
+class TestShutdownEndpoint(AbstractApiTest):
```
xoriole: Please rename this endpoint and the docstring on the method test_upgrader_skip()

ichorid: Nice (end)point! :wink:

Tribler/Test/Core/Modules/RestApi/test_upgrader_endpoint.py
```diff
@@ -0,0 +1,36 @@
...
+from Tribler.Test.tools import trial_timeout
+
+
+class TestUpgraderEndpoint(AbstractApiTest):
+
+    @trial_timeout(10)
+    @inlineCallbacks
+    def test_upgrader_skip(self):
+        """
+        Testing whether the API triggers a Tribler shutdown
```
xoriole: Also this one

ichorid: Oops..

PR #4493, opened 2019-05-13
Added support for adding/moving/removing torrents in My Channel

xoriole:
- Add torrents in downloads page to My channel with a right click
- Support multiple torrents selection in My Channel to delete
- Move downloaded torrent to a different location

### DIFF
--- a/TriblerGUI/widgets/downloadspage.py
+++ b/TriblerGUI/widgets/downloadspage.py
@@ -422,6 +422,22 @@ class DownloadsPage(QWidget):
         else:
             self.window().tray_show_message("Torrent file exported", "Torrent file exported to %s" % dest_path)
 
+    def on_add_to_channel(self):
+        self.request_mgr = TriblerRequestManager()
+        for selected_item in self.selected_items:
+            infohash = selected_item.download_info["infohash"]
+            name = selected_item.download_info["name"]
+            post_data = {"uri": u"magnet:?xt=urn:btih:%s&dn=%s" % (infohash, name)}
+            self.request_mgr.perform_request("mychannel/torrents",
+                                             lambda response, _name=name: self.on_added_to_channel(_name, response),
+                                             method='PUT', data=post_data)
+
+    def on_added_to_channel(self, name, response):
+        if not response:
+            return
+        if 'added' in response and response['added']:
+            self.window().tray_show_message("Channel update", "%s torrent successfully added to your channel" % name)
+
     def on_right_click_item(self, pos):
         item_clicked = self.window().downloads_list.itemAt(pos)
         if not item_clicked or self.selected_items is None:
@@ -435,6 +451,7 @@ class DownloadsPage(QWidget):
         start_action = QAction('Start', self)
         stop_action = QAction('Stop', self)
         remove_download_action = QAction('Remove download', self)
+        add_to_channel_action = QAction('Add to My Channel', self)
         force_recheck_action = QAction('Force recheck', self)
         export_download_action = QAction('Export .torrent file', self)
         explore_files_action = QAction('Explore files', self)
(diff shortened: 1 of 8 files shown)


TriblerGUI/widgets/downloadspage.py
```diff
@@ -380,6 +381,31 @@ def on_explore_files(self):
...
+        }
+
+        request_mgr = TriblerRequestManager()
+        request_mgr.perform_request("downloads/%s" % _infohash,
+                                    lambda res, _, name=_name, target=dest_dir: self.on_files_moved(res, name, target),
+                                    data=data, method='PATCH')
+
+    def on_files_moved(self, response, name, dest_dir):
+        if "modified" in response and response["modified"]:
+            self.window().tray_show_message(name, "Moved to %s" % dest_dir)
```
devos50: Could you make this message a bit more descriptive? Something like:

"Moved download big-buck-bunny-1080p.avi to /my/new/path"

(so adding the name of the download)

TriblerGUI/widgets/downloadspage.py
```diff
@@ -422,6 +448,22 @@ def on_export_download_request_done(self, filename, data):
         else:
             self.window().tray_show_message("Torrent file exported", "Torrent file exported to %s" % dest_path)
 
+    def on_add_to_channel(self):
+        self.request_mgr = TriblerRequestManager()
+        for selected_item in self.selected_items:
+            infohash = selected_item.download_info["infohash"]
+            name = selected_item.download_info["name"]
+            post_data = {"uri": u"magnet:?xt=urn:btih:%s&dn=%s" % (infohash, name)}
```
devos50: Don't we have a utility method to construct a magnet URL from info hash and name?

TriblerGUI/widgets/downloadspage.py
```diff
@@ -431,13 +473,16 @@ def on_right_click_item(self, pos):
             self.selected_items.append(item_clicked)
 
         menu = TriblerActionMenu(self)
+        menu.setMinimumWidth(CONTEXT_MENU_WIDTH)
```
devos50: Cool, I didn't know about this method 👍 . Should we add this to the `TriblerActionMenu` object itself maybe?

TriblerGUI/widgets/downloadspage.py
```diff
@@ -431,13 +473,16 @@ def on_right_click_item(self, pos):
...
 
         start_action = QAction('Start', self)
         stop_action = QAction('Stop', self)
         remove_download_action = QAction('Remove download', self)
+        add_to_channel_action = QAction('Add to My Channel', self)
```
devos50: Minor comment: we should be consistent in the spelling of "my channel". I think it's better to not capitalise the `M` and `C` since it may look like some piece of technology that we named.

TriblerGUI/widgets/downloadspage.py
```diff
@@ -431,13 +473,16 @@ def on_right_click_item(self, pos):
...
         force_recheck_action = QAction('Force recheck', self)
         export_download_action = QAction('Export .torrent file', self)
         explore_files_action = QAction('Explore files', self)
+        move_files_action = QAction('Move file storage  ', self)
```
devos50: There seems to be an (unnecessary) space here at the end of the string?

TriblerGUI/widgets/downloadspage.py
xoriole: The name of the torrent is already present as the title of the tray message

TriblerGUI/widgets/downloadspage.py
xoriole: Good point. Updated

TriblerGUI/widgets/downloadspage.py
xoriole: Moved to TriblerActionMenu

TriblerGUI/widgets/downloadspage.py
xoriole: sure, done

PR #4503, opened 2019-05-18
WIP: integrates bloom filter in tribler (from py-ipv8)

jonay2000:
As far as i could see, this resolves all references to the bloom filter from pyipv8 and redirects them to an implementation in tribler. I am not completely sure where i should put it (it's in core now). Input is appreciated. As promised in https://github.com/Tribler/py-ipv8/pull/477

### DIFF
--- a/Tribler/Core/bloomfilter.py
+++ b/Tribler/Core/bloomfilter.py
@@ -1,0 +1,290 @@
+# disable E0611, No name %r in module %r.  pylint is unable to correctly assess the content of hashlib
+# pylint: disable=E0611
+
+# disable E1103, %s %r has no %r member (but some types could not be inferred).  Because pylint can not correctly assess
+# the content of hashlib is can not assess its members either.
+# pylint: disable=E1103
+
+"""
+This module provides the bloom filter support.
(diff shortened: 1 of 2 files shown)


devos50:
ok to test

jonay2000:
oh whoops, i totally forgot to also add the tests from py-ipv8 for the bloom filter. Right now only the code itself is added (that is why coverage fails). Should i add now or wait until you figure out if it should be there at all / refactored (https://github.com/Tribler/py-ipv8/pull/477)

qstokkink:
@devos50 as the market is still in, should we merge this now?

devos50:
@jonay2000 could you move the Bloomfilter class to the `Tribler.community.market` module (since the market is the only component that actually uses the bloomfilter)?

Also, I will remove the market implementation from Tribler and add [the submodule](https://github.com/tribler/anydex) instead as soon as possible. 👍

jonay2000:
I am currently on holiday so I'll try to get it done by Monday. I'm sorry.

devos50:
Superseded by #4547

PR #4505, opened 2019-05-20
Channel updates for 7.3

xoriole:
Includes various GUI updates on My Channel page and plus button at the navigation bar.


Tribler/Core/Config/config.spec
```diff
@@ -68,6 +68,7 @@ seeding_mode = string(default='ratio')
 seeding_ratio = float(default=2.0)
 seeding_time = float(default=60)
 channel_download = boolean(default=False)
+add_download_to_channel = boolean(default=False)
```
ichorid: I feel like this option should belong to the GUI config, and not to the core config. Instead, the downloading REST API should provide a URL `&` parameter for a download to do it.

It is a cleaner design choice to move user convenience options into the GUI part completely.

TriblerGUI/widgets/settingspage.py
```diff
@@ -60,6 +61,17 @@ def initialize_settings_page(self):
...
+        checkbox_style = get_checkbox_style()
+        for checkbox in [self.window().family_filter_checkbox, self.window().channel_autocommit_checkbox,
+                         self.window().always_ask_location_checkbox, self.window().developer_mode_enabled_checkbox,
+                         self.window().use_monochrome_icon_checkbox, self.window().download_settings_anon_checkbox,
+                         self.window().download_settings_anon_seeding_checkbox, self.window().lt_utp_checkbox,
+                         self.window().watchfolder_enabled_checkbox, self.window().allow_exit_node_checkbox,
+                         self.window().credit_mining_enabled_checkbox, self.window().developer_mode_enabled_checkbox,
+                         self.window().checkbox_enable_network_statistics, self.window().checkbox_enable_resource_log,
+                         self.window().download_settings_add_to_channel_checkbox]:
+            checkbox.setStyleSheet(checkbox_style)
```
ichorid: We should instead use CSS stylesheets in the `.ui` file. It is especially important as we are going to make the GUI's CSS  totally configurable. Is it possible to move this thing to the stylesheet?

TriblerGUI/widgets/ClickableLineEdit.py
```diff
@@ -0,0 +1,20 @@
+from PyQt5.QtCore import pyqtSignal
+from PyQt5.QtWidgets import QLineEdit
+
+
+class ClickableLineEdit(QLineEdit):
+
+    clicked = pyqtSignal()
+    focussed = pyqtSignal(bool)
```
ichorid: should be `focused` instead :wink:

TriblerGUI/widgets/editchannelpage.py
```diff
@@ -126,7 +127,17 @@ def initialize_with_channel_overview(self, overview):
...
+
+        # Channel name
+        self.window().edit_channel_name_label.setText(self.channel_overview["name"])
+        self.window().edit_channel_name_label.setReadOnly(True)
+        self.window().edit_channel_name_label.clicked.connect(self.on_click_channel_name)
+        self.window().edit_channel_name_label.returnPressed.connect(self.on_update_channel_name)
+        self.window().edit_channel_name_label.focussed.connect(self.on_focus_channel_name)
+
+        # Channel public key
+        self.window().edit_channel_cid_label.setText(self.channel_overview["public_key"][:74] + "...")
```
ichorid: :+1:

TriblerGUI/widgets/editchannelpage.py
```diff
@@ -97,6 +89,9 @@ def initialize_edit_channel_page(self, gui_settings):
             self.controller.table_view.setColumnHidden(4, True)
             self.model.exclude_deleted = False
 
+    def showEvent(self, QShowEvent):
```
ichorid: :+1:

TriblerGUI/dialogs/createtorrentdialog.py
```diff
@@ -30,11 +30,7 @@ def __init__(self, parent):
 
 class CreateTorrentDialog(DialogContainer):
 
-    button_clicked = pyqtSignal(int)
-    received_metainfo = pyqtSignal(dict)
-
-    signal_torrent_created = pyqtSignal(dict)
-    signal_torrent_added_to_channel = pyqtSignal(dict)
+    signal_create_torrent_updates = pyqtSignal(dict)
```
ichorid: Please, rename the signal to something more specific, like `create_torrent_tick` or `create_torrent_finished`, etc.

Tribler/Core/Config/config.spec
xoriole: It is a GUI config but it is also specific to download defaults which I think is a setting that should be in tribler.conf so we can support portable config when we do.

TriblerGUI/widgets/settingspage.py
xoriole: The way our ui is structured, it does not seem to support class-based styles so I needed either to repeat the styles for each checkboxes or apply programmatically. We should refactor it when write global CSS for Tribler.

PR #4514, opened 2019-05-23
READY: Fixes for beta3

ichorid:
Small fixes for the latest beta

Fixes #4506 
Fixes #4510

PR #4517, opened 2019-05-23
READY: VSIDS-based channel rating system

ichorid:
This PR implements a [VSIDS](https://arxiv.org/pdf/1506.08905.pdf)-like heuristics for channel popularity.
In short, whenever a gossip with a channel is received, the corresponding vote count is increased by +1, and then *all the other votes* are **multiplied** by a constant *decay factor*. As a result, the heuristic works like a temporal PageRank, bumping the most actively spread around channels to the top.

The decay factor for this version is 2.71 per 24 hours of Tribler online time. This means that when some channel gets a vote today, tomorrow by the same time the vote will be 2.71x times smaller. The decay is performed on per-vote basis, meaning that it happens gradually over time, as the votes come in. The algorithm is O(1) in regards to the number of DB calls for each vote update.

We keep each individual vote in the DB, along with its last update time and the amount that it added to the total vote count for the channel. Whenever a repeating vote is received, the previous vote amount for the same vote is *subtracted* from the vote sum, and the new amount is added instead. This results in votes *refresh* instead of *accumulation*, meaning that a single peer cannot aggressively bump their channel by bombarding another peer with gossip. Instead, persistence in vote support is key.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import os
 import random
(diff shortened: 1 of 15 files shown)


devos50:
Nice mechanism! 👍 

Dynamic decaying is something for next releases I guess?

ichorid:
> Dynamic decaying is something for next releases I guess?
Most probably, yes. For this release, I'll have to work out the normalization and QT delegate for displaying the rating.

synctext:
Great stuff!
Can you please schedule a performance evaluation? Like using 250k real votes from the old AllChannel and test the retrieval time plus vote magic cost. Nightly Gumby experiment is deemed useful?

ichorid:
@synctext , I'll think about how to the experiment properly.

ichorid:
Note that this breaks the database format. The old database should be removed manually unless we merge #4514 first, and then change this PR to use the beta-deletion method.

devos50:
Adding an additional [validation experiment](https://jenkins-ci.tribler.org/job/validation_experiments/)] would indeed be helpful, not only for correctness at scale but also to have a solid foundation to run some performance/accuracy tests on more advanced voting mechanisms in the future.

ichorid:
retest this please

Issue #2901, opened 2017-04-16
[Feature Request] Move complete downloads to new folder


startailcoon:
This is a feature request.

When a download is complete, move it to a new folder to separate it from incomplete downloads.
Many modern clients has this feature.

Dmole:
While implementing this, one might consider groups (like rtorrent) where multiple watch/downloading/complete folders can be specified... If Tribler were optimised a bit it might gain usage as a seeding tool (not just a hit and run tool) where it is advantageous to group torrents by tracker/type/genre/etc.

A "move Storage" option per torrent in the UI is another related feature.

Thatoo:
> A "move Storage" option per torrent in the UI is another related feature.

Is the real needed feature. Being able to move a downloaded file we want to keep to a desired location and keeping seeding it is really important.
Files are first all downloaded in the download folder to check them. Once checked and validated as valuable file, it should be possible to move it to a new folder (each file in different folders depending on how one classify his/her computer) and yet keep seeding it.

Dmole:
> ...needed...
In case you are interested a workaround is to use [hard links](https://en.wikipedia.org/wiki/Hard_link), which has the added advantages or renaming files, splitting folders, etc.

Thatoo:
Thank you for this fast reply.
However, I'm not sure to understand...
I just think it should be good to be able to right click on a torrent in Trible, and click on something like "move to...", chose a location and it moves the folder and files of this specific torrent to a the chosen location.
I hope I'm clear...

xoriole:
Support for moving torrent is recently added and will be available from 7.3 beta 3.
![2019-05-17-112431_3840x1291_scrot](https://user-images.githubusercontent.com/1442867/57918331-dd048800-7896-11e9-9d1a-f3491fd30a2d.jpg)

qstokkink:
@xoriole can we close this issue now then?

Dmole:
The request was for a torrent to be automatically moved on completion, so unless that is also implemented I think that this issue should technically stay open.

xoriole:
Not yet, we still don't have move on complete. The previous PR only offers an option to move the torrent storage but its a user action and does not do automatically on complete. While this is relatively easy to do but we need to introduce something like a move directory in settings. We should discuss about it.

mrjonlovem:
[...]

PR #4521, opened 2019-05-25
Documentation improvements

devos50:
In this PR, I made various improvements to the (outdated) documentation:
- Updated REST API docs, removed non-existing endpoints and included new endpoints. For our convenience, we should have a check that automatically includes these endpoints.
- Fixed all warnings when building the documentation.
- Added doc build status badge to README.md.
- Added [a PR job](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_build_docs/) that builds the documentation and fails when the documentation cannot be built successfully.

Note that I consider the extensive mocking in `conf.py` not pretty, but necessary since we cannot build dependencies like `libtorrent` in the `virtualenv` environment of readthedocs.


devos50:
retest this please

heldersepu:
Would be nice if the REST API docs was using swagger/OAS

heldersepu:
I could not find this section:
https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Modules/restapi/downloads_endpoint.py#L444
anywhere in the documentation

qstokkink:
> 
> 
> Would be nice if the REST API docs was using swagger/OAS

We are currently creating a proof of concept using Swagger. It will take a little while for this to make it into Tribler though.

Dmole:
How about "Self-documenting code"?

README.rst
```diff
@@ -2,11 +2,11 @@
 Tribler
 *******
 
-|jenkins_build| |contributors| |pr_closed| |issues_closed| |downloads_7_0| |downloads_7_1| |downloads_7_2| |doi| |openhub|
+|jenkins_build| |docs| |contributors| |pr_closed| |issues_closed| |downloads_7_0| |downloads_7_1| |downloads_7_2| |doi| |openhub|
 
 *Towards making Bittorrent anonymous and impossible to shut down.*
 
-Developers usually hang out in the official IRC channel #tribler @ FreeNode (click `here <http://webchat.freenode.net/?channels=tribler>`_ for direct a webchat window)
+Developers usually hang out in the official IRC channel #tribler @ FreeNode (click `here <http://webchat.freenode.net/?channels=tribler>`__ for direct a webchat window)
```
ichorid: Except they really don't... :wink:

Tribler/Core/Modules/restapi/search_endpoint.py
```diff
@@ -16,8 +16,8 @@
 
 class SearchEndpoint(BaseMetadataEndpoint):
     """
-    This endpoint is responsible for searching in channels and torrents present in the local Tribler database. It also
-    fires a remote search in the Dispersy communities.
+    This endpoint is responsible for searching in channels and torrents present in the local Tribler database.
+    It also fires a remote search in the IPv8 channel community.
```
ichorid: Shouldn't this be "GigaChannel" community instead?

Dmole:
I think the target for API documentation would be a power user trying to implement some automation without having to waste time reading 3994+ lines of code, when it could be reduced to ~81 lines with some trivial 4 line script;
```
while read -r F ; do
	echo "$F" | perl -pe 's/.*\///g;s/_.*//g'
	grep  -P "in (request.args|parameters)" "$F" | perl -pe 's/if //g;s/^[ \t]*/\t\t/g;s/ in .*//g'
done < <(find Tribler/Core/Modules/restapi/ -name '*_endpoint.py')
```

qstokkink:
> 
> 
> How about "Self-documenting code"?
> (Just have a script build the documentation from the API code)

This is exactly why we need something like Swagger.

Currently all our REST endpoints are custom magic, which cannot be automatically read ([I did, at one point, document the REST API in our networking library using a custom Swagger-like specification to deal with our own magic code](https://github.com/Tribler/py-ipv8/files/2708742/output.pdf)).

By porting our REST API to something like Swagger, it will also force us to clean the API and make it a bit easier to work with 😃

Dmole:
Also all UI communication to the core should be forced to use the API (or have the HTTP API auto updated from whatever the UI API is using)

For example the new move feature
https://github.com/Tribler/tribler/issues/2901#issuecomment-493388260
I don't see that in the API
https://github.com/Tribler/tribler/blob/master/Tribler/Core/Modules/restapi/downloads_endpoint.py
but maybe I'm looking on the wrong branch or something.

qstokkink:
@Dmole From reverse engineering this unit test:

https://github.com/Tribler/tribler/blob/562b8d86b4dcf4c821b1df5039276734116a6dd3/Tribler/Test/Core/Modules/RestApi/test_downloads_endpoint.py#L487-L499

Moving downloads should be done with an `HTTP PATCH` request to `'downloads/<HEX ENCODED INFOHASH>'` and a request body of:

```
{
"state": "move_storage", 
"dest_dir": <NEW DIRECTORY> 
}
```

Dmole:
Ah it's on devel not master, thanks.

heldersepu:
> What is our documentation's target audience? Developers? Personally, I don't use it. Extensive in-code comments and proper code structure are much more useful to me. Users? Haven't met one who reads the docs. Power users? Maybe. Fresh developers and external contributors? It is hard to say.
> 

@qstokkink  Consider me part of that target audience... What am I? Developer? SRE? Power user? 
But yes the docs is the first place I go when interacting with a REST api, that is all you get on non OpenSource projects, I had a "hard time" putting together a script to queue downloads:
https://github.com/heldersepu/hs-scripts/blob/master/Bash/tribler_queue.sh

ichorid:
As @heldersepu pointed out, the hottest point of our documentation is the REST API because it is, well, *the* interface for Tribler. Remind that we had some discussions about moving to a proper Web API toolkit, like Flask's `asyncio`-based counterpart [Quart](https://hackernoon.com/3x-faster-than-flask-8e89bfbe8e4f). I wonder if it is possible to use Flask's robust Swagger-integration features ([RESTPlus](http://michal.karzynski.pl/blog/2016/06/19/building-beautiful-restful-apis-using-flask-swagger-ui-flask-restplus/)) with Quart...

PR #4526, opened 2019-05-28
Change VSIDS scaling

ichorid:
Change VSIDS scaling to max of all votes.


qstokkink:
The `External import checker` is failing because this PR is not rebased onto the latest Tribler head.

Issue #4524, opened 2019-05-28
The filename, directory name, or volume label syntax is incorrect


xoriole:
From error reporter - Tribler 7.3.0 - beta3
```python
Traceback (most recent call last):
File "TriblerGUI\core_manager.py", line 118, in on_received_state
RuntimeError: [Failure instance: Traceback: <type 'exceptions.ValueError'>: The filename, directory name, or volume label syntax is incorrect
run_tribler.py:85:start_tribler
Tribler\Core\Session.py:432:start
lib\site-packages\twisted\internet\defer.py:321:addCallback
lib\site-packages\twisted\internet\defer.py:310:addCallbacks
--- <exception caught here> ---
lib\site-packages\twisted\internet\defer.py:653:_runCallbacks
Tribler\Core\Session.py:429:load_checkpoint
Tribler\Core\Session.py:384:load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:570:load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:567:do_load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:599:resume_download
Tribler\Core\TorrentDef.py:115:load_from_dict
Tribler\Core\TorrentDef.py:68:__init__
]
```

```
Traceback (most recent call last):
File "TriblerGUI\core_manager.py", line 118, in on_received_state
RuntimeError: [Failure instance: Traceback: <type 'exceptions.ValueError'>: La syntaxe du nom de fichier, de r\xe9pertoire ou de volume est incorrecte
run_tribler.py:85:start_tribler
Tribler\Core\Session.py:432:start
lib\site-packages\twisted\internet\defer.py:321:addCallback
lib\site-packages\twisted\internet\defer.py:310:addCallbacks
--- <exception caught here> ---
lib\site-packages\twisted\internet\defer.py:653:_runCallbacks
Tribler\Core\Session.py:429:load_checkpoint
Tribler\Core\Session.py:384:load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:570:load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:567:do_load_checkpoint
Tribler\Core\APIImplementation\LaunchManyCore.py:599:resume_download
Tribler\Core\TorrentDef.py:115:load_from_dict
Tribler\Core\TorrentDef.py:68:__init__
]
```

ichorid:
[...]

PR #4531, opened 2019-06-04
READY: Unicode-related fixes for 7.3

ichorid:
This PR contains some fixes to make Tribler working directory path and `.state` files support non-ASCII characters.


qstokkink:
Does this fix #4524?

ichorid:
@qstokkink , I don't know. Probably, not. But it fixes some related bugs that I discovered trying to reproduce #4524 .

PR #4532, opened 2019-06-04
READY: Free-for-all unsigned gossip

ichorid:
This PR adds unsigned gossip support to GigaChannel. To achieve this, we add an ability to process and send unsigned ("free-for-all", FFA) payloads . Zeroed out public key and signatures are space-efficient because:

1. these are never serialized to disk in .mdblob files, but instead only exist in Pony DB;
2. on the wire zeroes are very efficiently compressed by LZ4;
3. in the DB the fields are left empty.

Legacy entries are now stored as FFA and shared through the gossip.

Also, this PR adds basic support for nested channels to REST endpoints and `share` column to DB. This will help us avoid DB conversion on 7.4 when the nested channels GUI is ready.


Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
```diff
@@ -80,11 +86,24 @@ def __init__(self, *args, **kwargs):
...
+                    raise InvalidChannelNodeException(
+                        ("Attempted to create %s free-for-all (unsigned) object without specifying id_ : " %
+                         str(self.__class__.__name__)))
+
             # For putting legacy/test stuff in
-            skip_key_check = False
-            if "skip_key_check" in kwargs and kwargs["skip_key_check"]:
-                skip_key_check = True
-                kwargs.pop("skip_key_check")
+            skip_key_check = kwargs.pop("skip_key_check") if "skip_key_check" in kwargs else skip_key_check
```
qstokkink: You could shorten this to `skip_key_check = kwargs.pop("skip_key_check", skip_key_check)`.

qstokkink:
Make sure to strip the `READY: ` from the PR title before you merge.

ichorid:
> Make sure to strip the `READY: ` from the PR title before you merge.

oops, I did it again...

ichorid:
I propose using READY/WIP tags instead of name changes to solve this problem.

qstokkink:
Only admins can assign tags. Then normal users can no longer contribute.

PR #4538, opened 2019-06-07
Automatically add torrents to free-for-all channels

ichorid:
Whenever the user checks out a magnet link or opens a torrent, a corresponding is added to free-for-all channel. These entries are only spread around via remote search requests.


ichorid:
retest this please

PR #4547, opened 2019-06-10
Integrated Bloomfilter in market, updated IPv8 pointer

devos50:
Replaces #4503.


devos50:
Depends on https://github.com/Tribler/py-ipv8/pull/522

Issue #4438, opened 2019-03-26
Magnet link not working


Nubro01:
##### Tribler version/branch+revision: #####
7.3 march 26th build
##### Operating system and version: #####
Windows 10 Insider 18362
##### Steps to reproduce the behavior: #####
Find a Torrent and click the magnet link
##### Expected behavior: #####
Tribler is accepting the link and opens a download location and file selection dialog window
##### Actual behavior: #####
A second tribler task is started and a blank screen is shown, after a while the screen closes, but nothing is started or downloaded.
The last couple of builds also did not show the files to select for download, if I just select ok, then torrent is started and all files are downloaded.

xoriole:
@Nubro01 Thank you for reporting. I'm not sure if I can understand what you meant.
> A second tribler task is started and a blank screen is shown, after a while the screen closes, but nothing is started or downloaded.

When you try to download a torrent using magnet link, Tribler should show a dialog box either with a list of files inside the torrent or display a timeout message as shown in the pic below. I'm not sure what you are referring as blank screen and closing of the screen after a while. Could you possibly attach a (safe) screenshot?
![timeout](https://user-images.githubusercontent.com/1442867/55062200-bdf73000-5075-11e9-9cc2-dfcf22a6d2c9.png)

Nubro01:
I know how Tribler works, I have been testing it for ages! When I select a Magnet Link with the latest 7.3 versions under windows 10 build 1903 (!!!) I get a blank white screen for a few seconds, no time and no cause to print  just a white screen, I would think. The above dialog is not appearing!
If I manage to download the torrent and add it manually to Tribler, the dialog indeed appears, but the files to select are never show, when I just select download, the torrents is started and all files are downloaded, when I try to exclude a file, in the previous versions of 7.3, Tribler would abend. 
I'll try it again as soon as I have a new file to download.

synctext:
Complex issue. Trying to understand..
Are you opening a magnet link from a browser, while Tribler is already running? So we messed up the 2nd instance check logic. You have tested a 7.3 version that did work?
A white screen glitch..

Btw it would be really efficient to fix your bugs with instant visualisation of the Tribler flaw. Any Ubuntu download as a test case.
Animated Gif creation Tool: https://github.com/phw/peek
.gif uploads are supported by Github.

devos50:
I see several problems here.

[...]

PR #4556, opened 2019-06-11
Fix for loading torrent or magnet link in Tribler via external click

xoriole:
This PR fixes the issues mentioned in https://github.com/Tribler/tribler/issues/4438.

Fixes the failure to load torrent or magnet link when Trible is already running. Additionally, adds an option to retry fetching the meta info in download dialog if it times out to load the files on the first request.

### DIFF
--- a/TriblerGUI/tribler_window.py
+++ b/TriblerGUI/tribler_window.py
@@ -707,6 +707,8 @@ class TriblerWindow(QMainWindow):
                 self.fetch_settings()
                 self.dialog = ConfirmationDialog.show_error(self, "Download Error", "Tribler settings is not available\
                                                                    yet. Fetching it now. Please try again later.")
+                # By re-adding the download uri to the pending list, the request is re-processed
+                # when the settings is received
                 self.pending_uri_requests.append(uri)
                 return
             # Clear any previous dialog if exists
--- a/TriblerGUI/tribler_window.py
+++ b/TriblerGUI/tribler_window.py
@@ -532,6 +532,12 @@ class TriblerWindow(QMainWindow):
         if 'error' in settings:
             raise RuntimeError(TriblerRequestManager.get_message_from_error(settings))
 
+        # If there is any pending dialog (likely download dialog or error dialog of setting not available),
+        # close the dialog
+        if self.dialog:
+            self.dialog.close_dialog()
+            self.dialog = None
+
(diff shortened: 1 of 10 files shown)


xoriole:
retest this please

TriblerGUI/tribler_window.py
```diff
@@ -532,6 +532,12 @@ def received_settings(self, settings):
         if 'error' in settings:
             raise RuntimeError(TriblerRequestManager.get_message_from_error(settings))
 
+        # If there is any pending dialog (likely download dialog or error dialog of setting not available),
+        # close the dialog
+        if self.dialog:
```
ichorid: I don't get it, why do we need this? In what kind of situation this should trigger? This can become a source of race conditions in the GUI and make its behavior non-deterministic.

run_tribler.py
```diff
@@ -115,16 +115,18 @@ def start_tribler():
             from TriblerGUI.tribler_window import TriblerWindow
 
             app = TriblerApplication("triblerapp", sys.argv)
-
+            if app.is_running():
+                for arg in sys.argv[1:]:
+                    if os.path.exists(arg) and arg.endswith(".torrent"):
+                        app.send_message("file:%s" % arg)
```
ichorid: Maybe we should add a test for this? If it is not very hard to do, of course.

TriblerGUI/tribler_window.py
xoriole: I was able to get to this situation trying to reproduce nubro's issue. This happens if you're opening a torrent file with Tribler by clicking it. Tribler will try to load the download dialog but if the setting is not fetched it, it will show an error dialog and the UI gets stuck on it. Note the changes in made in `start_download_from_uri()`. By re-adding the download uri to the pending list, the request is re-processed when the settings is received. With the above code, it makes sure that error dialog is closed and proper download dialog is shown.

run_tribler.py
xoriole: ok

TriblerGUI/tribler_window.py
ichorid: Excellent!
Please, put this:

>By re-adding the download uri to the pending list, the request is re-processed when the settings is received.

in a comment in the code.

It was not obvious for me during the review, so it should be commented so we do not forget why we need this :wink:

PR #4562, opened 2019-06-12
Make cast_to_unicode_utf8 safer

ichorid:
This little hack makes `cast_to_unicode_utf8` robust against wrong input types, such as `unicode`. This can happen, say, on Windows where paths are unicode.

Overall, this kind of hacks is really bad and only masks problems. However, until we move to Python3 we cannot really start fixing the unicode mess.


ichorid:
retest this please

egbertbouman:
@ichorid Please don't forget that I'm currently working on Python 3, and the more files you touch, the more problems I'll get when I have to do a rebase.

ichorid:
@egbertbouman , I'll try to make it as non-invasive as possible.

ichorid:
retest this please

ichorid:
retest this please

ichorid:
After spending two days fighting encodings on python2.7 on Windows, trying to make Tribler work in a non-ASCII-named dir, I say:
:imp: **to hell with it.** :imp: 

It is impossible to solve it and keep one's sanity without moving to either Python3 or some magic intermediate package like `pathlib`.
Thus, I revoke the part of this patch that adds non-ASCII characters to TMP dir path. I was able to make all but one test work with non-ASCII paths, but fighting this last test I realized it really does not worth it.

We should move to Python3 (and/or `pathlib`) ASAP.

PR #4568, opened 2019-06-14
READY: Make personal channel use timestamp for id_

ichorid:
This PR adds further support for nested channel feature. Before, user's channel always got a 0 `id_`. This is compatible with the planned nested channels scheme, but is inconvenient, because after moving to nested channels users will have their personal channels contents right at their root channel. Instead, we're now creating a nested personal channel with timestamp-based `id_` that resides in the user's root (`id_==0`) channel.

There is no real change in the database schema, but we still have to clean the previous beta database, because otherwise, our beta testers will end up with messy channels after the transition to nested channels in 7.4.


ichorid:
retest this please

ichorid:
retest this please

PR #4572, opened 2019-06-17
READY: Add GUI notifications to personal channel conversion process

ichorid:
This adds progress notifications to the personal channel conversion process, the same as we use for converting legacy torrents to FFA entries.

PR #4574, opened 2019-06-17
Fix torrent health conversion

ichorid:
This adds conversion of the health info from 7.2.2 database to Pony. Also, it fixes a bug with torrents conversion introduced in #4572 that resulted in skipping conversion of most entries.

### DIFF
--- a/Tribler/Core/Upgrade/db72_to_pony.py
+++ b/Tribler/Core/Upgrade/db72_to_pony.py
@@ -17,6 +17,7 @@ from twisted.internet import reactor
 from twisted.internet.defer import inlineCallbacks
 from twisted.internet.task import deferLater
 
+from Tribler.Core.Modules.MetadataStore.OrmBindings.channel_metadata import BLOB_EXTENSION
 from Tribler.Core.Modules.MetadataStore.OrmBindings.channel_node import LEGACY_ENTRY, NEW
 from Tribler.Core.Modules.MetadataStore.OrmBindings.torrent_metadata import infohash_to_id
 from Tribler.Core.Modules.MetadataStore.serialization import REGULAR_TORRENT, int2time, time2int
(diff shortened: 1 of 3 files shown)


ichorid:
retest this please

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -292,23 +302,22 @@ def convert_async(self, add_to_pony, get_old_stuff, total_to_convert, offset=0,
...
-            except (TransactionIntegrityError, CacheIndexError):
-                pass
+            with db_session:
+                for (torrent_dict, health) in batch:
+                    try:
+                        torrent = add_to_pony(torrent_dict)
+                        if torrent and health:
+                            torrent.health.set(**health)
+                    except:
+                        pass
```
qstokkink: Could you hook in a logging message here for easier debugging in the future?

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -172,49 +174,58 @@ def get_old_torrents(self, personal_channel_only=False, batch_size=10000, offset
...
+                leechers = int(num_leechers or 0)
+                last_tracker_check = int(last_tracker_check or 0)
                 health_dict = {
-                    "seeders": int(num_seeders or 0),
-                    "leechers": int(num_leechers or 0),
-                    "last_check": int(last_tracker_check or 0)}
+                    "seeders": seeders,
+                    "leechers": leechers,
+                    "last_check": last_tracker_check
+                } if (last_tracker_check >= 0 and seeders >= 0 and leechers >= 0) else None
```
xoriole: Putting `and` might not be the correct condition here

ichorid: Why so? The reason for this check is to only convert health entries that have correct (non-negative) attribute values.

xoriole: You're right.

PR #4586, opened 2019-06-19
READY: Fix get() queries to support nested channels

ichorid:
Another fix to support recent changes in channel addressing.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -202,17 +202,18 @@ class MetadataStore(object):
         self._shutting_down = True
         self._db.disconnect()
 
-    def process_channel_dir(self, dirname, channel_id, external_thread=True):
+    def process_channel_dir(self, dirname, public_key, id_, external_thread=True):
         """
         Load all metadata blobs in a given directory.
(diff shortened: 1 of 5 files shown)


Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -202,17 +202,18 @@ def shutdown(self):
         self._shutting_down = True
         self._db.disconnect()
 
-    def process_channel_dir(self, dirname, channel_id, external_thread=True):
+    def process_channel_dir(self, dirname, public_key, id_, external_thread=True):
```
xoriole: Can `id_` be a keyword argument with default value instead?

ichorid: I don't like this idea, as it will add lots of confusion when we move to nested channels.
We have 2 options here:
 * default = 0
 * default = the channel with the highest id_

Zero channel should never be shared, but sometimes the second default option will still point to it. On the other hand, if 0 is the default, it is not very useful.

Why do you want this?

xoriole: My line of thought here was to make id_ optional with default behaviour to take the latest/highest value. Anyway, it is fine as is.

PR #4587, opened 2019-06-19
READY: Restart channel processing if it was stopped abruptly

### LINKED ISSUE #4578: Error on shutdown in latest devel
```python
ERROR   1560869708.24            tribler_window:74   (root)  Traceback (most recent call last):
  File "/home/quinten/Documents/tribler/TriblerGUI/event_request_manager.py", line 136, in <lambda>
    self.reply.error.connect(lambda error: self.on_error(error, reschedule_on_err=reschedule_on_err))
  File "/home/quinten/Documents/tribler/TriblerGUI/event_request_manager.py", line 81, in on_error
    raise RuntimeError("Could not connect with the Tribler Core within 20 seconds")
RuntimeError: Could not connect with the Tribler Core within 20 seconds
```

ichorid:
Before, if a channel was downloaded, but never processed, it would become stuck in `DOWNLOADING` state and never be processed by MetadataStore. This PR fixes this problem by introducing a channel dir processing queue in GigaChannel Manager and adding an additional check for downloaded, but unprocessed channels. Coincidentally, introducing this queue fixes potential race conditions and launching of more than one background channel-processing thread at a time.

Also, fixes #4578

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
@@ -433,7 +433,8 @@ def define_binding(db):
         @classmethod
         @db_session
         def get_updated_channels(cls):
-            return select(g for g in cls if g.subscribed and (g.local_version < g.timestamp))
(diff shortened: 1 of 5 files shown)


qstokkink:
@xoriole could you also review?

PR #4590, opened 2019-06-20
Fixed issues with homepage tiles

### LINKED ISSUE #4588: resize_event_timer is not single shot
https://github.com/Tribler/tribler/blob/8c74fb9559afd714f557505384136541532e74b6/TriblerGUI/widgets/homepage.py#L95-L102

This causes the homepage to be resized continuously. Thankfully, this also isn't really noticeable anyway.

qstokkink:
Fixes the annoying streching of tiles on tab switch:
afbeelding

Fixes #4588.


qstokkink:
Seems like the Mac is sleeping again, marking this as ready.

PR #4617, opened 2019-06-25
Twistified metainfo lookups

### LINKED ISSUE #4614: Metainfo lookup not done when BEP33 is not available
The torrent health check is simply not done if BEP33 is not available. This means that torrents without any tracker attached will not be checked in that case. We should have a fallback on the 'old' torrent health lookup mechanism to prevent this.

devos50:
In this PR, I made a few changes to the metainfo lookup mechanism:
- Refactored the legacy code and twistified the whole lookup mechanism. A call to `get_metainfo` now returns a `Deferred` that fires with the requested metainfo, or `None` if no metainfo could be found.
- Removed the `fix_torrent` method. We are not sure about the consequences of this removal though but it should not impact overall stability.
- Implemented fallback on old torrent health lookup mechanism.
- Removed various unused methods and variables in the torrent checker session classes.

Fixes #4614


devos50:
Fixed the issue in the GUI, had to maintain a list of `Deferred` instances in the `self.metainfo_requests`, instead of a single one.

Issue #4639, opened 2019-06-27
Discovered channels shows incorrect number of torrents for own channel


qstokkink:
The "discovered" version of my own channel shows 1/3 torrents, when the "my channel" version only has 1 torrent (which is correct).

![afbeelding](https://user-images.githubusercontent.com/3630389/60257925-5fd82400-98d4-11e9-9b67-efc6dcaf44c3.png)

Issue #4627, opened 2019-06-25
Channel reports completion but seems incomplete


devos50:
On my Mac at home, I subscribed to a channel with many torrents. It downloaded successfully and then started to process mdblobs (this process actually became slower over time). When it appeared to be done, it reports completed, even though the counter seems to indicate that I do not have all the processed items.

![Schermafbeelding 2019-06-25 om 21 35 42](https://user-images.githubusercontent.com/1707075/60127852-8477b300-9791-11e9-9dc8-26ecf87f1fe5.png)

xoriole:
I think this is because of the bug in committing the channel. This is related to https://github.com/Tribler/tribler/issues/4639 where discovered channels show an incorrect number of torrents for own channel. I believe fixing #4639 should also fix this.

qstokkink:
Also overlap with #4638

Issue #4601, opened 2019-06-24
GUI issues for 7.3


xoriole:
- [x] Torrent count in the channel header is not correct [assigned to @xoriole]
![incorrect_stats](https://user-images.githubusercontent.com/1442867/60005350-9a408780-966e-11e9-8e12-56381abfc838.png)

- [x] Channel options still visible when there is no channel
![channel_options_issue](https://user-images.githubusercontent.com/1442867/60005407-b8a68300-966e-11e9-9fcd-2b0899a3085c.png)

- [x] Resizing the torrent details leaves some blank space in the channel page. Scrolling the list fixes it. This should be auto resizing itself.
![resize_issue](https://user-images.githubusercontent.com/1442867/60006026-f48e1800-966f-11e9-823f-412fc81a13d6.png)

- [x] There is a 10px offset to the right of the discovered channels list
- [x] The sorting indicator should be removed from the first column in the channels list, and the last column of the torrents table (with the action buttons).
- [x] On Mac, various characters are not rendering correctly
- [x] The progress bar of the trust graph blends with the background on the trust page
![0dfdf6478093c4d2d4da2b6f2e561e3c](https://user-images.githubusercontent.com/1707075/60015248-4be9b380-9683-11e9-85bd-0b4bdd210e10.png)

xoriole:
Note on Unicode support: Python 2 on Windows and MacOS only support basic multilingual characters by default but not wide characters. Wide characters support requires python to be compiled with `--enable-unicode=ucs4`. Our builders are not compiled with ucs4 support. This could be one of the reasons why we get Unicode crashes sometimes. 
This is also the reason why full circle symbol is not displayed in the Trust Graph in Windows & MacOS while working fine in Linux.

PR #4641, opened 2019-06-28
Various Giga channel fixes

xoriole:
Various Giga channel fixes:
- Refresh button for personal channel is not shown anymore
- Shows if the channel is outdated and caps the torrents count to max available
- Removed 'Add to My Channel' option from personal channel showed in discovered page
- Added unsortable columns in different table views

Fixes https://github.com/Tribler/tribler/issues/4639
Fixes https://github.com/Tribler/tribler/issues/4627
Fixes https://github.com/Tribler/tribler/issues/4601

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -177,6 +177,11 @@ class MetadataStore(object):
             # Decay only happens while Tribler is running
             default_vsids.last_bump = datetime.utcnow()
 
+        # Flag to skip processing of personal channel torrent metadata. This is True by default because we assume
+        # the user always has the latest channel version and no need to update his own channel with gossipped data
+        # from the network. We need this flag in testing the processing of metadata.
+        self.skip_received_personal_metadata_payload = True
+
     @db_session
     def upsert_vote(self, channel, peer_pk):
         voter = self.ChannelPeer.get(public_key=peer_pk)
@@ -318,7 +323,15 @@ class MetadataStore(object):
 
             # We separate the sessions to minimize database locking.
             with db_session:
+                my_channel = self.ChannelMetadata.get_my_channel()
                 for payload in batch:
+                    # If we received our metadata payload of torrent we have in our channel, we simply ignore it since
+                    # We always have the latest version ourselves.  This prevents from adding already deleted torrents
+                    # again in My channel.
+                    if self.skip_received_personal_metadata_payload and my_channel \
+                            and payload.public_key == str(my_channel.public_key) \
+                            and payload.metadata_type == REGULAR_TORRENT:
+                        continue
                     result.extend(self.process_payload(payload))
             if external_thread:
                 sleep(self.sleep_on_external_thread)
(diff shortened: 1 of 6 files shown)


ichorid:
Using `contents_len` is super-slow, as it fires a new SQL query on each invocation. If you want to cap the shown number of entries in the channel, you should do it in the GUI.

qstokkink:
While I definitely think we should never accept our own information from third parties, as patched in this PR, I feel there is a deeper issue.

It seems discovered/subscribed channels fail to capture any sort of _state_. It is very difficult for anyone (even ourselves) to determine the contents of a channel at any particular point in time. We should have some method to determine what is supposed to be represented by a channel at a particular point in time, properly taking into account deletions.

I might turn that into a separate issue.

ichorid:
> While I definitely think we should never accept our own information from third parties, as patched in this PR, I feel there is a deeper issue.
> 
> It seems discovered/subscribed channels fail to capture any sort of _state_. It is very difficult for anyone (even ourselves) to determine the contents of a channel at any particular point in time. We should have some method to determine what is supposed to be represented by a channel at a particular point in time, properly taking into account deletions.
> 
> I might turn that into a separate issue.

Yes please, make this into a separate issue.

qstokkink:
Moved it to #4645

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -177,6 +177,11 @@ def sqlite_disable_sync(_, connection):
             # Decay only happens while Tribler is running
             default_vsids.last_bump = datetime.utcnow()
 
+        # Flag to skip processing of personal channel torrent metadata. This is True by default because we assume
+        # the user always has the latest channel version and no need to update his own channel with gossipped data
+        # from the network. We need this flag in testing the processing of metadata.
+        self.skip_received_personal_metadata_payload = True
```
ichorid: When you use the parent object-level flag, you still leave yourself open to the possibility of getting old personal metadata during the moments this flag is true. Unless you make it an argument for mdblob processing functions, the following scenario can happen:
 1. You set `skip_received_personal_metadata_payload`  to `True` to restore your channel from disk;
 2. You call `process_channel_dir`. It starts processing mdblobs and eventually `yield`s to Twisted reactor.
 3. Someone from the network sends you your old metadata. This metadata gets processed by the GigaChannel community and because the `skip_received_personal_metadata_payload ` flag is `True`, this metadata is added to the database.
 4. Channel dir processing is resumed.

As a result, you've successfully restored your channel from the backup but also got some unexpected old entries from the network. T

Please, change the flag to be an argument.

xoriole: My line of thought here was if we're importing the user's channel, it should be synchronous or at least the channel should be locked so no extra payload can be processed.

ichorid: Good point. However, this will require additional effort (special starting mode, stopping IPv8, etc). And the proper usage is totally not obvious to anyone who did not read this thread. It is much safer and convenient to make it an argument (though the code becomes a bit uglier).

PR #4648, opened 2019-07-03
READY: GigaChannel Community gossip caching

### LINKED ISSUE #4636: GigaChannel Community 100% CPU
GigaChannel Community somehow causes up to 100% load on the core and lots of DB i/o. I suspect this is caused by inefficient Pony queries on push-based gossip.

ichorid:
This makes GigaChannel Community cache the results of the SQL query that selects the channel contents to gossip around. The resulting blob is then saved for reuse in the Community object. The blob will be sent to 30 peers, and then the new query will be performed. This significantly reduces the background DB activity.

Fixes #4636

### DIFF
--- a/Tribler/Test/Community/gigachannel/test_community.py
+++ b/Tribler/Test/Community/gigachannel/test_community.py
@@ -12,7 +12,8 @@ from six.moves import xrange
 
 from twisted.internet.defer import inlineCallbacks
 
-from Tribler.Core.Modules.MetadataStore.OrmBindings.channel_node import COMMITTED, NEW
+from Tribler.Core.Modules.MetadataStore.OrmBindings.channel_node import NEW
+from Tribler.Core.Modules.MetadataStore.serialization import REGULAR_TORRENT
 from Tribler.Core.Modules.MetadataStore.store import MetadataStore
 from Tribler.Core.Utilities.random_utils import random_infohash
 from Tribler.Test.Core.base_test import MockObject
@@ -72,9 +73,11 @@ class TestGigaChannelUnits(TestBase):
         """
         Test whether sending a single channel with a multiple torrents to another peer works correctly
         """
+        # We set gossip renewal period to 2 for this test, so the 3rd packet will renew the cache
+        self.nodes[0].overlay.gossip_renewal_period = 2
         with db_session:
             channel = self.nodes[0].overlay.metadata_store.ChannelMetadata.create_channel("test", "bla")
-            for _ in xrange(20):
+            for _ in xrange(10):
                 self.add_random_torrent(self.nodes[0].overlay.metadata_store.TorrentMetadata)
             channel.commit_channel_torrent()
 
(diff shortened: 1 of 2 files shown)


Tribler/Test/Community/gigachannel/test_community.py
```diff
@@ -83,9 +85,31 @@ def test_send_random_multiple_torrents(self):
...
+        self.nodes[0].overlay.send_random_to(Peer(self.nodes[1].my_peer.public_key, self.nodes[1].endpoint.wan_address))
+
+        yield self.deliver_messages(timeout=0.5)
+        with db_session:
+            torrents2 = self.nodes[1].overlay.metadata_store.TorrentMetadata.select()[:]
+            self.assertEqual(len(torrents1), len(torrents2))
+
+        self.nodes[0].overlay.send_random_to(Peer(self.nodes[1].my_peer.public_key, self.nodes[1].endpoint.wan_address))
+
+        # Warning! The following test can fail once in a blue moon because the second batch of gossiped torrents
```
qstokkink: Please, dear god, no racy tests. It's better to have no test, than a test which randomly fails.

ichorid: The blue moon is not going to happen soon. The probability of the test failing with the set of 40 items of which we randomly select 4 is: ` 4! / (40!/36!)`, which is 0,000010942.
If that is not good enough for you, I can make the probability of the fail as high as you like :wink:

devos50: Windows probably ignores this blue moon, it's really better to have uncovered code rather than having tests that might fail, even if the probability appears to be small.

ichorid: > Windows probably ignores this blue moon, it's really better to have uncovered code rather than having tests that might fail, even if the probability appears to be small.

I do not agree. If a test that should fail with some marginally low probability fails more than once in our lifetime, that is a good indication of something fishy going on that should be investigated.

qstokkink: https://hackernoon.com/flaky-tests-a-war-that-never-ends-9aa32fdef359
https://martinfowler.com/articles/nonDeterminism.html

ichorid: @qstokkink , I hate non-determinism as much as you do. However, as a trained physicist, I humbly understand that every thing in life is only determined to a point :wink: 

From the hackernoon article:
> "Usually with these tests a non-deterministic failure is relatively common, so you end up shrugging your shoulders when these tests go red."

A year has 365 days. Let's say we run 100 tests a day. Let's say we continue development for 20 years. If I add 200 torrents, we go to `0,000000015*100*36500*20 = 1` expected fail during these 20 years. For me, that is a fair price for having a working integration test (that can become deterministic at some point, when we learn how to control RNG seeds in Python tests...).

So if you think that the price is too high - I'll remove the test.

ichorid:
> Instead of relying on the randomness, the number of torrents that can be sent can be played with. For example, a test scenario could be.
> 
>     * Say max_entries that can fit in a gossip is 4, renewal period is 3 and Node 0 has 3 (<max_entries) at time t, then
> 
>     * At time t+1, Node 0 gossips random torrents (all 3 in this case) to Node1. Node 1 confirms receival of those 3 torrents.
> 
>     * At time t+2, Node 0 adds more 3 more torrents, totalling 6. With this at least one more random torrent will be there for a new gossip. But since we are caching gossipping data, the next gossip would be a cached one and be equal to the first gossip. Assertion could be for the equality here.
> 
>     * At time t+3, gossip_renewal_period has reached so the new gossip should be sent. In this case, at least one torrent should be different from the previous gossips which could be checked.
>       This should always pass. Just a thought.

Good point! This points me to another solution: just delete the old torrents on the sender and add new ones.

PR #4649, opened 2019-07-03
Disable the broken "revert state" feature in the GUI

### LINKED ISSUE #4643: Reverting entry status in GigaChannel is broken
When channel autocommit is disabled, we show "advanced" editing column to the user. This column includes tri-stat display/switch that shows the current status of an entry:
 * "Committed"
 * "To delete"
 * "New"
 * "Update" (this one has no icon, that is a bug)

The user then can click on this icon to either delete the entry or revert the change. However, this is only correct for the transition between "To delete" and "Committed" states, because the database holds no information about the previous state of an entry. Therefore, the entry can end up in a wrong "Committed" state in case of chains like: "Update" -> "To delete" -> "Committed". Also, updating a "New" entry also leads it to "Updated" state instead of "New" state, which is clearly wrong.

One solution to this problem is to disable the switching functions of the indicator completely.

UPDATE:
on adding torrents to your channel, there is the same broken logic. This should be fixed too.

ichorid:
Fixes #4643


qstokkink:
@xoriole you want to review this before merge?

Issue #4647, opened 2019-07-02
Cannot import channel metadata from file


xoriole:
Trying to import an exported channel from a mdblob file selected by  clicking the + button in Tribler fails with the following error.
```python
ERROR   1562070054.96              rest_manager:83   (RESTRequest)  [Failure instance: Traceback: <class 'ipv8.messaging.serialization.PackError'>: Could not pack item 9: varlenI
error: unpack_from requires a buffer of at least 1869967971 bytes
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/http.py:2121:_finishRequestBody
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/http.py:2196:allContentReceived
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/http.py:920:requestReceived
/home/tribler/tudelft/tribler/Tribler/Core/Modules/restapi/rest_manager.py:118:process
--- <exception caught here> ---
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/server.py:199:process
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/server.py:259:render
/home/tribler/.local/lib/python2.7/site-packages/twisted/web/resource.py:250:render
/home/tribler/tudelft/tribler/Tribler/Core/Modules/restapi/downloads_endpoint.py:347:render_PUT
/home/tribler/tudelft/tribler/Tribler/Core/Modules/MetadataStore/serialization.py:166:from_file
/home/tribler/tudelft/tribler/Tribler/Core/Modules/MetadataStore/serialization.py:135:from_signed_blob
/home/tribler/tudelft/tribler/Tribler/Core/Modules/MetadataStore/serialization.py:140:from_signed_blob_with_offset
/home/tribler/tudelft/tribler/Tribler/pyipv8/ipv8/messaging/serialization.py:328:unpack_multiple
/home/tribler/tudelft/tribler/Tribler/pyipv8/ipv8/messaging/serialization.py:320:unpack_multiple
/home/tribler/tudelft/tribler/Tribler/pyipv8/ipv8/messaging/serialization.py:297:unpack
/home/tribler/tudelft/tribler/Tribler/pyipv8/ipv8/messaging/serialization.py:160:unpack_from
]
Traceback (most recent call last):
  File "/home/tribler/.local/lib/python2.7/site-packages/twisted/web/server.py", line 199, in process
    self.render(resrc)
[...]

PR #4651, opened 2019-07-03
READY: Fixed export/import channel metadata

xoriole:
Fixes https://github.com/Tribler/tribler/issues/4647


xoriole:
I think we can safely ignore Pylint issue and the SonarCloud failure is not related to my PR so I think this PR is ready. @ichorid do you have any more comments?

Fix for SonarCloud can come in a separate PR.

Issue #4645, opened 2019-07-02
Show precise number of torrents in the channel


qstokkink:
Right now we have some issues regarding the determining of the state of a channel.
It seems to be impossible to say which torrents have been removed in the latest state of a channel, given gossip from a previous state.

Concretely, we appear to be unable to do the following:

 - _Upon receiving gossip (1) about channel A_: determine if (1) is part of A at timestamp t

This causes old information to be included in the views of channels and giving results like `300% completion`.

ichorid:
The ability to get excess torrents is a feature, and not a bug. For example, if you already got a channel complete and after that, you receive some new torrent from the newer version of the same channel, you do not drop it, but include it in the same channel. That is how you get 101/100 torrents.

This may be ugly, but this is better than just dropping useful and legit information. The proper way to fix it is to show only the stuff that is covered by the current timestamp range of the channel known to you and hide everything else. In this scenario, receiving any torrent with a timestamp that is higher than the highest timestamp of a channel with the same public key means its author had updated their channel. This means we need to take some *actions* to get the updated channel entry. This should be served by some form of global search such as DHT. Unfortunately, our DHT is still not ready.

So, my design decision was to keep showing excess entries until we get working DHT-based channel updates checks.

qstokkink:
Fair enough from a technical standpoint, but we need to communicate this properly to our users.

> Unfortunately, our DHT is still not ready.

It is available and used in IPv8, you should be able to hook into it.

ichorid:
> It is available and used in IPv8, you should be able to hook into it.
Good! I'll make this an issue for 7.4

qstokkink:
![afbeelding](https://user-images.githubusercontent.com/3630389/63033470-f4640900-beb7-11e9-99cb-2ac9131c4ce8.png)

`(Complete) 8/11` feels a bit weird. Should we maybe just hide the completion rate for the time being?

ichorid:
@qstokkink , IDK, really. It is hard to guarantee that the channel's author will not screw the count up by their actions.

PR #4653, opened 2019-07-03
READY: Check timestamp of entry against the channel ver

### LINKED ISSUE #4639: Discovered channels shows incorrect number of torrents for own channel

ichorid:
Timestamps of incoming GigaChannel payloads are now checked against the parent channel's `local_version`. This allows us to stop getting old deleted results from previous channel versions.
Also, fixed a severe bug in channel commit procedure. When committing the personal channel, mdblob files should only contain entries with timestamps <= to the mdblob filename, but the last mdblob must have the same timestamp as the channel entry itself. Before, this was only true if the channel entry was never edited (e.g. renamed) before commit.

Now we assign filenames based on the timestamp of the last entry in the mdblob, except for the last file. For this, we create a new timestamp and then assign it to _both_ the filename and the updated channel entry.

Fixes #4639 properly.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
@@ -25,7 +25,6 @@ from Tribler.Core.exceptions import DuplicateChannelIdError, DuplicateTorrentFil
 CHANNEL_DIR_NAME_LENGTH = 32  # Its not 40 so it could be distinguished from infohash
 BLOB_EXTENSION = '.mdblob'
 LZ4_END_MARK_SIZE = 4  # in bytes, from original specification. We don't use CRC
-ROOT_CHANNEL_ID = 0
 
 
 def chunks(l, n):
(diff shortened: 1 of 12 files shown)


qstokkink:
Also fixes #4645?

ichorid:
> Also fixes #4645?

In a sense, yes. To guarantee that *300% complete* will never happen, we have to strictly check the influx of gossip into the channel at all stages.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -189,7 +188,7 @@ def consolidate_channel_torrent(self):
 
             return self.commit_channel_torrent(new_start_timestamp=start_timestamp)
 
-        def update_channel_torrent(self, metadata_list):
+        def update_channel_torrent(self, metadata_list, final_timestamp):
```
xoriole: * Please add the argument in the docstring

ichorid: Good point!

ichorid:
retest this please

PR #4655, opened 2019-07-04
READY: Change channel torrent directory names to include id_

ichorid:
This is a necessary change to support nested channels feature in the upcoming 7.4 release.
Unfortunately, this breaks compatibility with the beta channel torrents.

Obviously, had to regenerate the channel sample files used for testing.


xoriole:
retest this please

PR #4664, opened 2019-07-07
READY: Remove redundant dirty channel queries

ichorid:
To determine if user modifications to their channel leave it in the "dirty" state that should be committed to torrent, we scan all the contents of the personal channel. This operation is very slow on big channels and should be avoided. The query is served each time the `.dirty` property of ChannelMetadata is accessed.

Previously, we used to do this scan on each query into MyChannel endpoint, which made scrolling through big personal channels very slow. This PR solves the problem by only querying the `.dirty` property if really necessary, making channel state tracking responsibility of the GUI.

Also, this adds some more indexes to the database to speed up common queries.

### DIFF
--- a/Tribler/Core/Modules/restapi/mychannel_endpoint.py
+++ b/Tribler/Core/Modules/restapi/mychannel_endpoint.py
@@ -165,8 +165,7 @@ class MyChannelTorrentsEndpoint(BaseMyChannelEndpoint):
                 "first": sanitized['first'],
                 "last": sanitized['last'],
                 "sort_by": sanitized['sort_by'],
-                "sort_asc": int(sanitized['sort_asc']),
-                "dirty": my_channel.dirty
+                "sort_asc": int(sanitized['sort_asc'])
             })
 
     def render_POST(self, request):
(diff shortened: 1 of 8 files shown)


Tribler/Core/Modules/restapi/mychannel_endpoint.py
```diff
@@ -424,15 +421,17 @@ def render_PATCH(self, request):
...
-            return json.twisted_dumps({"error": "torrent with the specified infohash could not be found"})
+        with db_session:
+            torrent = my_channel.get_torrent(self.infohash)
+            if not torrent:
+                request.setResponseCode(http.NOT_FOUND)
+                return json.twisted_dumps({"error": "torrent with the specified infohash could not be found"})
 
-        torrent.update_properties(dict([(attribute, parameters[attribute][0]) for attribute in
-                                        ['status', 'tags', 'title'] if attribute in parameters]))
+            status = torrent.update_properties(dict([(attribute, parameters[attribute][0]) for attribute in
```
qstokkink: Instead of `dict([(key, a[key]) for key in something])` you can do `{key: a[key] for key in something}`, which is faster and more beautiful. In this case:

```python
status = torrent.update_properties({
    attribute: parameters[attribute][0]
    for attribute in ['status', 'tags', 'title']
    if attribute in parameters
})
```

Issue #4104, opened 2018-12-09
Can we autoformat the Python code?


cclauss:
PyLint is very picky about import order, line lengths, indentation, etc.

Could we pick a tool(s) like [autopep8](https://github.com/hhatto/autopep8), [yapf](https://github.com/google/yapf), [black](https://github.com/ambv/black), [isort](https://github.com/timothycrosley/isort) that we could run on our PRs before submission so that there is not so much manual rework required to placate PyLint?

My person preference is __black__ because you can't mess with it.
* Any color you want as long as it is black.

qstokkink:
Personally I believe programmers should be in charge of styling code to be readable and tools should tell them when they go too far, instead of the tools dictating the style. I also believe that sometimes you need to ignore "correct" formatting/style.

Both programmers and automatic tools make mistakes though. I have seen autopep code that made my eyes water, but the same goes for humans.

cclauss:
There is Goolge's golang __fmt__ philosophy that says a lot of programmer time and code review attention goes into code formatting.  They have established one format language wide and it is normal operations to put fmt into precommit hooks and continuous integration tools so that programmer attention is focused on what the code does instead of on changing how it looks.  Their yapf was an effort to do the same for Python and black has an even stricter focus.  A lot of programmer time and effort is being sunk into placating linters.

qstokkink:
I understand where you're coming from, but I disagree with the "linter is always right" idea. Basically this: https://dev9.com/blog-posts/2015/1/the-myth-of-developer-productivity .

Humans make the best code for other humans and tools are only there to point where the human might have had a few too many beers during a late night programming session. If there is ever a conflict between tool and human, human should *investigate* the cause and *decide* who is right. Only in the most trivial cases should we adopt what the machine poops out.

ichorid:
I would argue that many previous generations of programmers on Tribler project have never heard about linting... Personally, I don't like to spend my finite human life in this world fighting lint errors, and then fighting other developers on should these lint errors be fixed or not. And as my current job here is refactoring, which is basically *repaying technical debt* of the previous generations, I would definitely like some of this debt to be repaid by a machine. Because that is what programming is about - automating boring tasks.

Therefore, I vote to define some clear (and not *too* strict) PEP8-based rules for the Tribler project and then **run autopep8 on the whole Tribler tree**.

[...]

PR #4720, opened 2019-07-31
READY: Nested channels refactoring

ichorid:
### Motivation
This PR started as an attempt to remove unused widgets from the GUI `.ui` file. Very soon, it became obvious that it is impossible without heavily refactoring all channels-related parts of the GUI. The code for the channel-showing widget was duplicated more than four times both in the `.ui` file and in `.py` files. In the end, small changes proved to be impossible.

**What started as a dead code removal, ended up as an effort to add nested channels support to both the GUI and the Core.**

 ### Main changes

#### GUI technical changes
 * All channels-related GUI stuff is now handled by a single widget `channelcontentswidget`.
 * `channelcontentswidget` features a single View into a stack of channel Models. It always shows the top model.
 * Each different channel-related page (e.g., Search, Discovered, etc.) runs its instance of `channelcontentswidget`. The root of the respective stack is initialized with a page-specific Root model.
 * Communication between the current (top of the stack) model and higher-level `channelcontentswidget` and its member objects are minimized to simplify model-switching and eliminate backreferences. There is only one signal for communicating changes from model to `channelcontentswidget`: the `info_changed` signal. It is reconnected on every change of the models' stack.
 * Non-top models in the stack remember their state (filters, sort order, etc.). They never get reloaded on hitting `<` button. (That make them super fast! :rocket: )
 * Row deletions are now handled in-place for some actions. The same is true for row info updates. This avoids costly channel view reloading :rocket:.

#### GUI feature changes
* :tada: **Full nested channels support!**  :tada: 
* The user can now have several top-level channels, with separate torrent for each channel.
* Sub-channels are called "folders" in the GUI. In the Core the corresponding type is called CollectionNode (as Channels inherit from it). Only true Channels get seeded with libtorrent. Folders are organizational units only, they reside in Channels.
* Ability to move/copy channels. Recursive, as expected.
* New dialog for selecting the target directory for moving/copying stuff into. The dialog supports **creating new channels/folders** :open_file_folder: :scissors: , lazy loading of the structure of the channels and tracks the updates to your channels' structure automatically :eyes: via the `info_changed` signal. 
* **Multiple selection** works as expected, with Ctrl-LMB and Shift-LMB!
* :file_cabinet: **Category filter drop-down menu** :file_cabinet:  is added to the channels view. It allows the user to filter the current channel on either torrent categories or channels/folders. This neatly unifies the Search page functionality with other pages.
* Torrent checking is now initiated only after 500ms after selection change. This makes the GUI much and the Core much more responsive during channels scrolling, as it avoids unwanted health checks.

#### Core changes
* Channels-related endpoints are refactored: everything is now (almost) completely compliant with the REST paradigm, entries are correctly identified by a combination of the public key and id instead of infohash.
* All REST data exchange with channels-related endpoints is going in JSON format now.
* Recursive commits / copying / moving / deletion algorithms are added for ORM-managed objects.
* `mychannel` endpoint is removed. Instead, `mychannel` is now just an alias for the user's public key.
* Ugly `count` endpoint is removed.
* Almost everything is now served by the single `metadata` endpoint that RESTfully PATCHes and POSTs objects. It is amazing how powerful this abstraction is! (However, this means some logic had to be moved to the GUI).

### Known problems / TODO:
 * Credit mining does not work for collections, because credit mining sources are identified by infohash, and collections got none. Instead, CreditMining Manager should identify sources by pk+id pair. GUI test for credit mining is disabled until this is fixed.
 * Changing root channel name is not supported, as there is no root channel now.
 * Changing channel name by clicking on it is not supported.
 * No button for copying the channel public key.
 * Searching by channel public key feature is deleted. It should be re-implemented properly as a generic RPC-like request to remote DB. (Searching for infohash is still there, as it can be useful to the user).
 * Infohash uniqueness requirement is removed, but the actual database format did not change. True database conversion is not required, but some indexes should be deleted, and some should be created. We don't have the code for this kind of upgrade yet.
 * Some minor signals/connections in the GigaChannels GUI might be missing.
 * Algorithms for recursive commits/deletion/copying are probably far from perfect. I've tried my best to make them work as fast as I could, but there is only so much one can do in a limited time. They are all tested, though.
 * Some buttons and their placement are UGLY (as always). Some can be in the wrong place.
 * View/controller stuff is all over the namesake objects and `channelcontentswidget`. There should be a better way to organize it, but even in its current state, it is much more idiomatic and clear than it was before.
 * Some form of garbage collection for ORM is required, as well as smarter sanity checks.
 * Overall, the code could be better in many places.


ichorid:
retest this please

ichorid:
Note that C0330 check is wrong and should be disabled. See https://github.com/Tribler/tribler/issues/4104#issuecomment-520938141

ichorid:
retest this please

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -1,255 +1,151 @@
...
+            entry.delete()
+            result = {"public_key": hexlify(public_key), "id": id_, "state": "Deleted"}
+            results_list.append(result)
+        return json.twisted_dumps(results_list)
 
-class ChannelPublicKeyEndpoint(BaseChannelsEndpoint):
 
+# /<public_key>
+#              /<id_>
+# Intermediate endpoint for parsing public_key part of the request
```
devos50: Why not place these comments in a docstring under the class definition?

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
```diff
@@ -0,0 +1,68 @@
...
+def convert_sort_param_to_pony_col(sort_param):
+    """
+    Convert an incoming sort parameter to a pony column in the database.
+    :return a string with the right column. None if there exists no value for the given key.
+    """
+
+    return json2pony_columns[sort_param] if sort_param in json2pony_columns else None
+
+
+class MetadataEndpointBase(resource.Resource):
```
devos50: Could you rename this to `BaseMetadataEndpoint` ?

Tribler/Core/Utilities/bootstrap_util.py
```diff
@@ -1,5 +1,9 @@
 from __future__ import absolute_import
 
+import random
```
devos50: Why are these imports added here? The `random` module is not used (anymore?) in this file. Maybe there's a rebase issue?

Tribler/Test/Core/Modules/MetadataStore/test_channel_download.py
```diff
@@ -59,6 +59,6 @@ def test_channel_update_and_download(self):
         with db_session:
             # There should be 4 torrents + 1 channel torrent
             channel2 = self.session.lm.mds.ChannelMetadata.get(public_key=database_blob(payload.public_key))
-            self.assertEqual(5, len(list(self.session.lm.mds.TorrentMetadata.select())))
-            self.assertEqual(1562257279010, channel2.timestamp)
+            self.assertEqual(8, self.session.lm.mds.ChannelNode.select().count())
```
devos50: Why are there 8 `ChannelNode` objects in this test? This is not consistent with the comment on line 60.

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
```diff
@@ -285,6 +266,82 @@ def test_commit_channel_torrent(self):
...
+    def test_recursive_commit_channel_torrent(self):
+        status_types = [NEW, UPDATED, TODELETE, COMMITTED]
+
+        def all_status_combinations():
+            result = []
+            for card in range(0, len(status_types) + 1):
+                result.extend(list(combinations(status_types, card)))
+            return result
+
+        def gen_coll(parent, collection_status, contents_statuses, recurse=False):
```
devos50: Can you just name this `generate_collection`?

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
```diff
@@ -285,6 +266,82 @@ def test_commit_channel_torrent(self):
...
+                title=parent.title + '->child_new_nonempty', origin_id=parent.id_, status=collection_status
+            )
+            for s in contents_statuses:
+                self.mds.TorrentMetadata(infohash=random_infohash(), origin_id=chan.id_, status=s)
+                if recurse:
+                    for status in status_types:
+                        gen_coll(chan, status, [NEW])
+            return chan
+
+        def gen_chan(recurse=False, status=NEW):
```
devos50: Can you just name this `generate_channel`?

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
```diff
@@ -374,3 +442,55 @@ def test_add_torrents_from_dir(self):
 
     def test_add_torrents_from_dir_recursive(self):
         self.check_add(11, 1, recursive=True)
+
+    @db_session
+    def test_make_copy(self):
```
devos50: I always add a docstring above every test to quickly explain what it is testing (and sometimes what I expect). This docstring is displayed when running the tests. This might be a matter of personal preference but I recommend annotating these tests with such comments. The reason is that the (limited length of the) method name might not always reflect the right intentions of the test.

Tribler/Test/Core/Modules/RestApi/test_channels_endpoint.py
```diff
@@ -0,0 +1,518 @@
...
+                )
+            for ind in xrange(5, 9):
+                _ = self.session.lm.mds.TorrentMetadata(
+                    origin_id=chan2.id_, title='torrentB%d' % ind, infohash=random_infohash()
+                )
+            return chan
+
+
+class TestChannelsEndpoint(BaseTestMetadataEndpoint):
+    def test_get_channels(self):
```
devos50: Newline?

Tribler/Test/Core/Modules/RestApi/test_debug_endpoint.py
```diff
@@ -31,19 +30,18 @@ def verify_response(response):
             response_json = json.twisted_loads(response)
             self.assertEqual(len(response_json["slots"]["random"]), 4)
 
-        self.should_check_equality = False
```
devos50: Why were these removed?

Tribler/Test/Core/Modules/RestApi/test_libtorrent_endpoint.py
```diff
@@ -4,7 +4,6 @@
 
 
 class TestLibTorrentSettingsEndpoint(AbstractApiTest):
-
```
devos50: Related to the formatter probably but I would prefer a newline between a class definition and the definition of its first method.

xoriole:
retest this please

Tribler/Test/Core/Modules/RestApi/test_torrentinfo_endpoint.py
```diff
@@ -52,10 +51,9 @@ def verify_valid_dict(data):
             # To fix it, we must switch to some encoding scheme that is able to encode and decode raw binary
             # fields in the dicts.
             # However, for this works fine at the moment because we never use pieces data in the GUI.
-            #self.assertTrue(TorrentDef.load_from_dict(metainfo_dict))
+            # self.assertTrue(TorrentDef.load_from_dict(metainfo_dict))
```
devos50: Commented-out code, remove or uncomment.

Tribler/Test/GUI/test_gui.py
```diff
@@ -278,112 +281,63 @@ def test_home_page_channels(self):
...
+        index = self.get_index_of_row(widget.content_table, 0)
+        widget.content_table.on_table_item_clicked(index)
+        self.wait_for_list_populated(widget.content_table)
+        self.screenshot(window, name="%s-channel_loaded" % widget_name)
 
+        # FIXME: credit mining for collections!
         # Toggle credit mining
-        QTest.mouseClick(window.credit_mining_button, Qt.LeftButton)
-        self.wait_for_signal(window.subscription_widget.credit_mining_toggled)
+        # QTest.mouseClick(widget.credit_mining_button, Qt.LeftButton)
```
devos50: Commented-out code, at this point I think you know what to do with it :)

Tribler/community/gigachannel/community.py
```diff
@@ -186,18 +213,18 @@ def on_search_request(self, peer, request):
...
-            "torrent": REGULAR_TORRENT
+            "torrent": REGULAR_TORRENT,
         }.get(request.metadata_type.decode('utf8'), REGULAR_TORRENT)
 
         request_dict = {
             "first": 1,
             "last": max_entries,
             "sort_by": request.sort_by.decode('utf8'),
-            "sort_asc": request.sort_asc,
+            "sort_desc": request.sort_desc,
```
devos50: What's the motivation behind changing the `sort_asc` to `sort_desc`?

TriblerGUI/defs.py
```diff
@@ -115,7 +171,7 @@
 
 # This list of content categories is used in drop-down menu when editing a personal channel
 # TODO: build this automatically and/or move it somewhere
-CATEGORY_LIST = [
+CATEGORY_LIST = (
```
devos50: Why is this a tuple now?

TriblerGUI/dialogs/addtopersonalchanneldialog.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, parent):
+        DialogContainer.__init__(self, parent)
+        uic.loadUi(get_ui_file_path('addtochanneldialog.ui'), self.dialog_widget)
+        self.dialog_widget.btn_cancel.clicked.connect(self.close_dialog)
+        self.dialog_widget.btn_add.clicked.connect(self.on_add_clicked)
+        self.dialog_widget.btn_move.clicked.connect(self.on_move_clicked)
+        self.dialog_widget.btn_new_channel.clicked.connect(self.on_create_new_channel_clicked)
+
+        # self.dialog_widget.setSizePolicy(QSizePolicy.Fixed, QSizePolicy.Expanding)
```
devos50: Commented-out code.

ichorid:
retest this please

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -100,18 +99,17 @@ class ChannelMetadata(db.TorrentMetadata):
...
+        ][1:]
 
-        @db_session
-        def update_metadata(self, update_dict=None):
-            channel_dict = self.to_dict()
-            channel_dict.update(update_dict or {})
-            channel_dict.update({'num_entries': self.contents_len})
-            self.set(**channel_dict)
-            self.sign()
+        # As channel metadata depends on public key, we can't include the infohash in nonpersonal_attributes
```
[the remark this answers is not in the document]
ichorid: Sorry for my Runglish ))

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -100,18 +99,17 @@ class ChannelMetadata(db.TorrentMetadata):
 
         votes_scaling = 1.0
 
+        # Special class-level properties
         _payload_class = ChannelMetadataPayload
         _channels_dir = None
         _category_filter = None
         _CHUNK_SIZE_LIMIT = 1 * 1024 * 1024  # We use 1MB chunks as a workaround for Python's lack of string pointers
+        payload_arguments = _payload_class.__init__.__code__.co_varnames[
```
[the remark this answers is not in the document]
ichorid: Added comment in the base class.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -30,7 +31,7 @@
 def chunks(l, n):
     """Yield successive n-sized chunks from l."""
     for i in range(0, len(l), n):
-        yield l[i:i + n]
+        yield l[i : i + n]
```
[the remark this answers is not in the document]
ichorid: It would be very hard for me to comply with our current Flake8 checks (especially in regards to imports) without using code autoformatting. Sorry about that.
If we had introduced code autoformatting along with introducing the checks, this problem would have never arisen.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -100,18 +99,17 @@ class ChannelMetadata(db.TorrentMetadata):
...
 
-        @db_session
-        def update_metadata(self, update_dict=None):
-            channel_dict = self.to_dict()
-            channel_dict.update(update_dict or {})
-            channel_dict.update({'num_entries': self.contents_len})
-            self.set(**channel_dict)
-            self.sign()
+        # As channel metadata depends on public key, we can't include the infohash in nonpersonal_attributes
+        nonpersonal_attributes = set(db.CollectionNode.nonpersonal_attributes)
```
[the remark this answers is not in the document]
ichorid: A non-personal attribute of an entry is an attribute that would have the same value regardless of where, when and by whom it was introduced for the entry. In other words, it does not depend on the *person* who created it. I agree, the name can be a bit misleading.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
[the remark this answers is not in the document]
devos50: Yeah we discussed this already yesterday. Although I must say that most of these cosmetic changes indeed improve readability 👍

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
```diff
@@ -7,10 +7,14 @@
 from ipv8.keyvault.crypto import default_eccrypto
 
 from pony import orm
-from pony.orm.core import DEFAULT
+from pony.orm.core import DEFAULT, db_session
 
 from Tribler.Core.Modules.MetadataStore.serialization import (
-    CHANNEL_NODE, ChannelNodePayload, DELETED, DeletedMetadataPayload)
+    CHANNEL_NODE,
```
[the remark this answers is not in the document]
ichorid: Actually, this style results in smaller and more readable diffs, as it makes changes to imports atomic. It is fixed in `black`.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_node.py
```diff
@@ -214,4 +231,50 @@ def from_payload(cls, payload):
...
+
+            if signed_attribute_changed:
+                if self.status != NEW:
+                    self.status = UPDATED
+                self.timestamp = self._clock.tick()
+                self.sign()
+
+            return self
+
+        def get_parents_ids(self, recursion_depth=15):
```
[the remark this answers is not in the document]
ichorid: The name can be a bit misleading, I agree. However, in that case, it would require creating another argument to represent the maximum recursion depth. I'll rename this one to `max_recursion_depth`

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+        _payload_class = CollectionNodePayload
+        payload_arguments = _payload_class.__init__.__code__.co_varnames[
+            : _payload_class.__init__.func_code.co_argcount
+        ][1:]
+        nonpersonal_attributes = db.MetadataNode.nonpersonal_attributes + ('num_entries',)
+
+        @property
+        @db_session
+        def state(self):
+            # TODO: optimize this by stopping doing blob comparisons on each call, and instead remember rowid?
```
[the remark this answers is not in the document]
ichorid: it is performed by `self.personal`. I'll copy the comment there and modify this one.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+                {
+                    "torrents": self.num_entries,
+                    "state": self.state,
+                    "total": self.contents_len,
+                    "dirty": self.dirty if self.is_personal else False,
+                }
+            )
+            return result
+
+        def make_copy(self, tgt_parent_id, recursion_depth=15, **kwargs):
```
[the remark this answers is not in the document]
ichorid: No, we don't. We don't even have a GUI facility to display this kind of information. To show it, we need a blocking dialog window with a progress bar *a la* TotalCommander. Estimating the time to copy can take as much time as the process itself. However, a good estimation can be taken from the channel size itself, since it includes a recursive count of contents.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+                    if issubclass(type(node), CollectionNode):
+                        node.make_copy(new_node.id_, recursion_depth=recursion_depth - 1)
+                    else:
+                        node.make_copy(new_node.id_)
+            return new_node
+
+        @db_session
+        def copy_torrent_from_infohash(self, infohash):
+            """
+            Create a new signed copy of the given torrent metadata
```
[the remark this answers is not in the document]
ichorid: Ok

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+                    torrent_metadata = old_torrent
+                else:
+                    raise DuplicateTorrentFileError()
+            else:
+                torrent_metadata = db.TorrentMetadata.from_dict(dict(origin_id=self.id_, **new_entry_dict))
+            return torrent_metadata
+
+        @db_session
+        def pprint_tree(self, file=None, _prefix="", _last=True):
+            print(_prefix, "`- " if _last else "|- ", (self.num_entries, self.metadata_type), sep="", file=file)
```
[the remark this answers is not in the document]
ichorid: It is super useful for debugging purposes. To cover it, I use it once in the testing suite.

What's our policy on debugging methods? If we are to remove them from the codebase, were/how do we keep them?

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+            results_stack = []
+            for subnode in self.contents:
+                if issubclass(type(subnode), CollectionNode):
+                    results_stack.extend(subnode.get_contents_recursive())
+                results_stack.append(subnode)
+            return results_stack
+
+        @db_session
+        def add_torrents_from_dir(self, torrents_dir, recursive=False):
+            # TODO: Optimize this properly!!!!
```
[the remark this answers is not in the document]
ichorid: Reasonable

Tribler/Core/Modules/MetadataStore/serialization.py
```diff
@@ -166,6 +168,7 @@ def from_file(cls, filepath):
             return cls.from_signed_blob(f.read())
 
 
+# fmt: off
```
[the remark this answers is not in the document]
ichorid: It is a black statement, indeed. It turns off black formatting. It is needed here to keep the nice hand-formatting for serializer arguments.

Tribler/Core/Modules/MetadataStore/serialization.py
```diff
@@ -195,17 +198,112 @@ def from_unpack_list(cls, metadata_type, reserved_flags, public_key,
...
-        })
+        dct.update(
+            {"id_": self.id_,
+             "origin_id": self.origin_id,
+             "timestamp": self.timestamp
+             })
         return dct
 
+
+class MetadataNodePayload(ChannelNodePayload):
```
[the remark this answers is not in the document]
ichorid: Reasonable

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -13,15 +13,31 @@
 from pony.orm import db_session
 
 from Tribler.Core.Modules.MetadataStore.OrmBindings import (
-    channel_metadata, channel_node, channel_peer, channel_vote, misc, torrent_metadata, torrent_state, tracker_state,
-    vsids)
+    channel_metadata,
```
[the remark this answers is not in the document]
ichorid: Again, it makes it:
 * more readable;
 * produce smaller diffs;
 * less prone to merge conflicts.

Tribler/Core/Modules/MetadataStore/serialization.py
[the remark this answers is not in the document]
ichorid: Hmm... At first, I thought you said that about the ORM classes. I'll provide comment on these.
However, do you really think providing comments that just copy-paste the names of the classes make the code more readable, or more maintainable? Really?

```python
class SignedPayload(Payload):
    """
    Payload for metadata.
    """

    format_list = ['H', 'H', '64s']

class ChannelNodePayload(SignedPayolad):
    """
    Payload for ChannelNode class entries.
    """

    format_list = ['H', 'H', '64s']
```

Really? 

Just say "Yes", and I'll gladly add them :wink:.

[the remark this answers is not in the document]
devos50: Ok, for payloads it's not really necessary I would say but I noticed that there was a comment for another (payload) class, so that's why I asked.

Tribler/Core/Modules/gigachannel_manager.py
```diff
@@ -43,34 +44,42 @@ def start(self):
         or subscribed channels require updating.
         """
 
+        # TODO: break this into smaller checks/functions
+        # TODO: account for all kinds of troubles: lost/wrong .mdblob/torrent, etc
```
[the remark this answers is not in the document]
ichorid: Have you noticed question marks on some of these TODOs? That is because it is still not clear if these are important.
I would rather not add issues for optimizing the code that is still not merged.

When we merge it, we can grep the codebase for TODOs and create a separate issue for each one, if it is worth it.

Tribler/Core/Modules/gigachannel_manager.py
```diff
@@ -43,34 +44,42 @@ def start(self):
...
-                    if not(tdef and tdef.infohash == bytes(my_channel.infohash)):
-                        regenerated = my_channel.consolidate_channel_torrent()
-                        # If the user created their channel, but added no torrents to it, the channel torrent will not
-                        # be created.
-                        if regenerated:
-                            tdef = TorrentDef.load_from_dict(regenerated)
-                    if tdef:
-                        self.updated_my_channel(tdef)
+                for my_channel in self.session.lm.mds.ChannelMetadata.get_my_channels():
+                    if (
```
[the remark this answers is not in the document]
ichorid: Indeed, it improves readability. On each line, you have a separate condition. The lines are started with condition arguments, meaning that *you can change conditions independent of each other*. Also, this means that a similar (e.g. copy-pasted) condition statement could be immediately identified by naked eye.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,412 @@
...
+
+        @staticmethod
+        @db_session
+        def get_children_dict_to_commit():
+            db.CollectionNode.collapse_deleted_subtrees()
+            upd_dict = {}
+            children = {}
+            # TODO: optimize me by rewriting in pure SQL with recursive CTEs
+
+            def upd_node_info(n):
```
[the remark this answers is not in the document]
ichorid: OK

Tribler/Core/Modules/MetadataStore/serialization.py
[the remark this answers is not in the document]
ichorid: It was you, who left that comment originally...

Tribler/Core/Modules/gigachannel_manager.py
```diff
@@ -258,9 +283,36 @@ def updated_my_channel(self, tdef):
...
-        if my_channel and my_channel.status == COMMITTED and not self.session.has_download(my_channel.infohash):
+            my_channel = self.session.lm.mds.ChannelMetadata.get(infohash=database_blob(tdef.get_infohash()))
+        if my_channel and my_channel.status == COMMITTED and not self.session.has_download(bytes(my_channel.infohash)):
             dcfg = DownloadConfig(state_dir=self.session.config.get_state_dir())
             dcfg.set_dest_dir(self.session.lm.mds.channels_dir)
             dcfg.set_channel_download(True)
             self.session.lm.add(tdef, dcfg)
+
+    @db_session
+    def clean_unsubscribed_channels(self):
```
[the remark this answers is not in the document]
ichorid: REST request to PATCH a channel entry to set `subscribed` to `False` does just that: changes the entry attribute in the database. It is GigaChannel manager that deletes the contents afterward. It is done so for three reasons:
 1. It is more RESTful.
 2. It guarantees that there will be no race conditions between updating the channel and removing its contents, especially for trigger-happy users.
 3. It helps with the separation of concerns in the architecture.

Tribler/Core/Modules/gigachannel_manager.py
```diff
@@ -258,9 +283,36 @@ def updated_my_channel(self, tdef):
...
+        for channel in unsubscribed_list:
+            self.channels_processing_queue[channel.infohash] = (
+                CLEANUP_UNSUBSCRIBED_CHANNEL,
+                (channel.public_key, channel.id_),
+            )
+
+    def cleanup_channel(self, to_cleanup):
+        self.processing = True
+        public_key, id_ = to_cleanup
+        # Maybe run it threaded?
```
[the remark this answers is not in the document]
ichorid: Right

Tribler/Core/Modules/gigachannel_manager.py
```diff
@@ -258,9 +283,36 @@ def updated_my_channel(self, tdef):
...
+        try:
+            with db_session:
+                channel = self.session.lm.mds.ChannelMetadata.get_for_update(public_key=public_key, id_=id_)
+                if not channel:
+                    return
+                channel.local_version = 0
+                channel.contents.delete(bulk=True)
+        finally:
```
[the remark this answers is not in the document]
ichorid: Reasonable, added warning.

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -1,255 +1,151 @@
...
+                return http.BAD_REQUEST, {"error": "Changing signed parameters in non-personal entries not supported."}
 
-        child_handler_dict = {
-            b"channels": ChannelsEndpoint,
-            b"torrents": TorrentsEndpoint
-        }
-
-        for path, child_cls in child_handler_dict.items():
-            self.putChild(path, child_cls(session))
+        # if "subscribed" in update_dict:
```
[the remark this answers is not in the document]
ichorid: Done

Tribler/Core/Modules/restapi/metadata_endpoint.py
```diff
@@ -1,255 +1,151 @@
...
-                return json.twisted_dumps({"error": "the limit parameter must be a positive number"})
-
-        with db_session:
-            popular_channels = self.session.lm.mds.ChannelMetadata.get_random_channels(limit=limit_channels)
-            results = [channel.to_simple_dict() for channel in popular_channels]
-        return json.twisted_dumps({"channels": results})
-
+                return json.twisted_dumps({"error": "Entry %i not found" % id_})
+            entry.delete()
+            result = {"public_key": hexlify(public_key), "id": id_, "state": "Deleted"}
```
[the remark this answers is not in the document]
ichorid: I'll leave it as it is for now. When we move to Swagger-based testing, we can redo it properly.

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
ichorid: I feel it is better to leave `Base` at the end to reflect the fact that this is *not* an endpoint for some object called `BaseMetadata`.

Tribler/Core/Utilities/bootstrap_util.py
ichorid: Probably so. I haven't run the pre-commit hooks after the rebase.
Don't worry about this small discrepancies. When we'll run the hooks on the whole repository, that will cleanup everything.

Tribler/Test/Core/Modules/MetadataStore/test_channel_download.py
ichorid: I've added some nested channels to the test .mdblobs and forgot to update the comment.

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
ichorid: Ok

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
ichorid: Ok

Tribler/Test/Core/Modules/MetadataStore/test_channel_metadata.py
ichorid: Ok. I'll add the strings to the tests with short names

Tribler/Test/Core/Modules/RestApi/test_debug_endpoint.py
ichorid: Because these are unnecessary. If `should_check_equality` is `True`, the result of the request will be checked against `expected_json` argument, which is None by default. I can't imagine a situation where the code would expect the returned JSON value to really be `None`. On the other hand, I can't imagine a situation in which one would set `expected_json` to something other than `None`, and will not set `should_check_equality`, either.

Therefore, it is perfectly fine to consider that `should_check_equality` is `False` in all cases where `expected_json` is `None`, and it is True if `expected_json` is not `None`.
And that is fewer code lines and less mental burden for the programmer.

Tribler/Test/Core/Modules/RestApi/test_libtorrent_endpoint.py
ichorid: Sorry, but `black` thinks otherwise...

Tribler/Test/Core/Modules/RestApi/test_torrentinfo_endpoint.py
ichorid: Have you read the FIXME above?

Tribler/Test/GUI/test_gui.py
ichorid: Again, that's part of the FIXME.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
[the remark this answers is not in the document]
devos50: We should decide on a policy for these methods. I personally don't like dead and untested code to float around. I vote to remove it if it is not used; the code base is already large enough. But we should also ask the other devs about this.

Tribler/community/gigachannel/community.py
ichorid: In SQL, by default stuff is sorted ascending. The same is true for the GUI. There is a modifier in ORM that makes the sorting key sort descending, like this: `desc(sort_expression)`. This change aligns the REST data format with the database commands. It simplifies the logic in between, by removing the unneccesary conversions.

TriblerGUI/defs.py
ichorid: Because tuples are immutable, which is safer and faster. Should have been like this from the beginning.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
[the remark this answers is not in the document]
ichorid: This is a tool that I, as a developer, use daily when dealing with nested channels. What's dead about that?

Tribler/Core/Modules/restapi/metadata_endpoint.py
ichorid: Ok

Tribler/Test/Core/Modules/RestApi/test_channels_endpoint.py
ichorid: `black` says `no` :wink:

Tribler/Core/Modules/MetadataStore/serialization.py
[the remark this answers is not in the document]
devos50: I wasn't referring to the Github comment. I saw that there was another docstring so I wondered why there was no docstring under the class for which I made the initial Github comment.

Tribler/Core/Modules/gigachannel_manager.py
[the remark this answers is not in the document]
devos50: > Have you noticed question marks on some of these TODOs? That is because it is still not clear if these are important.

I think we should decide whether they are important/viable or not before merging the PR. We already have many TODOs in the codebase and if we do not properly discuss them or be aware of them before they are merged, they will clutter the code base.

Tribler/Core/Modules/MetadataStore/serialization.py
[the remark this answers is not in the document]
ichorid: I removed you original comments for all the payloads but first, because I find those not to add any new information for the reader, but instead obscuring the code structure. For me, it looks like this:
sign

Tribler/Test/Core/Modules/RestApi/test_torrentinfo_endpoint.py
devos50: Oh sorry, I couldn't see the FIXME in the diff, nvm.

Tribler/community/gigachannel/community.py
devos50: Ok.

TriblerGUI/defs.py
devos50: Agreed, we should change the other definitions to tuples also (at least for most of the lists in `defs.py`). Let's not make these changes in this PR.

Tribler/Test/GUI/test_gui.py
devos50: Well, I think credit mining is something that should be fixed before this PR is merged. Maybe you can discuss this with @xoriole ?

ichorid: Yep, that's a good idea.

synctext:
This is quite an [epic](https://www.atlassian.com/agile/project-management/epics) features list, of multiple sprints (LoC +6,908 −8,128). Exciting list of core features and good starting point for documentation in this PR. Solid progress for Tribler. Reading up on my "agile process" understanding. Does the standard GIT workflow now call for credit mining repair to be a separate PR, PR-to-PR, or something?

Another long-term thought, also not suitable to be buried in a PR: this GUI is becoming tightly coupled to our core. This co-evolution has implication for our desire for Tribler to become a library and leaving "awesome&pretty GUI" work to other specialists. The GUI might be inherently complex and coupled. This PR clearly demonstrates this. As scientists we desire to shield ourselves for daily usage&bugs by making a library and only dealing with requests from other (GUI) Devs. We can then focus on our core strength: delivering world-first academically-pure decentral algorithms.

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
qstokkink: It would be a good idea to inherit from the IPv8 base endpoint; https://github.com/Tribler/py-ipv8/blob/master/ipv8/REST/base_endpoint.py

Tribler/community/gigachannel/payload.py
```diff
@@ -5,7 +5,7 @@
 
 class SearchRequestPayload(VariablePayload):
     format_list = ['I', 'varlenH', 'varlenH', 'varlenH', '?', '?']
-    names = ['id', 'query_filter', 'metadata_type', 'sort_by', 'sort_asc', 'hide_xxx']
```
qstokkink: MEMO: Check if this logic is reflected in the rest of the code.

TriblerGUI/widgets/homepage.py
```diff
@@ -108,4 +107,4 @@ def on_home_page_item_clicked(self, row, col):
             channel_info = cell_widget.channel_info
             self.window().channel_page.initialize_with_channel(channel_info)
             self.window().navigation_stack.append(self.window().stackedWidget.currentIndex())
-            self.window().stackedWidget.setCurrentIndex(PAGE_CHANNEL_DETAILS)
```
qstokkink: Why is this commented out? Should have a TODO or a FIXME.

TriblerGUI/widgets/lazytableview.py
```diff
@@ -1,46 +1,49 @@
...
+        # Mix-in connects
+        self.clicked.connect(self.on_table_item_clicked)
+        self.delegate.play_button.clicked.connect(self.on_play_button_clicked)
+        self.delegate.subscribe_control.clicked.connect(self.on_subscribe_control_clicked)
+        self.delegate.rating_control.clicked.connect(self.on_subscribe_control_clicked)
+        self.delegate.download_button.clicked.connect(self.on_download_button_clicked)
+        # TODO: status changing feature should remain turned off until we fix the undo mess
+        self.delegate.delete_button.clicked.connect(self.on_delete_button_clicked)
+
+    # def mousePressEvent(self, event):
```
qstokkink: This should be removed.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -455,17 +401,18 @@ def channel_state(self):
...
-                "subscribed": self.subscribed,
-                "votes": self.votes/db.ChannelMetadata.votes_scaling,
-                "state": self.channel_state
-            })
+            result = super(ChannelMetadata, self).to_simple_dict(**kwargs)
+            result.update(
+                {
+                    "state": self.state,
+                    "subscribed": self.subscribed,
+                    "votes": self.votes / db.ChannelMetadata.votes_scaling,
```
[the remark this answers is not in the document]
ichorid: No, it really can't. There is only one case when this could happen: if the user modifies their SQLite DB with some external tool directly to set it to 0.

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
ichorid: I don't see anything particularly appealing in that base endpoint now. In a few months, we're moving to Quart-based endpoints anyway.

TriblerGUI/widgets/lazytableview.py
ichorid: Done

TriblerGUI/widgets/homepage.py
ichorid: Fixed

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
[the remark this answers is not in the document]
qstokkink: Thanks, that saves me checking up 👍

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
qstokkink: If we move to Quart, its more convenient to port one base class than to port two base classes. I'm not a fan of duplicate code.

TriblerGUI/widgets/channelcontentswidget.py
```diff
@@ -0,0 +1,461 @@
...
+        else:
+            self.new_channel_button.setHidden(True)
+            self.channel_state_label.setHidden(False)
+            if "state" in self.model.channel_info:
+                self.channel_state_label.setText(self.model.channel_info["state"])
+                self.subscription_widget.update_subscribe_button(self.model.channel_info)
+
+        # self.total_items = response['total']
+        # if self.num_results_label:
+        #    self.num_results_label.setText("%d results" % self.total_items)
```
xoriole: Does not seem to be present in the .ui file (if I'm not wrong) so might as well remove it.

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
```diff
@@ -0,0 +1,416 @@
...
+            :param commit_queue:
+            :return:
+            """
+            for node in commit_queue:
+                # Avoid updating entries that must be deleted:
+                # soft delete payloads require signatures of unmodified entries
+                if issubclass(type(node), db.CollectionNode) and node.status != TODELETE:
+                    # Update recursive count of actual non-collection contents
+                    node.num_entries = select(
+                        (g.num_entries if g.metadata_type == COLLECTION_NODE else 1) for g in node.actual_contents
```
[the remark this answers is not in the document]
ichorid: `1` just means `1` here. It is not a constant, it just means that that there is only one sub-object to count. Of course, this could be changed to something like:
```python
DEFAULT_NUMBER_OF_OBJECTS_IN_NON_COLLECTION_NODE = 1
```
But I don't think it will make it more readable. I will just add a comment on this instead.

Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
```diff
@@ -164,11 +190,20 @@ def consolidate_channel_torrent(self):
...
+
+            def update_timestamps_recursive(node):
+                if issubclass(type(node), db.CollectionNode):
+                    for child in node.contents:
+                        update_timestamps_recursive(child)
+                if node.status in [COMMITTED, UPDATED, NEW]:
+                    node.status = UPDATED
+                    node.timestamp = self._clock.tick()
+                    node.sign()
+                    stuff.append(node)
```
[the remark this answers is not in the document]
ichorid: Oops, sorry for that :wink:

Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -233,33 +259,42 @@ def process_channel_dir(self, dirname, public_key, id_, skip_personal_metadata_p
...
-                               dirname, hexlify(bytes(channel.public_key)), channel.local_version,
-                               channel.timestamp)
+            self._logger.debug(
+                "Finished processing channel dir %s. Channel %s local/max version %i/%i",
+                dirname,
+                hexlify(bytes(channel.public_key)),
+                channel.local_version,
+                channel.timestamp,
+            )
+        return result
```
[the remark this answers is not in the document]
ichorid: Actually, the returned list is only used for debugging purposes. It is better to remove it to reduce the memory load on processing bigger channels.

Tribler/Test/Core/Socks5/test_conversion.py
```diff
@@ -2,12 +2,12 @@
 
 import struct
 
+from Tribler.Core.Socks5.conversion import decode_request
```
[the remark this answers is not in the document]
ichorid: Fixed.

TriblerGUI/widgets/channelcontentswidget.py
ichorid: Fixed

Tribler/Core/Modules/MetadataStore/OrmBindings/collection_node.py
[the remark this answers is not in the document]
xoriole: Right

Tribler/Core/Modules/restapi/metadata_endpoint_base.py
ichorid: What kind of code the IPv8 endpoint add anyway? Is it really necessary?

ichorid:
Closed to be continued as a Collective Effort PR  #4759

PR #4727, opened 2019-08-01
Fix personal channel status not changing on creation/edit/commit

ichorid:
At the personal channel entry creation its status should be NEW and if the title is edited it should become UPDATED. However, the channel status was always COMMITTED. We never noticed it because there was always only a single personal channel entry. On refactoring for nested channels this problem became apparent.


qstokkink:
Should the default setting for `status` on the `ChannelNode` class also be `NEW`, or does that not matter?

ichorid:
> Should the default setting for `status` on the `ChannelNode` class also be `NEW`, or does that not matter?

I was pondering hard on this for some time. Maybe it should. But then every instance incoming from the network should be created with `status=COMMITTED`. I guess it depends on the frequency of calls of `ChannelNode` related to externally added vs locally created. When I was designing this, I opted for default `COMMITED`, because there would be much more code lines related to processing external entries. From the logical standpoint, they should be `NEW`, though.

We can always change it if we want to.

PR #4733, opened 2019-08-09
Polished torrent creation dialog

devos50:
Fixed various layout inconsistencies, including a typo. Also moved the cancel button to the bottom of the dialog, which is more consistent with the layout in our other dialogs.

39db215431041ceb1319bf0a6a33c190


devos50:
retest this please

PR #4751, opened 2019-08-13
Fix false broken db warning in logger

ichorid:
This fixes the `Broken DB state!` false warnings fos some corner-case metadata update scenarios.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/store.py
+++ b/Tribler/Core/Modules/MetadataStore/store.py
@@ -436,11 +436,13 @@ class MetadataStore(object):
             if node.timestamp < payload.timestamp:
                 node.set(**payload.to_dict())
                 result.append((node, UPDATED_OUR_VERSION))
+                return result
             elif node.timestamp > payload.timestamp:
                 result.append((node, GOT_NEWER_VERSION))
+                return result
             # Otherwise, we got the same version locally and do nothing.
-            # The situation when something was marked for deletion, and then we got here (i.e. we have the same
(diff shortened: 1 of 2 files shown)


Tribler/Core/Modules/MetadataStore/store.py
```diff
@@ -436,11 +436,13 @@ def check_update_opportunity():
             if node.timestamp < payload.timestamp:
                 node.set(**payload.to_dict())
                 result.append((node, UPDATED_OUR_VERSION))
+                return result
             elif node.timestamp > payload.timestamp:
                 result.append((node, GOT_NEWER_VERSION))
+                return result
```
devos50: I think it is relatively easy to cover this line with a unit test. Could you please do so?

ichorid:
Note that C0330 check is wrong and we should disable it. See https://github.com/Tribler/tribler/issues/4104#issuecomment-520938141

Tribler/Core/Modules/MetadataStore/store.py
ichorid: Done.

Tribler/Test/Core/Modules/MetadataStore/test_store.py
```diff
@@ -5,10 +5,11 @@
 import string
 from binascii import unhexlify
 
+
```
devos50: Is this additional newline intended?

ichorid: No. I didn't run my pre-commit hooks on this thing, see...

PR #4759, opened 2019-08-19
Collective Effort: Nested channels support

ichorid:
(This PR is a collective effort continuation of #4720)

### Motivation
This PR started as an attempt to remove unused widgets from the GUI `.ui` file. Very soon, it became obvious that it is impossible without heavily refactoring all channels-related parts of the GUI. The code for the channel-showing widget was duplicated more than four times both in the `.ui` file and in `.py` files. In the end, small changes proved to be impossible.

**What started as a dead code removal, ended up as an effort to add nested channels support to both the GUI and the Core.**

 ### Main changes

#### GUI technical changes
 * All channels-related GUI stuff is now handled by a single widget `channelcontentswidget`.
 * `channelcontentswidget` features a single View into a stack of channel Models. It always shows the top model.
 * Each different channel-related page (e.g., Search, Discovered, etc.) runs its instance of `channelcontentswidget`. The root of the respective stack is initialized with a page-specific Root model.
 * Communication between the current (top of the stack) model and higher-level `channelcontentswidget` and its member objects are minimized to simplify model-switching and eliminate backreferences. There is only one signal for communicating changes from model to `channelcontentswidget`: the `info_changed` signal. It is reconnected on every change of the models' stack.
 * Non-top models in the stack remember their state (filters, sort order, etc.). They never get reloaded on hitting `<` button. (That make them super fast! :rocket: )
 * Row deletions are now handled in-place for some actions. The same is true for row info updates. This avoids costly channel view reloading :rocket:.

#### GUI feature changes
* :tada: **Full nested channels support!**  :tada: 
* The user can now have several top-level channels, with separate torrent for each channel.
* Sub-channels are called "folders" in the GUI. In the Core the corresponding type is called CollectionNode (as Channels inherit from it). Only true Channels get seeded with libtorrent. Folders are organizational units only, they reside in Channels.
* Ability to move/copy channels. Recursive, as expected.
* New dialog for selecting the target directory for moving/copying stuff into. The dialog supports **creating new channels/folders** :open_file_folder: :scissors: , lazy loading of the structure of the channels and tracks the updates to your channels' structure automatically :eyes: via the `info_changed` signal. 
* **Multiple selection** works as expected, with Ctrl-LMB and Shift-LMB!
* :file_cabinet: **Category filter drop-down menu** :file_cabinet:  is added to the channels view. It allows the user to filter the current channel on either torrent categories or channels/folders. This neatly unifies the Search page functionality with other pages.
* Torrent checking is now initiated only after 500ms after selection change. This makes the GUI much and the Core much more responsive during channels scrolling, as it avoids unwanted health checks.

#### Core changes
* Channels-related endpoints are refactored: everything is now (almost) completely compliant with the REST paradigm, entries are correctly identified by a combination of the public key and id instead of infohash.
* All REST data exchange with channels-related endpoints is going in JSON format now.
* Recursive commits / copying / moving / deletion algorithms are added for ORM-managed objects.
* `mychannel` endpoint is removed. Instead, `mychannel` is now just an alias for the user's public key.
* Ugly `count` endpoint is removed.
* Almost everything is now served by the single `metadata` endpoint that RESTfully PATCHes and POSTs objects. It is amazing how powerful this abstraction is! (However, this means some logic had to be moved to the GUI).

### Known problems / TODO:
 * Credit mining does not work for collections, because credit mining sources are identified by infohash, and collections got none. Instead, CreditMining Manager should identify sources by pk+id pair. GUI test for credit mining is disabled until this is fixed.
 * Changing root channel name is not supported, as there is no root channel now.
 * Changing channel name by clicking on it is not supported.
 * No button for copying the channel public key.
 * Searching by channel public key feature is deleted. It should be re-implemented properly as a generic RPC-like request to remote DB. (Searching for infohash is still there, as it can be useful to the user).
 * Infohash uniqueness requirement is removed, but the actual database format did not change. True database conversion is not required, but some indexes should be deleted, and some should be created. We don't have the code for this kind of upgrade yet.
 * Some minor signals/connections in the GigaChannels GUI might be missing.
 * Algorithms for recursive commits/deletion/copying are probably far from perfect. I've tried my best to make them work as fast as I could, but there is only so much one can do in a limited time. They are all tested, though.
 * Some buttons and their placement are UGLY (as always). Some can be in the wrong place.
 * View/controller stuff is all over the namesake objects and `channelcontentswidget`. There should be a better way to organize it, but even in its current state, it is much more idiomatic and clear than it was before.
 * Some form of garbage collection for ORM is required, as well as smarter sanity checks.
 * Overall, the code could be better in many places.


xoriole:
retest this please

xoriole:
@ichorid Could you please rebase the PR?

ichorid:
retest this please

ichorid:
@xoriole , rebased, good to go.

PR #4761, opened 2019-08-21
READY: Fixed issue with loading torrent with Unicode characters

xoriole:
Fixes the issue with loading torrent with Unicode characters in the name while loading by clicking the torrent file. The crash happened for the case when Tribler was not running already.


qstokkink:
Is there no related issue?

xoriole:
Reported by Vadim, no documented issue.

Issue #4767, opened 2019-08-24
tribler fails to start on debian9; 7.3.0-rc4_all.deb


imstanz:
#### Please, fill all relevant items: ####

- [ nope ] I have read CONTRIBUTING.rst

- [ yep ] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Tribler version/branch+revision: #####
tribler_7.3.0-rc4_all.deb

##### Operating system and version: #####
Debian 9,  Linux 4.9.0-9-amd64 #1 SMP Debian 4.9.168-1+deb9u5 (2019-08-11) x86_64 GNU/Linux
##### Steps to reproduce the behavior: #####
just install & try to start. both from menu or terminal
##### Expected behavior: #####
start & run
##### Actual behavior: #####
no response at 1st startup attempt. 
Checked for and Install remaining 'Recommends', and install for
development environment
[tribler_error.TXT](https://github.com/Tribler/tribler/files/3536685/tribler_error.TXT)

. 
blt (2.5.3+dfsg-3)
libtcl8.6 (8.6.6+dfsg-1+b1)
libtk8.6 (8.6.6-1+b1)
python-faulthandler (2.4-1)
python-tk (2.7.13-1)
tk8.6-blt2.5 (2.5.3+dfsg-3)
##### anything odd to report: #####
could not connect to core within 60 secs.
A few error reports have been submitted by me, after failed attempts.

qstokkink:
As per https://stackoverflow.com/questions/34819221/why-is-python-setup-py-saying-invalid-command-bdist-wheel-on-travis-ci, seems like `pip install wheel` is also required in our deb file.

synctext:
Ai, good we did another RC to test the installer. Still a missing dependency.

xoriole:
@imstanz could you write the output of pip version?
```
pip -V
```
also, try to install `wheel` followed by `pony` and `lz4`
```
pip install --user wheel
pip install --user pony lz4
```
Keep us informed if that works for you.

xoriole:
There seems to be issues with ubuntu 19.04 and Debian 10.
Many dependencies are missing because of Python 2.7
![ubuntu-19 04](https://user-images.githubusercontent.com/1442867/63694476-d763fa00-c816-11e9-9909-40046e7adf56.png)

xoriole:
@imstanz Could you try the latest release and let us know if it works for you?
https://github.com/Tribler/tribler/releases/tag/v7.3.0

[...]

PR #4769, opened 2019-08-26
Fixed pip package issues

xoriole:
Installing pip packages with `--user` with `sudo` previlege during installation is causing these packages to be installed for only root, while user simply runs tribler in user mode. This causes those packages not being available in user mode. With this PR, I'm removing the `--user` flag so it is installed globally and should be available to all users. Moreover, `wheel` pip dependency is also added which is necessary for systems with older version of pip.

Fixes https://github.com/Tribler/tribler/issues/4767


xoriole:
Yes, we do. Will update after the tests.

PR #4805, opened 2019-09-05
Fix white background on downloads page in Python 3

ichorid:
Some versions of Python 3.7 show wrong background colors for some widgets in PyQt 5.13.0.
Setting CSS properties explicitly makes for a workaround on this problem.


devos50:
retest this please

PR #4813, opened 2019-09-06
Removed tkinter

### LINKED ISSUE #4799: Drop Tkinter
At some point in time `tkinter` was shipped with Python.

However, this is no longer the case. 

Therefore we should not use `tkinter` anymore as a fallback option and switch to some other library we already use anyway (like Qt).

qstokkink:
Fixes #4799

### DIFF
--- a/check_os.py
+++ b/check_os.py
@@ -17,6 +17,22 @@ FORCE_RESTART_MESSAGE = "An existing Tribler core process (PID:%s) is already ru
                         "Do you want to stop the process and do a clean restart instead?"
 
 
+def _do_popup(title, text):
+    """
+    Create a native pop-up without any third party dependency.
+
+    :param title: the pop-up title
+    :param text: the pop-up body
+    """
+    try:
+        import win32api
+        win32api.MessageBox(0, text, title)
+    except ImportError:
+        os.system('xmessage -center "$(printf "%s")"' % (text))
+    sep = "*" * 20
+    print('\n'.join([sep, title, sep, text, sep]), file=sys.stderr)
+
+
 def error_and_exit(title, main_text):
     """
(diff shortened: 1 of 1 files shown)


qstokkink:
retest this please

check_os.py
```diff
@@ -17,53 +17,30 @@
...
+    Create a native pop-up without any third party dependency.
+
+    :param title: the pop-up title
+    :param text: the pop-up body
+    """
+    try:
+        import win32api
+        win32api.MessageBox(0, text, title)
+    except ImportError:
+        os.system('notify-send -u "critical" "%s" "$(printf "%s")"' % (title, text))
```
devos50: `notify-send` is not available on MacOS :(

check_os.py
```diff
@@ -17,53 +17,30 @@
...
+    Create a native pop-up without any third party dependency.
+
+    :param title: the pop-up title
+    :param text: the pop-up body
+    """
+    try:
+        import win32api
+        win32api.MessageBox(0, text, title)
+    except ImportError:
+        os.system('xmessage -center "$(printf "%s")"' % (text))
```
arvidn: I would like to point out that this appears to be a security risk. At least a ticking time-bomb. If an attacker can control the content of `text`, any arbitrary shell command can be executed.

arvidn: using `os.subprocess` with `shell=False` is probably sufficient to fix it btw

ichorid: I confirm that using `os.system` is a security risk and is considered to be a bad practice these days. `os.subprocess` with `shell=False` is the way to go safely.

devos50: Thanks for your feedback! I will change this as part of my PR (#4815)

PR #4815, opened 2019-09-06
Checking for existence of mediaplayer object

### LINKED ISSUE #4775: Video player issue on reset
Version: 7.3.0
Platform: Windows-10-10.0.18362-SP0
uptime: 58768.2129998 secs

```python
File "TriblerGUI\widgets\downloadspage.py", line 342, in on_remove_download_dialog
File "TriblerGUI\widgets\videoplayerpage.py", line 239, in reset_player
AttributeError: 'NoneType' object has no attribute 'stop'
```

Message from the reporter:

>I was setting up my channel and after I added some, I clicked on play within my channel. Player opened but nothing really happend after so I clicked on download and noticed Tribler is trying to download what I added to my channel. When I tried to remove it from download list the error occured.

devos50:
When removing a download and resetting the player.

Fixes #4775

### DIFF
--- a/check_os.py
+++ b/check_os.py
@@ -28,7 +28,8 @@ def _do_popup(title, text):
         import win32api
         win32api.MessageBox(0, text, title)
     except ImportError:
-        os.system('xmessage -center "$(printf "%s")"' % (text))
+        import subprocess
+        subprocess.Popen(['xmessage', '-center', text], shell=False)
     sep = "*" * 20
     print('\n'.join([sep, title, sep, text, sep]), file=sys.stderr)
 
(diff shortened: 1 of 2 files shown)


check_os.py
```diff
@@ -28,7 +28,8 @@ def _do_popup(title, text):
         import win32api
         win32api.MessageBox(0, text, title)
     except ImportError:
-        os.system('xmessage -center "$(printf "%s")"' % (text))
+        import subprocess
+        subprocess.Popen(['xmessage', '-center', text], shell=False)
```
qstokkink: Was this tested with newlines in the text?

ichorid: Just checked it, totally works with newlines.

PR #4829, opened 2019-09-09
Fix black-related pre-commit config

ichorid:
`black` [ignores](https://github.com/psf/black/issues/395) `include` and `exclude` config entries when running on a single exact file path. This is exactly how `pre-commit` works: running singular checks on singular changed files one-by-one. This means that to really exclude something from `black` scope we need to specify it in `.pre-commit-config.yaml`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,11 @@ repos:
     rev: 19.3b0
     hooks:
     - id: black
+      files: |
+          (?x)(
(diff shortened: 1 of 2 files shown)


qstokkink:
@egbertbouman seems like `Tribler.Test.Community.Tunnel.FullSession.test_tunnel_community.TestTunnelCommunity.test_anon_download` is also failing randomly.

ichorid:
retest this please

PR #4834, opened 2019-09-11
Limited trust graph

xoriole:
Set max nodes to 500 and max transactions to 2500 for trust graph rendering.


xoriole:
retest this please

Issue #4822, opened 2019-09-08
'NoneType' object has no attribute 'address'


devos50:
Saw this one in the error reporter:

```
Traceback (most recent call last):
File "TriblerGUI\event_request_manager.py", line 114, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'address'
lib\threading.py:801:__bootstrap_inner
lib\threading.py:754:run
lib\site-packages\twisted\_threads\_threadworker.py:46:work
lib\site-packages\twisted\_threads\_team.py:190:doWork
--- <exception caught here> ---
lib\site-packages\twisted\python\threadpool.py:250:inContext
lib\site-packages\twisted\python\threadpool.py:266:<lambda>
lib\site-packages\twisted\python\context.py:122:callWithContext
lib\site-packages\twisted\python\context.py:85:callWithContext
Tribler\Core\APIImplementation\LaunchManyCore.py:573:sesscb_states_callback
Tribler\community\triblertunnel\community.py:501:monitor_downloads
Tribler\community\triblertunnel\community.py:531:create_introduction_point
ipv8\messaging\anonymization\hidden_services.py:479:create_introduction_point
ipv8\messaging\anonymization\hidden_services.py:171:create_circuit_for_infohash
ipv8\messaging\anonymization\community.py:315:create_circuit
]
```

qstokkink:
Seems to be a duplicate of #4644. What version of Tribler is this?

egbertbouman:
Looks like a threading issue. I'll fix this in IPv8.

Issue #4821, opened 2019-09-08
b2a_hex() argument 1 must be string or buffer, not None


devos50:
Saw this one in the error reporter (26985):

```
Traceback (most recent call last):
File "TriblerGUI\event_request_manager.py", line 114, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.TypeError'>: b2a_hex() argument 1 must be string or buffer, not None
lib\threading.py:801:__bootstrap_inner
lib\threading.py:754:run
lib\site-packages\twisted\_threads\_threadworker.py:46:work
lib\site-packages\twisted\_threads\_team.py:190:doWork
--- <exception caught here> ---
lib\site-packages\twisted\python\threadpool.py:250:inContext
lib\site-packages\twisted\python\threadpool.py:266:<lambda>
lib\site-packages\twisted\python\context.py:122:callWithContext
lib\site-packages\twisted\python\context.py:85:callWithContext
Tribler\Core\APIImplementation\LaunchManyCore.py:573:sesscb_states_callback
Tribler\community\triblertunnel\community.py:463:monitor_downloads
Tribler\community\triblertunnel\community.py:557:get_lookup_info_hash
]
```

qstokkink:
This suggests `get_lookup_info_hash` was called with `None`.

Issue #4780, opened 2019-08-29
TypeError: must be string or buffer, not None


xoriole:
Version: 7.3.0
Platform: Windows-7-6.1.7601-SP1
Uptime:125.427000046
Report: [26571](http://reporter.tribler.org/stacktrace/26571)

```python
Traceback (most recent call last):
File "TriblerGUI\event_request_manager.py", line 114, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.TypeError'>: must be string or buffer, not None
lib\threading.py:810:__bootstrap_inner
lib\threading.py:763:run
lib\site-packages\twisted\_threads\_threadworker.py:46:work
lib\site-packages\twisted\_threads\_team.py:190:doWork
--- <exception caught here> ---
lib\site-packages\twisted\python\threadpool.py:246:inContext
lib\site-packages\twisted\python\threadpool.py:262:<lambda>
lib\site-packages\twisted\python\context.py:118:callWithContext
lib\site-packages\twisted\python\context.py:81:callWithContext
Tribler\Core\APIImplementation\LaunchManyCore.py:573:sesscb_states_callback
Tribler\community\triblertunnel\community.py:463:monitor_downloads
Tribler\community\triblertunnel\community.py:557:get_lookup_info_hash
]
```

Issue #4776, opened 2019-08-29
Unsupported scheme: ''


xoriole:
```python
Traceback (most recent call last):
File "TriblerGUI/event_request_manager.py", line 114, in on_read_data
RuntimeError: [Failure instance: Traceback: <class 'twisted.web.error.SchemeNotSupported'>: Unsupported scheme: ''
twisted/internet/defer.py:565:_startRunCallbacks
twisted/internet/defer.py:651:_runCallbacks
Tribler/Core/Utilities/utilities.py:76:_on_response
Tribler/Core/Utilities/utilities.py:87:http_get
--- <exception caught here> ---
twisted/web/client.py:1596:request
twisted/web/client.py:1580:_getEndpoint
twisted/web/client.py:1460:endpointForURI
]
```
Reporter ticket: [26548](http://reporter.tribler.org/stacktrace/26548)
Related issue: https://github.com/Tribler/tribler/issues/3828

devos50:
It seems that the information passed in the `location` header in the HTTP response is invalid. In this case, the `http_get` method returns a `Failure`, which apparently is not handled correctly.

PR #4836, opened 2019-09-11
Various fixes

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/4821, fixes https://github.com/Tribler/tribler/issues/4780, fixes https://github.com/Tribler/tribler/issues/4776, fixes https://github.com/Tribler/tribler/issues/4822. Possibly also fixes https://github.com/Tribler/tribler/issues/4757.

### DIFF
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -55,6 +55,7 @@ class BaseTestCase(unittest.TestCase):
 
     def __init__(self, *args, **kwargs):
         super(BaseTestCase, self).__init__(*args, **kwargs)
+        self.selected_ports = set()
         self._tempdirs = []
         self.maxDiff = None  # So we see full diffs when using assertEquals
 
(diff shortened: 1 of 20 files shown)


qstokkink:
Also #4757?

egbertbouman:
@qstokkink I'm not sure. Based on how libtorrent would write the file to disk, I calculated the block borders and determined the possible token errors that we would get. Out of the 7 tokens I tried, only 6 matched. Although it could be that Tribler tried to read the bootstrap file in the middle of writing a block to disk, or that the missing token is part of one of the older bootstrap downloads.

Tribler/Test/test_as_server.py
```diff
@@ -91,6 +92,36 @@ def temporary_directory(self, suffix='', exist_ok=False):
...
 
+    def get_bucket_range_port(self):
+        """
+        Return the port range of the test bucket assigned.
+        """
+        test_bucket = os.environ.get("TEST_BUCKET", None)
+        if test_bucket is not None:
+            test_bucket = int(test_bucket) + int(os.environ.get('TEST_BUCKET_OFFSET', 0))
+
+        min_base_port = 1000 if test_bucket is None else test_bucket * 2000 + 2000
```
ichorid: Ports 0-1023 are system ports and are unavailable to users. So, `min_base_port` should be at least 1024

PR #4841, opened 2019-09-18
Added check for dependencies during startup

xoriole:
I have updated `run_tribler.py` to check for missing dependencies before anything else is loaded. This will help us find out dependency issues. Note that, I added a `dependencies.py` file which includes a list of all dependencies. We should keep this file updated whenever we change any dependency.

Also, note that for each dependency `scope` and `optional` fields are set. This separates where the dependency is used ('gui', 'core' or 'both') and if it is optional so both GUI and Core can be separately tested for the dependencies.

Missing dependencies will be shown in a system popup as shown below:
missing-deps

### DIFF
--- a/Tribler/dependencies.py
+++ b/Tribler/dependencies.py
@@ -1,0 +1,72 @@
+"""
+This file lists the python dependencies for Tribler.
+
+Note that this file should not depend on any external modules itself other than builtin ones.
+"""
+from __future__ import absolute_import, print_function
+
+import importlib
+import sys
+
+dependencies = [
+    {'module': 'PyQt5', 'pip_install': 'pip3 install PyQt5', 'optional': False, 'scope': 'gui'},
+    {'module': 'twisted', 'pip_install': 'pip3 install Twisted', 'optional': False, 'scope': 'core'},
+    {'module': 'libtorrent', 'pip_install': 'apt install python-libtorrent', 'optional': False, 'scope': 'core'},
+    {'module': 'cryptography', 'pip_install': 'pip3 install cryptograpy>=2.3', 'optional': False, 'scope': 'core'},
+    {'module': 'libnacl', 'pip_install': 'pip3 install libnacl', 'optional': False, 'scope': 'core'},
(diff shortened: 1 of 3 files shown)


Tribler/dependencies.py
```diff
@@ -0,0 +1,72 @@
...
+    for dep in dependencies:
+        if scope == 'both' or dep['scope'] == 'both' or dep['scope'] == scope:
+            try:
+                importlib.import_module(dep['module'])
+            except ImportError:
+                if not dep['optional']:
+                    missing_deps.append(dep)
+
+    if missing_deps:
+        formatted_message = "\n ".join([miss_dep['pip_install'] for miss_dep in missing_deps])
```
ichorid: It is inconvenient to copy/paste and run pip commands one-by-one. Could you please form a single command for installing everything at once? like this: `pip install dep1 dep2` etc.

Tribler/dependencies.py
```diff
@@ -0,0 +1,72 @@
...
+    {'module': 'chardet', 'pip_install': 'pip3 install chardet', 'optional': False, 'scope': 'core'},
+    {'module': 'cherrypy', 'pip_install': 'pip3 install cherrypy', 'optional': False, 'scope': 'core'},
+    {'module': 'configobj', 'pip_install': 'pip3 install configobj', 'optional': False, 'scope': 'both'},
+    {'module': 'netifaces', 'pip_install': 'pip install netifaces', 'optional': False, 'scope': 'core'},
+    {'module': 'six', 'pip_install': 'pip install six', 'optional': False, 'scope': 'both'},
+    {'module': 'bitcoinlib', 'pip_install': 'pip install bitcoinlib', 'optional': True, 'scope': 'core'},
+]
+
+
+def show_system_popup(title, text):
```
qstokkink: Could you make this private? Nobody should ever use this outside of this file.

xoriole: Good point. Done

Tribler/dependencies.py
xoriole: Good point. Done!

xoriole:
After grouping the dependencies:
missing-deps-grouped

PR #4858, opened 2019-09-23
Add level2 category filter

### LINKED ISSUE #4855: Level2 category filter
Some kinds of information should be completely off-limits in Tribler. One research that can help with this is https://www.cl.cam.ac.uk/~nz247/publications/PAM12.pdf

ichorid:
Adds filtering of all incoming GigaChannel payloads for really bad stuff.
Fixes #4855 

The terms are extracted after a quick survey of scientific papers on the topic. For instance, [this](https://www.cl.cam.ac.uk/~nz247/publications/PAM12.pdf). Unfortunately, none of those papers provide a full list of terms or source code. So it is a pretty ad-hoc list.


xoriole:
@ichorid A separate filter file similar to existing filter `filter_terms.filter` might be better instead of specifying the keywords directly in a python file.

ichorid:
Have to fix Windows busy file on test

PR #4884, opened 2019-10-04
READY: Prioritize gossiping personal channel

ichorid:
This  PR makes GigaChannel community send two packets on each walk step instead of one:
the first packet contains gossip with the user's personal channel, the second one contains a random subscribed channel. The format remains the same, everything is compatible with the previous version of the GigaChannel community.

During the last month, observation of GigaChannel dynamics showed that the original content propagates slowly. One reason for that is that if a user subscribes for several channels, their personal channel has a really low chance to sent to others. User's personal channel should always get the highest priority in gossip. This is how interpersonal relationships work: we ask "How are *you* doing?" and tell others of things that matter to *us* first.

### DIFF
--- a/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
+++ b/Tribler/Core/Modules/MetadataStore/OrmBindings/channel_metadata.py
@@ -410,7 +410,8 @@ def define_binding(db):
             :rtype: list
             """
             query = db.ChannelMetadata.select(
-                lambda g: g.status not in [LEGACY_ENTRY, NEW, UPDATED, TODELETE] and g.num_entries > 0)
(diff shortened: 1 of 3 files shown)


ichorid:
retest this please

synctext:
Just thinking.. This means you can easily de-anonymise channel owners?

ichorid:
> Just thinking.. This means you can easily de-anonymise channel owners?

Channel owners are never anonymous. Each gossip element is signed with its creator's public key and includes the public key itself. Anonymity for channel owners never was a priority. If someone wants to spread around stuff anonymously in Tribler network, they can always create their channel, change the public key, and *then* spread the thing around.

synctext:
OK. Good to have this explicit here.

ichorid:
> Could you please add a test (or update existing) showing that a peer will receive two channels (personal and another a random)?

Added

ichorid:
retest this please

PR #4887, opened 2019-10-10
READY: Fix non-deterministic results in squash_mdblobs tests

ichorid:
Added `random.seed(123)` to the base test class to make everything as deterministic as possible.
On `tearDown` random generator is re-seeded with the system default generator.

### DIFF
--- a/Tribler/Test/test_as_server.py
+++ b/Tribler/Test/test_as_server.py
@@ -157,6 +157,7 @@ class AbstractServer(BaseTestCase):
         self.annotate(self._testMethodName, start=True)
         self.watchdog.start()
         yield reactor_deferred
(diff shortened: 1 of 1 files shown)


Tribler/Test/Core/Modules/MetadataStore/test_store.py
```diff
@@ -93,6 +93,7 @@ def test_process_channel_dir_file(self):
 
     @db_session
     def test_squash_mdblobs(self):
+        random.seed(123)
```
devos50: Note that this will affect the randomness of all tests. It's better to create an instance of `Random` and seed/use that one.

xoriole:
retest this please

Issue #4891, opened 2019-10-13
Log exception instead of ignoring it


Solomon1732:
- [x] I have read CONTRIBUTING.rst

##### Tribler version/branch+revision: #####
V7.3.1-RC3 and devel
##### Operating system and version: #####
All
##### Expected behavior: #####
Exception should at least be logged to aid developers' debugging and users' error submissions.
##### Actual behavior: #####
Exception ignored and swallowed.
##### Location In Codebase: #####
https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Upgrade/upgrade.py#L57

devos50:
I think logging the error you linked could indeed be helpful for both users and developers. @ichorid is there a reason why this exception is not logged?

ichorid:
The line @Solomon1732 is pointing out is in 7.3.0->7.3.1 conversion procedure. It checks `.pstate` files for the files not confromant to the channel torrent format and deletes them. This kind of files can only appear in two cases:
* the channel creator produced a non-conformant channel torrent file, and it was downloaded in 7.3.0 (7.3.1 checks the channels torrent for conformance *before* downloading);
* some local error resulted in malformed `.pstate` file.

The first case is handled by the upgrader script. The second case (malformed file) should be handled upstream by Libtorrent wrapper. Therefore, there is no reason for us to put something in the log in this case during the upgrade procedure: the same files will be read again and handled appropriately at the later stages of Tribler startup.

@Solomon1732 , however if you want, you can create a PR to add some logging statement.

Solomon1732:
Would "debug" level suffice? Or should it be something like "warning"?

Solomon1732:
Created pull request at "debug" level #4894

Solomon1732:
I'm closing the issue since #4894 , which was based on this issue, was merged

PR #4894, opened 2019-10-13
READY: Added debug level logging to pstate parsing exception

Solomon1732:
Added a debug level logging statement to the exception of a malformed .pstate file as per bug  #4891

### DIFF
--- a/Tribler/Core/Upgrade/upgrade.py
+++ b/Tribler/Core/Upgrade/upgrade.py
@@ -55,7 +55,8 @@ def cleanup_noncompliant_channel_torrents(state_dir):
                     if name and len(name) != CHANNEL_DIR_NAME_LENGTH:
                         os.unlink(file_path)
                 except (TypeError, KeyError, ValueError):
-                    logger.debug("Malformed .pstate file %s", file_path)
+                    logger.debug("Malfored .pstate file %s found during cleanup of non-compliant channel torrents.",
+                                 file_path)
 
 
 class TriblerUpgrader(object):
--- a/Tribler/Core/Upgrade/upgrade.py
+++ b/Tribler/Core/Upgrade/upgrade.py
@@ -55,7 +55,7 @@ def cleanup_noncompliant_channel_torrents(state_dir):
                     if name and len(name) != CHANNEL_DIR_NAME_LENGTH:
                         os.unlink(file_path)
                 except (TypeError, KeyError, ValueError):
-                    pass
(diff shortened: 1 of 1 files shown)


devos50:
ok to test

Tribler/Core/Upgrade/upgrade.py
```diff
@@ -55,7 +55,7 @@ def cleanup_noncompliant_channel_torrents(state_dir):
                     if name and len(name) != CHANNEL_DIR_NAME_LENGTH:
                         os.unlink(file_path)
                 except (TypeError, KeyError, ValueError):
-                    pass
+                    logger.debug("Malformed .pstate file %s", file_path)
```
devos50: Could you make this logging statement a bit more specific? E.g.: "Malfored .pstate file %s found during cleanup of non-compliant channel torrents."

Solomon1732:
Made logged message more descriptive as per @devos50 's feedback

Tribler/Core/Upgrade/upgrade.py
Solomon1732: Made it more descriptive per your feedback

Solomon1732: On a minor note, instead of hiding the message mentioning you I deleted it... Sorry about that...
For protocol its content was basically what I wrote above plus `@ name` of your name.

Solomon1732:
> Please, merge two commits together.
Two commits for a single-line change is one commit too many.

Sorry about this. I admit that I'm a novice at using git, so after downloading github for desktop and struggling with rebasing, I saw in this [github help article](https://help.github.com/en/articles/about-pull-request-merges) that it's possible to rebase on the receiving end of a pull request, unless I'm misunderstanding something?

What do you think?

ichorid:
@Solomon1732 , that will result in a "fast-forward" merge. We don't typically do this, but I guess that's fine. Also, I would recommend you going through an interactive git branching tutorial like [this](https://learngitbranching.js.org/). It really helps with understanding git branching model!

Solomon1732:
> 
> 
> @Solomon1732 , that will result in a "fast-forward" merge. We don't typically do this, but I guess that's fine. Also, I would recommend you going through an interactive git branching tutorial like [this](https://learngitbranching.js.org/). It really helps with understanding git branching model!

Thanks! Sorry about all the hassle @ichorid ! ^^;

Issue #4897, opened 2019-10-14
Enumerating iglob in for-each possibly not needed?


Solomon1732:
- [x] I have read CONTRIBUTING.rst
##### Tribler version/branch+revision: #####
devel
##### Operating system and version: #####
All

In the following line there's an `enumerate` used

https://github.com/Tribler/tribler/blob/887bd0499b27b8850ea890ff49fe09a4a592e354/Tribler/Core/Upgrade/config_converter.py#L50

Reading the iglob docs for [py 2](https://docs.python.org/2/library/glob.html#glob.iglob) and [py 3](https://docs.python.org/3/library/glob.html#glob.iglob) indicates to me that it returns an iterator. Since the index isn't used and no further transformations are done on the `iglob` iterator itself, is it possible it was necessary to enumerate the iterator in the past but as it stands it may no longer be necessary? Or am I missing something?

ichorid:
The reason was either some legacy code transition or something to do with threads. But `iglob` iterator is thread-safe. @devos50 what was the reason for this?

@Solomon1732 , you can change it to iterator and try it out for yourself. If there *is* a test for this, and it will not fail, assume it is safe to file a PR for the change. If there is *no* test for this, consider writing the test along with the proposed change (yep, that's not fair...).

Solomon1732:
https://github.com/Tribler/tribler/blob/887bd0499b27b8850ea890ff49fe09a4a592e354/Tribler/Test/Core/Upgrade/test_config_upgrade_70_71.py#L80

@ichorid I can see that there's *a* test for the function in question, but I can't tell if it's one that checks the functionality in question... I think it does, but I'm not sure 😅

Solomon1732:
Closing since #4902 fixed the bug

PR #4895, opened 2019-10-13
WIP: Handle resources with "with" statements

Solomon1732:
Resources (files, sockets, and database connections) in relatively old files in both the Tribler application and its testing files were handled manually or handled with "finally". This is due to them being written before "with" statements were added to python (per @devos50 's comment in issue #4890 )

They have been rewritten to be handled with "with" statements to handle them correctly in case of exceptions, avoid doubly closing them (some were closed manually after being handled with "with" block), and make the code more concise, readable, and less brittle.

PR #4902, opened 2019-10-19
Iterate iglob result directly

Solomon1732:
The result received from `iglob` has been iterated through `enumerate`. The need for an index has at some point been rendered not needed, and the index variable replaced with an underscore, so there's no need for `enumerate` and the result can be iterated directly. Fixes bug #4897.

### DIFF
--- a/Tribler/Core/Upgrade/config_converter.py
+++ b/Tribler/Core/Upgrade/config_converter.py
@@ -47,8 +47,7 @@ def convert_config_to_tribler71(current_config, state_dir=None):
 
     # We also have to update all existing downloads, in particular, rename the section 'downloadconfig' to
     # 'download_defaults'.
-    for _, filename in enumerate(iglob(
-            os.path.join(state_dir, STATEDIR_CHECKPOINT_DIR, '*.state'))):
+    for filename in iglob(os.path.join(state_dir, STATEDIR_CHECKPOINT_DIR, '*.state')):
         download_cfg = RawConfigParser()
         try:
             with open(filename) as cfg_file:


Solomon1732:
The reason that this is a draft and a WIP is that in #4897 @ichorid stated:
> @Solomon1732 , you can change it to iterator and try it out for yourself. If there is a test for this, and it will not fail, assume it is safe to file a PR for the change. If there is no test for this, consider writing the test along with the proposed change (yep, that's not fair...).

I replied with:
>https://github.com/Tribler/tribler/blob/887bd0499b27b8850ea890ff49fe09a4a592e354/Tribler/Test/Core/Upgrade/test_config_upgrade_70_71.py#L80
@ichorid I can see that there's *a* test for the function in question, but I can't tell if it's one that checks the functionality in question... I think it does, but I'm not sure 😅

If the test I found will be sufficient I will remove the draft status. Otherwise I will add a test.

devos50:
ok to test

Solomon1732:
retest this please

Solomon1732:
Hm... can't seem to understand what causes the exception...
https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux_py3_tester/264/console

Might be relevant for #4895

ichorid:
retest this please

ichorid:
retest this please

Solomon1732:
O...k, then...
Since latest retesting is successful I'm changing it from draft PR to Ready for review PR

ichorid:
retest this please

Issue #4909, opened 2019-10-23
Tribler crashes when Client terminates the torrent addition request prematurely. 


DanGraur:
#### Please, fill all relevant items: ####

- [x] I have read CONTRIBUTING.rst

- [x] I have tried with the latest pre-release version and I still can reproduce the issue.

##### Steps to reproduce the behavior: #####

1. Forward a torrent addition request via magnet link to the `mychannel/torrents` endpoint.
2. Terminate the connection on the Client side before the Server gets a chance to call `request.finish()`.

##### Relevant log file output: #####

```
Traceback (most recent call last):
  File "/data/software/tribler-7.3/tribler/TriblerGUI/event_request_manager.py", line 114, in on_read_data
    raise RuntimeError(json_dict["event"]["text"])
RuntimeError: [Failure instance: Traceback: <type 'exceptions.RuntimeError'>: Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this.
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:460:callback
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
--- <exception caught here> ---
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/software/tribler-7.3/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:288:_on_add_failed
/home/dan/.local/lib/python2.7/site-packages/twisted/python/failure.py:462:trap
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/software/tribler-7.3/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:285:_on_added
/home/dan/.local/lib/python2.7/site-packages/twisted/web/server.py:268:finish
/home/dan/.local/lib/python2.7/site-packages/twisted/web/http.py:1071:finish
```

ichorid:
Fixed by #4910

PR #4910, opened 2019-10-23
READY: Fix for torrent addition requests when Client terminates request prematurely

DanGraur:
This PR fixes issue #4909. Previously, when a Client would request the addition of a torrent by forwarding a `PUT` HTTP request towards the `mychannel/torrents` endpoint, if Tribler's Twisted handler (`render_PUT`) would follow an execution path which initially returns a `NOT_DONE_YET` (this happens for instance for magnet links or URLs), then if the requesting Client terminates the request prematurely, Tribler would crash, as it later calls `request.finish()` on a request which has been terminated prematurely.

To fix this, a flag variable (called `can_close`) is introduced. This is a list with one boolean element (initially `True`), which indicates whether the server can close the connection itself, or if it has already been closed. In order to be able to tell when a request has been terminated (either gracefully or via unexpected circumstances), a callback is attached to it via the `notifyFinish()` method. The callback method is named `_faulty_req_termination` (nested in `render_PUT`). Its purpose is to change the value in `can_close` to `False`, indicating that the server cannot call `request.finish()`. Moreover, when the callback is called due to a `twisted.internet.error.ConnectionDone` error, it will trap it, and report its occurrence via the error logs. The callback can be triggered by the server itself, when calling `request.finish()` (when `can_close[0] == True`), however, this is perfectly normal.

This idea should be reusable for other Twisted based asynchronous request handlers, which may be susceptible to premature request termination by the Client.

### DIFF
--- a/Tribler/Core/Modules/restapi/mychannel_endpoint.py
+++ b/Tribler/Core/Modules/restapi/mychannel_endpoint.py
@@ -15,6 +15,7 @@ from six import viewitems
 from six.moves.urllib.parse import unquote
 
 from twisted.internet.defer import Deferred
+from twisted.internet.error import ConnectionDone
 from twisted.web import http, resource
 from twisted.web.error import SchemeNotSupported
 from twisted.web.server import NOT_DONE_YET
@@ -262,13 +263,27 @@ class MyChannelTorrentsEndpoint(BaseMyChannelEndpoint):
 
         title = parameters['title'][0] if 'title' in parameters and parameters['title'] else None
 
+        # This is required to determine if the connection can be closed by the server
+        can_close = [True]
+
+        def _faulty_req_termination(failure, can_close):
+            # If this callback was triggered, then the connection is guaranteed to have been closed.
+            can_close[0] = False
+            if failure is not None:
+                failure.trap(ConnectionDone)
+                self._logger.exception("Connection did not close properly: %s %s", failure.getErrorMessage(),
+                                       failure.type)
+
+        request.notifyFinish().addBoth(_faulty_req_termination, can_close)
+
         def _on_url_fetched(data):
             return TorrentDef.load_from_memory(data)
 
         def _on_magnet_fetched(meta_info):
             if not meta_info:
                 request.write(self.return_500(request, RuntimeError("Metainfo timeout")))
-                request.finish()
+                if can_close[0]:
+                    request.finish()
                 return
 
             return TorrentDef.load_from_dict(meta_info)
@@ -284,13 +299,15 @@ class MyChannelTorrentsEndpoint(BaseMyChannelEndpoint):
 
         def _on_added(added):
             request.write(json.twisted_dumps({"added": added}))
-            request.finish()
+            if can_close[0]:
+                request.finish()
 
         def _on_add_failed(failure):
             failure.trap(ValueError, DuplicateTorrentFileError, SchemeNotSupported)
             self._logger.exception(failure.value)
             request.write(self.return_500(request, failure.value))
-            request.finish()
+            if can_close[0]:
+                request.finish()
 
         # First, check whether we did upload a magnet link or URL
         if 'uri' in parameters and parameters['uri']:


ichorid:
Yep, that would be a better solution. And the ultimate one would be to use cancellation scopes from Python 3.8.

ichorid:
@DanGraur , could you please try out @qstokkink 's solution with `request.finished`?

DanGraur:
I already tried it a couple of times prior to coming up with this solution and it didn't work, as I would still get the `Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this` error. I don't know if the `request.finished` field is a reliable means of indicating whether the request has truly been finished when the client terminates the request abruptly. I can try it again, just to make sure I didn't miss anything the first time, and if successful, I'll forward a new PR, but I don't think I'm wrong on this one.

DanGraur:
I tried using the `request.finished` field approach once more, to see if it can fix this problem, and it did not work. I got the same error as before:

```
  File "/data/projects/python/tribler/TriblerGUI/event_request_manager.py", line 114, in on_read_data
RuntimeError: [Failure instance: Traceback: <type 'exceptions.RuntimeError'>: Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this.
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:460:callback
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/projects/python/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:306:_on_add_failed
/home/dan/.local/lib/python2.7/site-packages/twisted/python/failure.py:462:trap
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/projects/python/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:303:_on_added
/home/dan/.local/lib/python2.7/site-packages/twisted/web/server.py:268:finish
/home/dan/.local/lib/python2.7/site-packages/twisted/web/http.py:1071:finish
]
```

I'm not quite sure why `request.finished` is not reliable in this case, as I didn't dig too deeply into Twisted's code. Perhaps there's a race condition in Twisted which is causing this unreliability. 

I should also mention that there was a `request.write()` immediately prior to every `if not request.finished: request.finish()` block, which would have thrown a `RuntimeError` itself had the `request.finished` field been set to `True`. 

The only thing I can conclude here is that simply using `request.finished` does not work for this case.

qstokkink:
Just looked at the code, apparently there are two flags, `_disconnected` AND `finished`, both should be `False`:

https://github.com/twisted/twisted/blob/twisted-19.7.0/src/twisted/web/http.py#L1069-L1075

DanGraur:
Indeed, this seems to solve the issue. By changing `if not request.finished:` to `if not request.finished and not request._disconnected:`, also solves this problem. `request._disconnected` is however a protected member of the `Request` class, so accessing it directly might be considered bad practice. That being said, I can forward a PR which changes the code such that it uses the `request._disconnected` approach in lieu of the `notifyFinish()` one. Should I do so?

qstokkink:
Seeing as we will have to switch out all our `if not request.finished:` statements throughout the codebase, I'd personally prefer the short version 😃

DanGraur:
Ok, sure thing. If it's ok with you, then I'll forward a PR which changes all the `if not request.finished:` to `if not request.finished and not request._disconnected:`. Should be pretty straight forward.

ichorid:
This discussion looks especially amusing in the light of the upcoming move to AIOHTTP, which is planned for the next week. Poor @egbertbouman ... :skull:

PR #4911, opened 2019-10-24
Fix channels not downloading on Python3

ichorid:
Fix mismatch in text types in Python3 in GigaChannel Manager routine that checks channels for correction. Without this fix, all channels would be rejected in Python3 as incorrect.


ichorid:
retest this please

Issue #4912, opened 2019-10-24
Better defensive programming required for finishing Twisted requests


DanGraur:
As discussed in PR #4910 (which stemmed from issue #4909), using the condition `if not request.finished:` (where `request` is of the `twisted.web.http.Request` type)  as a guard prior to calling `request.finish()`, does not always prove to be safe and sufficient. In certain scenarios, such as when the client abruptly terminates the connection, the `request.finished` field is not set to `True`, and the `request.finish()` method will be called in unsafe circumstances, triggering an error of the following type:

```
Traceback (most recent call last):
  File "/data/projects/python/tribler/TriblerGUI/event_request_manager.py", line 114, in on_read_data
    raise RuntimeError(json_dict["event"]["text"])
RuntimeError: [Failure instance: Traceback: <type 'exceptions.RuntimeError'>: Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this.
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:460:callback
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
--- <exception caught here> ---
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/projects/python/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:306:_on_add_failed
/home/dan/.local/lib/python2.7/site-packages/twisted/python/failure.py:462:trap
/home/dan/.local/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/data/projects/python/tribler/Tribler/Core/Modules/restapi/mychannel_endpoint.py:303:_on_added
/home/dan/.local/lib/python2.7/site-packages/twisted/web/server.py:268:finish
/home/dan/.local/lib/python2.7/site-packages/twisted/web/http.py:1071:finish
]
``` 

In such scenarios, Twisted will update the `request._disconnected` protected field, by setting it to `True`. Hence, one solution to this issue is to update the aforementioned `if` clauses from `if not request.finished:` to `if not request.finished and not request._disconnected:`, in order to improve the robustness of Tribler in such scenarios.

ichorid:
[...]

PR #4913, opened 2019-10-24
WIP: Defensive programming approach for terminating requests Server-Side safely

DanGraur:
As detailed in issue #4912, Tribler is sometimes susceptible to crashing, when it attempts to close the connection by calling `request.finish()` when the request has actually been unexpectedly terminated in the meantime. This happens even in scenarios where the call to `request.finish()` is guarded by an `if not request.terminated:` clause. As outlined in the aforementioned issue this does not suffice, as the clause needs to be extended to `if not request.terminated and not request_disconnected:` to safely avoid calling `request.finish()` when the situation demands it. The code forwarded in this PR does just this: extends the aforementioned clauses.

### DIFF
--- a/Tribler/Core/Modules/restapi/create_torrent_endpoint.py
+++ b/Tribler/Core/Modules/restapi/create_torrent_endpoint.py
@@ -119,7 +119,7 @@ class CreateTorrentEndpoint(resource.Resource):
 
             request.write(json.twisted_dumps({"torrent": base64.b64encode(result['metainfo']).decode('utf-8')}))
             # If the above request.write failed, the request will have already been finished
-            if not request.finished:
+            if not request.finished and not request._disconnected:
                 request.finish()
 
         def _on_create_failure(failure):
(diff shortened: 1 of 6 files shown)


DanGraur:
I should also mention that I've noticed the existence of a [method called `finish_request`](https://github.com/Tribler/tribler/blob/c292ada2c0cbe226fad831b9036e3bd867455adb/Tribler/Core/Modules/restapi/torrentinfo_endpoint.py#L34), which wraps a `request.finish()` call within a `try .. except` block that catches `RuntimeError`s. I don't know if this is safe to do. I've removed two of it's usages in this PR, since it fell into the scope of the task I was trying to accomplish here, however, there are 3 other usages (as far as I've seen). Do you think it would be wise to use `if not request.terminated and not request_disconnected:` clauses in lieu of the aforementioned `finish_request` method?

DanGraur:
Since issue #4912 was closed, I will also close this PR.

PR #4918, opened 2019-10-28
Builder update

xoriole:
- Added show console in Pyinstaller build
- Removed Matplotlib as dependency on all platforms
- Updated VLC bindings
- Fixed VLC stale plugins warning
- Added six also as python3 dependency


xoriole:
`vlc.py` is generated python binding provided by VLC with some custom changes to load the DLL from within Tribler. The changes to this file in this PR only address the updates on the bindings for the newer version of VLC (3.0.8).

PR #4926, opened 2019-10-31
Updated IPv8 pointer

xoriole:
Updated IPv8 pointer


xoriole:
```
ModuleNotFoundError: No module named 'pyasn1'
```
Seems `pyasn1` module is missing. I wonder how IPv8 tests were passing.

qstokkink:
`pyasn1` used to be installed with the `cryptography` dependency.

PR #4945, opened 2019-11-11
State directory update for rollback support

xoriole:
Whenever the code version (present in `version.py`) and config version (present in `triblerd.conf`) are different, state directory is backed up. This is done by copying the current `.Tribler` directory under the version name present in `triblerd.conf`, also inside the `.Tribler` directory. Since only the copy of the state directory is done, the state directory remains always forward compatible. When Tribler starts it tries to load the state directory from the versioned path. If it cannot find the version path then it falls back to `.Tribler`. 

The directory structure for backing up the state directory looks like the following:
*For stable release, eg. 7.4.0*
```
/home/<user>/.Tribler/7.4.0
```
*For pre-release, eg. 7.4.0-exp1*
```
/home/<user>/.Tribler/7.4.0/exp1
```

### DIFF
--- a/Tribler/Core/Upgrade/upgrade.py
+++ b/Tribler/Core/Upgrade/upgrade.py
@@ -16,7 +16,10 @@ from Tribler.Core.Modules.MetadataStore.store import MetadataStore
 from Tribler.Core.Upgrade.config_converter import convert_config_to_tribler71, convert_config_to_tribler74
 from Tribler.Core.Upgrade.db72_to_pony import DispersyToPonyMigration, cleanup_pony_experimental_db, should_upgrade
 from Tribler.Core.Utilities.configparser import CallbackConfigParser
-from Tribler.Core.simpledefs import NTFY_FINISHED, NTFY_STARTED, NTFY_UPGRADER, NTFY_UPGRADER_TICK
+from Tribler.Core.osutils import dir_copy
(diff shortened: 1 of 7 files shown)


Tribler/Core/Upgrade/upgrade.py
```diff
@@ -174,6 +178,34 @@ def upgrade_config_to_71(self):
...
+                src_sub_dirs = os.listdir(src_state_dir)
+                for backup_dir in backup_dirs:
+                    if backup_dir in src_sub_dirs:
+                        dir_copy(os.path.join(src_state_dir, backup_dir), os.path.join(dest_state_dir, backup_dir))
+                    else:
+                        os.makedirs(os.path.join(dest_state_dir, backup_dir))
+
+                # Backup keys and config files
+                backup_files = ['ec_multichain.pem', 'ecpub_multichain.pem', 'ec_trustchain_testnet.pem',
+                                'ecpub_trustchain_testnet.pem', 'triblerd.conf']
```
qstokkink: Does this work if one of the files does not exist?

xoriole: Yes.  If the file does not exist it will be created during session initialization and later copied during upgrade.

qstokkink: Alright, I thought testnet keys were not generated by default.

PR #4948, opened 2019-11-14
Remove some dead code

ichorid:
Found some dead code both manually and with `vulture`


Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -633,17 +631,9 @@ def start_download(self, torrentfilename=None, infohash=None, tdef=None, dconfig
         # the priority of the parameters is: (1) tdef, (2) infohash, (3) torrent_file.
         # so if we have tdef, infohash and torrent_file will be ignored, and so on.
         if tdef is None:
-            if infohash is not None:
```
qstokkink: This seems like it should stay.

ichorid: No, it can be safely removed. `get_collected_torrent` is long dead. The code block under this condition is basically never executed - and that's why we get no errors from the thing being dead.

qstokkink: Agreed. Seems fine.

Tribler/Core/APIImplementation/LaunchManyCore.py
```diff
@@ -763,31 +756,6 @@ def early_shutdown(self):
             self.session.notify_shutdown_state("Shutting down IPv8...")
             yield self.ipv8.stop(stop_reactor=False)
 
-        if self.channelcast_db is not None:
```
devos50: Hmm, interesting that we missed this one. Most likely because `LaunchManyCore.py` is just too complex...

Tribler/Core/Libtorrent/LibtorrentMgr.py
devos50: Also, we should get rid of these nested `is None` or `is not None` combinations. There are methods that receive like four or five parameters and depending on whether they are set or not, the executed logic is completely different.

PR #4973, opened 2019-11-27
READY: Move seeding ratio settings to global config

### LINKED ISSUE #4716: No seeding setting broken
RC1 on Mac.

Seed and see the upload continue.

ichorid:
Fixes #4716  (mostly)

Our default download settings system is broken. To un-break it (at least in some way), this PR moves default download settings to be managed by top-level Tribler config. Also, `user_stopped` option in download configs was used by LibtorrentManager to stop downloads that reached their seeding limit. This is semantically wrong, as it was not *the user* who caused the stop. This PR fixes the problem.

The only problem remaining after applying this PR is that, if user changes seeding setting, Tribler will only automatically restart downloads that were stopped because of reaching seeding limit on the next restart. I do not dare to touch LibtorrentManager *that deep*. :mushroom:  :see_no_evil: :mushroom:


ichorid:
retest this please

ichorid:
retest this please

Tribler/Core/Modules/restapi/settings_endpoint.py
```diff
@@ -89,19 +89,7 @@ def parse_setting(self, section, option, value):
...
-                    # Do not interfere with credit mining downloads
-                    continue
-                elif option == "seeding_mode":
-                    download.config.set_seeding_mode(value)
-                elif option == "seeding_time":
-                    download.config.set_seeding_time(value)
-                elif option == "seeding_ratio":
-                    download.config.set_seeding_ratio(value)
-        elif section == 'credit_mining' and option == 'enabled' and \
+        if section == 'credit_mining' and option == 'enabled' and \
```
egbertbouman: Shouldn't we be setting the download defaults here? Otherwise, nothing will happen.

ichorid: No, we can leave that out. All settings are automatically copied by
https://github.com/Tribler/tribler/blob/2cc456f4c3f8d4e94cd8acfda8235fedba535ef8/Tribler/Core/Modules/restapi/settings_endpoint.py#L83-L86

I've checked it out, it works as expected.

PR #4978, opened 2019-11-28
Fix unicode escape chars in download dialog filenames

### LINKED ISSUE #4977: Start download dialog: wrong characters encoding in torrent files list
release-7.4.0 Windows an Linux.
When opening a torrent that contains files with non-Latin character names, the names are displayed as unicode codes.

Screenshot from 2019-11-28 15-44-22

ichorid:
Fixes #4977


ichorid:
oops, wrong branch...

Issue #4988, opened 2019-12-02
TrustGraph object has no attribute node


xoriole:
Reported on `Tribler 7.4.0-exp2`
![trust-graph-issue-2019-12-02](https://user-images.githubusercontent.com/1442867/69945957-f09dd600-14ea-11ea-881b-ce25c5ddcae3.png)
![trust-graph-issue-2019-12-02-2](https://user-images.githubusercontent.com/1442867/69945960-f3003000-14ea-11ea-932f-a1a2bd8c3fb1.png)

xoriole:
This is because of the update in `networkx` package from `2.3` to `2.4`. There are some incompatible changes.

PR #4989, opened 2019-12-02
Trustgraph networkx compatibility

xoriole:
Fixes https://github.com/Tribler/tribler/issues/4988

### DIFF
--- a/Tribler/Core/Modules/restapi/trustview_endpoint.py
+++ b/Tribler/Core/Modules/restapi/trustview_endpoint.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 import logging
 import math
 from binascii import unhexlify
+from distutils.version import LooseVersion
 
 import networkx as nx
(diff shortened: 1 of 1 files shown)


Tribler/Core/Modules/restapi/trustview_endpoint.py
```diff
@@ -18,6 +18,14 @@
 MAX_PEERS = 500
 MAX_TRANSACTIONS = 2500
 
+networkx_version = float('.'.join(nx.__version__.split('.')[:2]))
```
qstokkink: Why not use `LooseVersion`?

xoriole: There is no need to check for patch version in networkx so float comparison was sufficient. For consistency, you're right, it is a good idea to use LooseVersion. I have updated the PR to use LooseVersion.

PR #4990, opened 2019-12-02
Fix misleading VOD message

### LINKED ISSUE #4987: Downloads always start in VOD mode
Every download seems to start in VOD mode. This is reflected in the log, and can be seen as well in the download pieces map prioritizing the beginning of the torrent. This reduces download performance.

Instead, VOD should be used **only** if the user clicked the "Play" button.

```python
INFO    1575206079.86             LibtorrentMgr:671  (LibtorrentMgr)  start_download: Starting in VOD mode
```

ichorid:
Fixes #4987

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentMgr.py
+++ b/Tribler/Core/Libtorrent/LibtorrentMgr.py
@@ -668,7 +668,7 @@ class LibtorrentMgr(TaskManager):
                 self.tribler_session.update_trackers(tdef.get_infohash(), new_trackers)
             return d
 
-        self._logger.info('start_download: Starting in VOD mode')
+        self._logger.info('start_download: calling start_download_from_tdef')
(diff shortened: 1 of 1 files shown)


Tribler/Core/Libtorrent/LibtorrentMgr.py
```diff
@@ -666,9 +666,7 @@ def start_download(self, torrentfilename=None, infohash=None, tdef=None, dconfig
                 d.get_def().get_trackers_as_single_tuple()))
             if new_trackers:
                 self.tribler_session.update_trackers(tdef.get_infohash(), new_trackers)
-            return d
```
qstokkink: Was this removal intentional?

ichorid: oops. It slipped from another PR. That's why we do reviews...

PR #5011, opened 2019-12-23
F log exceptions

Solomon1732:
A pull request added after this: https://github.com/Tribler/tribler/pull/4895/#issuecomment-544174366

Bare exception was changed to `except Exception as e:` (see above comment for reasoning). Logging was also added (see comment above).


qstokkink:
ok to test

ichorid:
@Solomon1732 your commit is not based on the latest `devel`. That's why this PR accidentally includes more commits than just your changes. We cannot merge it as is. Please, rebase it on top of current `devel`.
To do it, first checkout the current `devel`, and then make sure your local `devel` is **exactly the same as the upstream one**. If you're not sure how to do this with `git` commands, it will be easier for you to just `git clone` the repo again in the new directory and start everything from a clean slate.

Solomon1732:
Closing pull request and starting from scratch

Solomon1732:
Replaced by #5013

PR #5012, opened 2019-12-23
Cleanup LibtorrentMgr

egbertbouman:


ichorid:
Does `add_torrent` really have to be `async`? I mean, there is only a single `await` in it, and it returns a `LibtorrentDownloadImpl`, that's `__init__` is not even `async` itself!

Async is a really virulent pattern. I would prefer terminating `async` call chains as early as possible and keeping `async` stuff to a necessary minimum.

egbertbouman:
@ichorid Your first remark makes no sense to me. The `await` is there to ensure we remove any previously started metainfo downloads with the same infohash as the new download.

Yes, `async` is a virulent pattern, and you may remember I've complained about that a lot these last few weeks. Since a seem very keen on removing it for the `start_download`, I'll take another look.

ichorid:
Is it finished?

egbertbouman:
No, not yet.

PR #5013, opened 2019-12-23
Log exceptions

Solomon1732:
Added logging to exceptions.

Exceptions are now `except Exception` instead of bare exceptions to prevent catching errors such as `SystemExit`. This prevents unexpected behavior.

Replaces #5011.


ichorid:
ok to test

PR #5014, opened 2019-12-23
READY: F handle resources with statement

Solomon1732:
Closes #4890 and replaces #4895

PR #5018, opened 2019-12-24
Handle db with conn manager

Solomon1732:
As mentioned in #5014 , this PR is about migrating sqlite3 db connections to be used with "with" statements instead of being handled manually.

### DIFF
--- a/Tribler/Core/Upgrade/db72_to_pony.py
+++ b/Tribler/Core/Upgrade/db72_to_pony.py
@@ -1,4 +1,5 @@
 import base64
+import contextlib
 import datetime
 import logging
 import os
@@ -35,25 +36,30 @@ def final_timestamp():
 
 
 class DispersyToPonyMigration(object):
-    select_channels_sql = "SELECT id, name, dispersy_cid, modified, nr_torrents, nr_favorite, nr_spam " \
-                          + "FROM Channels " \
-                          + "WHERE nr_torrents >= 3 " \
-                          + "AND name not NULL;"
-
+    select_channels_sql = (
+        "SELECT id, name, dispersy_cid, modified, nr_torrents, nr_favorite, nr_spam "
+        "FROM Channels "
+        "WHERE nr_torrents >= 3 "
+        "AND name not NULL;"
+    )
     select_trackers_sql = "SELECT tracker_id, tracker, last_check, failures, is_alive FROM TrackerInfo"
 
-    select_full = "SELECT" \
-                  " (SELECT ti.tracker FROM TorrentTrackerMapping ttm, TrackerInfo ti WHERE " \
-                  "ttm.torrent_id == t.torrent_id AND ttm.tracker_id == ti.tracker_id AND ti.tracker != 'DHT' " \
-                  "AND ti.tracker != 'http://retracker.local/announce' ORDER BY ti.is_alive ASC, ti.failures DESC, " \
-                  "ti.last_check ASC), chs.dispersy_cid, ct.name, t.infohash, t.length, t.creation_date, " \
-                  "t.torrent_id, t.category, t.num_seeders, t.num_leechers, t.last_tracker_check " \
-                  "FROM _ChannelTorrents ct, Torrent t, Channels chs WHERE ct.name NOT NULL and t.length > 0 AND " \
-                  "t.category NOT NULL AND ct.deleted_at IS NULL AND t.torrent_id == ct.torrent_id AND " \
-                  "t.infohash NOT NULL AND ct.channel_id == chs.id"
-    select_torrents_sql = " FROM _ChannelTorrents ct, Torrent t WHERE " + \
-                          "ct.name NOT NULL and t.length>0 AND t.category NOT NULL AND ct.deleted_at IS NULL " + \
-                          " AND t.torrent_id == ct.torrent_id AND t.infohash NOT NULL "
+    select_full = (
+        "SELECT"
+        " (SELECT ti.tracker FROM TorrentTrackerMapping ttm, TrackerInfo ti WHERE "
+        "ttm.torrent_id == t.torrent_id AND ttm.tracker_id == ti.tracker_id AND ti.tracker != 'DHT' "
+        "AND ti.tracker != 'http://retracker.local/announce' ORDER BY ti.is_alive ASC, ti.failures DESC, "
+        "ti.last_check ASC), chs.dispersy_cid, ct.name, t.infohash, t.length, t.creation_date, "
+        "t.torrent_id, t.category, t.num_seeders, t.num_leechers, t.last_tracker_check "
+        "FROM _ChannelTorrents ct, Torrent t, Channels chs WHERE ct.name NOT NULL and t.length > 0 AND "
+        "t.category NOT NULL AND ct.deleted_at IS NULL AND t.torrent_id == ct.torrent_id AND "
+        "t.infohash NOT NULL AND ct.channel_id == chs.id"
+    )
+    select_torrents_sql = (
+        "FROM _ChannelTorrents ct, Torrent t WHERE "
+        "ct.name NOT NULL and t.length>0 AND t.category NOT NULL AND ct.deleted_at IS NULL "
+        " AND t.torrent_id == ct.torrent_id AND t.infohash NOT NULL"
+    )
 
     def __init__(self, tribler_db, notifier_callback=None, logger=None):
         self._logger = logger or logging.getLogger(self.__class__.__name__)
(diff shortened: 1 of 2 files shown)


devos50:
ok to test

Solomon1732:
The tests seem to fail since I changed the flow of logic.

In the following functions the db connections are _not_ closed at all paths:
```python
def old_db_version_ok(old_database_path):
    # Check the old DB version
    connection = sqlite3.connect(old_database_path)
    with connection:
        cursor = connection.cursor()
        cursor.execute('SELECT value FROM MyInfo WHERE entry == "version"')
        version = int(cursor.fetchone()[0])
        if version == 29:
            return True
    connection.close()
    return False

# Some code...

def new_db_version_ok(new_database_path):
    # Let's check if we converted all/some entries before
    connection = sqlite3.connect(new_database_path)
    with connection:
        cursor = connection.cursor()
        cursor.execute('SELECT value FROM MiscData WHERE name == "db_version"')
        version = int(cursor.fetchone()[0])
        if version != CURRENT_DB_VERSION:
            return False
    connection.close()
    return True
```

These two functions are called in [this](https://github.com/Tribler/tribler/blob/devel/Tribler/Test/Core/Upgrade/test_db72_to_pony.py#L106) test and [this](https://github.com/Tribler/tribler/blob/devel/Tribler/Test/Core/Upgrade/test_db72_to_pony.py#L150) test at lines 118 and 163 respectively. Since the second assertion at each test tests the path that returns false, the db connection is never closed, and thus closing it through `contextlib.closing` changes the flow of logic, causing the tests to fail.

While I can leave the two functions in db72_to_pony.py as they are, personally leaking db connections is something that I find ill-advised 😅 I can change the tests to fit a new flow of logic, but what new flow and how to change the tests I don't know 😅 

Don't worry about the commit mess. I'll squash it when I'm done 😄

Solomon1732:
Found my bug: closing does not commit changes.
https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.close

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -75,56 +76,53 @@ def initialize(self, mds):
...
-                                 "public_key": b"",
-                                 "timestamp": final_timestamp(),
-                                 "origin_id": 0,
-                                 "size": 0,
-                                 "subscribed": False,
-                                 "status": LEGACY_ENTRY,
-                                 "votes": -1,
-                                 "num_entries": int(nr_torrents or 0)})
-        connection.close()
+        with contextlib.closing(sqlite3.connect(self.tribler_db)) as connection, connection:
```
ichorid: Why is there `as connection, connection`? What kind of idiom is this? I don't get it.

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -134,89 +132,88 @@ def get_old_torrents_count(self, personal_channel_only=False):
...
-        connection = sqlite3.connect(self.tribler_db)
-        cursor = connection.cursor()
-        cursor.execute("PRAGMA temp_store = 2")
-
-        personal_channel_filter = ""
-        if self.personal_channel_id:
-            personal_channel_filter = " AND ct.channel_id " + \
-                                      (" == " if personal_channel_only else " != ") + \
-                                      (" %i " % self.personal_channel_id)
+        with contextlib.closing(sqlite3.connect(self.tribler_db)) as connection, connection:
```
ichorid: Again, why two `connection`s? Am I missing something?

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -134,89 +132,88 @@ def get_old_torrents_count(self, personal_channel_only=False):
...
-                        (not torrent_id or int(torrent_id) == 0) or \
-                        (not length or (int(length) <= 0) or (int(length) > (1 << 45))) or \
-                        (is_forbidden(name)) or \
-                        not name:
+                try:
+                    invalid_decoding = len(base64.decodestring(infohash.encode('utf-8'))) != 20
+                    invalid_id = not torrent_id or int(torrent_id) == 0
+                    invalid_length = not length or (int(length) <= 0) or (int(length) > (1 << 45))
+                    invalid_name = not name or is_forbidden(name)
+                    if invalid_decoding or invalid_id or invalid_length or invalid_name:
```
ichorid: Nice refactoring! This really makes the code more readable :+1:

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -430,22 +427,17 @@ def mark_conversion_finished(self):
 
 def old_db_version_ok(old_database_path):
     # Check the old DB version
-    connection = sqlite3.connect(old_database_path)
-    with connection:
+    with contextlib.closing(sqlite3.connect(old_database_path)) as connection, connection:
```
ichorid: Double `connection`?

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -430,22 +427,17 @@ def mark_conversion_finished(self):
...
-    return False
+    return True if version == 29 else False
 
 
 def cleanup_pony_experimental_db(new_database_path):
     # Check for the old experimental version database
     # ACHTUNG!!! NUCLEAR OPTION!!! DO NOT MESS WITH IT!!!
-    connection = sqlite3.connect(new_database_path)
-    with connection:
+    with contextlib.closing(sqlite3.connect(new_database_path)) as connection, connection:
```
ichorid: Again, double `connection`?

Tribler/Core/Upgrade/db72_to_pony.py
```diff
@@ -454,36 +446,30 @@ def cleanup_pony_experimental_db(new_database_path):
...
     # We're looking at the old experimental version database. Delete it.
     if delete_old_pony_db:
         os.unlink(new_database_path)
 
 
 def new_db_version_ok(new_database_path):
     # Let's check if we converted all/some entries before
-    connection = sqlite3.connect(new_database_path)
-    with connection:
+    with contextlib.closing(sqlite3.connect(new_database_path)) as connection, connection:
```
ichorid: Well, you get it...

Tribler/Test/Core/Upgrade/test_db72_to_pony.py
```diff
@@ -110,11 +111,9 @@ def test_old_db_version_check(self):
         # Wrong old database version
         old_db = os.path.join(self.session_base_dir, 'old.db')
         shutil.copyfile(OLD_DB_SAMPLE, old_db)
-        conn = sqlite3.connect(old_db)
-        with conn:
+        with contextlib.closing(sqlite3.connect(old_db)) as conn, conn:
```
ichorid: Also, use either `connection` or `conn` everywhere. If the stuff should co-exist in the same context, give names like `connection1`, etc.

Tribler/Test/Core/Upgrade/test_db72_to_pony.py
```diff
@@ -133,11 +132,9 @@ def test_cleanup_pony_experimental_db(self):
         mds.shutdown()
         shutil.copyfile(pony_db, pony_db_bak)
 
-        connection = sqlite3.connect(pony_db)
-        with connection:
+        with contextlib.closing(sqlite3.connect(pony_db)) as connection, connection:
```
ichorid: Ok, I admit. In two years of full-time Python programming, I've never seen context managers unfold to tuples. And I have never seen tuples unfolding like this (i.e. to the same variable). If one field of the tuple should be discarded, the convention is to assign it to `_`, like `as _, connection`.

Tribler/Test/Core/Upgrade/test_db72_to_pony.py
```diff
@@ -153,13 +150,13 @@ def test_new_db_version_ok(self):
...
         self.assertTrue(new_db_version_ok(pony_db))
-
-        connection = sqlite3.connect(pony_db)
-        with connection:
-            cursor = connection.cursor()
-            cursor.execute("UPDATE MiscData SET value = 12313512 WHERE name == 'db_version'")
-        connection.close()
+        
+        # Wrong new database version
+        with contextlib.closing(sqlite3.connect(pony_db)) as connection, connection:
```
ichorid: Bla-bla, doulbe `connection` :speaking_head:

Tribler/Core/Upgrade/db72_to_pony.py
Solomon1732: What happens is that using a db connection with a bare "with" statement doesn't close it -- instead it commits whatever transaction it had in the block (and keeps the connection open). So, to close it, `contextlib.closing` is used. However, as the docs states (can be seen in the link I posted above), closing doesn't commit the transactions. Basically, two "with" blocks are needed to handle errors:
```python
# This closes the connection
with contextlib.closing(sqlite3.connect(self.tribler_db)) connection
    # This commits the transactions
    with connection:
        # Code
```
However, python allows to do this instead of nesting with statements:
```python
with f() as a, g() as b:
    # Code
```
And it closes them in the reverse order to their initial setup. Basically, it's syntax sugar to nested "with" statements.
So `with contextlib.closing(sqlite3.connect(self.tribler_db)) as connection, connection:` is equivalent to nesting them.

While going back to closing them manually is possible, it has a danger of a connection leaking when an exception happens since it won't be closed. This way allows for more robust error handling.

Tribler/Core/Upgrade/db72_to_pony.py
Solomon1732: Thank you 😄

Tribler/Test/Core/Upgrade/test_db72_to_pony.py
Solomon1732: It's not a tuple, it's syntax sugar for nested "with" statements. Full answer is in the first conversation 😅

Tribler/Test/Core/Upgrade/test_db72_to_pony.py
Solomon1732: Done

Solomon1732:
Is this inline comment removable? It seems to refer to something which doesn't exist anymore 
https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Upgrade/db72_to_pony.py#L86

Tribler/Core/Upgrade/db72_to_pony.py
ichorid: Ok, I see now. This is a corner case of syntactic sugar usage that just happen to look weird.

ichorid:
retest this please

ichorid:
As usual, don't forget to squash the commits!

Solomon1732:
> As usual, don't forget to squash the commits!

Right-o!

Tribler/Core/Upgrade/db72_to_pony.py
Solomon1732: Yes, exactly

ichorid:
The strings do not pass the test because our pylint tester is not aware of the new Py3 syntax. We still have to change our pylint tester to Py3. So, that's our fault.

Solomon1732:
Oh, ok. Got it. For now I'll leave them as-is so the tests will pass. Thanks 😄

ichorid:
Pylint is not a test. It only checks the syntax. The new `f`-syntax is cleaner. Please, bring it back. (Hint: use either `git rebase -i` or `git revert`)

Solomon1732:
Got it. I thought it was a blocker for you. Will bring them back.

ichorid:
retest this please

ichorid:
See, the testers are still unstable after py3 migration.

ichorid:
Well, it is a little bit faster to first run the tests in question on a local machine, you know...

Solomon1732:
I would, but for some reason I can't install the python3 dependencies. I don't know why 🤷‍♂

ichorid:
Hmm... Do you use Pycharm? If you do, you should use its "pyenv" feature to setup a custom interpreter for the project, with custom packages set.

Solomon1732:
As so happens I do use pycharm. I'll try setting up a custom environment

Solomon1732:
How do I run the tests? `run_nose.py`?

Solomon1732:
I'm tempted to give up on adding f-strings. I've been trying for two and a half hours to set up a local environment for testing, so far with no end in sight. I also have no lead as to what causes the errors to begin with -- when I look at the code I changed I can't find what's wrong.

ichorid:
Ok, so how did you try to set up the environment? Maybe, [this video](https://www.youtube.com/watch?v=HJ9bTO5yYw0) can help?

Solomon1732:
Well, seems like I solved the reason for the errors. Regardless, however, locally I'll set up an environment for the future.

Solomon1732:
Basically I created a virtualenv environment through pycharm and installed packages via that environment's pip script. I put the libtorrent.pyd package in that environment site-packages dir. For some reason the environment failed to pick it up.

ichorid:
Well, I completely forgot that you're on Windows! And on Windows you have to compile Libtorrent yourself, with binding it to proper Python environment. That's not trivial. Maybe, we could share our pre-compiled Libtorrent with external developers? Please, create an issue for that.

Solomon1732:
Opened 👍 #5020

PR #5023, opened 2019-12-25
READY: Remove lambda and file

### LINKED ISSUE #5022: Refractor util.py

devel
Change assigned lambda into a function. Also drop support for python 2.
##### Actual behavior: #####
Code assigns a lambda -- an anonymous function -- is assigned to a variable, thus making the whole point of it being a lambda moot. In addition, it supports python 2, which is a version that the codebase, afaik, seeks to stop supporting in favor of python 3

https://github.com/Tribler/tribler/blob/devel/Tribler/util.py

Solomon1732:
~~Before a lambda was assigned to a variable. This made the whole point of lambdas being anonymous functions moot. Instead python 2 support was dropped, and the lambda was refractored to a function.~~

~~In addition instead of checking the type inside the function, single dispatch is used.~~

File held only a single lambda, and said lambda was a one-liner. In addition it was used only in one place. Hence the file the lambda was in was removed and the one-liner copied to the location it was used in.

Fixes #5022

### DIFF
--- a/Tribler/Core/Modules/restapi/downloads_endpoint.py
+++ b/Tribler/Core/Modules/restapi/downloads_endpoint.py
@@ -25,7 +25,6 @@ from Tribler.Core.Utilities.unicode import ensure_unicode, hexlify
 from Tribler.Core.exceptions import InvalidSignatureException
 from Tribler.Core.simpledefs import DLMODE_VOD, DOWNLOAD, UPLOAD, dlstatus_strings
 from Tribler.pyipv8.ipv8.messaging.anonymization.tunnel import CIRCUIT_ID_PORT
-from Tribler.util import cast_to_unicode_utf8
 
 
 def _safe_extended_peer_info(ext_peer_info):
@@ -453,7 +452,7 @@ class DownloadsEndpoint(RESTEndpoint):
             elif state == "recheck":
                 download.force_recheck()
             elif state == "move_storage":
-                dest_dir = cast_to_unicode_utf8(parameters['dest_dir'])
+                dest_dir = parameters['dest_dir']
                 if not os.path.exists(dest_dir):
                     return RESTResponse({"error": "Target directory (%s) does not exist" % dest_dir})
                 download.move_storage(dest_dir)
--- a/Tribler/util.py
+++ b/Tribler/util.py
@@ -1,9 +1,0 @@
-"""
-This file contains various utility methods.
-"""
-import sys
-
-if sys.version_info.major > 2:
-    cast_to_unicode_utf8 = lambda x: "".join([chr(c) for c in x]) if isinstance(x, bytes) else str(x)
-else:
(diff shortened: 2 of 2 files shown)


Solomon1732:
Hmm, on second thought both `byte` and `bytearray` can be converted into `str` and the latter will decode it, so there's no need for single dispatch. At this point the utility becomes a thin wrapper for `str` and can be removed entirely. What do you think?

Solomon1732:
Also there's only one reference to the utility: this line
https://github.com/Tribler/tribler/blob/devel/Tribler/Core/Modules/restapi/downloads_endpoint.py#L456

Solomon1732:
Ignore my last two comments. They're irrelevant. Bytes can't be converted into string just by `str`

qstokkink:
ok to test

devos50:
add to whitelist

Solomon1732:
I've seen this exception returning in more than one PR. Should this be investigated perhaps?
https://gist.github.com/Solomon1732/ccea65e767ed78a32655cb9a8f613ce0
https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_win64_py3/593/artifact/output/test_runners_output/4.err/*view*/

ichorid:
Hmm... I see that the only place `cast_to_unicode` is used is `downloads_endpoint.py`. A better solution would be to completely remove the function. It was used only through the transitional period from Py2 to Py3.

In general, it is always a good idea to first look at how a function is used before refactoring it. Often, the best solution is to completely *delete* the thing.

Solomon1732:
I can remove the entire file, then, since `util.py` has only one utility – the removable lambda

ichorid:
That would be perfect! The best skill a programmer can possess is the ability to *delete* code :wink:

Tribler/Core/Modules/restapi/downloads_endpoint.py
```diff
@@ -453,7 +452,8 @@ def get_files_info_json(self, download):
             elif state == "recheck":
                 download.force_recheck()
             elif state == "move_storage":
-                dest_dir = cast_to_unicode_utf8(parameters['dest_dir'])
+                dest_dir = parameters['dest_dir']
+                dest_dir = dest_dir.decode() if isinstance(dest_dir, (bytes, bytearray)) else str(dest_dir)
```
ichorid: Could you try passing `parameters['dest_dir']` as is, without any conversions? After Py3 and AIOHttp all the parameters should be already unicode.

Solomon1732: Ah, ok. Gotcha. Will try

ichorid:
Squash the thing?

Solomon1732:
Ah, ok. I thought to keep them seperate. Will do, though 👍

PR #5025, opened 2019-12-28
READY: Integrated Pathlib

### LINKED ISSUE #4563: Solving 10 years of unicode problem?
Investigate possible libraries for universal support of unicode in Tribler. The library should work on all platforms we support (Windows, MacOS and Linux).

Note that there are issues in windows with availability of codecs and encoding. We already remap the codecs to support `'mbcs` and `cp65001`. So this should be tested with various encodings to avoid future issues.

I suggest creating a utility to encode and decode the python strings (and paths) and using it across Tribler. This utility could depend on some other library but Tribler should not directly depend on the library itself so we can switch the underlying library later if necessary.

Note that UTF-8 character set should be `utf8mb4` where a character is represented by one to four bytes instead of `utf8mb3` which misses some supplementary characters.

ichorid:
This replaces all the `os.path` calls with compatible `pathlib` methods wrapped in the compatibility module `path_util`.
Ideally, at some point, we would want to remove the compatibility object and do everything directly for semantic rigour and better-reading code. While it is possible to make a `sed` script to redo the bulk of the `join`s, a lot of tedious work should still be done by hand.

EDIT: well, I've done what could have been done with `sed` magic. The rest it up to the team to check manually :hand: :alien: :hand: 

Fixes #4563

### DIFF
--- a/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
+++ b/Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
@@ -5,7 +5,6 @@ Author(s): Arno Bakker, Egbert Bouman
 """
 import base64
 import logging
-import os
 import shutil
 import sys
 import time
@@ -18,14 +17,21 @@ from Tribler.Core.Config.download_config import DownloadConfig, get_default_dest
 from Tribler.Core.DownloadState import DownloadState
 from Tribler.Core.Libtorrent import check_handle, require_handle
 from Tribler.Core.TorrentDef import TorrentDef, TorrentDefNoMetainfo
-from Tribler.Core.Utilities.torrent_utils import get_info_from_handle
+from Tribler.Core.Utilities import path_util
+from Tribler.Core.Utilities.torrent_utils import commonprefix, get_info_from_handle
 from Tribler.Core.Utilities.unicode import ensure_unicode, hexlify
 from Tribler.Core.Utilities.utilities import bdecode_compat, succeed
 from Tribler.Core.exceptions import SaveResumeDataError
 from Tribler.Core.osutils import fix_filebasename
-from Tribler.Core.simpledefs import DLMODE_VOD, DLSTATUS_SEEDING, DLSTATUS_STOPPED, NTFY_ERROR, NTFY_FINISHED,\
-    NTFY_TORRENT
-from Tribler.pyipv8.ipv8.taskmanager import TaskManager, task
+from Tribler.Core.simpledefs import (
+    DLMODE_VOD,
+    DLSTATUS_SEEDING,
+    DLSTATUS_STOPPED,
+    NTFY_ERROR,
+    NTFY_FINISHED,
+    NTFY_TORRENT,
+)
+from Tribler.pyipv8.ipv8.taskmanager import TaskManager
 from Tribler.pyipv8.ipv8.util import int2byte
 
 if sys.platform == "win32":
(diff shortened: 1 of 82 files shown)


Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
```diff
@@ -934,6 +940,8 @@ def checkpoint(self):
                     b'info-hash': self.tdef.get_infohash()
                 }
                 self.post_alert('save_resume_data_alert', dict(resume_data=resume_data))
+            else:
+                print("either file does not exist or is not file")
```
qstokkink: Wait, why not use `logging` here?

Tribler/Core/Utilities/install_dir.py
```diff
@@ -16,7 +17,7 @@ def is_frozen():
     """
     try:
         # PyInstaller creates a temp folder and stores path in _MEIPASS
-        base_path = sys._MEIPASS
+        sys._MEIPASS
```
qstokkink: I guess this was assigned to a variable before as it could get optimized out if not assigned. We should compile this change in PyInstaller with optimization enabled and see what happens before removing.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
ichorid: hmm...

Tribler/Core/Utilities/install_dir.py
ichorid: I'll just change it back.

Tribler/Core/Libtorrent/LibtorrentDownloadImpl.py
qstokkink: Oh, seems like the print statement PR check is broken.

qstokkink: As well as the `External import checker`. They are still on Python 2 and dying.

qstokkink: I'll make an issue.

qstokkink: #5029

qstokkink:
Hmm.. the `/` syntax is 33% slower than just joining in a `Path.__init__`:

```python
>> import timeit
>> print(timeit.timeit('pathlib.Path("a") / pathlib.Path("b") / pathlib.Path("c") / pathlib.Path("d")', setup="import pathlib", number=100000))
14.19430541699694
>> print(timeit.timeit('pathlib.Path(pathlib.Path("a"), pathlib.Path("b"), pathlib.Path("c"), pathlib.Path("d"))', setup="import pathlib", number=100000))
10.705111750998185
```

Is it worth the cost for the readability?

ichorid:
> Hmm.. the `/` syntax is 33% slower than just joining in a `Path.__init__`:
> 
> ```python
> >> import timeit
> >> print(timeit.timeit('pathlib.Path("a") / pathlib.Path("b") / pathlib.Path("c") / pathlib.Path("d")', setup="import pathlib", number=100000))
> 14.19430541699694
> >> print(timeit.timeit('pathlib.Path(pathlib.Path("a"), pathlib.Path("b"), pathlib.Path("c"), pathlib.Path("d"))', setup="import pathlib", number=100000))
> 10.705111750998185
> ```
> 
> Is it worth the cost for the readability?

You should see for yourself. The readability is much better. Also, most of the "/" syntax is in the tests.
We should consider the performance thing, though, for our core routines, like torrent manipulation, etc. Still, I'm sure its performance hit is nothing compared to e.g. libtorrent procedures, `await`s, etc.

ichorid:
```python
CATEGORY_TEST_DATA_DIR = path_util.abspath(path_util.join(FILE_DIR, "data", "Tribler", "Core", "Category"))
CATEGORY_TEST_DATA_DIR = path_util.realpath(__file__).parent / "data" / "Tribler" / "Core" / "Category"
```
for instance.

Issue #4953, opened 2019-11-19
Loose coupling and sane folders structure


ichorid:
### The Mess

Our source code directories structure is a mess. As is our import/modules structures (I don't even want to start on the topic of inconsistent dir names capitalization...). For some things, like REST endpoints, sources are grouped by *function*:
```
Tribler/Core/Modules/restapi/*_endpoint.py
```
For other things by *component*:
```
Tribler/Core/Modules/MetadataStore/*
Tribler/Core/CreditMininig/*
```
(Note the inconsistent placement of modules!)

Another problem is the Config. Every component depends on the base Config object. When anyone wants to change something there, they have to add `get_*/set_*` methods directly into the Config class. As a result, the Config became a garbage dump. Also, it creates tight coupling and prevents submodules (like Anydex) from getting really independent (which defies the whole idea of having a submodule).

### The Solution
We can make Tribler more modularized, and component less dependent on each other by changing the source folders structure and moving component-specific sections of the Config class into the corresponding modules/components.

#### Directories
The directory structure could look like this:
```
Tribler/Core/Session.py
  ... (and friends)
Tribler/Core/Modules/
Tribler/Core/Modules/MetadataStore - (component dir)
Tribler/Core/Modules/MetadataStore/restapi/channels_endpoint.py - (channels endpoint)
Tribler/Core/Modules/MetadataStore/config.py (sub-config)
...
Tribler/Core/Modules/CreditMining
Tribler/Core/Modules/Anydex
Tribler/Core/Modules/TorrentChecker
Tribler/Core/Modules/ ... (etc)
```

#### Subconfigs
Subconfigs will enhance the root Config class as member objects, like this:

```python
class TriblerConfig(object):

    def __init__(self, config=None):
        ...
        self.metadata_store_config = MetadataStore.config(self.config)
```

After [joining LaunchMany with Session](https://github.com/Tribler/tribler/issues/4949), we'll be able to loosen the coupling between Tribler components and make their initialisation sequences unified and straightforward, like this:
```python
class Session(object):
    def start(config=None):
        if self.config.metadata_store_config.enabled:
            self.mds = MetadataStore(config=self.config.metadata_store_config)
```

In this scheme, `ComponentConfig` objects could look like this:
```python
class ComponentConfig(object):
        def __init__(self, config):
            self._config = config
        
        @property
        def enabled():
             return self._config["enabled"]
        
        @property.setter
        def enabled(val):
             self._config["enabled"] = val
```

[...]

PR #5032, opened 2019-12-31
READY: Total dirs renaming

ichorid:
## !!! ACHTUNG !!!!
[Git *does not* preserve history through file renames.](https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history) This is a known Git shortcoming. Locally, `git blame` should still work, but GitHub will forever show yours truly as the sole author of the moved files.

Ok, this is bad. However, considering the amount of refactoring Tribler codebase underwent through the last year, and in the wake of Blackifiyng the whole repository, this is not *that* bad. Otherwise, we're stuck with :dizzy_face:  :mushroom: dRuG-aDdiCt-StyLe :mushroom: :dizzy_face: non-pep8 naming forever.


qstokkink:
Could you give us a quick overview of the performed renaming in the OP*? The LOC change is rather.. daunting.

\* And any changes that were non-trivial.

ichorid:
> Could you give us a quick overview of the performed renaming in the OP*? The LOC change is rather.. daunting.
> 
> * And any changes that were non-trivial.

Well, the best way to look at the new structure is to browse it: https://github.com/ichorid/tribler/tree/rename_dir
The principal idea of this PR is described in #4953. At this moment, this is still just `git mv` renames. The next step is to fix the imports with `sed` magic and make sure everything works. The last step will be the hardest one, as I'll have to adjust the infrastructure (Jenkins builds, etc).

egbertbouman:
I have to say, compared to previous versions of your branch, I looks a lot better now that you've cleaned up the root a bit and moved to tests into separate directories.

> However, considering the amount of refactoring Tribler codebase underwent through the last year, and in the wake of Blackifiyng the whole repository, this is not that bad. 

Just a quick comment on code reformatting, before you make another massive pr. As far as I know blackifying the _entire_ codebase is still up for discussion and we merely agreed it was ok to blackify gui and metadata code.

ichorid:
> Just a quick comment on code reformatting, before you make another massive pr. As far as I know blackifying the _entire_ codebase is still up for discussion and we merely agreed it was ok to blackify gui and metadata code.

Don't worry, I'm not going to do it without *every* developer's consent :wink:

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
 jobs = auto 
-max-line-length = 120
```
qstokkink: Why was this changed?

ichorid: To be in line with the settings in Black formatter file. You're right, it is better to change it the other way around...

ichorid:
@xoriole, I consider this good and ready for merging.

ichorid:
`run_tribler.py` works from PyCharm if individual `src/` subdirs are marked as source dirs. This way it is compatible with potential pip packaging. Core tests pass, though Jenkins test runners still have to be adjusted.

xoriole:
The main execution file `run_tribler.py` and the core tests, all seems to work fine. I think we can merge this, and fix the test runners and other issues one by one.
Also with this change, documentation and update to (now outdated) ReadMe become quite important.

synctext:
:open_mouth:  LoC: "***+70,962  -71,584***"

ichorid:
> open_mouth LoC: "_**+70,962 -71,584**_"

Yep, I'm holding the record for the biggest PR now :stuck_out_tongue:

egbertbouman:
Yes, truly amazing.

Issue #5030, opened 2019-12-30
Clean up path_util compatibility wrapper


ichorid:
[Moving](https://github.com/Tribler/tribler/pull/5025) to `Pathlib` required creating a compatibility wrapper module `path_util` implementing some `os.path`-like functions. Cleaning it up will make the code cleaner and will reduce the probability of masking path-handling errors.

So, everyone is welcome to try removing `path_util.*` lines from the code and replace them with proper Pathlib methods!

Solomon1732:
I'm willing to tackle it

Solomon1732:
There is a reference in the code to a `path_util.splitdrive` function which I can't find anywhere. What should I use instead?
https://github.com/Tribler/tribler/blob/bcf03e0ea696ce0c704426a5b4120bc5c675d6de/src/tribler-core/tribler_core/utilities/osutils.py#L122

Solomon1732:
Can I get unassigned from this bug report? This is getting too much for me.

devos50:
Note that I did some work on this in #5573. This migration is not completely finished though.

devos50:
The `fix_win_long_file` method in our wrapper should probably be preserved since we are using that in a few locations.

qstokkink:
It seems like this has now been sufficiently addressed.

PR #5047, opened 2020-01-10
WIP: Clean path_util usages

Solomon1732:
Partially deals with #5030

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -21,7 +21,16 @@ import sys
 
 from mock import Mock as MagicMock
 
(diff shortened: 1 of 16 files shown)


Solomon1732:
For the record the reason this PR is getting bigger is because things are so tightly coupled around `path_util.py` that changing 3 files has repercussions on 30 more xD

Solomon1732:
I think my PR breaks something in the tests. I can't tell what, though. Some of the windows tests don't seem to run at all, for example. It makes it hard for me to debug without a stack trace.

Solomon1732:
I think I've bitten off more than I can chew…

PR #5048, opened 2020-01-10
Fix some problems with events endpoint

ichorid:
Notifier refactoring breaked events endpoint. Damn lambdas...


devos50:
Why are we merging PRs without review?

ichorid:
> Why are we merging PRs without review?

I already did this kind of thing this year, remember? That was fine by everyone. This PR is a fix for a non-trivial error that I personally introduced earlier (remind that no review or test caught it then). Most important, I am the only person who works on `devel` right now, and I don't dare to disturb the Usenix team at this critical moment.

PR #5066, opened 2020-01-20
READY: State directory backup mechanism

xoriole:
Implements forward copy and upgrade as backup mechanism on state directory. This supports rollback to any version (after 7.3.0).

### DIFF
--- a/src/tribler-core/tribler_core/tests/test_osutils.py
+++ b/src/tribler-core/tribler_core/tests/test_osutils.py
@@ -1,6 +1,8 @@
 import os
 import shutil
 import sys
+import tempfile
+from pathlib import Path
 
 from tribler_core.tests.tools.test_as_server import BaseTestCase
 from tribler_core.utilities import path_util
@@ -13,7 +15,6 @@ from tribler_core.utilities.osutils import (
     get_picture_dir,
     is_android,
 )
-from tribler_core.utilities.path_util import mkdtemp
 
 if os.path.exists('test_osutils.py'):
     BASE_DIR = '..'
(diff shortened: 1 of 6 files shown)


devos50:
Wait, shouldn't this go to the `release-7.4.0` branch? Or are we going to cherry-pick it later?

ichorid:
This should go into `release-7.4.0` branch.

ichorid:
@xoriole , let me remind you of how it should work.

### Normal upgrade process
Say, user had `v7.6` state directory before upgrade. It was located in
```.Tribler/7.6.0```
Then, user upgrades to `v7.6.1`. This upgrade does not change anything in the state directory. Nonetheless, Tribler 7.6.1 creates a **new** directory `.Tribler/7.6.1` and then looks backwards for Tribler directory with latest version name. In this case, it finds `.Tribler/7.6.0`. Then it copies files from `.Tribler/7.6.0` to `.Tribler/7.6.1` and runs upgrade scripts on `.Tribler/7.6.1`.
Note how `.Tribler/7.6.0` remains **completely untouched**!

If user would want to go back to `v7.6.0` now, they will just install `v7.6.0` and `v7.6.0` will just look for `.Tribler/7.6.0` and totally ignore all **future** versions.

### Pre 7.4 - versions
There still remains a problem of older Tribler versions that do not support this protocol. For these, *everything* inside `.Tribler` directory that does not look like a version prefix (e.g. `7.6.0`) we treat like it is a single old ("pre- 7.4.0") version. This way, even if `x.x->7.4.0` upgrade process will screw up, **the user will not have to do anything to revert the state directory to the old version**. In other words, things will "just work", if users reinstall old versions back!

xoriole:
@devos50 Yes, it should go to `release-7.4.0`. I'll create a PR for it. Putting this on hold.
@ichorid Thank you for your explanation. Please go through the PR again. It does create a copy of the current state directory and upgrade on it. This was the exact change made over the last implementation.

PR #5068, opened 2020-01-22
[7.4.0] READY: State directory backup mechanism

xoriole:
Implements forward copy and upgrade as backup mechanism on state directory. This supports rollback to any version (after 7.3.0).

### DIFF
--- a/Tribler/Core/Upgrade/version_manager.py
+++ b/Tribler/Core/Upgrade/version_manager.py
@@ -66,7 +66,7 @@ class VersionManager(object):
         if not self.session.config.get_version_backup_enabled():
             return
 
-        code_version = version_id if version_id else tribler_version.version_id
+        code_version = version_id or tribler_version.version_id
         last_usage_version = self.version_history.get("last_version", None)
 
         # If the code and state directory version are different
--- a/Tribler/Core/Upgrade/version_manager.py
+++ b/Tribler/Core/Upgrade/version_manager.py
@@ -1,0 +1,83 @@
+from __future__ import absolute_import, division
+
+import logging
+import os
+import shutil
+import time
+from pathlib import Path
+
+from Tribler.Core import version as tribler_version
+from Tribler.Core.Utilities import json_util as json
+from Tribler.Core.osutils import dir_copy
+from Tribler.Core.simpledefs import STATEDIR_CHANNELS_DIR, STATEDIR_CHECKPOINT_DIR, STATEDIR_DB_DIR, STATEDIR_WALLET_DIR
+
+VERSION_HISTORY_FILE = "version_history.json"
+
+
+class VersionManager(object):
+
+    def __init__(self, session):
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+        self.version_history = {"last_version": None, "history": {}}
+        self.version_history_path = Path(self.session.config.get_root_state_dir(), VERSION_HISTORY_FILE)
+        self.read_version_history()
+
+    def read_version_history(self):
+        if self.version_history_path.exists():
+            self.version_history = json.loads(self.version_history_path.read_text().strip())
+
+    def update_version_history(self, version_id):
+        self.version_history["last_version"] = version_id
+        self.version_history["history"][time.time()] = version_id
+        self.version_history_path.write_text(json.dumps(self.version_history))
+
+    def fork_state_directory(self, version_id, base_state_directory):
+        """
+        Creates a new state directory for the given version based on the given base state directory. It basically
+        copies important files and directories to the new state directory. If the new state directory already
+        exists and tribler config file is present, it does not overwrite anything on the directory.
+        """
+        new_state_dir = Path(self.session.config.get_state_dir(version_id=version_id))
+
+        # If only there is no tribler config already in the new state directory,
+        # then we assume that this is a new version so we copy the last state directory and upgrade it.
+        conf_path = new_state_dir / 'triblerd.conf'
+        if base_state_directory.exists() and not conf_path.exists():
+            # Copy the selected directories from the current state directory to the new one.
+            upgrade_dirs = [STATEDIR_DB_DIR, STATEDIR_CHECKPOINT_DIR, STATEDIR_WALLET_DIR, STATEDIR_CHANNELS_DIR]
+            src_sub_dirs = os.listdir(base_state_directory)
+            for upgrade_dir in upgrade_dirs:
+                if upgrade_dir in src_sub_dirs:
+                    dir_copy(base_state_directory / upgrade_dir, new_state_dir / upgrade_dir, merge_if_exists=True)
+
+            # Copy other important files: keys and config
+            extra_files = ['ec_multichain.pem', 'ecpub_multichain.pem', 'ec_trustchain_testnet.pem',
+                           'ecpub_trustchain_testnet.pem', 'triblerd.conf']
+            for extra_file in extra_files:
+                if os.path.exists(base_state_directory / extra_file):
+                    shutil.copy(base_state_directory / extra_file, new_state_dir / extra_file)
+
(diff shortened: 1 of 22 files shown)


xoriole:
retest this please

xoriole:
retest this please

xoriole:
retest this please

Tribler/Core/Upgrade/version_manager.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class VersionManager(object):
+
+    def __init__(self, session):
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self.session = session
+
+        self.version_history = {"last_version": None, "history": {}}
+        self.version_history_path = Path(self.session.config.get_root_state_dir(), VERSION_HISTORY_FILE)
```
ichorid: Is Pathlib really included in 7.4 dependencies?

Tribler/Core/Upgrade/version_manager.py
```diff
@@ -0,0 +1,83 @@
...
+                    shutil.copy(base_state_directory / extra_file, new_state_dir / extra_file)
+
+    def setup_state_directory_for_upgrade(self, version_id=None):
+        """
+        Runs the version migration  of the state directory if there was an update in the code version.
+        """
+        if not self.session.config.get_version_backup_enabled():
+            return
+
+        code_version = version_id if version_id else tribler_version.version_id
```
ichorid: `code_version = version_id or tribler_version.version_id`

Tribler/Core/Upgrade/version_manager.py
```diff
@@ -0,0 +1,83 @@
...
+        last_usage_version = self.version_history.get("last_version", None)
+
+        # If the code and state directory version are different
+        # then fork the state directory for the new code version and update the history file.
+        if last_usage_version != code_version:
+            self.fork_state_directory(code_version, self.get_state_directory(last_usage_version))
+            self.update_version_history(code_version)
```
ichorid: `self.update_version_history` should be only run after Tribler starts completely, e.g. at the end of the Session initialization sequence. Otherwise, a crash in upgrader can mess up the versions usage logic.

Tribler/Test/Core/Upgrade/test_upgrader.py
```diff
@@ -119,16 +121,44 @@ def test_backup_state_directory(self):
...
+        backup_dirs = [STATEDIR_DB_DIR, STATEDIR_CHECKPOINT_DIR, STATEDIR_WALLET_DIR, STATEDIR_CHANNELS_DIR]
+        for backup_dir in backup_dirs:
+            self.assertTrue(backup_dir in version_state_sub_dirs)
+
+        # Creating keypairs to test that these key pairs are also copied
+        self.upgrader.session.init_keypair()
 
+        # Now lets increase the version and test that upgrader does backup the directory and copy proper files
+        original_version = version.version_id
+        version.version_id = '7.5.0'
```
ichorid: It is better to use something like 7.99.0 for the test value, so it will never clash with real ones and is easier to spot as a test result.

Tribler/Test/Core/Upgrade/test_upgrader.py
```diff
@@ -119,16 +121,44 @@ def test_backup_state_directory(self):
...
+        self.assertTrue(os.path.exists(version_state_dir))
         version_state_sub_dirs = os.listdir(version_state_dir)
         backup_dirs = [STATEDIR_DB_DIR, STATEDIR_CHECKPOINT_DIR, STATEDIR_WALLET_DIR, STATEDIR_CHANNELS_DIR]
         for backup_dir in backup_dirs:
             self.assertTrue(backup_dir in version_state_sub_dirs)
+
+        self.assertTrue(len(list(version_state_dir.glob("*.pem"))) > 1)
+
+        # Restore the original version
+        version.version_id = original_version
```
ichorid: This kind of logic belongs in `tearDown` method. Changing the version in the test should either not affect the other tests, e.g. by creating a throw-away module instance, or guarantee that the value will be restored *even if the test fails*. That's what `tearDown` is for.

Tribler/Test/Core/Upgrade/test_version_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        self.version_manager = VersionManager(self.session)
+
+    def test_read_write_version(self):
+        # If there is no version history file, no information about last version is available
+        self.assertFalse(os.path.exists(self.version_manager.version_history_path))
+        self.assertIsNone(self.version_manager.version_history['last_version'])
+        self.assertEqual(self.version_manager.version_history['history'], {})
+
+        # Saving the version
+        self.version_manager.update_version_history('7.0.0-GIT')
```
ichorid: `7.0.0-GIT` is the default value for the version. This means that when you run `update_version_history` it could somehow get into the file and your test will falsely pass. Please, change the value to something completely improbable, like `17.123.0`. In general, test should *never* reuse any values found anywhere in the system or in the wild.

Tribler/Test/Core/Upgrade/test_version_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(self.version_manager.version_history['last_version'], '7.0.0-GIT')
```
ichorid: Also, it is more readable to use local variables for this purpose. Otherwise, the reader is confused if `7.0.0-GIT` is expected because it is the default value, or because has been *set* to this value.

Tribler/Test/Core/Upgrade/test_version_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        default_version_id = tribler_version.version_id
+        self.assertEqual(default_version_id, '7.0.0-GIT')
+
+        # With the latest implementation of the version manager, a separate state directory is created for each code
+        # version including the default version.
+        root_state_dir = self.session.config.get_root_state_dir()
+        self.version_manager.setup_state_directory_for_upgrade(version_id=default_version_id)
+        self.assertTrue(default_version_id, os.listdir(root_state_dir))
+
+        # Also check that there are all important directories and files in the created state directory
```
ichorid: :+1:

Tribler/Test/Core/Upgrade/test_version_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        default_version_state_dir = self.version_manager.get_state_directory(default_version_id)
+        self.assertTrue(os.path.exists(default_version_state_dir))
+        self.assertTrue(len(list(default_version_state_dir.glob("*.pem"))) > 1)
+        default_version_state_sub_dirs = os.listdir(default_version_state_dir)
+        backup_dirs = [STATEDIR_DB_DIR, STATEDIR_CHECKPOINT_DIR, STATEDIR_WALLET_DIR, STATEDIR_CHANNELS_DIR]
+        for backup_dir in backup_dirs:
+            self.assertTrue(backup_dir in default_version_state_sub_dirs)
+
+        # Assuming, new upgrade to be done
+        current_version = '7.5.0'
```
ichorid: Again, better to use never-to-exist-fantastic-version.

Tribler/Test/Core/Upgrade/test_upgrader.py
xoriole: By default, version.version_id is always 7.0.0-GIT on the code. The original_version is used to reset to this GIT so it will never clash with any released version so 7.5.0 or anything else should be fine.

Tribler/Core/Upgrade/version_manager.py
xoriole: This need not be applied at the end of session initialization because the version in the `version_history` file is independent of the database version. The SQL and other upgrade scripts will still run the forked state directory as earlier. Also, we don't try to fork the state directory if there is a Tribler config already present in the versioned directory so updating version history at the end of the session initialization will not help.

PR #5076, opened 2020-01-27
WIP: Clean path_util usages

Solomon1732:


Solomon1732:
Closing since changing one file seems to require changing other files, leading to an ever-growing PR.

PR #5081, opened 2020-01-29
Delete unloadable checkpoint files

ichorid:


ichorid:
Better to handle it some other way...

PR #5153, opened 2020-02-21
Hacky fix for config upgrading

ichorid:


ichorid:
retest this please

Tribler/Test/Core/Upgrade/test_upgrader.py
```diff
@@ -67,8 +67,10 @@ def test_upgrade_pony_db_6to7(self):
         old_database_path = os.path.join(self.session.config.get_state_dir(), 'sqlite', 'metadata.db')
         shutil.copyfile(OLD_DB_SAMPLE, old_database_path)
 
+        print(self.session.config.get_chant_channels_dir())
```
qstokkink: Seems our print statement checker is broken.

devos50: In Jenkins, we have a separate pipeline for Tribler 7.4 PRs. It seems that all these jobs are still based on Python 2:

```
  File "f40e35cc02fdd11df7b6b830f9ca3a35/print_extractor.py", line 23
    print(f"print at line {subnode.lineno}, col {subnode.col_offset} in file {file_path}")
                                                                                        ^
SyntaxError: invalid syntax
```

Should we attempt to fix these Jenkins jobs? I think these print statements should be detected when making a PR, and not enter the main code base. Unfortunately, it might take some time before we can remove these old Jenkins jobs since 7.5 is not ready yet.

ichorid: :man_facepalming: 

7.4 is EOL, I hope. So let's not waste more time on it and instead focus on the future. I agree, we should fix our print statements checker. For 7.5.

PR #5158, opened 2020-02-22
GigaChannel RemoteQuery Community

ichorid:
RemoteQuery Community's purpose is to serve remote DB queries for metadata stored by GigaChannel. The community remotely sends JSON-serialized arguments for get_entries method of MetadataNode ORM class. The receiving side processes the query and sends back the results. Eventually, this community should replace the first-generation GigaChannel Community, as RemoteQuery Communtiy is more general.

Note that this community does no actions on its own now. The plan is to first test it as a part of an experimental release and then add an aggressive walker that will be used to do fast bootstraps of Tribler Channels system as stated in #3615. Also, the current version imposes no limits on query size on the receiving side, which is fine for experiments but unacceptable for the final release.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
@@ -7,8 +7,6 @@ from ipv8.lazy_community import lazy_wrapper
 from ipv8.messaging.lazy_payload import VariablePayload, vp_compile
 from ipv8.peer import Peer
 
-from pony.orm import CacheIndexError, TransactionIntegrityError, db_session
-
 from tribler_core.modules.metadata_store.orm_bindings.channel_metadata import entries_to_chunk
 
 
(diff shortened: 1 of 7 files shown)


src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -0,0 +1,87 @@
...
+
+from ipv8.community import Community
+from ipv8.lazy_community import lazy_wrapper
+from ipv8.messaging.lazy_payload import VariablePayload
+from ipv8.peer import Peer
+from ipv8.requestcache import RequestCache
+
+from tribler_core.modules.metadata_store.orm_bindings.channel_metadata import entries_to_chunk
+
+minimal_blob_size = 200
```
qstokkink: Some preemption for testing and the future: please put these global variables into some sort of "settings" object (like https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/trustchain/settings.py).

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -0,0 +1,87 @@
...
+maximum_payload_size = 1300
+max_entries = maximum_payload_size // minimal_blob_size
+max_query_peers = 5
+
+
+class RemoteSelectPayload(VariablePayload):
```
qstokkink: Adding `@vp_compile` to your `VariablePayload` definitions makes them much faster. See also https://py-ipv8.readthedocs.io/en/latest/reference/serialization/ .

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    master_peer = Peer(
+        unhexlify(
+            "4c69624e61434c504b3a667b8dee4645475512c0780990cfaca234ad19c5dabcb065751776"
+            "b75a4b4210c06e2eb4d8bbf4a775ed735eb16bbc3e44193479ad7426d7cd1067807f95b696"
+        )
+    )
+
+    SELECT_REQUEST = 1
```
qstokkink: What I do lately is add a `msg_id` field to the variable payloads instead of using community class fields.

Something like:

```python
@vp_compile
class SelectResponse(VariablePayload):
    format_list = ['I', 'raw']
    names = ['id', 'raw_blob']
    msg_id = 2

[...]

self.ezr_pack(SelectResponse.msg_id, SelectResponse(request.id, data)
```

It's not shorter, but it keeps everything in the same place (and if I add a shortcut for it in the future you're immediately compatible 😃 ).

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
ichorid: Good point. However, I would prefer to keep it simple for now, so how about just moving these into `__init__` instance variables?

qstokkink: Sure, that's a start.

PR #5160, opened 2020-02-25
Bring startup sequence and version manager back to sanity

ichorid:
Refactor `run_tribler.py` Core startup sequence to make Upgrader independent of Version Manager. Also, delete old 70->71 upgrade stuff.

Documentation comment from `version_manager.py`:
The main purpose of the Version Management mechanism is to provide a safe fallback for both users
and developers during upgrades of major/minor (non-patch) versions. The rules are:
1. Every (major.minor) X.Y version has a separate state directory inside the root state directory.
2. Patch versions (x.y.Z) do not get a separate state directory.
3. The history of Tribler versions is written into version_history.json file in the root state directory.
3. When a new Tribler version is launched, it looks into the root state directory and version_history.json to determine if copying the state from the previous version is necessary:
     a. if the last used version is the same version as the code version AND the corresponding state directory exists, do nothing.
     b. otherwise, look for the most recently used Tribler version that has lower major/minor version that has state directory and copy state files from it into the new state directory named according to code version.

If there already exists a state directory for the code version (the same major/minor numbers), but the history file indicates that the most recent used Tribler version with the same major/minor version has a lower patch version (x.y.<Z>),  we rename the existing dir to rename_DIRNAME_<timestamp>.
  This corresponds to a situation where the user tried some version of Tribler and then went back to the previous version, but now tries again the same series with a higher patch version. This is a most typical scenario in which the version management system comes into play.

Example:
   the user installed 7.5.0 and used it for some time, then tried to upgraded to 7.6.0 and failed. Then the user went back to 7.5.0. After 7.6.1 was released, the user tried upgrading again. When run, 7.6.1 looked at the version_history.json and detected that the last used version was 7.5.0. It tried to create the 7.6 directory but failed because 7.6 was already installed before. Then, version management procedure renamed the old 7.6 directory to 7.6_bak_<timestamp> and created 7.6 directory anew, and copied 7.5 contents into it.

In some sense, the system works as GIT does: it "branches" from the last-non conflicting version dir and "stashes" the state dirs with conflicting names by renaming them.

Note that due to failures in design pre-7.4 series and 7.4.x series get special treatment.


ichorid:
retest this please

PR #5164, opened 2020-02-28
READY: Check torrents on click in GUI

ichorid:


src/tribler-gui/tribler_gui/widgets/torrentdetailstabwidget.py
```diff
@@ -120,6 +120,8 @@ def update_with_torrent(self, index, torrent_info):
                 self.rest_request2.cancel_request()
             self.is_health_checking = False
         if torrent_info['last_tracker_check'] == 0:
+            if not self.healthcheck_timer.isActive():
+                self.on_check_health_clicked()
```
egbertbouman: Considering that we also restart the times, isn't this checking the health twice?

ichorid: Yes, it is. I've tried different approaches with it, even added a second timer... Checking twice is a bad drawback, but every other solution I've tried felt really bad in UI :frowning_face: 
I'll add an issue to solve double check problem.

ichorid: Oh no, it will not! `on_check_health_cliked` will only trigger the check it `is_health_checking` id `False`. And it is only set `False` by `update_with_torrent`. So, no double-checks. :v:

PR #5170, opened 2020-03-02
READY: Stream video files with aiohttp

egbertbouman:

PR #5172, opened 2020-03-02
READY: Fix include/exclude download files

### LINKED ISSUE #5169: Error on including/excluding files in a download
Reported by a user on Tribler forum for 7.5.0-exp1 and confirmed on `devel`.
When trying to include/exclude a file in the file list of a torrrent in the downloads list, Tribler produces an error message: "An error occurred during the request: ‘LibtorrentDownloadImpl’ object has no attribute ‘set_selected_file_indexes’."

ichorid:
fixes #5169

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/restapi/downloads_endpoint.py
@@ -434,7 +434,7 @@ class DownloadsEndpoint(RESTEndpoint):
             num_files = len(download.tdef.get_files())
             if not all([0 <= index < num_files for index in selected_files_list]):
                 return RESTResponse({"error": "index out of range"}, status=HTTP_BAD_REQUEST)
-            download.set_selected_file_indexes(selected_files_list)
+            download.config.set_selected_file_indexes(selected_files_list)
 
         if parameters.get('state'):
             state = parameters['state']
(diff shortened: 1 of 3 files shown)


src/tribler-core/tribler_core/modules/libtorrent/restapi/downloads_endpoint.py
```diff
@@ -434,7 +434,7 @@ def get_files_info_json(self, download):
             num_files = len(download.tdef.get_files())
             if not all([0 <= index < num_files for index in selected_files_list]):
                 return RESTResponse({"error": "index out of range"}, status=HTTP_BAD_REQUEST)
-            download.set_selected_file_indexes(selected_files_list)
+            download.config.set_selected_file_indexes(selected_files_list)
```
egbertbouman: The names are a bit messed up right now (something I intend to address in https://github.com/Tribler/tribler/pull/5170). Are u sure this works? Maybe it should be `LibtorrentDownloadImpl.set_selected_files`?

ichorid: It definitely works. I've checked it by hand several times by including/excluding files through the GUI. Everything was fine.

The error was exactly that: `set_selected_file_indexes` was moved to `config` everywhere but in the test and the endpoint. The test did not catch it because of extensive mocking :man_shrugging:

egbertbouman: That's unexpected. And Tribler actually downloads the right files?

ichorid: Yes

ichorid: Well, at least it is consistent in the GUI.

ichorid: Well, you were right. After waiting 'till the download finished, the thing did not have any effect at all. It seems I was just lucky to select exactly those files from the test download (some music album) that were downloaded the most :dizzy_face:

ichorid: I'll finish this one after you're done with `LibtorrentDownloadImpl`

Issue #5015, opened 2019-12-23
"New arrivals" feed


ichorid:
Currently, Tribler "Home" screen does not show anything useful, triggering [banner blindness](https://en.wikipedia.org/wiki/Banner_blindness) instead. That's a shame because it is literally the first thing the user sees when starting Tribler.

One important element of every social network that could be put on the main screen is the **new arrivals feed**. It typically looks like an infinitely scrollable list of objects of different kinds, each object providing simple controls to manipulate them. In our case that would be:
 * **popular torrents** received through Popularity Community;
 * **updates of subscribed channels** (number of new torrents in the channel since the last visit);
 * **new channels** discovered.
 
The corresponding GUI objects should include "instant satisfaction" controls to decide the fate of the shown object immediately (_a la_ "swipe right"): subscription / look inside for channels; download / watch later for torrents.

In time the same feed can be used for [collective editing](#4486) notifications.

qstokkink:
Maybe we should then remove the home screen and start on the discovered channels?

ichorid:
> Maybe we should then remove the home screen and start on the discovered channels?
That is one possibility, yes. I would still prefer something less overwhelming and more user-friendly and personalised. I guess we can remove the home screen and start from "Discovered" screen as the first step in that direction.

ichorid:
Related to #6214

PR #5181, opened 2020-03-05
Remove homepage from the GUI and fix the discovery screen

ichorid:
Also, disable Credit Mining by default, and properly hide its icon if it is disabled.
Also, use emojis for Channels and Folders indication.
Also, fix personal channel options displayed in wrong situationl.


ichorid:
> Seems fine, fixes #5015?

No, still does not truly fix it. It just removes the current starting screen. Tribler starts much faster now if you have a big database, as there is no costly query for random torrents.

#5015 should still be properly addressed later, some time after 7.5 I guess.

PR #5186, opened 2020-03-06
Fix selected files

ichorid:
Also, there seems to be a problem with `remove_torrent` in Libtorrent. Added a skipped test to fix it later.


ichorid:
retest this please

PR #5204, opened 2020-03-14
Fix memory leak in TorrentChecker

ichorid:
`session_stop_task_list` was used to track the coroutines for stopping tracker sessions. The list was never emptied which resulted in a memory leak of about 50 MB/hour.

This PR deprecates `session_stop_task_list` in favour of TaskManager, which stops the leak and makes the code cleaner.

### DIFF
--- a/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
+++ b/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
@@ -3,7 +3,7 @@ import logging
 import random
 import socket
 import time
-from asyncio import CancelledError, ensure_future, gather
+from asyncio import CancelledError, gather
 
 from ipv8.database import database_blob
 from ipv8.taskmanager import TaskManager, task
@@ -38,9 +38,6 @@ class TorrentChecker(TaskManager):
 
         self._should_stop = False
         self._session_list = {'DHT': []}
-
-        # Track all session cleanups
-        self.session_stop_task_list = []
 
         self.socket_mgr = self.udp_transport = None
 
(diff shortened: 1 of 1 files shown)


src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
```diff
@@ -85,15 +82,6 @@ def __init__(self, session):
 
         await self.shutdown_task_manager()
 
-        # kill all the tracker sessions.
-        # Wait for the defers to all have triggered by using a DeferredList
-        for tracker_url in self._session_list.keys():
-            for session in self._session_list[tracker_url]:
-                self.session_stop_task_list.append(session.cleanup())
```
egbertbouman: Doesn't removing this mean that we don't shutdown the TaskManager of each individual session?

ichorid: Each HTTPSession cleans itself on cancel and when it is done:
https://github.com/Tribler/tribler/blob/8fcba201280effb89dff424b557bbfa240abf16f/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py#L316

egbertbouman: I could be wrong, but looking at `connect_to_tracker` it seems that if there is a `CancelledError`, it doesn't call `clean_session`.
https://github.com/Tribler/tribler/blob/8fcba201280effb89dff424b557bbfa240abf16f/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py#L136-L147

ichorid: You're right, I'll fix it.

ichorid: added `.clean_session` call to `CancelledError`

PR #5210, opened 2020-03-16
Fix simplified personal channels view

ichorid:
Also, fix dialog option for playing a video showing for every torrent instead of video torrents only.


ichorid:
retest this please

PR #5217, opened 2020-03-17
Updated Ubuntu development docs

qstokkink:
I just did a fresh install on Ubuntu 19.10, this is what I required.


qstokkink:
Please ignore the spelling checker, it has found that `pyqtgraph` is not in the dictionary.. and it is correct.

PR #5222, opened 2020-03-18
Add Swagger support

egbertbouman:
This PR adds /docs to our REST API server that will serve the Swagger UI. From there, you will find a list of all API endpoints available in both Tribler and IPv8.

I've converted all the old doc strings to aiohttp_apispec decorators. I've also added some new documentation to a number of undocumented channels/metadata endpoints. However, since I'm not sure *exactly* what these functions do and what their input/output is, I've kept the documentation to a minimum. Despit this, I've probably made some mistakes here.

Swagger also supports API keys (you just need to click 'Authenticate' and copy-past the API key). The Swagger UI itself can be visited without the API key.

image


qstokkink:
Is this ready for review?

egbertbouman:
@qstokkink Yep, ready.

PR #5234, opened 2020-03-23
Fixed logger.yaml loading in frozen environments

### LINKED ISSUE #5230: logger.yaml not found in frozen environments
When starting `tribler.exe`/`Tribler.dmg`, the following error is printed (twice) in the logs:

```
Logger config not found in tribler-gui\tribler_gui\logger.yaml. Using default configs.
```

This leads to a less verbose format of our log output. Furthermore, it might hide valuable debug information. Could be related to #5229

devos50:
In a frozen environment, the path to `logger.yaml` is different than when running Tribler from source. We should check for this and adjust the path accordingly.

Fixes #5230

### DIFF
--- a/src/tribler-core/tribler_core/__init__.py
+++ b/src/tribler-core/tribler_core/__init__.py
@@ -21,5 +21,9 @@ def load_logger_config(log_dir):
     Loads tribler-core module logger configuration. Note that this function should be called explicitly to
     enable Core logs dump to a file in the log directory (default: inside state directory).
     """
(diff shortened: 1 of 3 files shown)


devos50:
I'm having some issues with PyInstaller and paths, it's going to take a little longer.

devos50:
retest this please

ichorid:
I don't know how hard would it be to test these lines of code because this can only be properly tested with deployment testing...

devos50:
I don't think it's worth testing these lines. If something is wrong with logging, we immediately find out by running Tribler (the changed lines in this PR are always covered).

PR #5235, opened 2020-03-23
Properly cancel future in BalanceRequestCache

### LINKED ISSUE #5220: Tribler crashes after 4-10 minutes
This happens both on Linux (witnessed myself many times) and on Windows (reported in 7.5.0-rc1). Just run Tribler, do nothing (or do anything), and after about 5 minutes it crashes.

```python
[PID:3451] 2020-03-18 00:43:50,235 - ERROR <tribler_window:107> root.on_exception(): Traceback (most recent call last):
  File "/home/vader/my_SRC/TRIBLER/tribler_ichorid/src/tribler-gui/tribler_gui/event_request_manager.py", line 114, in on_read_data
RuntimeError: Task was destroyed but it is pending!
```

Observations so far:
 * the problem has nothing to do with the recent introduction of `WeakValueDictionary`. As when I change it back to `dict` it happens still.
 * the problem is seemingly a [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug): enabling  `DEBUG` level of logging makes it go away (or I was just really lucky to not trigger in several dozens runs with debug enabled).
 * the problem does not manifest itself when running core-only through `run_tribler_headless` script.
 * the problem happens on both Linux and Windows system, both with Libtorrent 1.2.x and 1.1.x (as packaged with our installer)
 * the problem seems to be related to Tunnels, as the last log line before crash typically refers to Tunnels. Also, disabling Tunnels seems to prevent the problem from occurring.
 * the problem seems to be not related to GigaChannel Manager, Popularity Community and Torrent Checker, as the crash still happens with these disabled.

ichorid:
fixes #5220

### DIFF
--- a/src/tribler-core/tribler_core/modules/tunnel/community/caches.py
+++ b/src/tribler-core/tribler_core/modules/tunnel/community/caches.py
@@ -7,6 +7,7 @@ class BalanceRequestCache(NumberCache):
         super(BalanceRequestCache, self).__init__(community.request_cache, u"balance-request", circuit_id)
         self.circuit_id = circuit_id
         self.balance_future = balance_future
+        self.register_future(self.balance_future)
 
     def on_timeout(self):
         pass


qstokkink:
Oh wait, the tests didn't run.

devos50:
Still fixing the tests...

egbertbouman:
This will still still produce an error, but this should be fixed once https://github.com/Tribler/py-ipv8/pull/718 is merged.

qstokkink:
It's merged, please update the pointer.

devos50:
retest this please

devos50:
retest this please

ichorid:
updated the pointer

ichorid:
retest this please

PR #5236, opened 2020-03-23
Include full exception context in report

ichorid:
This includes the full `context` in the error report. Debugging complex asynchronous code is basically impossible without this.


ichorid:
retest this please

Issue #5229, opened 2020-03-23
Build from current devel branch not starting on Mac and Windows (32-bit and 64-bit)


devos50:
I'm getting the following error when starting `Tribler.dmg` or `tribler.exe`, after Tribler has been stuck on the loading screen for 60 seconds:

```
Traceback (most recent call last):
  File "tribler_gui/event_request_manager.py", line 136, in <lambda>
  File "tribler_gui/event_request_manager.py", line 78, in on_error
RuntimeError: Could not connect with the Tribler Core within 60 seconds
```

This is with a clean state directory.

devos50:
The actual error seems to be hidden by the following warning: `Logger config not found in tribler-core/tribler_core/logger.yaml. Using default configs.`. I think that the location of this file is different in a frozen environment.

devos50:
It also seems to fail on Windows, both 32-bit and 64-bit.

egbertbouman:
I hope I'm wrong here, but it could be related to aiohttp-apispec.

egbertbouman:
Yep, it was aiohttp-apispec. It wasn't able to find the static html/js files, and produced no error message. I'll make a PR.

devos50:
Nice find!

ichorid:
I've tested rc1 on Julia's PC before releasing it. There was no problem (used x64 version).

egbertbouman:
It's broken on devel only.

PR #5238, opened 2020-03-23
READY: Fix core not starting when running from binaries

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/5229.

I've only tested Windows 32/64 installers, as I don't have access to a Mac.

### DIFF
--- a/tribler.spec
+++ b/tribler.spec
@@ -5,6 +5,9 @@ import os
 import sys
 import shutil
 
+import aiohttp_apispec
+
+
 root_dir = os.path.abspath(os.path.dirname(__name__))
 src_dir = os.path.join(root_dir, "src")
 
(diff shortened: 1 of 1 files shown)


ichorid:
Is it ready for review?

egbertbouman:
@ichorid Let me quickly fix the videoplayer too, while I'm at it.

egbertbouman:
@ichorid Oops! I was right the first time. The aiohttp_apispec fix should go into devel, the other one should go into release-7.5. Will fix tomorrow.

egbertbouman:
@ichorid Should be OK now.

devos50:
I guess the `aio_httpspec` bug fix should be included in the 7.5 release? Should cherry-pick this on the `release-7.5` branch after merge?

ichorid:
> I guess the `aio_httpspec` bug fix should be included in the 7.5 release? Should cherry-pick this on the `release-7.5` branch after merge?

It should work the other way around: this change should be merged into `release-7.5` branch and then `release-7.5` branch should be merged into `devel`.

@egbertbouman , please rebase this PR to go into `release-7.5`. When RC2 is out we'll merge all the bugfixes from `release-7.5` into `devel`. The reason for this is to minimize the number of merge conflicts.

p.s.
Cherry-picking stuff is tolerable only for *really* important bugfixes that should go into *every* branch ASAP.

ichorid:
To be clear: if this problem is affecting only release-7.5, it should go onto release-7.5. If it affects *only* devel, it should go into devel.

egbertbouman:
@devos50 @ichorid The release branch does not use aiohttp_apispec, so this should go into devel.

ichorid:
Is it fine that `aiohttp_apispect` is used in ipv8 and mentioned in our `requirements.txt`, `dependencies.py`, etc. on `release-7.5` branch? How Tribler is even able to start without it?

egbertbouman:
Because of the IPv8 import, the `aiohttp_apispec` library is a dependency in both `release-7.5` and `devel`. No one is questioning that.

When you build an installer for 7.5 it does seems to include `aiohttp_apispec`, as it should. It's just not including the html/js files, which aren't needed in 7.5 anyway. We only started serving the Swagger UI after https://github.com/Tribler/tribler/pull/5222 got merged, and the `release-7.5` was created before that.

ichorid:
@egbertbouman , thanks for the detailed explanation! Merging now.

PR #5242, opened 2020-03-23
Fix crash on upgrader done

ichorid:
2 characters!
I guess that's my shortest PR yet...


qstokkink:
retest this please

ichorid:
> The lambda is indeed not necessary here as you can use the `self.upgrader_finished.emit` as dictionary value immediately +1

But then I won't set a record for shortest bugfix PR! :disappointed:

devos50:
@ichorid I'm sure that you will have an opportunity to break your record before the release of Tribler 8!

PR #5246, opened 2020-03-25
fix head request to push all streaming content.

hbiyik:
Kodi Player does bunch of head request before playing a web stream.

Currently rest server iterates the stream object and then handles the HEAD & GET request on the same handler.

This causes stuttering on first launch of the stream, this small fix handles HEAD requests more efficiently and prevents server to bottleneck and fixes the stuttering problem.

PS:
Also it may be beneficial to provide Content-Length to the HEAD response to initialize the Player Buffering Mechanism more efficiently, but i faced a problem when consequent HEAD requests are made, the 2nd HEAD response was not been sent to client by AOI server when content length is provided in HEAD response. This may be another bug.


qstokkink:
ok to test

ichorid:
I believe this should be merged into `release-7.5` branch instead.

hbiyik:
I am heavily modifying sream endpoint to work it fine with kodi. I think I should close this for now, not to cause confussion on your end.

Issue #5249, opened 2020-03-26
Problem on Streaming with Kodi


hbiyik:
Hello Tribler Team

Since tribler is getting more and more stable everyday i started kodi addon development.
https://github.com/hbiyik/repository.biyik/tree/master/plugin.program.tribler
But i have a problem when streaming an unfinished torrent with kodi player.

When kodi starts streaming the torrent, the torrent sets the state DLSTATUS_SEEDING and progress hits to %100 and the download does not download anymore but still keeps seeding. However the torrent is in reality halfly downloaded.

To better describe the problem let me explain how kodi player streams an http media.

1) Send a HEAD request to check if media is available and get basic meta info like content length and type
2) Send a 2nd HEAD request to either validate the previous HEAD request or just some bug in kodi. idk.
3) Start GET request with rage 0- and get the header information of the media, as long as header is received close the TCP session. Generally reset around 1-2 mb of data.
4) Start GET request with range 0- and play the media with the header information parsed from previous request. This will continuously download the media until playback is stopped.
5) After GET request in step 4 is sent, send another GET request in another thread with range %99-%100 of file. Basically the footer of the file to parse metadata of some specific formats.

For HEAD requests i fixed the issues at:
https://github.com/hbiyik/tribler/blob/fdc589488238c93ea9a4d0b9da69af1b9d2fe37b/src/tribler-core/tribler_core/modules/libtorrent/restapi/downloads_endpoint.py#L622

But i speculatively think that step 5 causes the problem I described above (DLSTATUS_SEEDING %100 issue). But not sure, i fiddled the code a lot but could not identify the root cause.

You can duplicate the problem with.
1) Install kodi >= 17 to a linux machine where tribler is installed. And run tribler and kodi at the same time
2) Download my repo at https://github.com/hbiyik/repository.biyik/blob/master/repository.biyik.dist/repository.biyik/repository.biyik-0.0.2.zip
3) On kodi, navigate to addons, install from zip file, allow unknown sources, browse the repo zip file and install.
4) On kodi, navigate to addons, install from repo, select "biyik kodi repo", program addons, and tribler, select install.
5) Then navigate to addons, programm addons, tribler.
6) Add a torrent from url or local file.
7) Go to downloads, downloading, right click the newly added torrent and select play.

Thanks for your help :)

devos50:
@egbertbouman could you have a look at this?

egbertbouman:
[...]

PR #5253, opened 2020-03-27
Fix torrent checker timeout value is sent from the rest endpoint as string 

hbiyik:
This commit fixes the bug where torrentchecker endpoint sends a string
type timeout value to the TorrentChecker class,

If this variable is string then a suppressed error accours in each query
with the hidden trace below. This causes all torrenct health checks
requested from the endpoint to fail, since the argument is passed over
querystrings and query string arguments always are string types.

With this change timeout value is used when arguemnt is passed and a
number.

```
File "/home/boogie/src/tribler/src/tribler-core/tribler_core/modules/torrent_checker/torrentchecker_session.py",
line 137, in connect_to_tracker
    async with
self._session.get(url.encode(\'ascii\').decode(\'utf-8\')) as response:
    File "/home/boogie/.local/lib/python3.7/site-packages/aiohttp/client.py",
line 1012, in __aenter__
    self._resp = await self._coro
    File "/home/boogie/.local/lib/python3.7/site-packages/aiohttp/client.py",
line 405, in _request
    handle = tm.start()
    File "/home/boogie/.local/lib/python3.7/site-packages/aiohttp/helpers.py",
line 530, in start
    if self._timeout is not None and self._timeout > 0:

```

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/restapi/metadata_endpoint.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/restapi/metadata_endpoint.py
@@ -12,6 +12,9 @@ from tribler_core.restapi.rest_endpoint import HTTP_BAD_REQUEST, HTTP_NOT_FOUND,
 from tribler_core.utilities.unicode import hexlify
 
 
+TORRENT_CHECK_TIMEOUT = 20
+
+
 class UpdateEntryMixin(object):
     @db_session
     def update_entry(self, public_key, id_, update_dict):
@@ -147,7 +150,13 @@ class MetadataEndpoint(MetadataEndpointBase, UpdateEntryMixin):
 
             :statuscode 404: if the torrent is not found in the database
         """
-        timeout = request.query.get('timeout', 20)
+        timeout = request.query.get('timeout')
+        if not timeout:
+            timeout = TORRENT_CHECK_TIMEOUT
+        elif timeout.isdigit():
+            timeout = int(timeout)
+        else:
+            return RESTResponse({"error": "Error processing timeout parameter '%s'" % timeout}, status=HTTP_BAD_REQUEST)
         refresh = request.query.get('refresh', '0') == '1'
         nowait = request.query.get('nowait', '0') == '1'
 
--- a/src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
@@ -3,6 +3,7 @@ from ipv8.keyvault.crypto import default_eccrypto
 from pony.orm import db_session
 
 from tribler_core.modules.metadata_store.orm_bindings.channel_node import COMMITTED, TODELETE, UPDATED
+from tribler_core.modules.metadata_store.restapi.metadata_endpoint import TORRENT_CHECK_TIMEOUT
 from tribler_core.modules.torrent_checker.torrent_checker import TorrentChecker
 from tribler_core.restapi.base_api_test import AbstractApiTest
 from tribler_core.tests.tools.base_test import MockObject
@@ -213,7 +214,7 @@ class TestTorrentHealthEndpoint(AbstractApiTest):
             await self.http_tracker.stop()
         await super(TestTorrentHealthEndpoint, self).tearDown()
 
-    @timeout(20)
+    @timeout(TORRENT_CHECK_TIMEOUT + 5)
     async def test_check_torrent_health(self):
         """
         Test the endpoint to fetch the health of a chant-managed, infohash-only torrent
@@ -228,7 +229,7 @@ class TestTorrentHealthEndpoint(AbstractApiTest):
             self.session.mds.TorrentMetadata(
                 infohash=infohash, title='ubuntu-torrent.iso', size=42, tracker_info=tracker_url, health=torrent_state
             )
-        url = 'metadata/torrents/%s/health?timeout=10&refresh=1' % hexlify(infohash)
+        url = 'metadata/torrents/%s/health?timeout=%s&refresh=1' % (hexlify(infohash), TORRENT_CHECK_TIMEOUT)
 
         # Initialize the torrent checker
         self.session.torrent_checker = TorrentChecker(self.session)
@@ -252,3 +253,12 @@ class TestTorrentHealthEndpoint(AbstractApiTest):
 
         json_response = await self.do_request(url + '&nowait=1')
         self.assertDictEqual(json_response, {u'checking': u'1'})
+
+    @timeout(TORRENT_CHECK_TIMEOUT)
+    async def test_check_torrent_query(self):
+        """
+        Test that the endpoint responds with an error message if the timeout parameter has a wrong value
+        """
+        infohash = b'a' * 20
+        await self.do_request("metadata/torrents/%s/health?timeout=wrong_value&refresh=1" % infohash,
+                              expected_code=400)
--- a/src/tribler-core/tribler_core/tests/tools/tracker/http_tracker.py
+++ b/src/tribler-core/tribler_core/tests/tools/tracker/http_tracker.py
@@ -33,7 +33,8 @@ class HTTPTracker(object):
         """
         Stop the HTTP Tracker, returns a deferred that fires when the server is closed.
         """
-        return await self.site.stop()
+        if self.site:
+            return await self.site.stop()
 
     async def handle_scrape_request(self, request):
         """
--- a/src/tribler-core/tribler_core/tests/tools/tracker/udp_tracker.py
+++ b/src/tribler-core/tribler_core/tests/tools/tracker/udp_tracker.py
@@ -96,4 +96,5 @@ class UDPTracker(object):
         """
         Stop the UDP Tracker, returns a deferred that fires when the server is closed.
         """
-        self.transport.close()
+        if self.transport:
+            self.transport.close()


qstokkink:
ok to test

qstokkink:
retest this please

devos50:
Shouldn't this go to the `release-7.5` branch?

ichorid:
This should definitely go into `release-7.5` branch

hbiyik:
do you want me to change the PR branch to release7.5 or are you jsut chatting?

qstokkink:
> do you want me to change the PR branch to release7.5

Yes, please rebase it on that branch.

synctext:
great bug fix thnx!

hbiyik:
ok, ill do the rebase when i get to access my environment

hbiyik:
rebased

src/tribler-core/tribler_core/modules/metadata_store/restapi/metadata_endpoint.py
```diff
@@ -147,7 +147,8 @@ def setup_routes(self):
 
             :statuscode 404: if the torrent is not found in the database
         """
-        timeout = request.query.get('timeout', 20)
+        timeout = request.query.get('timeout')
+        timeout = int(timeout) if timeout and timeout.isdigit() else 20
```
ichorid: Please, don't just silently ignore wrong input because it will hide usage errors in the future. Instead, make the endpoint produce an error (something like "Wrong request"), see our tests. Also, produce a test for it. It is easy to do it by copy-pasting the existing tests.

hbiyik: done.

hbiyik: which also makes me think, the test should also fail here because the test already provide timeout value in the query.

https://github.com/Tribler/tribler/blob/09fbd720a6c251ffa76bfb2d2560b60a6920994d/src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py#L231

was this test even passing? I am confused :S

hbiyik: Ah i think i found the root cause,

https://github.com/Tribler/tribler/blob/09fbd720a6c251ffa76bfb2d2560b60a6920994d/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py#L302

The returned res obejct should be iterated and exception types should not be returned as succesfull results. Otherwise all extentions in the coro will be examined as successful, this is why the test scripts didnt work in this case. @ichorid

ichorid: Well, TorrentChecker was always kind of a lost child... No one owns the code and no one wants to touch it without good reason. So, can you fix it? :wink:

hbiyik: Sure no problem, but where to redirect the exceptions and their tracebacks? Is logging.error ok?

hbiyik: I checked the code of torrent checker, there are some ugly stuff there, feels like i need to pull the forest itself to pull the banana.

I suggest to limit this pr rest end point limited. I will do a final check of the test script, i dont think it will work yet but it will cover the scenario of the changes. I will request a review when done hopefully in the afternoon.

ichorid: Exceptions should be raised as normal to trigger a bug report. However, incorrect input into REST endpoint should never raise an exception but instead, send an "error" response and log the error as a warning.

hbiyik:
the tests have been written in Python3.7 i could not execute the new one, but it should be fine now, can you give a peek @ichorid

ichorid:
Well, tests are failing...

hbiyik:
Ugh, then i need a new environment with python3.7 to run the tests and see whats goin on, is it ok to keep pr until weekend?

ichorid:
Well, we were going to release Tribler 7.5.0 this week, as RC3 is pretty stable now. You can still base your changes on `release-7.5` branch, as I'm completely sure that, like, 10 days after we produce the release we'll have to produce the hotfix release 7.5.1. Your changes will end up there if you'll be on time.

hbiyik:
@ichorid tested with python3.8, tests are working now

hbiyik:
there were no commit so merged back the head of release 7.5. Flake8 thing is now fixed

hbiyik:
is this rebasing thing for build checks? why rebasing when there is no conflict?

devos50:
@hbiyik I was hoping that that would save the print statement issue/uncovered lines in the tests but it doesn't.

src/tribler-core/tribler_core/modules/metadata_store/restapi/metadata_endpoint.py
```diff
@@ -12,6 +12,9 @@
 from tribler_core.utilities.unicode import hexlify
 
 
+default_timeout = 20
```
ichorid, line 15: Please, make this ALL_CAPITAL_LETTERS. Module-level constants should be all caps.

ichorid: Also, the variable name should reflect *what kind* of timeout that is. Like, `TORRENT_CHECK_TIMEOUT`.
These measures are necessary to prevent name clashes and enhance readability.

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
```diff
@@ -228,7 +229,7 @@ def setUpPreSession(self):
             self.session.mds.TorrentMetadata(
                 infohash=infohash, title='ubuntu-torrent.iso', size=42, tracker_info=tracker_url, health=torrent_state
             )
-        url = 'metadata/torrents/%s/health?timeout=10&refresh=1' % hexlify(infohash)
+        url = 'metadata/torrents/%s/health?timeout=%s&refresh=1' % (hexlify(infohash), default_timeout)
```
ichorid, line 232: why not use f-string here? We're Python 3.7.

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
```diff
@@ -252,3 +253,12 @@ def setUpPreSession(self):
 
         json_response = await self.do_request(url + '&nowait=1')
         self.assertDictEqual(json_response, {u'checking': u'1'})
+
+    @timeout(default_timeout)
+    async def test_check_torrent_query(self):
+        """
+        Test the endpoint to check if correct error message is processed if wrong timeout is provided
```
ichorid, line 260: I don't get what this description means. Please, be more clear. Something like:
"Test that the endpoint responds with an error message if the timeout parameter has a wrong value."

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
```diff
@@ -252,3 +253,12 @@ def setUpPreSession(self):
...
+        """
+        infohash = b'a' * 20
+        await self.do_request("metadata/torrents/%s/health?timeout=wrong_value&refresh=1" % infohash,
```
ichorid, line 263: f-string?

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
hbiyik: yeah you can or you can not, but you should understand that few lines above there is not such  idiom as wel.

For this lines context i dont see such technical need.

https://github.com/Tribler/tribler/blob/fd39697c59b4a67f063591ed9d2621cc1b614d51/src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py#L184

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
hbiyik: I think the sentence is fine and clear.

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
hbiyik: same as above

src/tribler-core/tribler_core/modules/metadata_store/restapi/metadata_endpoint.py
hbiyik: done.

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
ichorid: "if correct error message is processed" is passive voice. Turning in to active voice would read like: "something processes the correct error message". This means "something" (the system?) is doing something to the already existing "correct error message".

"To process" means "to do something to some existing thing". The endpoint does not "process" any "error messages". Instead, it processes requests. Then, it possibly *responds* with error messages.
:man_shrugging:

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
ichorid: This is not a strict requirement, but a good opportunity to modernise the code.

synctext:
> I checked the code of torrent checker, there are some ugly stuff there, feels like i need to pull the forest itself to pull the banana.

@hbiyik Apologies for putting you through this painful PR process. We need to '_volunteer_' somebody to fix this code and adopt responsibility at some point. Slowly we're getting there, probably less then another 15 years :laughing:

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
hbiyik: ok, done

src/tribler-core/tribler_core/modules/metadata_store/restapi/tests/test_metadata_endpoint.py
hbiyik: it would be more ideal to have a separate work package or PR for that.

hbiyik:
@synctext I would rather focus on https://github.com/Tribler/tribler/issues/5249 which will bring more benefit imho. However I want to close this PR first which fixes the visible part of the problem.

To manage it effectively I will work on my fork repo which has less strict rules to be productive from my time point of view.

PR #5256, opened 2020-03-28
WIP: Implement versioned statedirs to headless daemon and write config

hbiyik:
This commit implements versioned statedirs to headless daemon and writes the triblerd.conf if rest api is enabled.

I am planning to use this daemon as internal kodi server that will be
build for different platforms

### DIFF
--- a/src/tribler-core/run_tribler_headless.py
+++ b/src/tribler-core/run_tribler_headless.py
@@ -11,11 +11,13 @@ from asyncio import ensure_future, get_event_loop, sleep
 from datetime import date
 from socket import inet_aton
 
-from tribler_core.config.tribler_config import TriblerConfig
+from tribler_core.config.tribler_config import TriblerConfig, CONFIG_FILENAME
 from tribler_core.modules.process_checker import ProcessChecker
 from tribler_core.session import Session
-from tribler_core.utilities.osutils import get_appstate_dir
+from tribler_core.utilities.osutils import get_root_state_directory
 from tribler_core.utilities.path_util import Path
+from tribler_core.upgrade.version_manager import get_versioned_state_directory, fork_state_directory_if_necessary
+from tribler_core.version import version_id
 
 
 class IPPortAction(argparse.Action):
@@ -27,7 +29,7 @@ class IPPortAction(argparse.Action):
         ip, port = parsed.group(1), int(parsed.group(2))
         try:
             inet_aton(ip)
-        except:
+        except Exception:
             raise argparse.ArgumentError("Invalid server address")
 
         if not (0 < port < 65535):
@@ -68,12 +70,21 @@ class TriblerService(object):
         signal.signal(signal.SIGINT, lambda sig, _: ensure_future(signal_handler(sig)))
         signal.signal(signal.SIGTERM, lambda sig, _: ensure_future(signal_handler(sig)))
 
-        config = TriblerConfig(options.statedir or Path(get_appstate_dir(), '.Tribler'))
+        root_state_dir = Path(options.statedir) if options.statedir else get_root_state_directory()
+        if not root_state_dir.exists():
+            root_state_dir.mkdir()
+        elif root_state_dir.is_file():
+            print("Statedir is not a directory %s" % root_state_dir)
+            get_event_loop().stop()
+            return
+        fork_state_directory_if_necessary(root_state_dir, version_id)
+        state_dir = get_versioned_state_directory(root_state_dir)
+        config = TriblerConfig(state_dir, config_file=state_dir / CONFIG_FILENAME)
 
         # Check if we are already running a Tribler instance
         self.process_checker = ProcessChecker()
         if self.process_checker.already_running:
-            print("Another Tribler instance is already using statedir %s" % config.get_state_dir())
+            print("Another Tribler instance is already using statedir %s" % state_dir)
             get_event_loop().stop()
             return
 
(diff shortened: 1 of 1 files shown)


devos50:
ok to test

src/tribler-core/run_tribler_headless.py
```diff
@@ -82,6 +93,7 @@ def log_incoming_remote_search(self, sock_addr, keywords):
         if options.restapi > 0:
             config.set_http_api_enabled(True)
             config.set_http_api_port(options.restapi)
+            config.write()  # only write config when rest api is enabled
```
ichorid: From the user's standpoint, it is not obvious that the config should be written if REST API is enabled. Overall, we want the config to be read-only and command-line options to override the config options.

Please, remove this line.

src/tribler-core/run_tribler_headless.py
```diff
@@ -68,12 +70,21 @@ def log_incoming_remote_search(self, sock_addr, keywords):
         signal.signal(signal.SIGINT, lambda sig, _: ensure_future(signal_handler(sig)))
         signal.signal(signal.SIGTERM, lambda sig, _: ensure_future(signal_handler(sig)))
 
-        config = TriblerConfig(options.statedir or Path(get_appstate_dir(), '.Tribler'))
+        root_state_dir = Path(options.statedir) if options.statedir else get_root_state_directory()
```
ichorid: I see you're copy-pasting this stuff from `run_tribler.py`. Maybe there is a way to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this stuff by `import`ing it from `run_tribler.py`?

src/tribler-core/run_tribler_headless.py
hbiyik: how to get the rest apikey then? there is no option to push custom apikey to rest api key from run time arguments, afaik it is generated by the daemon and written stored in the config object.

The purpose is to receive the apikey so the users can connect.

I am referencing this comment

> # Immediately write the API key so other applications can use it

https://github.com/Tribler/tribler/blob/09fbd720a6c251ffa76bfb2d2560b60a6920994d/src/run_tribler.py#L82

src/tribler-core/run_tribler_headless.py
hbiyik: may be I can use run_tribler.py directly but instead of using environment variables like

https://github.com/Tribler/tribler/blob/09fbd720a6c251ffa76bfb2d2560b60a6920994d/src/run_tribler.py#L107

I need to add argparser directly and switch to loading core only with this argument options or-ed with environment variables. This way I can also push apikey as well.

This way i ca DRY it however i need to change run_tribler.py, which is kinda fundemental I prevented this in the first phase.

What do you think? If you think its cool then this PR goes to depths of hell :)

ichorid: I was thinking about unifying `run_tribler` with `run_tribler_headless` too. Note though, that `run_tribler` is a top-level script. It is not packaged alongside `tribler_core`. On the other hand, `run_tribler_headless` belongs to `tribler_core` package.

 The correct way to solve this is to make `run_tribler` use `run_tribler_headless` as an external script. The biggest problem is in packaging. Let's merge this PR as it is.

Nonetheless, I dare you to try doing it the proper way (i.e. running `run_tribler_headless` from `run_tribler`) :wink:

src/tribler-core/run_tribler_headless.py
ichorid: @egbertbouman , could you please comment on the API key usage?

src/tribler-core/run_tribler_headless.py
hbiyik: If you execuse me with the new test cases, i can do that, if there is already an existing test cases i can modify them though.

Projected files to modify:
Run_tribler.py
Run_tribler_headless.py
Triblerd.spec

ichorid: That's the problem - there are no real tests for these, except for deployment testers (which run only nightly and/or release builds).

If you figure out how to test these properly, that would be of great service to Tribler project!

src/tribler-core/run_tribler_headless.py
egbertbouman: The comment you're referring to is just a really ugly hack so that developers have easy access to the API key. You don't need to do this in `run_tribler_headless.py`.

By default the core does not use an API key. So, if you want to use one you have to set is yourself (using `config.set_http_api_key(apikey)`).

src/tribler-core/run_tribler_headless.py
hbiyik: Hmm, i think best approach i do a fork my own from devel and work this case on my own branch.

i can keep up with the code itself but integrating tests is beyond my resource availability.

src/tribler-core/run_tribler_headless.py
hbiyik: Ok, i understood the story, since this part is on production i thought it was ok to write api key, but on broader scale this is definetely a security leak to write pass to filesystem like that.

egbertbouman: I wasn't trying to say that you can't store API key in the config file, just that you don't need to write the config file here.

hbiyik: Yep, got it, config obj is on ram unless u write it. But headless does not have argument to pass it. Nevermind i think everyone are in the same page.

hbiyik:
gonna work on seperate branch with this. May in future we can merge. Closing this PR. @ichorid

PR #5257, opened 2020-03-29
WIP:Use uvloop for reactor

ichorid:


ichorid:
> Can we make sure that the `uvloop` is also executed in the unit tests? Using a different reactor in the tests and when running Tribler through the script might bite us later when debugging any error caused by the `uvloop` reactor.
> 


Good point! I was thinking about that too. I'll try to install the reactor in the tests base.

ichorid:
retest this please

ichorid:
There is no stable Windows implementation of `uvloop`. Closing this.

PR #5258, opened 2020-03-30
WIP: triblerd docker build sys for linux

hbiyik:
This script provides a sterile way of generating headless tribler daemon using containers.
The container is based on manylinux1 (https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy) abi compatability to satisfy broadrange of linux distrubtion systems.

Only dependecy generated by the executable is glibc v2.5 and zlin. zlib is alreay bundled in the executable. The complier used in he container is gcc 4.8.2.

 0x00000001 (NEEDED)                     Shared Library: [libdl.so.2]
 0x00000001 (NEEDED)                     Shared Library: [libz.so.1] (Bundled in the executable)
 0x00000001 (NEEDED)                     Shared Library: [libc.so.6]

The basis for the container can be found at (https://github.com/hbiyik/shared-manylinux1) which is a fork of original manylinux1 repo with python to be compiled shared libraries enabled.

To use the script you need docker installed inn your system. Then simply run the script.

First Results:
Drop that below nasty 45MB of binary to any linux distro and it will work.

https://github.com/hbiyik/repository.biyik/blob/master/plugin.program.tribler/bin/manylinux1-x86/triblerd?raw=true


hbiyik:
I was too sleepy to mention, of course not only kodi is the target here also all kind of embeddded platforms are a use case for this.

qstokkink:
ok to test

hbiyik:
Hello

I pushed this PR just to show the approach and I agree with the duplicate spec files are lame.

Do you think I should rely on https://github.com/Tribler/tribler/blob/09fbd720a6c251ffa76bfb2d2560b60a6920994d/src/run_tribler.py#L107 CORE_PROCESS env variable which is used for android (afaik) and not even related to PyInstaller (afaik) or should I create a new env variable like CORE_DAEMON. I prefer to have different variable just to be on the safe side.

Also I agree with the docker files but I will leave this task to the end which is very time demanding and cancerous for me.

I also need to create another docker for ARM probably with arm_v6 only that will work with all arms.

But i need to understand some versioning dependecies with openssl  & libtorrent.

Long story short this 2 are hard to make uniform for all bindings (including python itself as well) is there already agreed upon versions of them. Currently I am using libtorrent 1.1.5 which has issues with OpenSSL 1.1.1 and seem to work with OpenSSL 1.0.1 (at least for i386), also cryptography does not work with SSL 1.0.1 and needs SSL 1.1.1, so I would aprreciate either an update libtorrent may be 1.2.x and OpenSSL?

hbiyik:
Any comment on openssl and libtorrent versions?

PR #5268, opened 2020-04-02
Various minor GUI fixes

devos50:
In this PR, I made the following changes:
- Made sure that the width of all tables in the GUI match the window width.
- Moved the 'new channel' and 'channel options' button to the right of the screen.
- Renamed 'Personal channel root' to 'My channels'.
- Removed unnecessary margin around the discovered channels table.
- Customized the `QComboBox` dropdown and the 'new channel' buttons.
- Renamed 'results' to 'items' (the label left to the filter).

I also updated the contribution guidelines to include the rule that screenshots have to be attached when making GUI changes.

Schermafbeelding 2020-04-02 om 12 15 26

Schermafbeelding 2020-04-02 om 12 15 36

Schermafbeelding 2020-04-02 om 12 15 55

Schermafbeelding 2020-04-02 om 12 17 10

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -34,7 +34,7 @@ Reporting bugs
 Pull requests
 =============
 
-When creating a new Pull request, please observe the following:
+When creating a new Pull request, please take note of the following:
   * New features always go to ``devel``.
   * If there is an unreleased ``release-X.Y.Z`` branch, fixes go there.
   * Otherwise, fixes go to ``devel``.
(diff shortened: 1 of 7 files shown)


ichorid:
chann

devos50:
I've seen this issue before with this type of buttons. For some reason, the font size on Ubuntu does not adhere to the font size as specified in the ui file. I guess we should simply make the button bigger, but I think there are more buttons in the UI where the text content does not fit.

qstokkink:
I know the font size in the buttons follows DPI scaling, which also causes a lot of.. interesting.. font size issues on Windows.

Issue #4854, opened 2019-09-23
Add toggle button to control anonymity in the download dialog


ichorid:
There are two problems with our "Download torrent" dialog window:
 1. Anonymity control is done by a small checkbox that is hard to notice. Instead, the current anonymity mode should be instantly visible and toggleable.
 2. DHT requests go through the regular connection, which are [observable](https://iknowwhatyoudownload.com) (see #4021). Instead, if anonymity is on, the DHT request should go through the tunnels as well. But if the user toggles anonymity mode, the DHT request should be immediately sent again directly. This is non-intuitive with the checkbox.

The solution is to use a toggle button, like this:
![LHLiq](https://user-images.githubusercontent.com/2509103/65426842-59910f80-de11-11e9-82e5-e909bf97782c.gif)

One implementation of a toggle control in pure QT: https://stackoverflow.com/a/51825815/5553928

The button for anonymity should be "ON" by default, and the file list dialog should show a ticker with time till DHT request timeout. As soon as the timeout is reached, a suggestion to try the request again in non-anonymous mode should be shown to the user.

hbiyik:
for 2: %100 agree, I was trying to tweak out this behavoir for a while, but the code is still to complex for my mortal brain. tunnel dht please

ichorid:
Well, turned out tunneled DHT requests are very unreliable. This makes adding this toggle control top priority, so users who do not use anonymity will not suffer from bad DHT performance over the tunnels.

qstokkink:
This should be resolved now: anonymity is on by default and metadata retrieval when adding torrents should be tunneled.

PR #5288, opened 2020-04-12
Load metainfo directly when unchecking 'anonymous' checkbox

ichorid:
When unchecking :ballot_box_with_check:  "Download anonymously", trigger direct metainfo fetch, and vice-versa. Modify status label text accordingly to indicate in what mode the metainfo is being fetched currently.

Partially solves #4854

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/restapi/torrentinfo_endpoint.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/restapi/torrentinfo_endpoint.py
@@ -45,6 +45,15 @@ class TorrentInfoEndpoint(RESTEndpoint):
         """
 
         args = request.query
+
+        hops = None
+        if 'hops' in args:
+            try:
+                hops = int(args['hops'])
+            except ValueError:
+                return RESTResponse({"error": f"wrong value of 'hops' parameter: {repr(args['hops'])}"},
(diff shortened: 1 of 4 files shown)


devos50:
retest this please

src/tribler-core/tribler_core/modules/libtorrent/restapi/torrentinfo_endpoint.py
```diff
@@ -45,6 +45,14 @@ def setup_routes(self):
         """
 
         args = request.query
+
+        hops = None
+        if 'hops' in args:
+            try:
+                hops = int(args['hops'])
+            except ValueError:
+                return RESTResponse({"error": "wrong value of 'hops' parameter"}, status=HTTP_BAD_REQUEST)
```
devos50, line 54: Can we please add the erroneous value to the response message? I noticed that sometimes we just throw errors without giving an exact specification of what's wrong, so in order to make our life easier when we have to debug something, I argue it's helpful to also include the wrong value.

ichorid: Agree

ichorid:
> Also, please cover the missing `ValueError` code, it should be quite easy do to so with a unit test.

Done

PR #5304, opened 2020-04-26
Update run_market_helper.py

Dmole:


devos50:
ok to test

devos50:
Not sure why that mac test is failing...

PR #5311, opened 2020-05-05
Auto generate Swagger docs

egbertbouman:
This PR convinces readthedocs to auto-generate swagger.yaml for us.

### DIFF
--- a/doc/extract_swagger.py
+++ b/doc/extract_swagger.py
@@ -1,4 +1,6 @@
+import json
 from asyncio import get_event_loop
+from io import StringIO
 from unittest.mock import Mock
 
 import aiohttp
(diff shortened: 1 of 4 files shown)


doc/extract_swagger.py
```diff
@@ -8,21 +10,21 @@
...
     session = Mock()
     session.config.get_http_api_key = lambda: 'apikey'
     session.config.get_http_api_port = lambda: 8085
     api_manager = RESTManager(session)
     await api_manager.start()
 
+    fp = StringIO()
+    proto = aiohttp.web_protocol.RequestHandler(api_manager.site._runner._server, loop=get_event_loop())
+    proto.connection_made(Mock(is_closing=lambda: False, write=lambda b: fp.write(b.decode())))
+    proto.data_received(b'GET /docs/swagger.json HTTP/1.1\r\n'
```
qstokkink, line 23: I looked into alternatives for sending the HTTP plain text on this line:

```python
from io import BytesIO
from http.client import HTTPConnection

output = BytesIO()
con = HTTPConnection('localhost')
con.send = output.write
con.putrequest('GET', '/docs/swagger.json', True, True)
con.putheader('Connection', 'close')
con.endheaders()
proto.data_received(output.getvalue())
```

Though I'm not a fan of plain text HTTP GET requests, given the alternative, I think a plain text HTTP message is acceptable here. @devos50 your thoughts?

PR #5331, opened 2020-05-26
Add limited HTTP support to TunnelCommunity

egbertbouman:


src/tribler-core/tribler_core/modules/tunnel/community/triblertunnel_community.py
```diff
@@ -563,6 +578,95 @@ async def unload(self):
...
+        try:
+            response_decoded = lt.bdecode(response.split(b'\r\n\r\n')[1])
+        except (IndexError, RuntimeError):
+            response_decoded = None
+
+        if response_decoded is None:
+            self.logger.warning('Tunnel HTTP request not allowed')
+            return
+
+        num_cells = math.ceil(len(response) / MAX_HTTP_PACKET_SIZE)
```
qstokkink, line 624: Why doesn't this crash like DeepCode's report suggests?

egbertbouman: I've seen the warning, but in Python 3 `Math.ceil` seems to produce an integer (also see the [docs](https://docs.python.org/3/library/math.html#math.ceil)).

qstokkink: Accepted, checks out, weird though:

```python
import math
print(3/5, type(3/5))
> 0.6 <class 'float'>
print(math.ceil(3/5), type(math.ceil(3/5)))
> 1 <class 'int'>
```

According to the docs: "**If x is not a float**, delegates to x.__ceil__(), which should return an Integral value." The input to `math.ceil` is clearly a float though..

Issue #5348, opened 2020-06-02
Brainstorm ticket _The Plan_


synctext:
(as communicated by @ichorid on 31May 2020) **The Plan** ```Having said all that, I propose```:

- Short-term:
  - [x] Remove the entirely useless CreditMining subsystem ~~and "Trust Graph" panel from 7.5.~~
  - [x] ~~Release 7.5 w/o fixing the "0kb/s" bug (but fixing the remaining usability bugs).~~ Fixed
  - [ ] Postpone any activity resulting in increasing the codebase size until new team members arrive in September.
- Mid-term:
  - [ ] This Autumn, focus on DevOps and infrastructure, type annotations, modern Python development process.
  - [ ] Start doing "early access" builds with enabled "performance watchdog" sending logs automatically to get more users engagement (this will enable us to catch the "0kbs bug" and its kind).
  - [ ] Move AnyDex into a separate app with a web-based UI.
  - [ ] Review the necessity of every dependency in our codebase.
  - [ ] For 7.6, focus on building a working social score system that does not instantly piss off new users by showing them negative balance as our current token system does.
- Long-term:
  - [ ] Hire outsources to develop a proper UX/GUI.
  - [ ] For 2021, focus on building a working community-trust-score-based collective seeding system (e.g. people collectively seeding stuff from a subscribed channel).

hbiyik:
May be releasing with libtorrent 1.1.x branch is a safe approach for 0kbps bug, I also had some funny connection issues with libtorrent 1.2.x where I can not give concerete details, i just downgraded libtorrent 1.1.x and lived happily.

ichorid:
@hbiyik , I was considering this option. We figured out the 0kb/s bug was there at least since 7.4. There were no complains (though it does not mean there was no problem). Probably, the complains were triggered by problems with tunnel downloads (which were very real).

devos50:
> Start doing "early access" builds with enabled "performance watchdog" sending logs automatically to get more users engagement (this will enable us to catch the "0kbs bug" and its kind).

For this, I would recommend starting with a basic, separate Tribler debug build on all platforms, and add more advanced debugging features on the go. I think an important first step would be to include a libtorrent version with debug symbols enabled when building the debug version. Other non-invasive, yet useful debugging practices I can think of:
- Add a checkbox to the crash report dialog in the GUI to also send (parts of) the log to us. This requires minor modifications to the error reporter backend, which hasn't been touched in years.
[...]

PR #5350, opened 2020-06-03
Remove CreditMining feature

ichorid:
:hourglass: :skull: The time has come :skull:  :hourglass:


devos50:
Even though I'm in favour of removing it, did we reach consensus on its removal?

ichorid:
> Even though I'm in favour of removing it, did we reach consensus on its removal?

Considering @synctext  himself put it in #5348, and he did not cross the thing as he sadly did with the proposed removal of the trust graph screen, I would say yes, we reached the consensus.

devos50:
Ah, I see. Didn't notice that the ticket has been updated.

PR #5353, opened 2020-06-08
Added IPv8 health monitor to debug window

qstokkink:
output

This new tab shows how congested the main thread is, for debugging.

Disclaimer: could this have just been an overview of keys and values? Yes. In my defense, the average user will now also be able to visually spot when something is wrong with the core without any technical expertise.

### DIFF
--- a/src/tribler-gui/tribler_gui/debug_window.py
+++ b/src/tribler-gui/tribler_gui/debug_window.py
@@ -22,6 +22,7 @@ from tribler_gui.event_request_manager import received_events as tribler_receive
 from tribler_gui.tribler_request_manager import TriblerNetworkRequest, performed_requests as tribler_performed_requests
 from tribler_gui.utilities import format_size, get_ui_file_path
 from tribler_gui.widgets.graphs.timeseriesplot import TimeSeriesPlot
+from tribler_gui.widgets.ipv8health import MonitorWidget
 
 try:
     from meliae import scanner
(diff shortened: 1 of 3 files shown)


qstokkink:
I'm guessing the tunnel changes were not merged into devel yet.

qstokkink:
I'll mark this on hold until the `devel` branch get the latest fixes from `release-7.5` merged.

ichorid:
How about labelin the axes and adding a legend? Also, using a log scale could be a nice idea.

qstokkink:
> 
> 
> How about labelin the axes and adding a legend? Also, using a log scale could be a nice idea.

If we were publishing this, I agree. However, the purpose of this health tab is not to be a scientific plot. It is just for normal human beings to get a reading on how healthy their application core is: big irregular spikes bad, small frequent spikes good.

If you get a reading that shows irregular and/or big spikes*, it's time to hook up yappi and do a (manual) detailed rundown of I/O, memory and CPU use (for which users should seek our help).

*For which we _could_ hook up pattern recognition, but I don't really want to 😃

qstokkink:
Ideally, I would like to be able for the average/advanced users to _have fun_ in trying to break our core and point us to performance optimizations.

Trying to cause cardiac arrest (in software!) is more fun than seeing some random metric increase or decrease.

synctext:
Nice visualisation of IPv8 main thread. We could use some p95 and p99 testing in our infrastructure.

egbertbouman:
retest this please

PR #5357, opened 2020-06-11
Fixed DHT and updated debug panel

egbertbouman:


egbertbouman:
@devos50 Could you see if the `DHTHealthManager` fix is OK?

ichorid:
The tests produce no errors, so it should be fine.

egbertbouman:
It turns out that the BEP33 thing hasn't been running since at least v7.4. Now that it is, let's hope that we're not creating more problems.

PR #5366, opened 2020-06-13
Fix iteration on keys in request manager. Again.

### LINKED ISSUE #5363: AttributeError: 'int' object has no attribute 'time'
Caught by the app tester:

```
[PID:44363] 2020-06-12 17:46:47,261 - ERROR <tribler_window:107> root.on_exception(): Traceback (most recent call last):
  File "tribler_gui/widgets/downloadspage.py", line 148, in load_downloads
  File "tribler_gui/tribler_request_manager.py", line 178, in __init__
  File "tribler_gui/tribler_request_manager.py", line 100, in add_request
  File "tribler_gui/tribler_request_manager.py", line 95, in evict_timed_out_requests
AttributeError: 'int' object has no attribute 'time'

ERROR:Executor:********** TRIBLER CRASHED **********
ERROR:Executor:Tribler crashed after uptime of 3090.547427892685 sec!
```

I suspect that line 94 in `tribler_request_manager.py` is incorrectly enumerating over dictionary keys, instead of values. The keys of the `self.requests_in_flight` are integers, which will fail with the above error.

ichorid:
fixes #5363 
:facepalm:

### DIFF
--- a/src/tribler-gui/tribler_gui/tribler_request_manager.py
+++ b/src/tribler-gui/tribler_gui/tribler_request_manager.py
@@ -86,12 +86,12 @@ class TriblerRequestManager(QNetworkAccessManager):
         error_dialog.show()
 
     def clear(self):
-        for req_id, req in list(self.requests_in_flight.items()):
+        for req in list(self.requests_in_flight.values()):
             req.cancel_request()
 
     def evict_timed_out_requests(self):
         t = time()
-        for req in list(self.requests_in_flight.items()):
+        for req in list(self.requests_in_flight.values()):
             if t - req.time > self.request_timeout_interval:
                 req.cancel_request()
 


src/tribler-gui/tribler_gui/tribler_request_manager.py
```diff
@@ -91,7 +91,7 @@ def clear(self):
 
     def evict_timed_out_requests(self):
         t = time()
-        for req in list(self.requests_in_flight.items()):
+        for _, req in list(self.requests_in_flight.items()):
```
devos50, line 94: Why not use ‘.values’?

ichorid: Because I'm old-school :eyeglasses: .

Seriously, just copy-pasted the stuff...

PR #5381, opened 2020-06-17
Fix race condition in tablecontentmodel

ichorid:
fixes error from reporter

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/tablecontentmodel.py
+++ b/src/tribler-gui/tribler_gui/widgets/tablecontentmodel.py
@@ -202,7 +202,8 @@ class RemoteTableModel(QAbstractTableModel):
                 self.info_changed.emit(response['results'])
             elif remote and uuid.UUID(response.get('uuid')) == CHANNELS_VIEW_UUID:
                 # This is a discovered channel (from a remote peer), update the total number of channels and the labels
-                self.channel_info["total"] += len(response["results"])
+                if self.channel_info.get("total"):
+                    self.channel_info["total"] += len(response["results"])
                 self.info_changed.emit(response['results'])
         return True
 


devos50:
@ichorid next time, please create a new issue or link to an existing one (I think there already is a closed issue). We want to have this kind of stuff documented in GitHub, if there is a need to trace the history of this bug later.

ichorid:
@devos50 , could you please merge single-commit changes using rebase instead of merge? It looks much cleaner that way. Thanks!
изображение

qstokkink:
> 
> 
> @devos50 , could you please merge single-commit changes using rebase instead of merge? It looks much cleaner that way. Thanks!
> изображение

Whereas there is an entire discussion to be had around this (see https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase) I'd prefer if we use merge everywhere and all the time. Seeing as we are not a closed source project, that is the safer option.

ichorid:
>Rebasing the commits is considered "unsafe," such as when a rebase is possible without merge conflicts but would produce a different result than a merge would.

This means it is perfectly safe to rebase, no race condition is possible.

Why don't we use the best of both worlds? Rebase small, single commit PRs, merge PRs that contain *interdependent* commits. Like, to show that what you merge is a *single, indivisible piece of work*.

qstokkink:
> Why don't we use the best of both worlds? Rebase small, single commit PRs, merge PRs that contain interdependent commits. Like, to show that what you merge is a single, indivisible piece of work.

Once you start doing this, you break with the *append-only* git branching model. The whole **commits are merged cleanly from a feature branch to a main branch** turns into **commits are merged cleanly from a feature branch to a main branch, unless they are small -- in which case we rewrite history**.

Next to this fundamental issue I have, I also want to address the reason why we have a commit history in the first place. When I look through the history, I want to see what was merged and why it was merged. With the merge model this is cleanly separated in commits, I can instantly see who merged what, when and why. If we do the rebase thing, this meta structure is lost and some commits will have history inside of them and some will not.

ichorid:
The entire "merge" idea was created for *very big* projects, like the Kernel, where it is physically impossible to linearize the history. In our case, the rule to always use merge commits is an unnecessary evil cargo cult.

qstokkink:
We have over 15k commits, over 150 contributors, over 150k lines, 378 forks and 28 local branches in this repo. This seems sufficiently big to me to warrant some best practices to avoid horrible rebase scenarios.

ichorid:
>We have over 15k commits ...

The question is, how many times a single line is edited on average? How many *non-linear* merge commits do we have? For instance, in IPv8 there is only a couple of overlapping merges. Most of IPv8 looks like this:
изображение


That's completely linear.
In Tribler, the situation is similar.

qstokkink:
🤷‍♂️ and that's fine. The reason we use Git is to enable the non-linear exceptions to the rule. If you want a perfect commit history you should use Subversion. And indeed Tribler used to live on svn. But, it turned into a dumpster fire precisely due to these types of exceptions.

ichorid:
GitHub lets us have the best from both worlds. It checks that there are no race conditions, that the result of rebase-merge is the same as merge-merge. It is unwise not to use the feature.

qstokkink:
The resulting code is indeed the same. But, the resulting structure is not: rebasing rewrites the merged commit (see also https://sethrobertson.github.io/GitBestPractices/#pubonce).

ichorid:
> But, the resulting structure is not: rebasing rewrites the merged commit

We never change the history by applying "rebase and merge" in **GitHub**, because there is no previous history. You merge from your local repo. This is the same as merging from nowhere.

qstokkink:
So this is directly from the GitHub documentation, https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#rebase-and-merge-your-pull-request-commits:

> Rebase and merge on GitHub will **always** update the committer information and **create new commit SHAs**

So, unless you are talking about manually pushing to the main branch. I don't understand how "We never change the history by applying "rebase and merge" in GitHub" applies.

ichorid:
Ok, I agree, that creates new commit SHAs. However, there were not *prior* commits in the public repository with some other SHA. So that's not rewriting from a technical standpoint, this can't hurt other users subscribed to the repository.

qstokkink:
So yes, if you have one commit on a feature branch of your fork and you remove the branch after this single commit and nobody else (including yourself) works off your feature commit, there is no issue. Up to there, I agree. But, why risk nasty rebase situations just for an aesthetic reason, if there is a zero-risk alternative?

ichorid:
> you have one commit on a feature branch of your fork and you remove the branch after this single commit and nobody else (including yourself) works off your feature commit ...

thats and exact description of 98% of commits to Tribler repo.

>why risk nasty rebase situations.

There is no risk. GitHub logic guarantees that the rebase will result in exactly the same code state as merge. That is nobody will race your change.

qstokkink:
> that is an exact description of 98% of commits to Tribler repo.

Yes, but that is entirely not my point 😃 

> There is no risk. 

I drew out a complete scenario for you (X is the hash rewrite of commit B):

```
==== Scenario 1: Merge ====
- Initial -
MAIN ... <- [A]
FEATURE ... <- [A] <- [B]

- Merge 1 -
MAIN ... <- [A] <- [B] <- [MERGEB]
FEATURE ... <- [A] <- [B]

- Update 1 -
MAIN ... <- [A] <- [B] <- [MERGEB]
FEATURE ... <- [A] <- [B] <- [C]

- Merge 2 (fast forward!)-
MAIN ... <- [A] <- [B] <- [MERGEB] <- [C]  <- [MERGEC]
FEATURE ... <- [A] <- [B] <- [C]

==== Scenario 2: Rebase ====
- Initial -
MAIN ... <- [A]
FEATURE ... <- [A] <- [B]

- Merge 1 -
MAIN ... <- [A] <- [X]
FEATURE ... <- [A] <- [B]

- Update 1 -
MAIN ... <- [A] <- [X]
FEATURE ... <- [A] <- [B] <- [C]

- Merge 2 (merge conflict!)-
MAIN ... <- [A] <- [X]
FEATURE ... <- [A] <- [B] <- [C]
```

This is the easiest scenario I could come up with. This becomes worse if you have collaborators on the same feature branch and features on features. The risk is quite real.

ichorid:
Of course, conflict is possible. Though, with our project guidelines ("GitHub merges everything") and team size, the danger is negligible. Due to the same guidelines, resolving *concurrency* conflicts is a breeze.

In the rebase-based workflow, conflicts only occur in GIT when it can't automatically merge stuff on the line-based level. It does not require hashes of parent commits to be equal to hashes of stuff in your branch. It only requires the fast-forward to be possible.

We always require people to rebase their stuff on top of the current branch. That is only possible due to the little amount of concurrent contributions to the repository. This would not be feasible in e.g. the Kernel.

Our workflow typically looks like this:
 1. update the local copy of the upstream branch from upstream origin
 2. create a separate branch from it
 3. do some changes to it
 4. push it to the personal repository
 5. rebase it on top of other people's changes that were accumulated on the upstream branch
 6. try to merge it using the GitHub interface

The time between 5 and 6 is typically so small, that *no other person* is able to push something on that branch. Of course, this does not account for PR reviews that could add quite some time between 5 and 6. But after *every* change the contributor must go through the whole process again: tests, review, etc. The reviewers are those who push the merge button, BTW. So they always *know* something could go in the way of their contribution.

The main point:
we do not use Git to merge stuff. Ever. We merge using GitHub. It is safe to rebase stuff with GitHub when contributing from a personal repository, because **GitHub** merge does not care about hashes.

qstokkink:
My point is that what will work most of the time in your scenario is based on a small team and strict control on the manner of introducing features and the manner of creating commits.

As the team grows and with contributions of external developers, as the "merger" of Pull Requests we have no insights into (or frankly no business meddling with) the feature branches of the "mergee". As contributions become more complex and more frequent (even though right now this may be perfectly fine) this risk is no longer negligible.

My gripe with your approach is the same as my gripe with race conditions. No, it will not happen a lot. But when it does, it's very annoying.

PR #5390, opened 2020-06-18
Disable market by default, added GUI toggle

### LINKED ISSUE #5389: Disable the market community by default + add toggle to settings menu
The market community is important for Tribler since it is an integral part of our self-replicating botnet research. However, the market is not actively maintained and therefore, there might be some bugs that could affect Tribler as a whole. As discussed on Slack with @ichorid, we will disable the market community by default. The market should be enabled through a toggle in the settings panel.

devos50:
This PR disables the market by default and adds an option to the settings panel to enable it.

Fixes #5389 

Schermafbeelding 2020-06-18 om 14 03 06

### DIFF
--- a/src/tribler-core/tribler_core/config/tribler_config.spec
+++ b/src/tribler-core/tribler_core/config/tribler_config.spec
@@ -12,7 +12,7 @@ random_slots = integer(default=5)
 competing_slots = integer(default=15)
 
 [market_community]
-enabled = boolean(default=True)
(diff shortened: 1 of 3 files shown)


devos50:
Seems that one of the channel test hangs on Windows:

```
test_recursive_commit_channel_torrent (test_channel_metadata.TestChannelMetadata) ... + FAILURE=127
```

devos50:
retest this please

Issue #5409, opened 2020-06-25
End manual backporting


qstokkink:
Tribler has had a release and a development branch for many years now. As per the _git flow_ model fixes are made onto the latest release branch and then also merged into the development branch. This is fine.

However, applying fixes from the release branch to the development branch **has to be done manually**. This comes at the cost of the sanity of whomever has to actually perform this merge and invariably leads to errors, such as these:
https://github.com/Tribler/tribler/issues/5090#issuecomment-648778883
https://github.com/Tribler/tribler/pull/581#issuecomment-42945250

To get rids of these types of errors and consequently **having to fix the same bug twice** we should **automate as much as possible**:

 - GitHub supports webhooks which allow our Jenkins machine to see when a new commit has been pushed to the release branch.
 - The jenkins-ci user can then automatically apply this merge in 99% of the cases (I'll get to the other 1% in a bit) and automatically open a PR for this change with the original PR body onto the development branch. By also copying the original PR body and merging into the development branch issues that are marked as "fixed" in the original PR will automatically be closed by GitHub without human intervention (which is also completely manual right now).

In this manner all branches stay up to date and functional, developers can actually develop (and not worry about the process) and we will have less errors.

*What about the 1%?* So yes, merging does not always succeed from the release branch to the development branch. So our options are then to _(a)_ fire-and-forget and hope someone (who has no in-depth knowledge of what the code does - or forgot by the time he arrives at the scene) fixes it manually for us in a big backport merge at a later date or _(b)_ open the conflicting PR automatically so the one who has made the code has to jump in and fix it.
In light of code ownership and the *you break it for everyone, you fix it for everyone* mentality of monorepos (see https://github.com/Tribler/tribler/issues/5386#issuecomment-645989434) I believe the fully automatic approach is the better one.

ichorid:
I am perfectly willing to sacrifice my sanity every time I have to merge `release` into `devel`. I did it many times, and still have some sanity left. Doing the stuff automatically will result in an ungodly amount of auto-generated merge commits. Also, I don't want *any* automated system getting write access to the repository.

My solution to the problem of merging stuff would be less technical and more organizational: just start doing releases more often, so there are no big backlogs and merge conflicts are easy to resolve.

qstokkink:
> Also, I don't want any automated system getting write access to the repository.

[...]

PR #5427, opened 2020-07-01
Merge release-7.5 into devel

devos50:


ichorid:
изображение

:facepalm: 

the merge must be done directly from `release-7.5` branch into `devel` branch, by hand in git. And pushed manually through the command line (with checking it as a PR of course). Otherwise, history becomes a complete mess and you can't merge stuff without conflicts anymore.


That's how it should look:
изображение

devos50:
Ok, missed that one. Even though I really don't like to push directly to `devel`, if that is the best practice I will do it like that next time.

qstokkink:
> if that is the best practice

The official Git best practice is (see https://git-scm.com/book/en/v2/Distributed-Git-Distributed-Workflows#ch05-distributed-git):
"make it as easy on you and the project maintainer as possible"

The Atlassian (Bitbucket) best practice (see https://www.atlassian.com/git/tutorials/merging-vs-rebasing) is:
"If you would prefer a clean, linear history free of unnecessary merge commits, you should reach for git rebase [...] if you want to preserve the complete history of your project and avoid the risk of re-writing public commits, you can stick with git merge"

I prefer the risk-averse approach in IPv8, but the Tribler clean-branch approach is also excellently documented by Atlassian. But, as usual: "it depends" is the answer to what your project-specific best practice should be. We've already has lengthy discussions on this topic 😃

qstokkink:
Oh, by the way, I created issue #5409 specifically for this reason. If the branch merge magic is done automatically, there can be no mistakes with merging vs. rebasing.

PR #5434, opened 2020-07-03
Update snap core to core20

xoriole:
Update snap package core base from core18 to latest core20. With the core changed, the command script had to be slightly changed.

### DIFF
--- a/build/debian/snap/snapcraft.yaml
+++ b/build/debian/snap/snapcraft.yaml
@@ -5,7 +5,7 @@ description: >
   Peer-to-peer Bittorrent client with enhanced privacy. Search and download torrents with less worries or censorship.
 grade: devel
 confinement: devmode
-base: core18
+base: core20
 
 architectures:
   - build-on: amd64
@@ -103,7 +103,7 @@ parts:
 
 apps:
   tribler-bittorrent:
-    command: desktop-launch ${SNAP}/usr/share/tribler/tribler
+    command: bin/desktop-launch $SNAP/usr/share/tribler/tribler
     plugs:
       - desktop
       - desktop-legacy


devos50:
@xoriole do you have the tools to test the Snap package? I think this would be a valuable asset to include in our upcoming release(s) 👍

xoriole:
@devos50 I have been using snap package of Tribler on my machine (Ubuntu 20.04) for quite some time. The snap package is built using the Debian package so the only way I know to test it works is by installing the snap package itself. 

The script to build the snap package should already be set up in Jenkins (unless the builder jobs has changed). The same Debian package builder script also builds the snap package but I remember that we disabled the flag because it was taking over 10 minutes to complete the build. Recently, I notice that the build time has improved quite a lot with the latest core (at least on my local machine) so might be worth reviving snap build. For the next release, I can build the package and upload to snap store if you like.

For reference on building the snap package for Tribler, check out these two files.
- [build/debian/makedist_debian.sh](https://github.com/Tribler/tribler/blob/devel/build/debian/makedist_debian.sh) (Line 27+)
- [build/debian/build-snap.sh](https://github.com/Tribler/tribler/blob/devel/build/debian/build-snap.sh)

devos50:
> Recently, I notice that the build time has improved quite a lot with the latest core (at least on my local machine) so might be worth reviving snap build.

Nice! Yes, this is quite a low-hanging fruit, since all the builders and other infrastructure are already in place.

devos50:
@xoriole I just tried running the `Build-snap` job, this is the output:

```
Could not find snap/snapcraft.yaml. Are you sure you are in the right directory?
```

It seems that the path in the Jenkins job is somehow incorrect. Any idea how to fix that?

xoriole:
I see. Ubuntu builder job should build the snap package as well. Can you check `Configure` for `Build-Ubuntu64`?

devos50:
I changed the machine, but I'm getting the same error :)

xoriole:
Screenshot from 2020-07-03 10-14-26

I meant this job. There should be an environment variable commented out in the build script of the job configuration. Could you check that and run the job?

devos50:
Ah I see, thanks for pointing that out! (I thought the `Build-snap` job would do the job) 👍

devos50:
@xoriole New error:

```
The plugin used by part 'desktop-qt5' does not support snaps using base 'core20'.
```

xoriole:
I see. There should not have been any issues with 'desktop-qt5' and 'core20'. Yet there is. The issue I suspect is the mismatch between the `triblertester/snap_builder:core18` docker image and latest core20. The builder is using Ubuntu 18.04 and the new core is using Ubuntu 20.04. LibC versions are different there. I was building the snap package natively (line 36) so I had no issues. I'll update the docker image and submit another PR.

For reference
https://github.com/Tribler/tribler/blob/5e58ee71cbb4bcd8f308700cdf524b0dcb899f42/build/debian/makedist_debian.sh#L31-L37

devos50:
Yeah, it's probably failing because of incompatibility issues there. Should we set up an Ubuntu 20.04 builder instead? Would that still be backwards compatible?

xoriole:
> Should we set up an Ubuntu 20.04 builder instead?

Yes, if you would like a faster builder but it can work using just docker image (a bit slower only :D).

>Would that still be backwards compatible?

I'm not sure. Do we want to support snap package for 18.04? There are probably no users using snap package in 18.04 so in my opinion, it is okay to move to 20.04 LTS version directly.

Issue #4189, opened 2019-02-01
[Feature request] Separate server and client.


rb07:
This is a feature request.

##### Tribler version/branch+revision: #####
7.2.0
##### Operating system and version: #####
Linux (Debian 3.10.105)
##### Steps to reproduce the behavior: #####
Execute tribler.sh
##### Expected behavior: #####
Feature: run only server.

Let's say I want to have a seedbox running only the server, and a client somewhere else controlling that server.

It would be useful in addition of tribler.sh to have a tribler-server.sh (or tribler-daemon.sh) which sets the 2 environment variables, CORE_BASE_PATH, CORE_API_PORT, and then executes tribler.sh .

Similar for the client (tribler-client.sh), just don't run the server, and set the server's IP address, and port.

##### Actual behavior: #####
Runs both server and client.

xoriole:
@rb07 Thank you for the feature request. We had discussions about it in the past. Currently, it is possible to run the server only using Twisted plugin. However, documentation is lacking on how to do it. 

Our focus atm is on stabilizing the integrated client and improving the performance. We'll revisit this feature request at a later point.

hbiyik:
Related

https://github.com/Tribler/tribler/issues/5327

drew2a:
If I understand the original request correctly (I'm a bit confused with the terms "client" and "server"), the issue is a duplicate of:
- #5303

PR #5458, opened 2020-07-08
Added HTTPS support to REST API

### LINKED ISSUE #3078: Config option for http api interface
Currently the RESTfull api binds to localhost. If you want to run headless on an external server(nas for example) it would be nice to be able to access this. Therefor i suggest either listening on 0.0.0.0 by default or by adding a config option to change the default.

### LINKED ISSUE #3607: Make REST API available from other machines
Currently the REST API is only available on the same machine (localhost).

It would be nice to be able to control tribler from another machine.
I think for this all that's needed is to add a configuration option for http api binding interface, and use this here:
https://github.com/Tribler/tribler/blob/f40349a187f3cfa4010ef5c67a44d5a739497b12/Tribler/Core/Modules/restapi/rest_manager.py#L32
Where it replaces '127.0.0.1'.

Ofcourse it could also be always set to '0.0.0.0', but I guess that would be not preferable security wise :)

egbertbouman:
fixes #3078
fixes #3607


egbertbouman:
retest this please

egbertbouman:
retest this please

egbertbouman:
retest this please

src/tribler-core/tribler_core/modules/tunnel/tests/test_triblertunnel_community.py
```diff
@@ -22,25 +21,28 @@
...
 from tribler_core.tests.tools.base_test import MockObject
+from tribler_core.tests.tools.test_as_server import BaseTestCase
 from tribler_core.tests.tools.tools import timeout
 from tribler_core.tests.tools.tracker.http_tracker import HTTPTracker
 from tribler_core.utilities.path_util import mkdtemp
 from tribler_core.utilities.utilities import succeed
 
 
-class TestTriblerTunnelCommunity(TestBase):
+class TestTriblerTunnelCommunity(TestBase, BaseTestCase):  # pylint: disable=too-many-public-methods
```
qstokkink, line 31: This is not something that should change right now, but more of a note/memo: `BaseTestCase` seems to have high duplication with `TestBase` and should probably be consolidated.

src/tribler-core/tribler_core/upgrade/config_converter.py
```diff
@@ -102,3 +102,17 @@ def convert_config_to_tribler75(state_dir):
         config.set_selected_files([tdef.get_index_of_file_in_files(fn)
                                    for fn in config.config['download_defaults'].pop('selected_files')])
         config.write(str(filename))
+
+
+def convert_config_to_tribler76(state_dir):
```
ichorid, line 107: :clap:

src/tribler-core/tribler_core/upgrade/tests/test_config_upgrade_to_75.py
```diff
@@ -0,0 +1,27 @@
+import os
+import shutil
+
+from tribler_core.config.tribler_config import TriblerConfig
+from tribler_core.tests.tools.base_test import TriblerCoreTest
+from tribler_core.tests.tools.common import TESTS_DATA_DIR
+from tribler_core.upgrade.config_converter import convert_config_to_tribler76
+
+
+class TestConfigUpgradeto75(TriblerCoreTest):
```
ichorid, line 10: Should be UpgradeTo**76** and in a separate file, like `test_config_upgrade_to76.py`

src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
```diff
@@ -2398,6 +2398,177 @@ color: white;</string>
...
+                  <item row="19" column="1">
+                   <widget class="QLabel" name="label_65">
+                    <property name="sizePolicy">
+                     <sizepolicy hsizetype="MinimumExpanding" vsizetype="Preferred">
+                      <horstretch>0</horstretch>
+                      <verstretch>0</verstretch>
+                     </sizepolicy>
+                    </property>
+                    <property name="text">
+                     <string>While the GUI/Core communicate over HTTP with the REST API server bound to localhost, this setting allows you to run the REST API over HTTPS, and binds to 0.0.0.0. This option can be used to remotely control Tribler. Please only use this option if you realize its security implications!</string>
```
ichorid, line 2564: I guess the explanation part about the GUI/Core should be omitted, as it has nothing to do with this option. It is better to change the text to something like:


>This setting enables Tribler REST API over HTTPS, which allows you to control Tribler remotely. It binds to every available IP interface. CAUTION! Do not enable this option unless you realize the security implications!

src/tribler-core/tribler_core/upgrade/tests/test_config_upgrade_to_75.py
egbertbouman: Oops, copy-paste error!

src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
egbertbouman: It's there so people know that HTTPS isn't going replace HTTP, and GUI/Core communication remains unchanged.

I know there are separate HTTP and HTTPS port settings, which already imply that it runs separately, but I just wanted to be clear.

ichorid: People looking at the GUI don't care about what kind of communication protocol is used between the Core and the GUI.
This text is here to inform the advanced users that they now got an option to enable remote access through by means of HTTPS. It is not "release notes" or Readme.txt, or changes.md, etc.

We want this kind of texts in the GUI to be as non-technical as possible.

egbertbouman: So, it's an option for advanced users that should be as non-technical as possible. OK, that explains it. I'll remove it 🤷

ichorid: If the GUI-oriented user can't use the feature in a few clicks, and the feature requires some technical knowledge to be used properly, and some external tool (to e.g. generate a certificate), omitting it from the GUI makes sense. Otherwise, it could cause us more trouble than good: users blaming bad usability and security.

Please, do not misunderstand me: I liked the feature initially. And I made this comment about the text to make the feature more accessible. However, your last comment, while clearly ironical, made me re-think it. As you said the feature is probably too complex for the non-technical user, especially because of the certificate thing. (Hell! Even I don't know how to get that certificate!) Our little clash just made the problem more obvious.

So, I think it is better to remove this feature from the GUI.

egbertbouman: I wasn't actually suggesting to remove the feature entirely from the GUI, just that bit of text concerning GUI/Core communication. Otherwise, users will never know it's even there.

ichorid: BTW, is there an easy way to generate a certificate from Windows? What's the process for setting the remote access to the Core?

egbertbouman: You can just use OpenSSL. I considered generating the certificate automatically with cryptography, but it's probably better to let the user handle this. Maybe the average user doesn't know how to do this, but look at it as a safety precaution: if you can't generate a certificate, you probably shouldn't be using this feature.

ichorid: Well, the kind of users who knows how to use OpenSSL directly to create a certificate needs no GUI. Or even Tribler at all...

I feel the GUI option should only be there when we start providing at least a rudimentary web interface. As it is, adding a GUI option to enable a REST interface aimed at programmers makes no sense. Unless you're enabling a remote GUI, but that's a different story..

egbertbouman: OK.

egbertbouman:
I initially wanted HTTPS support because I was developing a debug web app that I wanted to use to gain insight into remotely running Tribler instances. However, since I've stopped development on the web app, I don't really see a use for HTTPS support anymore. So, I'm closing this PR, and I'll make a new PR for the bug fixes that are in this one.

ichorid:
We want to be able to [run the GUI and the Core separately](https://github.com/Tribler/tribler/issues/4189), and you PR is a major stepping stone for this. The web-based debug panel is even more important to us. Why did you stop developing the web app?

egbertbouman:
After some internal discussion, I'm resurrecting this PR.

While we don't have a direct need for HTTPS support without the debug web app, we might end up using it when connecting to the GUI in future releases.

ichorid:
@egbertbouman , could you please move the `fix libtorrent 1.2.4+` into a separate PR and address that to `release-7.5` branch? We really need that for 7.5.2.

egbertbouman:
@ichorid It's already in the `release-7.5` branch. I just cherry-picked it because we need it in devel as well.

ichorid:
@egbertbouman , don't worry, I'll merge release into devel as soon as this one will get merged.

ichorid:
Merged release into devel. The libtorrent fix is there now. No need for cherry-picking.

egbertbouman:
@ichorid OK, thanks!

Issue #5450, opened 2020-07-08
The identity database is created in ./sqlite when running Tribler


devos50:
When running Tribler, the `IdentityDatabase` is created in the current work directory (`./sqlite/identity.db`), and not in the state directory where the other databases are also located (`~/.Tribler/sqlite/identity.db)`. This behaviour is resulting in an error when running the tunnel helpers, since the current working directory is not writable for the `tunnel_helper` process. I'm not sure what happens in a frozen environment and where the identity database is being written away.

Furthermore, I found out that the `CommunicationManager` (which in turns creates the `IdentityManager` which creates the `IdentityDatabase`) is loaded in the `initialize` method of the `IdentityEndpoint` class. I'm not sure if this is a preferred design since the sqlite database is now managed by the `REST` endpoint.

@qstokkink can you comment on the above points?

qstokkink:
The whole Identity framework is _supposed_ to be a lazy-load mechanism. It _should_ only produce files if the community is loaded or a call is made.

qstokkink:
You could temporarily alleviate this issue by setting the `working_directory` of the IPv8 instance to the appropriate folder: `~/.Tribler/sqlite/`. However, still, this database file should not be there in the first place.

qstokkink:
I can reproduce this when running IPv8 standalone, it is not a configuration error.

devos50:
We probably have to pass the right state directory to [here](https://github.com/Tribler/py-ipv8/blob/13d933461d5cdbe6224066799186b377c905786f/ipv8/attestation/communication_manager.py#L198) when starting the REST API.

> The whole Identity framework is supposed to be a lazy-load mechanism. It should only produce files if the community is loaded or a call is made.

If I start Tribler (either in release mode or locally), the `identity.db` file is created even when I do not explicitly make calls to the identity framework.

qstokkink:
> We probably have to pass the right state directory to here when starting the REST API.

Yes, that pulls from the IPv8 `working_directory` field.

> If I start Tribler (either in release mode or locally), the identity.db file is created even when I do not explicitly make calls to the identity framework.

I can reproduce this when running IPv8 standalone, it is not a configuration error.

PR #5462, opened 2020-07-09
Updated IPv8 pointer

qstokkink:
Related to #5450

This PR:
 - Updates the IPv8 pointer, resolving #5450.
 - Explicitly sets `working_directory` of the IPv8 configuration.

### DIFF
--- a/src/tribler-core/tribler_core/session.py
+++ b/src/tribler-core/tribler_core/session.py
@@ -424,6 +424,7 @@ class Session(TaskManager):
             ipv8_config['address'] = self.config.get_ipv8_address()
             ipv8_config['overlays'] = []
             ipv8_config['keys'] = []  # We load the keys ourselves
+            ipv8_config['working_directory'] = str(self.config.get_state_dir())
 
             if self.config.get_ipv8_bootstrap_override():
                 import ipv8.community as community_file


src/tribler-core/tribler_core/session.py
```diff
@@ -424,6 +424,7 @@ async def start(self):
             ipv8_config['address'] = self.config.get_ipv8_address()
             ipv8_config['overlays'] = []
             ipv8_config['keys'] = []  # We load the keys ourselves
+            ipv8_config['working_directory'] = str(self.config.get_state_dir() / 'sqlite')
```
devos50, line 427: Isn't the `sqlite` directory appended by the IPv8 database logic?

qstokkink: Good point, is the `self.config.get_state_dir()` safe to write to for IPv8-based overlays or should this go into `self.config.get_state_dir() / 'ipv8'`?

devos50: `self.config.get_state_dir()` should be safe to write to. Please test it with Tribler to see if it ends up in the right directory (e.g., by reverting the lazy loading commit).

qstokkink: @devos50 I made that change and verified:

 - No `identity.db` or `attestations.db` is created on a normal run.
 - When lazy loaded in, `identity.db` and `attestations.db` are created in `.Tribler/7.5/sqlite`

qstokkink: Assuming that is expected behavior, I'll mark this as READY.

PR #5465, opened 2020-07-12
Added more info to OverflowError

### LINKED ISSUE #5369: [7.5.0-RC5] Python int too large to convert to C long
From the error reporter (7.5.0-RC5):

```
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 114, in on_read_data
RuntimeError: Python int too large to convert to C long
--LONG TEXT--
File "src\run_tribler.py", line 95, in start_tribler
File "src\tribler-core\tribler_core\session.py", line 452, in start
File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 95, in initialize
File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 245, in get_session
File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 230, in create_session
File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 565, in set_session_settings
OverflowError: Python int too large to convert to C long
--CONTEXT--
{'message': 'Task exception was never retrieved', 'exception': OverflowError('Python int too large to convert to C long'), 'future': <Task finished coro=<start_tribler_core.<locals>.start_tribler() done, defined at src\run_tribler.py:64> exception=OverflowError('Python int too large to convert to C long')>}
```

devos50:
When applying new libtorrent settings. This will help us to debug future OverflowErrors that occur, either because of a bug or a config modified by users.

I consider this enough to fix #5369 for the moment being. First, it only happened once on Windows 32-bit so it's not a severe issue. Second, this is most likely the result of users tampering with their configuration files since we have a built-in check whether the integers added in the settings screen are not overflowing. This PR will now help us to catch this kind of errors in the future, enabling us to make a call whether we should ignore it or fix it.

Fixes #5369

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/download_manager.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/download_manager.py
@@ -574,10 +574,13 @@ class DownloadManager(TaskManager):
             self.ltsettings[lt_session] = lt_session.get_settings()
         self.ltsettings[lt_session].update(new_settings)
 
-        if hasattr(lt_session, "apply_settings"):
-            lt_session.apply_settings(new_settings)
-        else:
-            lt_session.set_settings(new_settings)
+        try:
+            if hasattr(lt_session, "apply_settings"):
+                lt_session.apply_settings(new_settings)
+            else:
+                lt_session.set_settings(new_settings)
+        except OverflowError:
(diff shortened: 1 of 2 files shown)


ichorid:
Maybe you meant some other issue, not my older PR #5364

devos50:
Yes, it's #5369, good catch!

Issue #5474, opened 2020-07-15
Anonymous one-hop downloads not working with libtorrent 1.2.7 (Windows 32-bit and 64-bit and Ubuntu)


devos50:
When starting an anonymous one-hop download with libtorrent 1.2.7, on either Windows 32-bit or Windows 64-bit, the download does not start. It also does not find peers. When changing the hop count to 2, it seems to find peers and progress. Note that it seems to work with libtorrent 1.2.1. I would not like to downgrade libtorrent but instead try to find the problem and make sure Tribler works with libtorrent 1.2.7.

Some additional info:
- libtorrent was built with boost 1.73.
- I verified this both with a torrent from a random channel and with [my all-time favourite testing torrent](http://www.frostclick.com/torrents/video/animation/Big_Buck_Bunny_1080p_surround_frostclick.com_frostwire.com.torrent).

@egbertbouman could you debug/investigate this?

devos50:
Since the upgrade to libtorrent 1.2.7 (and Python 3.8), our hidden seeding e2e test is also failing, see https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_win64/825/artifact/output/test_runners_output/5.err/*view*/. This should help to debug the issue I think?

egbertbouman:
After a long bug hunt, the problem turned out to be annoyingly simple: when a socks proxy is used, libtorrent's `listen_port()` now returns 0 instead of the actual port number.

Because of this hidden services will set `SocksUDPConnection.remote_udp_address` incorrectly for the default hop count, causing all 1-hop tunnels to fail.

I'll make a PR for this tomorrow.

egbertbouman:
I've fixed this issue on Windows by listening for `listen_succeeded_alert`, which has the correct port. On Mac, however, we never get this alert, and instead `listen_port()` gives us the correct port. On Linux, both methods seem to work.

devos50:
@egbertbouman We are not using libtorrent 1.2.7 (yet?) on Mac. Do you want me to upgrade the version of libtorrent in our builder?

egbertbouman:
@devos50 No upgrade needed, it should just work.

PR #5475, opened 2020-07-15
Add default download path to recently used dirs cache

### LINKED ISSUE #5470: Download location option does not work
To reproduce, change the option through the GUI, then try downloading something. Observe the download going into the default (e.g. "Downloads/TriblerDownloads") dir. 

I experienced it on Ubuntu 20.04 with the target dir path containing Cyrillic characters.

ichorid:
fixes #5470 by adding the redefined default download path to the list of the most recently used download dirs.


devos50:
It's a bit annoying that the e2e seeding test fails until #5474 is fixed...

PR #5478, opened 2020-07-16
Fix libtorrent 1.2.4+ compatibility

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/5474.

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/download_manager.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/download_manager.py
@@ -63,6 +63,7 @@ class DownloadManager(TaskManager):
         self.ltsettings = {}  # Stores a copy of the settings dict for each libtorrent session
         self.ltsessions = {}
         self.dht_health_manager = None
+        self.listen_ports = {}
 
         self.notifier = tribler_session.notifier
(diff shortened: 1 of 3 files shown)


egbertbouman:
retest this please

devos50:
It looks like the `test_hidden_service` is still failing under libtorrent 1.2.4 and 1.2.5: https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_libtorrent_compatibility/581/console. @egbertbouman do you want to fix this one?

egbertbouman:
While hunting for the bug I only tried it with the latest version (1.2.7).

It seems there are 2 completely different issues at play here:
* In libtorrent 1.2.6/1.2.7 `listen_port()` returns 0 when used with a SOCKS5 proxy on Windows. That's fixed.
* In libtorrent 1.2.4/1.2.5 there are known issues with SOCKS5 proxies. I'm not sure what we can do about this. I think we may have to accept that this isn't going to work.

devos50:
Ok, I'm fine with deeming libtorrent 1.2.4/1.2.5 incompatible with Tribler them and removed them from the validation experiment 👍

PR #5484, opened 2020-07-17
Various fixes

### LINKED ISSUE #5481: Remove run_tribler.py cp65001 support
For some reason the `cp65001` fix has been moved to after the imports, but it should be the first thing in the file https://github.com/Tribler/tribler/blob/devel/src/run_tribler.py:

https://github.com/Tribler/tribler/blob/1aea0280026841fc740e7e302156616473dafaa7/src/run_tribler.py#L7-L34

Interestingly enough the comment is still at the top of the file though, maybe some formatting tool automatically changed this?

UPDATE: This code can now be completely removed, `cp65001` is supported by default since Python 3.3.

### LINKED ISSUE #5482: Our dependency check on startup is broken
Our dependency check on startup is not working. This is because a mis-indented `continue` statement prevents the main logic from working, see [here](https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/dependencies.py#L68).

The dependency checker also contains some wrong dependencies, which I will also fix.

devos50:
Fixes #5481, fixes #5482, partially addresses #5485


src/tribler-core/tribler_core/dependencies.py
```diff
@@ -64,14 +61,14 @@ def check_for_missing_dependencies(scope='both'):
     for dep in dependencies:
         if not is_scope_both and dep['scope'] != 'both' and dep['scope'] != scope:
             continue
-            try:
```
ichorid, line 67: :facepalm:

src/run_tribler.py
```diff
@@ -29,6 +30,8 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir):
 
     trace_logger = None
 
+    asyncio.set_event_loop(asyncio.SelectorEventLoop())
```
ichorid, line 33: Why?

devos50: Because using the `ProactorEventLoop` on Windows seems to break peer discovery in IPv8. For the moment being, and maybe even for the 7.5.2 release, I suggest to use the `SelectorEventLoop`, even though we know it has deficiencies and that this is the primary reason we moved to Python 3.8.

I leave the debugging of this behaviour/finding a bug to someone who is more experienced with IPv8 peer discovery.

Also see #5485

ichorid: Please, leave a comment in the code. And limit the change to the Windows platform.

devos50:
Ready for review. The failing checks are expected.

ichorid:
don't forget to update IPv8 pointer to include the latest tunnels fixes!

devos50:
@ichorid done!

ichorid:
What's the problem with Windows GUI tests?

ichorid:
retest this please

src/run_tribler.bat
```diff
@@ -0,0 +1,2 @@
+set PYTHONPATH=%~dp0tribler-core;%~dp0tribler-common;%~dp0tribler-gui
```
qstokkink, line 1: This doesn't need `anydex` and `ipv8` on the path, like in https://github.com/Tribler/tribler/blob/devel/src/tribler.sh#L13 ?

ichorid: Yep, these should probably be on PATH

devos50: It actually worked without it, since in the `__init__.py` file in the `tribler-core` module, we add the `pyipv8` directory to the pythonpath. Anydex did not have to be imported since the market community is disabled by default.

Just to be sure, I've added both of them to the bat file 👍

PR #5495, opened 2020-07-19
Totally ignore gaierrors

ichorid:
:fire: :godmode: :fire:

### DIFF
--- a/src/tribler-core/tribler_core/session.py
+++ b/src/tribler-core/tribler_core/session.py
@@ -79,11 +79,7 @@ IGNORED_ERRORS = {
     (OSError, 10022): "Failed to get address info. Error code: 10022",
     (OSError, 16): "Socket error: Device or resource busy. Error code: 16",
     (OSError, 0): "",
-    (gaierror, 11001): "Unable to perform DNS lookup.",
-    (gaierror, 11004): "Unable to perform DNS lookup.", # Windows specific ?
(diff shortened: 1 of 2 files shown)


src/tribler-core/tribler_core/session.py
```diff
@@ -79,11 +79,7 @@
     (OSError, 10022): "Failed to get address info. Error code: 10022",
     (OSError, 16): "Socket error: Device or resource busy. Error code: 16",
     (OSError, 0): "",
-    (gaierror, 11001): "Unable to perform DNS lookup.",
-    (gaierror, 11004): "Unable to perform DNS lookup.", # Windows specific ?
-    (gaierror, 8): "Unable to perform DNS lookup.", # MacOS specific
-    (gaierror, -2): "",
-    (gaierror, -5): "No address associated with hostname",
+    gaierror: "Unable to perform DNS lookup."
```
qstokkink, line 82: Please use `-1` as a default value: `(gaierror, -1)`.

Mixing dictionary key types (`tuple` and `Exception` in this case) causes several `dict` operations to fail ([that depend on sorting for example](https://realpython.com/python-sort/#lists-with-non-comparable-data-types-cant-be-sorted)) and it seems easily avoidable in this case.

qstokkink: This isn't really imporant enough to request changes though. Updating my review 😃

PR #5496, opened 2020-07-20
Split testnet settings into multiple ones

devos50:
For the application tester I would like to "mix and match" my testnet settings. For example, I want to download using exit nodes on the testnet but use the channels discovered in the chant mainnet. To do so, I split the testnet settings into multiple ones, which can be enabled separately. To fully enable the testnet, you can pass the --testnet flag to Tribler and all communities will automatically load the testnet version.


devos50:
retest this please

PR #5500, opened 2020-07-21
Not starting libtorrent session in download tests

devos50:
This PR fixes the failing tests on Mac, where shutting down a libtorrent session takes considerably longer after the upgrade to libtorrent 1.2.7, triggering timeouts and causing deadlocks. This PR specifically modifies the download tests such that we are not starting a libtorrent session anymore. Instead, we mock a Download/DownloadState/TorrentDef where applicable. This should make these tests faster and more reliable. This PR will also pave the way for a potential migration to `pytest`.

Note that there are many more locations in the tests where we are unnecessarily starting a libtorrent session.

In addition, I made the following changes:
- Logging exceptions in the REST API to ease debugging.
- Created a `get_torrent` method in the `Download` class that returns a bdecoded torrent representation. This is used by the `export` endpoint.
- Fixed a bug where a libtorrent session would be created when a non-initialized `DownloadManager` object is shutdown.
- Fixed a bug where a `DownloadManager` would not correctly shutdown if no metadata directory was set.

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/download.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/download.py
@@ -36,6 +36,7 @@ class Download(TaskManager):
         self._logger = logging.getLogger(self.__class__.__name__)
 
         self.session = session
+        self.config = None
         self.tdef = tdef
         self.handle = None
         self.state_dir = self.session.config.get_state_dir() if self.session else None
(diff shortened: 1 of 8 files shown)


devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
Ready for review. The uncovered lines are hard to cover with a unit test, without extensively mocking libtorrent.

PR #5516, opened 2020-07-22
Fix RemoteQuery not using model properties

### LINKED ISSUE #5514: Getting family-unfriendly material in my search results
When searching for 'big buck', I'm getting family-unfriendly results in the GUI, even though the family filter is enabled:

Schermafbeelding 2020-07-22 om 14 52 40

Related: the category label in the table view and in the detail pane is not the same.

ichorid:
fixes #5514


qstokkink:
[Memo] Windows test suite locked up on `test_stream_full (test_downloads_endpoint.TestStreamingEndpoint) ... `.

PR #5521, opened 2020-07-23
Running fake Tribler API in a subprocess

### LINKED ISSUE #5494: Windows GUI tests randomly failing
The GUI tests on Windows are randomly failing after our upgrade to Python 3.8. See for example [here](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_GUI_win64/795/console). From the output, it seems that it still uses an `IocpProactor` proactor, even though this is disabled on Windows.

devos50:
It seems that Python 3.8 does not like a Qt event loop and an asyncio event loop in the same process. To address this, I now spawn a subprocess that serves HTTP requests. This subprocess is killed when the tests are done.

I also noticed a few code errors which are also addressed in this PR.

Fixes #5494


devos50:
Well, that does not seem to solve the issue 😢 https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_GUI_macOS/858/console

Issue #5523, opened 2020-07-25
Separate release URL for Tribler besides GitHub for download statistics


xoriole:
Right now, we do release via GitHub which is great but we lack some important insights on release statistics. 

For example, from Github API, 
- We know how many downloads were done for each of the release artifacts but we don't know the distribution over time. 
- Daily/Weekly/Monthly downloads data is not available. 
- Global spread information is also not available. 
- How long does it take the majority of the users to download/switch to the newly released version? We don't know.

I believe release statistics will be useful to better understand our users to some extent and improve our release cycle. Therefore, as the first step towards https://github.com/Tribler/tribler/issues/5510, I propose to develop a separate release website/service/API that shows all the release statistics based on public information.

Update: https://releases.tribler.org (available now)

devos50:
@xoriole with #5524 merged, is this issue considered solved?

xoriole:
@devos50  yes, it is resolved. We can close this.

PR #5524, opened 2020-07-29
Add release url in the version check manager

xoriole:
With this PR, a new release URL (https://release.tribler.org) is added to the version check manager. Swagger docs for the currently available endpoints are available at /docs.

Fixes https://github.com/Tribler/tribler/issues/5523

### DIFF
--- a/src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
+++ b/src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
@@ -3,6 +3,8 @@ from asyncio import sleep
 
 from aiohttp import web
 
+from asynctest import Mock
+
 from tribler_core.modules import versioncheck_manager
 from tribler_core.modules.versioncheck_manager import VersionCheckManager
 from tribler_core.restapi.rest_endpoint import RESTResponse
@@ -14,17 +16,23 @@ class TestVersionCheck(TestAsServer):
 
     async def setUp(self):
         self.port = self.get_port()
-        self.site = None
+        self.runner = None
         self.should_call_new_version_callback = False
         self.new_version_called = False
         versioncheck_manager.VERSION_CHECK_URLS = ['http://localhost:%s' % self.port]
         await super(TestVersionCheck, self).setUp()
         self.session.version_check_manager = VersionCheckManager(self.session)
 
-        self.session.notifier.notify = self.notifier_callback
+        self.session.notifier.notify = Mock()
 
     def notifier_callback(self, subject, *args):
         self.new_version_called = True
+
+    async def tearDown(self):
+        if self.runner:
+            await self.runner.shutdown()
+
+        await super(TestVersionCheck, self).tearDown()
 
     async def setup_version_server(self, response, response_code=200):
         self.response = response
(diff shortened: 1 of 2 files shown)


src/tribler-core/tribler_core/modules/versioncheck_manager.py
```diff
@@ -29,23 +30,29 @@ async def stop(self):
         await self.shutdown_task_manager()
 
     async def check_new_version(self):
+        for version_check_url in VERSION_CHECK_URLS:
+            if await self.check_new_version_api(version_check_url):
```
ichorid, line 34: it is better to wrap the thing in `asyncio.wait_for` and catch timeout error here. Otherwise, the check could get stuck.

src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
```diff
@@ -97,3 +97,40 @@ async def test_non_json_response(self):
...
+        self.assertTrue(self.new_version_called)
+
+        # Setting a timeout of 1ms, version checks should fail
+        versioncheck_manager.VERSION_CHECK_TIMEOUT = 0.001
+
+        self.new_version_called = False
+        await self.session.version_check_manager.check_new_version()
+        self.assertFalse(self.new_version_called)
+
+        await self.site.stop()
```
ichorid, line 117: Cleanup stuff must be in the `tearDown` method. Otherwise, the line will not be called on test failure and the failed test can affect the rest of the tests that will run in the same process

src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
```diff
@@ -97,3 +97,40 @@ async def test_non_json_response(self):
...
         self.assertTrue(versioncheck_manager.check_failed)
+
+    @timeout(10)
+    async def test_version_check_timeout(self):
+        await self.setup_version_server(json.dumps({'name': 'v1337.0'}))
+
+        self.new_version_called = False
+        # Default timeout is 5 seconds so under normal circumstance, we don't expect timeout
+        await self.session.version_check_manager.check_new_version()
+        self.assertTrue(self.new_version_called)
```
ichorid, line 108: Thanks for the wonderful `Mock` magic, it is now much easier to check if the notifier was called as the result of version checking procedure. See 
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

Please, rewrite the version checker tests using `Mock` object.

ichorid:
@xoriole , are you going to finish this soon? I can finish it for you if you don't have time.

src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
xoriole: Good point

src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
xoriole: updated

xoriole:
Functionally this should be ready. I've tried to minimize the code changes considering the title and actual purpose of this PR. Re-requesting review.

Issue #4623, opened 2019-06-25
Crash when clicking debug tab


qstokkink:
```python
ERROR   1561469701.07            tribler_window:74   (root)  Traceback (most recent call last):
  File "/home/quinten/Documents/tribler/TriblerGUI/tribler_window.py", line 915, in clicked_menu_button_debug
    self.debug_window = DebugWindow(self.tribler_settings, self.tribler_version)
  File "/home/quinten/Documents/tribler/TriblerGUI/debug_window.py", line 156, in __init__
    self.window().debug_tab_widget.setTabEnabled(2, settings and settings['trustchain']['enabled'])
TypeError: setTabEnabled(self, int, bool): argument 2 has unexpected type 'NoneType
```

qstokkink:
```python
>>> print {0: None} and {0: None}[0]
None
```

xoriole:
This should already be fixed by now

PR #5529, opened 2020-08-03
Refactoring: moving multiple is_frozen definitions to tribler_common

### LINKED ISSUE #5233: The is_frozen function is defined three times
The `is_frozen` function that returns a boolean whether we are operating in a frozen environment, is defined in three files: `tribler_gui/utilities.py`, `vlc.py` and `install_dir.py`. We should move this method to the `tribler_common` package and import it from there.

mrcruz:
This implements  #5233.

It was a simple issue, but I don't think I fully understand the is_frozen function. I did some basic testing and it appears nothing changed but I would like someone else's perspective.

Fixes #5233

### DIFF
--- a/src/tribler-common/tribler_common/utilities.py
+++ b/src/tribler-common/tribler_common/utilities.py
@@ -1,0 +1,13 @@
+import sys
+
+
+def is_frozen():
+    """
+    Return whether we are running in a frozen environment
+    """
+    try:
+        # PyInstaller creates a temp folder and stores path in _MEIPASS
+        sys._MEIPASS
+    except Exception:
+        return False
+    return True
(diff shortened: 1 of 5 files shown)


devos50:
ok to test

devos50:
retest this please

devos50:
@mrcruz If you are interesting in fixing another (relatively easy) issue, please have a look at #4623. You can reproduce it by change the `triblerd.conf` file in your Tribler state directory.

qstokkink:
> there is most likely no easier way to check if we are in a frozen environment.

Challenge accepted:

```python
def is_frozen():
    return hasattr(sys, "_MEIPASS")
```

devos50:
@qstokkink Ok, there is definitely an easier way 👍 . However, I'm not completely sure if that would work for all frozen environments, given the magic inner workings of PyInstaller and custom logic for cross-platform support. At least we are confident that the current method functions correctly.

PR #5530, opened 2020-08-10
Disable video play buttons

ichorid:
Disables video buttons in the GUI for 7.5.2, because our VLC player is disabled too.

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/tableiconbuttons.py
+++ b/src/tribler-gui/tribler_gui/widgets/tableiconbuttons.py
@@ -61,7 +61,9 @@ class PlayIconButton(IconButton):
     icon = QIcon(get_image_path("play.png"))
 
     def should_draw(self, index):
-        return index.model().data_items[index.row()][u'category'] == u'Video'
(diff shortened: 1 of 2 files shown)


src/tribler-gui/tribler_gui/widgets/tableiconbuttons.py
```diff
@@ -61,7 +61,8 @@ class PlayIconButton(IconButton):
     icon = QIcon(get_image_path("play.png"))
 
     def should_draw(self, index):
-        return index.model().data_items[index.row()][u'category'] == u'Video'
+        # return index.model().data_items[index.row()][u'category'] == u'Video'
```
devos50, line 64: I would prefer to only have commented-out code if it is accompanied by a comment (do we have this in the `contributing.md` guidelines?). Could you add a comment to this code, both here and in `triblertablecontrollers.py`, so we know why it has been commented out?

ichorid: Good point

PR #5531, opened 2020-08-11
Use new IPv8 constructs for PopularityCommunity

qstokkink:
This PR uses some more modern IPv8 constructs to lower the line count of the `PopularityCommunity`, while staying wire-format compatible.


xoriole:
interesting

qstokkink:
> 
> 
> interesting

Not really actually 😃, these are just aesthetic changes. All of the functionality and wire-format is exactly the same.

PR #5552, opened 2020-09-08
macOS documentation update

drew2a:
## Readme

1. I removed ```git clone --recursive git@github.com:Tribler/tribler.git``` but left ```git clone --recursive https://github.com/Tribler/tribler.git``` as more common way to clone the repo.
1. "Running from repo" section has been restructured, because there were several points of "how to install dependencies".

## Development on osx

1. Added missed commands to PyQt5 section
1. Added workaround for libtorrent installation invented by @devos50: https://github.com/Homebrew/homebrew-core/pull/43858
1. Replace custom pip installation by "pip3 install -r requirements.txt"

## Requirements.txt

1. Missing dependencies have been added
1. `faulthandler` has been moved down because it is a builtin module since Python 3.3 (and its installation may cause an error).
1. Os-dependent function "_show_system_popup" has been fixed (it was falling on macOS)

PR #5555, opened 2020-09-14
READY: Should not eat exception silently

kozlovsky:
It is better to not eat exceptions silently, as it can mask some problems.

### DIFF
--- a/src/tribler-core/tribler_core/dependencies.py
+++ b/src/tribler-core/tribler_core/dependencies.py
@@ -4,7 +4,7 @@ This file lists the python dependencies for Tribler.
 Note that this file should not depend on any external modules itself other than builtin ones.
 """
 import importlib
-import os
+import platform
 import sys
 
 dependencies = [
@@ -45,15 +45,18 @@ def _show_system_popup(title, text):
     sep = "*" * 80
     print('\n'.join([sep, title, sep, text, sep]), file=sys.stderr)
 
-    if os.name == 'Windows':
+    system = platform.system()
+    if system == 'Windows':
         import win32api
         win32api.MessageBox(0, text, title)
-    elif os.name == 'Linux':
+    elif system == 'Linux':
         import subprocess
         subprocess.Popen(['xmessage', '-center', text])
-    elif os.name == 'Darwin':
+    elif system == 'Darwin':
         import subprocess
         subprocess.Popen(['/usr/bin/osascript', '-e', text])
+    else:
+        print(f'cannot create native pop-up for system {system}')
 
 
 def check_for_missing_dependencies(scope='both'):
--- a/src/tribler-core/tribler_core/dependencies.py
+++ b/src/tribler-core/tribler_core/dependencies.py
@@ -52,10 +52,10 @@ def _show_system_popup(title, text):
         elif os.name == 'Darwin':
             import subprocess
             subprocess.Popen(['/usr/bin/osascript', '-e', text])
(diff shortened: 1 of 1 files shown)


drew2a (dismissed):
Cool, thank you!

qstokkink:
ok to test

qstokkink:
Note that the purpose of this code is to communicate a message to the user no matter what goes wrong. Even in the event `xmessage` is not installed (notably arch linux), a message is at least printed to the terminal (line 60-61). With this change the source error is no longer communicated. I'd suggest moving line 60+61 to preceed the `try .. except`.

drew2a:
@qstokkink correct. 

But if I understand the code correctly, the `text` (the source error) will be printed in both cases.

Moving `print line` to the beginning of `def _show_system_popup` function is a good idea. But in this PR @kozlovsky just fixed my misstep from: https://github.com/Tribler/tribler/pull/5552/files

qstokkink:
> 
> 
> @qstokkink correct.
> 
> But if I understand the code correctly, the `text` (the source error) will be printed in both cases.
> 
> Moving `print line` to the beginning of `def _show_system_popup` function is a good idea. But in this PR @kozlovsky just fixed my misstep from: https://github.com/Tribler/tribler/pull/5552/files

If that is the intention, it is providing duplicate functionality: a secondary way to format the text in a print statement. In that case, I would recommend the same thing I suggested before and to also print the actual exception in the exception handler. In effect, what it was doing before.

qstokkink (approved):

kozlovsky:
I moved the print statement above as @qstokkink suggested and removed try/except block

Also I use `platform.system()` instead of `os.name` as it looks like a correct way according to doc:
https://docs.python.org/3/library/platform.html#platform.system
https://stackoverflow.com/questions/1854/python-what-os-am-i-running-on

qstokkink:
I like this solution, I'll quickly have it retested to see if the Windows GUI test failure is persistent.

qstokkink:
retest this please

qstokkink:
retest this please.. seems like the Windows machine is experiencing abnormal slowdown.

devos50:
add to whitelist

kozlovsky:
retest this please

devos50 (approved):

devos50:
@kozlovsky please prefix the PR with `READY` if it is ready for review/merge 👍 

Also, note that our tests are currently unstable, and the GUI tests on Windows are occasionally failing. We should do some proper investigation of this behaviour soonish.

kozlovsky:
> please prefix the PR with `READY` if it is ready for review/merge

@devos50 you mean as a prefix of PR name, like `"[READY] ..."`?

Maybe it is more convenient to create PRs that are not ready as drafts?
https://github.blog/2019-02-14-introducing-draft-pull-requests/

devos50:
@kozlovsky normally I try to look at that too, but in this case, I did not see that the PR was converted from a draft state into a non-draft state.

Thinking about this, different devs are using different approaches. I consider a PR ready if it is either prefixed with `READY` or when it converts to a non-draft state. We should opt for one approach: either use prefixes in the title or, as you suggest, use draft/non-draft states.

kozlovsky:
retest this please

drew2a:
@kozlovsky @devos50 from my point of view, prefixes were good way to sort `ready/draft` pull request until `draft pull requests` have been introduced.

Let's discuss it :)

PR #5557, opened 2020-09-14
Move TriblerRequestManager error callback to a separate slot

ichorid:
Beware!
Calls to PyQT signal callbacks **drop unexpected arguments**!

E.g.
```
s = Signal(int, int)

def foo(a):
    print(a)

s.connect(foo)
s.emit(1,2)
```
will perfectly print `1` and just **ignore** the second argument :exploding_head: 

I should probably report it to PyQT list...


devos50 (approved):
Nice find!

PR #5585, opened 2020-09-25
use real Core for GUI tests

### LINKED ISSUE #4750: Automate fake Tribler API responses generation
Currently, we are forced to duplicate our endpoints code and program REST API responses for the GUI tests by hand. Fake Tribler API code quality is horrible. It is basically the old copy-pasted Tribler code patched with ad-hoc mock objects. Also, whenever someone changes something in the Core, they have no way to say **if** and **how** it will affect the endpoints responses.

Instead, we can just run the GUI script against a real Tribler Core once and **record the responses**. The Fake Tribler API then essentially becomes a [Chinese Room](https://en.wikipedia.org/wiki/Chinese_room) for REST requests. Switching to a fixed playbook will have the following effects:
 * GUI tests become deterministic;
 * developers don't have to tediously copy-paste code into Fake Tribler API;
 * less code, fewer errors;
 * we get the ability to **get diffs** between a REST playbook generated by a pull request and a an existing one;
 * we gain the ability to **get visual diffs** between the GUI screenshots to see the changes immediately.

It should be straightforward to implement this, as the change requires just recording the JSON exchange between the Core and the GUI during the GUI tests.

ichorid:
Fixes #4750 

### The Problem

Testing with FakeTriblerAPI does not test the real *information exchange* between the Core and the GUI. Also, every change to the Core REST API requires the programmer to change FakeAPI accordingly, which doubles the effort for every change.

### The Solution
This PR throws away FakeAPI in favour of using the real Core running in a special "Dummy" mode that does not connect to the network, does (almost) no disk writes, creates a fake MetadataStore DB, etc. The mode is activated by setting the `TRIBLER_CORE_TEST_MODE` environment var to `1`.

Note that I had to comment out GUI debug panel logger tab tests and Market tests, because these require a more work test-mode content filling in the Core.


ichorid:
retest this please

ichorid:
retest this please

ichorid:
retest this please

src/tribler-core/tribler_core/session.py
```diff
@@ -456,13 +468,15 @@ async def start(self):
         if self.config.get_libtorrent_enabled():
             self.readable_status = STATE_START_LIBTORRENT
             from tribler_core.modules.libtorrent.download_manager import DownloadManager
-            self.dlmgr = DownloadManager(self)
+            self.dlmgr = DownloadManager(self, dummy_mode=self.core_test_mode)
             self.dlmgr.initialize()
             self.readable_status = STATE_LOAD_CHECKPOINTS
             await self.dlmgr.load_checkpoints()
+            if self.core_test_mode:
+                await self.dlmgr.start_download_from_uri("magnet:?xt=urn:btih:0000000000000000000000000000000000000000")
```
devos50, line 476: I think this start a libtorrent session. Could you check this? If so, can we somehow avoid starting libtorrent during the GUI tests?

src/tribler-gui/tribler_gui/tests/test_gui.py
```diff
@@ -25,8 +20,9 @@
 from tribler_gui.tribler_window import TriblerWindow
 from tribler_gui.widgets.loading_list_item import LoadingListItem
 
-if sys.platform.startswith('win'):
-    asyncio.set_event_loop(asyncio.SelectorEventLoop())
+RUN_TRIBLER_PY = Path(tribler_gui.__file__).parent.parent.parent / "run_tribler.py"
+
+SAMPLE_STATEDIR = RUN_TRIBLER_PY.parent / "tribler-core/tribler_core/tests/tools/data/state_dir_guitests"
```
devos50, line 25: Is this `SAMPLE_STATEDIR` actually used?

src/tribler-gui/tribler_gui/tests/test_gui.py
```diff
@@ -45,48 +41,40 @@ def window(api_port):
     QTest.qWaitForWindowExposed(window)
 
     screenshot(window, name="tribler_loading")
-    wait_for_signal(window.core_manager.events_manager.tribler_started, no_args=True)
+    wait_for_signal(window.core_manager.events_manager.tribler_started, timeout=50, no_args=True)
```
devos50, line 44: Is this really necessary?

src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
```diff
@@ -187,26 +187,6 @@ def draw_channel_state(self, painter, option, index, data_item):
         return True
 
 
-class SubscribedControlMixin(object):
```
devos50, line 190: Are the `SubscribedControlMixin` and `SubscribeToggleControl` unused?

devos50 (changes requested):
I like this solution 👍 

I have a few comments/questions though.

src/tribler-core/tribler_core/session.py
ichorid: I'll try to disable starting a real libtorrent session.

src/tribler-gui/tribler_gui/tests/test_gui.py
ichorid: Oops, thanks for catching it. Forgot from the previous iteration

src/tribler-gui/tribler_gui/tests/test_gui.py
ichorid: oops, again, more stuff from testing

src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
ichorid: Yes! I found this unused code by running the GUI against the real Core. Though, I'll probably readd it shortly, better to delete it now as it is really unused (voting column does the subscription management).

devos50: Ah good! Yeah I'm in favour of removing any dead code, even it is re-added later on. We can always fetch it from the VCS histories.

src/tribler-core/tribler_core/session.py
ichorid: I managed to `Mock` the libtorrent session in-code

devos50 (changes requested):
Thanks for mocking libtorrent!

One more thing though, there are a few fixable `pylint` errors. Can you address them? (e.g., pointless statement, line too long, ...). I missed them in my last review :/

ichorid:
Done. I got so used to `pre-commit` doing its magic that I forget that it only checks the MetadataStore module and main GUI files...

devos50 (changes requested):
I think you missed one:

```
31: W0612: (unused-variable), generate_channel: Unused variable 'ind'
```

ichorid:
Argh! :pirate_flag:

devos50 (approved):
Looks good!

ichorid:
Thanks for the fast and thorough (:wink:) review!

PR #5597, opened 2020-10-02
READY: Ported Gumby's modular Community loading

### LINKED ISSUE #5593: Modularize IPv8 community loading
As the Tribler repository now features an `experiment/` folder, we are seeing the same staging procedures being implemented as already available in Gumby. One of these features is the staging of IPv8 communities. In order for both repositories to benefit, I propose moving this from Gumby into the Tribler experiment infrastructure.

Loading communities in the `Session` is currently one big line of `if .. else` statements, see https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/session.py#L146
Instead, in Gumby, this is completely modularized, see https://github.com/Tribler/gumby/blob/devel/gumby/modules/gumby_session.py

This modularization of loading communities improves the Tribler code in four ways:

 - It shortens the `Session` code (making it more of a manager and less of a god class).
 - Allows for isolation and staging of community loading.
 - Reduces the complexity of staging components.
 - Allows for declarative programming.

There is also a downside: a big `if .. else` construction is slightly faster than a modular loading system. But, considering the direction we are heading, that seems like an acceptable sacrifice.

qstokkink:
Fixes #5593

This is a straight port of Gumby's modular Community loading.

Community loading is now declarative, which inflates the codebase slightly. Session is now less of a god-class.

### DIFF
--- a/src/tribler-core/tribler_core/conftest.py
+++ b/src/tribler-core/tribler_core/conftest.py
@@ -46,6 +46,7 @@ def tribler_config(tribler_state_dir, tribler_download_dir):
     config.set_default_destination_dir(tribler_download_dir)
     config.set_torrent_checking_enabled(False)
     config.set_ipv8_enabled(False)
+    config.set_ipv8_walk_scaling_enabled(False)
     config.set_libtorrent_enabled(False)
     config.set_libtorrent_dht_readiness_timeout(0)
     config.set_api_http_enabled(False)
--- a/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py
+++ b/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py
@@ -1,0 +1,295 @@
+from abc import ABC
+
+from ipv8.loader import CommunityLauncher
+from ipv8.peer import Peer
+from ipv8.peerdiscovery.discovery import EdgeWalk, RandomWalk
+
+
+# The lazy-loading of Community-specific files triggers Pylint, this is expected:
+# pylint: disable=C0415
+
+
+class IPv8CommunityLauncher(CommunityLauncher, ABC):
+
+    def get_my_peer(self, ipv8, session):
+        return (Peer(session.trustchain_testnet_keypair) if session.config.get_trustchain_testnet()
+                else Peer(session.trustchain_keypair))
+
+
+class IPv8DiscoveryCommunityLauncher(IPv8CommunityLauncher):
+
+    def get_overlay_class(self):
+        from ipv8.peerdiscovery.community import DiscoveryCommunity
+        return DiscoveryCommunity
+
+    def should_launch(self, session):
+        return True
+
+    def get_kwargs(self, session):
+        return {}
+
+    def get_walk_strategies(self):
+        from ipv8.peerdiscovery.churn import RandomChurn
+        from ipv8.peerdiscovery.community import PeriodicSimilarity
+        return [(RandomChurn, {}, -1), (PeriodicSimilarity, {}, -1), (RandomWalk, {}, 20)]
+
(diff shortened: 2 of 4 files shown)


devos50:
Nice! Will you also remove this code from Gumby then?

One consequence of removing the community loading code from Gumby is that we now have a hard requirement on Tribler for any experiment where we are using IPv8 and modular community loading. Specifically, AnyDex and Bami currently bypass the requirement for Tribler by making their own launcher. We would then require Tribler to run experiments since AnyDex and Bami are using modular community loading.

qstokkink:
> Specifically, AnyDex and Bami currently bypass the requirement for Tribler by making their own launcher.

Ah, as far as I knew the `ipv8_provider` was always a `GumbySession` (which uses Tribler anyway). Could you link me to one of those?

devos50:
@qstokkink see [here](https://github.com/Tribler/gumby/blob/devel/gumby/modules/anydex_module.py#L89)

qstokkink:
@devos50 ah, sneaky 😄 In that case it makes sense to move this modular loading into a common base repo.. which I fear will be IPv8.

devos50:
@qstokkink that would be nice, but if you think it is better to move it to Tribler, please go ahead. Having to clone/copy Tribler is a small sacrifice for preventing code duplication across many experiments 👍

qstokkink:
@devos50 no, let's do this to our best ability immediately. You have presented a clear case for moving the launcher/loader base-classes into IPv8.

qstokkink:
This PR will be "on hold" until these base classes are available in IPv8.

qstokkink:
No longer "on hold", base classes are available in IPv8.

qstokkink:
@devos50 I think pytest's timeout mechanism is broken.

devos50:
That's weird. I would expect this to be stable, but it seems that again, we cannot trust external libraries...

qstokkink:
retest this please

devos50:
For now, I've added a hard timeout of three minutes to the tests. Still, I would have expected `pytest-timeout` to do its job.

qstokkink:
👍 seems reasonable

devos50:
@qstokkink any idea why the GUI test would fail with this PR?

qstokkink:
@devos50 interesting.. they didn't fail before. Random timeouts?

devos50:
@qstokkink  When running your branch on my computer, the core does not start with the following error:

```
[PID:44252] 2020-10-02 13:34:50,809 - INFO - IPv8CommunityLoader(181) - Finished loading communities!
[PID:44252] 2020-10-02 13:34:50,809 - ERROR <session:259> Session.unhandled_error_observer(): Unhandled exception occurred! 'NoneType' object has no attribute 'wait_for_socks_servers'
--LONG TEXT--
  File "/Users/martijndevos/Documents/tribler/src/run_tribler.py", line 81, in start_tribler
    await session.start()
  File "/Users/martijndevos/Documents/tribler/src/tribler-core/tribler_core/session.py", line 348, in start
    await self.tunnel_community.wait_for_socks_servers()
AttributeError: 'NoneType' object has no attribute 'wait_for_socks_servers'
```

qstokkink:
Oops I might have screwed up the rebase, one moment.

qstokkink:
Ah yes, pleasing the linters killed the PR 😄

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
🙏 the tunnel tests and metadata tests aligned and both passed this run.

devos50 (approved):

PR #5611, opened 2020-10-05
Crawl random nodes

drew2a:
This PR includes:

1. Refactoring of `experiments`
1. New `crawl_torrents.py` tool

## crawl_torrents.py

Source: [crawl_torrents.py](crawl_torrents.py)

### Description

This script crawl first 100 torrents from random nodes in the network.

Result will be stored in a `sqlite` db:
```
class RawData(db.Entity):
    peer_hash = Required(str)
    torrent_hash = Required(str)
    torrent_title = Required(str)
    torrent_votes = Required(int)
    torrent_position = Required(int)
    date_add = Required(datetime)
```

Where:
* `position` is a torrent's rank inside single `query response`, ordered by `HEALTH`

### Usage

```
export PYTHONPATH=${PYTHONPATH}:`echo ../.. ../../src/{pyipv8,tribler-common,tribler-core} | tr " " :`

python3 crawl_torrents.py [-t <timeout_in_sec>] [-f <db_file.sqlite>]
```

Where:
* `timeout_in_sec` means the time that the experiment will last
* `db_file.sqlite` means the path to `sqlite` db file. 
    If file doesn't exists, then new file will be created. 
    If file exists, then crawler will append it.
 

### Example

```
python3 crawl_torrents.py -t 600
python3 crawl_torrents.py -t 600 -f torrents.sqlite
```


experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,244 @@
...
+
+        session.ipv8.overlays.append(session.remote_query_community_community)
+        session.ipv8.strategies.append((RandomWalk(session.remote_query_community_community,
+                                                   timeout=ExperimentConfiguration.RANDOM_WALK_TIMEOUT_IN_SEC,
+                                                   window_size=ExperimentConfiguration.RANDOM_WALK_WINDOW_SIZE,
+                                                   reset_chance=ExperimentConfiguration.RANDOM_WALK_RESET_CHANCE),
+                                        UNLIMITED))
+
+
+def _parse_argv(argv):
```
devos50, line 206: You can use the `argparse` library to simplify the parsing of arguments, for example, see [here](https://github.com/Tribler/application-tester/blob/master/main.py#L10).

drew2a: 👍 thanks!

synctext:
Thank you for removing [swear words from our code](https://github.com/Tribler/tribler/pull/5611/commits/85236e5fe9e523eae4dcd9abf312718509c4dcf2). That makes us sound like amateur juveniles.

drew2a:
retest this please

drew2a:
retest this please

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+    finally:
+        loop.run_until_complete(loop.shutdown_asyncgens())
+        loop.close()
+
+
+if __name__ == "__main__":
+    _arguments = _parse_argv()
+    print(f"Arguments: {_arguments}")
+
+    if _arguments.verbosity:
```
ichorid, line 212: Better to rewrite with: `logging.basicConfig(level=logging.DEBUG if _arguments.verbosity else logging.INFO`

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+    parser.add_argument('-t', '--timeout', type=int, help='a time in sec that the experiment will last', default=0)
+    parser.add_argument('-f', '--file', type=str, help='result file path (json)', default='result.json')
+    parser.add_argument('-v', '--verbosity', help='increase output verbosity', action='store_true')
+
+    return parser.parse_args()
+
+
+def _run_tribler(arguments):
+    service = Service(arguments.file, arguments.timeout,
+                      working_dir=os.path.join(
```
ichorid, line 195: Please use `pathlib` instead of `os` module. `pathlib` is much safer, modern and convenient.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+
+        config.set_chant_enabled(True)
+        return config
+
+    async def on_tribler_started(self):
+        await super().on_tribler_started()
+        session = self.session
+        peer = Peer(session.trustchain_keypair)
+
+        session.remote_query_community_community = Crawler(peer, session.ipv8.endpoint,
```
ichorid, line 172: That's an... inventive name for a variable :wink:

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.register_task("request_torrents", self.request_torrents_from_new_peers, interval=10)
+        self.register_task("save_torrents", self.save, interval=CRAWLER_SAVE_INTERVAL_IN_SEC)
+
+    def request_torrents_from_new_peers(self):
+        """Send a request for gathering torrent information
+
+        Behaviour:
+
+        For all new pears: ```self.get_peers() - self.polled_peers```
```
ichorid, line 93: New :pear: s! Yummy!

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    def __init__(self, my_peer, endpoint, network, metadata_store, out_file_path):
+        super().__init__(my_peer, endpoint, network, metadata_store)
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self._output_file_path = out_file_path
+        self.max_peers = UNLIMITED
+        self.queried_peers_limit = UNLIMITED
+
+        self.torrent_list = defaultdict(list)
```
kozlovsky, line 79: The name is a bit misleading, as it is a dictionary of lists and not a list

drew2a: Right. Will fix it.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+        self._logger.info(f"Collect torrent item for {key}: {title}")
+        self.torrent_list[key].append((votes, infohash, title))
+
+    # this method has been overloaded only for changing timeout
+    def send_select(self, peer, **kwargs):
+        request = CrawlerSelectRequest(self.request_cache, hexlify(peer.mid), kwargs)
+        self.request_cache.add(request)
+        self.ez_send(peer, RemoteSelectPayload(request.number, json.dumps(kwargs).encode('utf8')))
+
+    def save(self):
```
ichorid, line 138: You're rewriting the file completely instead of appending lines to the end. This is dangerous since in case of a crash during the dump the data will be lost. Also, it could be pretty slow in case of big files. Please consider appending new lines to the end of the save file, or better yet, create a simple Pony-based DB to dump your data.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+        self._logger.info(f"Save results to : {self._output_file_path}")
+
+        with open(self._output_file_path, 'w') as json_file:
+            json.dump(self.torrent_list, json_file)
+
+    def introduction_response_callback(self, peer, dist, payload):
```
ichorid, line 144: Original `introduction_response_callback` triggers `send_remote_select_subscribed_channels`. You don't want that.
So, you have to either stop calling the base class method here, or overload `send_remote_select_subscribed_channels` with empty method, like:

`def send_remote_select_subscribed_channels(...):
    pass
`

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,214 @@
...
+
+UNLIMITED = -1  # this magic number is use inside many Tribler modules
+
+IPV8_WALK_INTERVAL = 0.05
+
+RANDOM_WALK_TIMEOUT_IN_SEC = 5
+RANDOM_WALK_WINDOW_SIZE = 0
+RANDOM_WALK_RESET_CHANCE = 30
+
+CRAWLER_REQUEST_TIMEOUT_IN_SEC = 60
```
ichorid, line 49: :+1:  for `_IN_SEC` thing. I should start doing that myself...

experiment/popularity_community/crawl_torrents.py
drew2a: Fixed

experiment/popularity_community/initial_filling.py
```diff
@@ -0,0 +1,136 @@
...
+from ipv8.peerdiscovery.discovery import RandomWalk
+from tool.tiny_tribler_service import TinyTriblerService
+from tribler_core.modules.popularity.popularity_community import PopularityCommunity
+
+_logger = logging.getLogger(__name__)
+
+TARGET_PEERS_COUNT = 20  # Tribler uses this number for some reason
+
+
+class ObservablePopularityCommunity(PopularityCommunity):
```
ichorid, line 20: Why is this class even in this PR? I thought it was a part of another PR?

experiment/popularity_community/crawl_torrents.py
drew2a: Code has been improved.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,217 @@
...
+        Note:   it is possible to request more, but it requires a significant
+                increase of code complexity.
+        Ref: https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py#L19
+        """
+        peers_ready_to_request = set(self.get_peers()) - self.polled_peers
+        self._logger.info(f"Request torrents from {peers_ready_to_request}")
+
+        for peer in peers_ready_to_request:
+            self._logger.debug(f"Requesting torrents from peer: {peer}")
+            self.send_select(peer, metadata_type=[REGULAR_TORRENT], first=0, last=100)
```
ichorid, line 105: Could you please add at least one unit test to this thing? It will be really easy to copy the code from tests of RemoteQueryCommunity.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,217 @@
...
+        Behaviour:
+
+        For all new pears: ```self.get_peers() - self.polled_peers```
+        the crawler request 100 most popular torrents.
+
+        Note:   it is possible to request more, but it requires a significant
+                increase of code complexity.
+        Ref: https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py#L19
+        """
+        peers_ready_to_request = set(self.get_peers()) - self.polled_peers
```
ichorid, line 100: @qstokkink , is it the best way to query peers aggressively?

@grimadas , you've built an aggressive crawler some time ago. Did you use something like this?

ichorid (changes requested):

experiment/popularity_community/crawl_torrents.py
qstokkink: @ichorid Well, it is definitely very aggressive for the peers you know every 10 seconds. However, if the intent of this thing is to actually _crawl_, you'll also want to call `self.network.remove_peer()` to get rid of peers you the have collected information for (the ones in `self.polled_peers`) to get some new peers.

experiment/popularity_community/crawl_torrents.py
drew2a: Yes, misprint.

experiment/popularity_community/crawl_torrents.py
drew2a: I don't like ternary operators in python. They bulky.
I'll think.

experiment/popularity_community/crawl_torrents.py
drew2a: Ok, but give me proof first, please.

experiment/popularity_community/crawl_torrents.py
qstokkink: One disclaimer: setting the `max_peers` to infinity will also eventually yield you all of the information, just slower. So, is this the most aggressive? No. Will it work? Yes.

experiment/popularity_community/crawl_torrents.py
devos50: Good one. We have tons of timers/timeouts which can be enhanced by adding the unit to the variable name.

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,217 @@
...
+from ipv8.lazy_community import lazy_wrapper
+from ipv8.peer import Peer
+from ipv8.peerdiscovery.discovery import RandomWalk
+from tool.tiny_tribler_service import TinyTriblerService
+from tribler_core.modules.metadata_store.community.remote_query_community import RemoteQueryCommunity, \
+    SelectResponsePayload, SelectRequest, RemoteSelectPayload
+from tribler_core.modules.metadata_store.serialization import REGULAR_TORRENT
+from tribler_core.utilities.unicode import hexlify
+
+UNLIMITED = -1  # this magic number is use inside many Tribler modules
```
devos50, line 41: Can you give a specific instance where it it used in Tribler? We should document these occurrences better.

That said, we usually use `-1` to indicate "random" (e.g., when setting ports), or to indicate an "unlimited value".

experiment/popularity_community/crawl_torrents.py
```diff
@@ -0,0 +1,217 @@
...
+
+    It has been added only for having an ability to increase `timeout_delay`
+    """
+
+    @property
+    def timeout_delay(self):
+        return CRAWLER_REQUEST_TIMEOUT_IN_SEC
+
+
+class Crawler(RemoteQueryCommunity):
```
devos50, line 64: Minor recommendation: change the class name to something more specific. A crawler, at least in the context of Tribler, can have many different meanings (torrent crawler? TrustChain crawler? peer crawler?)

experiment/popularity_community/initial_filling.py
devos50: Seems to be moved from experiment/popularity_community/initial_filling/initial_filling.py

src/requirements.txt
```diff
@@ -18,3 +18,5 @@ marshmallow
 Pillow
 netifaces
 pyqtgraph
+pytest~=6.1.0
```
devos50, line 21: I'm not sure if we should add `pytest` to the requirements, since it is not strictly required to run Tribler.

@grimadas pointed me to [Poetry](https://python-poetry.org), which has a nice separation between main and dev dependencies. We could consider using that at a later stage.

experiment/popularity_community/crawl_torrents.py
ichorid: https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

experiment/popularity_community/crawl_torrents.py
ichorid, line 105: Ah! How could I miss that?
You're not **sorting** the torrents on health! Therefore, you will get essentially semi-random torrents. To enable sorting, you must add `sort_by=HEALTH` to the list of `send_select` arguments.

ichorid (changes requested):

experiment/popularity_community/crawl_torrents.py
ichorid: Also, sorting on health makes the *order* of torrents in the packets important. You can write that down to establish a partial order on the set of torrents. Later you can compare and combine partial orders from different peers.

experiment/popularity_community/crawl_torrents.py
kozlovsky: Probably it is better to have all timeouts at the same scale (seconds), to simplify arithmetic on it. In that case, it is not necessary to specify a unit as part of the name. And longer timeout may be specified as
```python
SOME_LONG_TIMEOUT = 10 * HOUR  #  where HOUR = 60 * MINUTE, MINUTE = 60 * SECOND, and SECOND = 1
```

devos50: I also like that idea. I think all of our timeouts are in seconds already. 👍

experiment/popularity_community/crawl_torrents.py
drew2a: Fixed

experiment/popularity_community/crawl_torrents.py
drew2a: Fixed

experiment/popularity_community/crawl_torrents.py
drew2a: fixed

src/requirements.txt
drew2a: Make sense. 

Does the build script use `src/requirements.txt`?

experiment/popularity_community/crawl_torrents.py
drew2a: Fixed

src/requirements.txt
devos50: To build Tribler, we use `pyinstaller`, which automatically scans the source code for any dependency.

experiment/popularity_community/crawl_torrents.py
drew2a: You know...  I was wrong. `-1` used not in "many" but in "few":

Explicit `-1` examples: 
* https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/trustchain/caches.py#L50
* https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/wallet/caches.py#L108
* [fixed already](https://github.com/Tribler/py-ipv8/blob/ad2e0e0697e2dac4917e5c4a8ede8c0aa6646bf4/ipv8_service.py#L156)

Implicit `0` or `-1` example:
* https://github.com/Tribler/py-ipv8/blob/master/ipv8/community.py#L271

My proposition here is to use this numbers more explicit
```
RANDOM = -1
UNLIMITED = 0
```

experiment/popularity_community/crawl_torrents.py
drew2a: Maybe I came from "strange dev-place", but I have seen that different modules uses different defaults for the time.
(seconds and microseconds, basically)

So, every time, i see "delay" or "interval" or "time", I can't realise, what exactly this means. 

Therefore, I prefer explicit naming.

xoriole:
How about moving `experiment/popularity_community/tool/tiny_tribler_service.py` to one directory up `experiment/tools/tiny_tribler_service.py`?

src/requirements.txt
drew2a: Is that means, content of `src/requirements.txt` does not reflects on users, but only on devs?

I've removed this string. 
But just want to be ensure, that we are not fighting with the ghosts.

devos50: Yes, the content of `requirements.txt` is only for devs currently.

experiment/popularity_community/crawl_torrents.py
drew2a: I've rewrite code:

```
    logging_level = logging.DEBUG if _arguments.verbosity else level = logging.INFO
    logging.basicConfig(level=logging_level)
```

thank for the suggestion

experiment/popularity_community/crawl_torrents.py
drew2a: @qstokkink cool! thank you for `self.network.remove_peer()` tip.

experiment/popularity_community/crawl_torrents.py
drew2a: @ichorid cool. Will fix it.

drew2a: fixed

experiment/popularity_community/crawl_torrents.py
drew2a: Good point.
I've replaced it by pony

Firstly, @kozlovsky offered me beautifully decomposed structure:
```
class Torrent(db.Entity):
    hash = Required(str)
    scraped_torrents = Set(lambda: ScrapedTorrent)


class Peer(db.Entity):
    hash = Required(str)
    scrapes = Set(lambda: Scrape)


class ScrapedTorrent(db.Entity):
    torrent = Required(lambda: Torrent)
    title = Required(str)
    votes = Required(int)
    position = Required(int)
    scrape = Required(lambda: Scrape)


class Scrape(db.Entity):
    peer = Required(lambda: Peer)
    torrents = Set(lambda: ScrapedTorrent)
    time = Required(datetime)

class RawData(db.Entity):
    peer_hash = Required(str)
    torrent_hash = Required(str)
    torrent_title = Required(str)
    torrent_votes = Required(int)
    torrent_position = Required(int)
    date_add = Required(datetime)
```

But I preferred more straightforward `RawData`
```
class RawData(db.Entity):
    peer_hash = Required(str)
    torrent_hash = Required(str)
    torrent_title = Required(str)
    torrent_votes = Required(int)
    torrent_position = Required(int)
    date_add = Required(datetime)
```

His said that you will be angry. So. 
I'm sorry.

src/requirements.txt
drew2a: Got it :)

drew2a:
@xoriole that is a good idea! Will implement soon.

experiment/popularity_community/crawl_torrents.py
ichorid: I propose to make a convention that in Tribler all time-related values should be in *seconds*. When we work with some external library we should add a time scale prefix, like @drew2a did.
What do you think, guys?

experiment/popularity_community/crawl_torrents.py
ichorid: How would you state `None` then? by using `None` explicitly?

experiment/popularity_community/crawl_torrents.py
qstokkink: I support using seconds as the default, as it is the grounding for all time in Python (stemming from `time.time()`).

experiment/popularity_community/crawl_torrents.py
drew2a: With all my love for the tests, I'm not sure that they are appropriate here.

drew2a:
Please take a look on PR one more time. 
I'm going to merge it within hour.

qstokkink:
You are failing the Pylint check. Please resolve that.

devos50:
retest this please

devos50:
retest this please

experiment/popularity_community/crawl_torrents.py
drew2a: I'm not sure I understood the question.

drew2a:
@qstokkink `103: C0301: (line-too-long), : Line too long (151/120)` because I post URL link:
```       
 Ref: https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py#L19
```  

Are you sure we need to cut it somehow just because the line is too long?

#
`106: W1203: (logging-fstring-interpolation) Use lazy % formatting in logging functions`


What reason for that?

devos50:
retest this please

qstokkink:
> @qstokkink `103: C0301: (line-too-long), : Line too long (151/120)` because I post URL link:
> 
> ```
>  Ref: https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py#L19
> ```
> 
> Are you sure we need to cut it somehow just because the line is too long?

Yes please, long lines have a tendency to warp viewports (IDE, GitHub, etc.). The 120 characters are chosen to fit GitHub (strict Python styling would be 80 characters).

> `106: W1203: (logging-fstring-interpolation) Use lazy % formatting in logging functions`
> 
> What reason for that?

All logging statements have string substitution built in, the linter is telling you here that you're applying string substitution inside of string substitution. In other words, it wants this:

```python
logging.error(f"Hello {World}!")  # Format inside of a format.
logging.error("Hello %s!", World)  # Pylint approved.
```

Now, personally, I don't mind the first format-in-format. But, we would have to discuss this with the team (and add an `ignore` to Pylint).

drew2a:
@qstokkink C0301 has been fixed.

The advantage of W1203 is not clear to me.

Let's discuss this.

https://github.com/PyCQA/pylint/issues/2354
https://github.com/PyCQA/pylint/issues/2395

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
The flake8 checker is fixed 👍

experiment/popularity_community/crawl_torrents.py
ichorid: :shrug:

ichorid: Consider it an excellent opportunity in learning how to write IPv8 tests :wink:

experiment/popularity_community/crawl_torrents.py
ichorid: Let's say there is a value that shows a limit of *something*, say a number of peers a community can have. In that case, as a programmer, I want to be able to use it to:
 * limit the number of peers a community can have to some *exact* number, say `10`
 * force the commuity to stop taking peers at all, e.g. by setting the value to `0`
 * indicate that there is no limit to the number of peers that a community can have, e.g. by setting it to `UNLIMITED`
 * indicate that the community can decide for itself, or use the default, e.g. by settin the value to `None` or omitting the keyword

Now, we're facing a design choice: should we indicate "nothing at all" with `0` or with `-1` or with `None`?

ichorid:
Basically, f-strings vs `%s` logging boils down to:

>if you follow the Pylint suggestion and don't use f-strings you're making your code slower in some places and maybe slightly faster in others

vs 

>many, many people think that logging with f-strings results in deferred evaluation.
> Sentry API won't work


a tough choice. On one hand, I definitely prefer f-strings readability. `%s` notation makes me shudder in C-horror. On the other hand, immediate evaluation can lead to *interesting* effects, save for Sentry compatibility (which we considered using)...

Seems like a tie.

experiment/popularity_community/crawl_torrents.py
drew2a: Got it.

Do you really need this quite complex behavior? 

Let say, we allow passing values `only` throw constructor. Then we easy can control the defaults.

```
class CustomCommunity(Community):
    def __init__(self, max_peers=30):
        self._max_peers = max_peers
```

And that should solve all the points on your list, shouldn't it?


Then, if you want to give an opportunity to change the behavior of your class, add a public method:

```
class CustomCommunity(Community):
    def __init__(self, max_peers=30):
        self._max_peers = max_peers

    def stop_to_receive_peers(self):
        # actions
```

qstokkink:
So far we have 2/6 reviewers in favor of _allowing_ f-strings to be used in logging statements (i.e. ignoring `W1203`). @devos50, @kozlovsky, @xoriole, @grimadas could you make your stance on this clear, so we can reach a majority vote?

xoriole:
I would go for making the Pylint god happy and green. Either we disable this conflicting rule completely or we all follow the rule fully in favor of consistency. I'm fine with both of them as long we agree that the tool should stay green.

devos50:
The question is whether we _allow_ then right? If so, I'm fine with disabling `W1203`.

kozlovsky:
So, as I understand it, the arguments for lazy evaluations of log strings (as originally intended by logging library):
- the code doesn't spend time on string interpolation (and don't allocate memory for a new string) if the logging message is suppressed due to the current logging level;
- Sentry can aggregate different log messages with the same message template;
- if someone hastily adds a log message with incorrect format (like using `%d` for strings), it will not generate a new exception in the code, and the error will be processed by the logging library.

The arguments for allowing f-strings interpolation in log messages (and interpolate non-lazily):
- f-string formatting is fast, and the performance difference is negligible comparing to lazy formatting;
- it is not always important to beautifully aggregate log messages in Sentry;
- the code looks more readable.

I think we can turn off W1203 and allow f-string interpolation in log messages, as it sometimes makes code more readable

qstokkink:
Alright gents, thank you for your input. We have a majority vote for ignoring `W1203` 🎉 

@drew2a please edit the `.pylintrc` in your PR so Pylint is placated and you get a shiny green checkmark.

devos50:
Pylint throws the following stack trace:

```
  File "/usr/local/bin/diff-quality", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/diff_quality_tool.py", line 271, in main
    ignore_whitespace=arg_dict["ignore_whitespace"],
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/diff_quality_tool.py", line 200, in generate_quality_report
    reporter.generate_report(output_file)
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 251, in generate_report
    report = template.render(self._context())
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 300, in _context
    context = super().report_dict()
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 183, in report_dict
    src_stats = {src: self._src_path_stats(src) for src in self.src_paths()}
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 85, in src_paths
    for src, summary in self._diff_violations().items()
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 178, in _diff_violations
    for src_path in self._diff.src_paths_changed()
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/report_generator.py", line 178, in <dictcomp>
    for src_path in self._diff.src_paths_changed()
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/violationsreporters/base.py", line 158, in violations
    output = execute(command, self.driver.exit_codes)
  File "/usr/local/lib/python3.7/dist-packages/diff_cover/command_runner.py", line 45, in execute
    raise CommandError(stderr)
diff_cover.command_runner.CommandError: Traceback (most recent call last):
  File "/usr/local/bin/pylint", line 10, in <module>
    sys.exit(run_pylint())
  File "/usr/local/lib/python3.7/dist-packages/pylint/__init__.py", line 22, in run_pylint
    PylintRun(sys.argv[1:])
  File "/usr/local/lib/python3.7/dist-packages/pylint/lint/run.py", line 311, in __init__
    linter.load_config_file()
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option_manager_mixin.py", line 316, in load_config_file
    self.global_set_option(option, value)
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option_manager_mixin.py", line 186, in global_set_option
    self._all_options[opt].set_option(opt, value)
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/options_provider_mixin.py", line 54, in set_option
    value = _validate(value, optdict, optname)
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option.py", line 105, in _validate
    return _call_validator(_type, optdict, name, value)
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option.py", line 86, in _call_validator
    return VALIDATORS[opttype](optdict, option, value)
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option.py", line 74, in <lambda>
    "choice": lambda opt, name, value: _choice_validator(opt["choices"], name, value),
  File "/usr/local/lib/python3.7/dist-packages/pylint/config/option.py", line 31, in _choice_validator
    raise optparse.OptionValueError(msg % (name, value, choices))
optparse.OptionValueError: option logging-format-style: invalid value: 'fstr', should be in ['old', 'new']
```

kozlovsky:
fstr option added in Pylint 2.4 and removed in 2.5:
> The `logging-fstring-interpolation` message has been brought back to allow multiple types of string formatting to be used. The type of formatting to use is chosen through enabling and disabling messages rather than through the `logging-format-style` option. *The `fstr` value of the `logging-format-style` option is not valid*.

drew2a:
@kozlovsky thank you for the clarification. It makes sense. Will disable W1203 then.

experiment/popularity_community/crawl_torrents.py
ichorid: > And that should solve all the points on your list, shouldn't it?

No, it doesn't say which *number* means :infinity: , and which means :zero: 

Anyways, that's worth a separate dicscussion. Typical programmer stuff :nerd_face:

experiment/popularity_community/crawl_torrents.py
ichorid: We want to be sure that future changes in RemoteQueryCommunity or any IPv8 related stuff, or *your* minor changes to the crawler will still leave your class working, at least in regards to inter-IPv8 communication. Of course, that's not a proper integration test. That would be *huge*. It is just a small sanity check for anyone who will come later, to make sure the most basic stuff does not break on minor changes.

If you think my comments are slowing you down, by all means, just merge it! This is not some essential core code to put it to such scrutiny.

ichorid (dismissed):

xoriole (dismissed):

devos50 (dismissed):
Don't forget to squash the commits when merging 👍

ichorid:
Well, we only squash commits that are parts of the bigger change. E.g. we *don't* squash "flake8 - config - change" commit with the codebase changing commits. On the other hand, we always squash cosmetic-changes commits.

ichorid:
@drew2a , I highly recommend using `pre-commit` to you. It will automatically `isort` imports. That saves *a lot* of manual labor.

drew2a:
For history:

How to make friends Flake8, PyCharm, and Tribler: https://gist.github.com/drew2a/83efb21eeec3080a9067cdf14aca7d5f

ichorid:
> For history:
> 
> How to make friends Flake8, PyCharm and Tribler: https://gist.github.com/drew2a/83efb21eeec3080a9067cdf14aca7d5f

:+1: 
Just put it into our dev docs folder in the main repo...

ichorid:
Also, we already have `pre-commit` hooks for that :wink:

Issue #5605, opened 2020-10-05
Segmentation fault on startup


da2x:
#### Please, fill all relevant items: ####

- [x] I have NOT read CONTRIBUTING.rst (no such file)
- [x] I have tried with the latest pre-release version and I still can reproduce the issue. (no prerelease)

##### Tribler version/branch+revision: #####
7.5.2 from source

##### Operating system and version: #####
Fedora 32

##### Steps to reproduce the behavior: #####
0. Downloadtribler-7.5.2, unpack, and change into its directory.
1. `sudo dnf python3-configobj python3-aiohttp python3-aiorestapi python3-lz4 python3-pynacl python3-netifaces python3-networkx python3-psutil python3-pyqtgraph python3-pillow-qt`
2. `pip3 install --user -r src/requirements.txt pyipv8` (for anything missing in Fedora repos)
3. `src/tribler.sh`

##### Actual behavior: ##### 
```
[PID:54667] 2020-10-05 14:19:59,721 - INFO - root(24) - __init__()
[PID:54667] 2020-10-05 14:19:59,762 - INFO - root(50) -   self._outSocket.error() = '2'
[PID:54667] 2020-10-05 14:19:59,762 - INFO - root(60) - __init__(): returning
src/tribler.sh: line 43: 54667 Segmentation fault      (core dumped) python3 $TRIBLER_SCRIPT "$@"
```

##### Relevant log file output: #####
gdb  backtrace: https://gist.github.com/da2x/84519514a65ca0e7a7c42e9fb123c918

qstokkink:
Let me preface this by saying we don't have any Fedora developers, so you're basically on your own 😃 Nonetheless, I'll try to help you as best I can.

>  I have NOT read CONTRIBUTING.rst (no such file)

Touché, you can find it here: https://github.com/Tribler/tribler/blob/devel/doc/contributing.rst

> I have tried with the latest pre-release version and I still can reproduce the issue. (no prerelease)

You can find the current `7.5.x`-series release code (which will be `7.5.3`) here: https://github.com/Tribler/tribler/archive/release-7.5.zip

Now, for the real issue.

From your gdb backtrace I see you have some PyQt4 font database initialization issues going on. Which is interesting, as we use PyQt5. This is probably a side effect from installing `python3-pyqtgraph`, but who knows what is actually going on there. If you want a fighting chance at getting this running, I'd recommend removing the packages you installed and sticking to this guide:

[...]

PR #5618, opened 2020-10-06
Set PYQTGRAPH_QT_LIB="PyQt5"

da2x:
PyQtGraph [prefers PyQt4](https://github.com/pyqtgraph/pyqtgraph/blob/0a8ad6b1aae733a2dbd5026b6b1011238aa75548/pyqtgraph/Qt.py#L29) when it and PyQt5 are installed side-by-side.
[This variable overrides](https://github.com/pyqtgraph/pyqtgraph/blob/0a8ad6b1aae733a2dbd5026b6b1011238aa75548/pyqtgraph/Qt.py#L22) that preference to PyQt5 as required by Tribler.

Resolves issue #5605.

### DIFF
--- a/src/tribler.sh
+++ b/src/tribler.sh
@@ -15,6 +15,9 @@ export PYTHONPATH
 
 TRIBLER_SCRIPT=$SRC_DIR/run_tribler.py
 
+PYQTGRAPH_QT_LIB="PyQt5"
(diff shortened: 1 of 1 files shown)


xoriole:
ok to test

devos50 (approved):
I've tested it and it seems to work fine 👍 

Thank you for your contribution!

Issue #5614, opened 2020-10-06
Newly created torrents ignore anonymous seeding preference


da2x:
##### Tribler version/branch+revision: #####
7.5.2

##### Operating system and version: #####
* Fedora Linux 32
* Windows 10

##### Steps to reproduce the behavior: #####
0. Open settings and ensure anonymous seeding is enabled (default)
1. Click + and select Create torrent from files
2. Give it a great name
3. Add some unique creative work (e.g. a screenshot or three)
4. Leave the default checkboxes (including seed this torrent) and click Create torrent

##### Expected behavior: #####
Should respect anonymous seeding preference and seed the torrent anonymously.

##### Actual behavior: #####
Ignores settings and creates a zero-hop torrent.

##### Workaround: #####
Uncheck the "Seed this torrent after creation" checkbox in torrent creator dialog, and manually import the torrent file afterwards.

devos50:
Thank you for your report!

da2x:
What’s up with the naming difference in download_config vs tribler_config?

https://github.com/Tribler/tribler/blob/c98871b033ac9e1ccc9cecd242734e8b9bf3e1d0/src/tribler-core/tribler_core/modules/libtorrent/download_config.py#L90
https://github.com/Tribler/tribler/blob/1900ab766b25166196fe38f469a128a63eb732d7/src/tribler-core/tribler_core/config/tribler_config.py#L530

devos50:
@da2x at a first glance, that looks like a bug to me, probably introduced when we refactored (parts of) the configuration mechanism.

@egbertbouman can you have a look at this?

egbertbouman:
The naming in the config is weird, for sure. Different names for `safe_seeding`, but also the naming of `download_defaults` in `DownloadConfig`, which actually aren't defaults at all.

Having said that, that's not the issue here. It's the `CreateTorrentEndpoint` that's not setting the right hop count.

PR #5619, opened 2020-10-07
Set safe_seeding default to True

da2x:
Partially resolves issue #5614 by enabling safe seeding by default. This change aligns the default download config with [the core `safeseeding_enabled` option](https://github.com/Tribler/tribler/blob/1900ab766b25166196fe38f469a128a63eb732d7/src/tribler-core/tribler_core/config/tribler_config.py#L530). The new default is a safer choice for people who need the anonymity offered by Tribler.

This doesn’t fix the real issue of this being two different settings for the same thing with only one of them being toggleable in the UI.


xoriole:
ok to test

devos50:
retest this please

devos50:
retest this please

da2x:
Fixed in #5621.

Issue #4656, opened 2019-07-04
documentation is very out of date, especially Trustchain


qstokkink:
https://github.com/Tribler/tribler/blob/devel/doc/trustchain.rst

We still refer to Dispersy even 😃

synctext:
Got feedback from an excellent technical writer on working for us. Writing down priorities and schedule for a 4 month documentation refactoring.

|  Duration | activity | Audience |
|-----|-----|---|
| ??? weeks | Conduct deep-dive into Tribler as preparation; [rewriting the main wiki project overview page](https://github.com/Tribler/tribler/wiki#tribler-an-attack-resilient-micro-economy-for-media) plus short [readme.md](https://github.com/Tribler/tribler#tribler) | General Public |
| ? weeks | 1 web page fluffy promo. Accomplishments of past 17 years, [self-replication AI and alternative Big Tech](https://github.com/Tribler/tribler/issues/4659) in future | General Public | 
| ? weeks | promo: "tokens for cultural heritage". From Trustchain, [credit mining](https://github.com/Tribler/tribler/issues/3337), with [artist income of 100%](https://github.com/Tribler/tribler/issues/3326#issuecomment-507672326) to [trust function against spam and evilness](#10)  | General Public |
| ? weeks | [Trustchain](https://github.com/Tribler/tribler/wiki/Multichain-specifications) mild overview | General Public |
| ? weeks | [Trustchain IETF](https://tools.ietf.org/id/draft-pouwelse-trustchain-01.html) technical documentation |  technical |
| ? weeks | [IPv8](https://github.com/Tribler/py-ipv8/tree/master/doc) | developers |

drew2a:
We do not have `trustchain.rst` anymore.

Issue #3784, opened 2018-08-14
Switch from jenkins-ci.tribler.org to jenkins.tribler.org


qstokkink:
Our main Jenkins should live at `jenkins.tribler.org`. The student environment should be something else.

Also, we'll have to update the documentation.

xoriole:
`jenkins-ci.tribler.org` is switched to `jenkins.tribler.org` and both domains are valid so no links (in the documentation) with any of the domain names should break. Closing this issue for now.

Issue #2719, opened 2017-01-14
Detect PR documentation spelling regression


qstokkink:
Right now the documentation spelling checker only compiles a report of the spelling errors in a PR's documentation. Ideally we would like to have this fail if the amount of spelling errors increases.

devos50:
That would be nice indeed :)

qstokkink:
@devos50 Should we include the spelling checker as a submodule so people can check their spelling offline (instead of having to fire a PR)?

devos50:
@qstokkink I don't think people that are working on documentation will do that frequently so imo it's fine to keep it in the PR check. Also, I would like to keep the stuff in the repo Tribler-related and not pollute it with external tools.

qstokkink:
Alright, the PR check now shows the added violations and fails the spelling check if new unknown words are introduced:
![check_preview](https://cloud.githubusercontent.com/assets/3630389/21955442/9363206a-da6b-11e6-8b68-e7a22c5ea22f.png)

PR #5622, opened 2020-10-08
Add requirements-test.txt

drew2a:
## The problem

We have two environments: 
* Pure Tribler
* Test Tribler

Pure Tribler installs only necessary dependencies, they located in [src/requirements.txt](https://github.com/drew2a/tribler/blob/requirements-dev/src/requirements.txt).

But there is no documentation and no `requirements.txt` for the Test Tribler.

## Solution

I've added [src/requirements-dev.txt](https://github.com/drew2a/tribler/blob/requirements-dev/src/requirements-dev.txt) and [small instruction](https://github.com/drew2a/tribler/blob/requirements-dev/src/README.md) on how to run tests.

Maybe `requirements-dev.txt` is a bad name, and better to use `requirements-test.txt`.

What do you think?

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,32 @@
+# Run Tribler
+
+Full instruction: [install requirements](https://github.com/Tribler/tribler#setting-up-your-development-environment)
+
+```
+python3 -m pip install -r requirements.txt
+
+tribler.sh
+```
+
+# Run Tests
+
+Install all necessary dependencies:
+```
+python3 -m pip install -r requirements-test.txt
+```
+Note: `requirements-test.txt` already contains all requirements 
+from` requirements.txt`.
+
+##
+
+Export to PYTHONPATH the following directories:
+
+* tribler-common
+* tribler-core
+* tribler-gui
+
+Execute:
+```
+python3 -m pytest tribler-core
+python3 -m pytest tribler-gui --guitests
+```
--- a/src/requirements-test.txt
+++ b/src/requirements-test.txt
@@ -1,0 +1,11 @@
+-r requirements.txt
+
+pytest
+pytest-asyncio                        
+pytest-cov                            
+pytest-mock
+pytest-randomly                        
+pytest-timeout                         
+pytest-xdist
+
+asynctest


kozlovsky:
If these requirements are for tests specifically, I like the name `requirements-test.txt` better. But what about things like `pre-commit`, maybe it should also be specified? Then it probably should be `requirements-dev.txt`.

Maybe a good idea to specify all necessary tools for new developers to easily join the project.

src/requirements-dev.txt
```diff
@@ -0,0 +1,15 @@
+-r requirements.txt
+
+pytest
+pytest
```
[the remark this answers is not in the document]
drew2a: Thanks, fixed

qstokkink:
I'm impartial towards `-test.txt` or `-dev.txt`, both are used by other projects.

I would like to see the documentation in the `docs/` folder though, fragmented documentation strewn about in `README.md` files in `src` subdirectories is very hard to maintain.

Nit: <code>export PYTHONPATH=${PYTHONPATH}:`echo {pyipv8,tribler-common,tribler-core,tribler-gui} | tr " " :`</code> does not work on Windows, we should make a Python script to launch the tests in an OS-independent manner.

devos50:
Note that I added `run_tests.sh` and `run_tests.bat` in the `tribler/src` directory recently.

src/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Run tests:
+
+```
+export PYTHONPATH=${PYTHONPATH}:`echo {pyipv8,tribler-common,tribler-core,tribler-gui} | tr " " :`
+
+python3 -m pytest tribler-core
+python3 -m pytest tribler-common
+python3 -m pytest tribler-gui
```
devos50, line 26: Running the GUI tests requires the `--guitests` flag, since they are heavy and spawn separate processes.

src/README.md
```diff
@@ -0,0 +1,27 @@
...
+python3 -m pip install -r requirements-dev.txt
+```
+
+Run tests:
+
+```
+export PYTHONPATH=${PYTHONPATH}:`echo {pyipv8,tribler-common,tribler-core,tribler-gui} | tr " " :`
+
+python3 -m pytest tribler-core
+python3 -m pytest tribler-common
```
devos50, line 25: Are there any tests in `tribler-common`? 🤔

src/requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
+-r requirements.txt
+
+pytest
+pytest-asyncio                        
+pytest-cov                            
+pytest-flakefinder
```
devos50, line 6: As far as I'm aware, we are not using `pytest-flakefinder` since it conflicts with `pytest-asyncio`.

src/requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
+-r requirements.txt
+
+pytest
+pytest-asyncio                        
+pytest-cov                            
+pytest-flakefinder                     
+pytest-forked
```
devos50, line 7: `pytest-forked` is a dependency of `pytest-xdist` and therefore does not not need to be listed explicitly in the requirements.

src/requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
...
+pytest-cov                            
+pytest-flakefinder                     
+pytest-forked                          
+pytest-mock                            
+pytest-randomly                        
+pytest-timeout                         
+pytest-xdist
+
+asynctest
+coverage
```
devos50, line 14: Why is `coverage` required?

src/README.md
drew2a: fixed

src/README.md
drew2a: not yet

src/requirements-dev.txt
drew2a: fixed

src/requirements-dev.txt
drew2a: If it is `requirements.txt` for developers, it makes sense.
Have changed requirements to `requirements.txt` for tests only.

src/requirements-dev.txt
drew2a: fixed

src/README.md
devos50: Then I don't see a need to run the tests in `tribler-common` :)

drew2a: Removed.

drew2a:
@kozlovsky I think it's a common courtesy to have such documentation in a project. I'm not sure we are the best persons for doing this, but we can try to write it.

@qstokkink good to know you are worried about documentation maintaining.

First of all: the most valuable part of this PR is `requirements-test.txt`, not documentation.

Next, I'm also worried about maintaining the documentation and I'd like to see the documentation changed every time the code changed. I understand that this is difficult. How to do this is less painful is to keep the documentation as close to the code as possible.

Note: current project documentation is quite outdated and fragmented.

xoriole (approved):

src/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+tribler.sh
+```
+
+# Run Tests
+
+Install all necessary dependencies:
+
+```
+python3 -m pip install -r requirements-test.txt
```
kozlovsky, line 16: Nitpick: from the description it is not clear, does `requirements-test.txt` include dependencies from `requirements.txt`, or dependences from `requirements.txt` should be installed as well.

kozlovsky (approved):

qstokkink:
> @qstokkink good to know you are worried about documentation maintaining.
> [...]
> Next, I'm also worried about maintaining the documentation and I'd like to see the documentation changed every time the code changed. I understand that this is difficult. How to do this is less painful is to keep the documentation as close to the code as possible.
> 
> Note: current project documentation is quite outdated and fragmented.

Historically, nobody gives a poop about the documentation. If you can somehow turn this around, I am all ears. Right now, I just perform a check-up manually every so often (apparently yearly): #2719, #3770, #3784, #4656.

src/README.md
drew2a: Fixed

qstokkink (approved):

drew2a:
@qstokkink we can accept our [Definition of Done](https://www.leadingagile.com/2017/02/definition-of-done).
One item in `DoD` will be "All changes are reflected in the documentation".

Looks like a good topic for further discussion :)

Examples:
* https://docs.gitlab.com/ee/development/contributing/merge_request_workflow.html#definition-of-done
* https://github.com/Privatix/dappctrl/blob/master/CONTRIBUTING.md#pull-request-process
* https://github.com/arvidn/libtorrent/blob/RC_2_0/CONTRIBUTING.rst#pull-request-checklist

qstokkink:
@drew2a good point, we should probably mention the documentation here somewhere (at least): https://github.com/Tribler/tribler/blob/devel/doc/contributing.rst#pull-requests.

From your examples, this is the only one that I would actually expect developers to take the time to read: https://github.com/Privatix/dappctrl/blob/master/CONTRIBUTING.md#pull-request-process. It's categorized and concise. It would be cool if we could converge to something similar.

PR #5626, opened 2020-10-11
Added simple bandwidth accounting mechanism

### LINKED ISSUE #5255: Token balance graph error
Somehow the first and last datapoint is connected with 2nd "Bytes taken" line.
Screenshot from 2020-03-28 13-12-08

### LINKED ISSUE #5623: A simple mechanism for bandwidth accounting
#4719 discussed the excellent insight that bandwidth accounting does not require hard security requirements for our upcoming economy. In particular, an adequate reputation mechanism already (partially) addresses many security concerns, including the Sybil Attack and fake transaction creation.

To this end, @grimadas and I have devised a very simple, yet effective approach to account bandwidth contributions in Tribler. The upcoming algorithm does not impose a personal ledger structure, therefore significantly simplifying the design and implementation. A compelling advantage is the storage overhead of our algorithm is kept to a minimum. At the minimum, individuals are only required to store the last transaction for every counterparty they have interacted with. Each transaction contains two signatures, one from each counterparty. Each transaction represents a single edge in the work graph and contains the weight of the edge between the interacting parties. Transactions contain an ever-growing counter, indicating the total amount of bandwidth token with a specific counterparty. Therefore, transactions can easily be merged when a state inconsistency is detected. That's all there is.

A key design decision is that transactions will be relatively small, e.g., at most 50MB. Therefore, it is not a big problem if a counterparty goes offline, or when a peer refuses to sign a specific transaction. Like TrustChain, liveness is guaranteed in the presence of network churn, and users can engage in multiple transactions at the same time.

Implementing this approach allows us to get rid of the TrustChain implementation. TrustChain will move to the AnyDex repository since it is still used to account trades in the decentralized market.

## The approach

The following algorithm is an essential step towards a secure micro-economy around bandwidth transfers. It requires minimal computation and state storage overhead. We exemplify our accounting approach using a simple example, where party A payouts 20MB to party B. It consists of the following seven steps:

```
1) A creates  T = (sig_a, EMPTY_SIG, total_taken_a_from_b = db.total_taken_a_from_b + 20)
2) A sends T to B
3) B verifies T
	3a) verify truthfulness of data in T
	3b) verify validity of signature sig_a
	3c) compare T against the last known state of A => total_taken_a_from_b > db.total_taken_a_from_b must hold
4)
	4a) If T is valid, create T’ = (sig_a, sig_b, total_taken_a_from_b) and commit to DB
	4b) If T is not valid, send the last (dual-signed) T’ to A
5) send T’ it to A
6) A verifies T’
	6a) verify validity of signature sig_a and sig_b
	6b) check that T’ is the last state => total_taken_a_from_b > db.total_taken_a_from_b
7)
	7a) if T’ is valid, commit to DB
	7b) if T’ is not valid, ignore T’
```

## Reconstructing the work graph

To devise input for the reputation mechanism, we require peers to reconstruct the work graph. This can easily be achieved by requesting the set of latest transactions  from a counterparty. This collection process can run at a very low speed (with minimal impact on other components in Tribler), and with low computational overhead. Eventually, each peer will build the work graph and devise a ranking of peers in the network through local computations.

## Subjective reputation mechanism

In parallel to the implementation of the above algorithm, an efficient reputation mechanism will be implemented. This progress will be tracked in a separate issue.

This ticket will be updated with results/issues arising as the implementation progresses.

devos50:
This PR implements a first, simple version of the bandwidth accounting mechanism introduced in #5623. Even though the mechanism is simple, the PR has grown a bit more than I anticipated since I had to also make various GUI changes, and I had to write/improve a few tests.

I also made a few modifications to the trust screen. Specifically, we now only show the evolution of token balances. This simplified the implementation and should be less confusing for the users. Since we only have one line now, I used the Tribler orange for the pen/brush colour 👍 

Schermafbeelding 2020-10-16 om 17 59 42

Fixes #5623, fixes #5255 

TODO:
- [x] ~Discuss/settle on the tunnel community upgrade~ We will do a hard upgrade of the tunnel community for Tribler 7.6. This will include a few breaking changes in the tunnels.
- [x] Update the database backend to use Pony.
- [x] Add sequence number + timestamp to each transaction.
- [x] Implement a basic Gumby validation experiment (can be found [here](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_payouts/))
- [x] Fix GUI statistics screen.
- [x] ~Either fix the GUI trust graph or remove it (to be decided)~ @xoriole will update the trust screen later. I will open a new issue when this PR has been merged.
- [x] Complete typing usage
- [x] Update the current Gumby experiments


src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
```diff
@@ -0,0 +1,90 @@
+import os
+
+from ipv8.database import Database
+
+from tribler_core.modules.bandwidth_accounting.transaction import BandwidthTransaction
+
+DATABASE_DIRECTORY = "sqlite"
+
+
+class BandwidthDatabase(Database):
```
ichorid, line 10: Why don't use Pony here? This is in Tribler, so we get that dependency for free.
Also, don't forget to add the indexes from the beginning. They really help.

grimadas: Overkill atm?

ichorid: Actually, Pony makes code much more readable and easy to modify. We can now even afford a luxury of asking @kozlovsky himiself to help with this :wind_chime:

devos50: I'm currently rewriting the database logic to use Pony instead, thanks for the advice! 👍

devos50:
retest this please

devos50:
(win64 timed out)

src/tribler-core/tribler_core/modules/bandwidth_accounting/transaction.py
```diff
@@ -0,0 +1,200 @@
...
+
+def define_binding(bandwidth_database):
+    db = bandwidth_database.database
+
+    class BandwidthTransaction(db.Entity):
+        """
+        This class describes a bandwidth transaction that resides in the database.
+        """
+        sequence_number = Required(int)
+        public_key_a = Required(bytes)
```
ichorid, line 166: Well, we **really** want indexes on public keys, sequence_number, timestamp, etc. Like on all of these fields, except for the signature. And just for safety, it is better to make signature fields `unique`

src/tribler-core/tribler_core/modules/payout_manager.py
```diff
@@ -25,29 +23,28 @@ async def do_payout(self, mid):
...
-                    await self.bandwidth_wallet.trustchain.sign_block(nodes[0],
-                                                                      public_key=nodes[0].public_key.key_to_bin(),
-                                                                      block_type=b'tribler_bandwidth',
-                                                                      transaction={b'up': 0, b'down': total_bytes})
-                except Exception as e:
-                    self.logger.error("Deferred errback fired: %s", e)
+        self.logger.debug("Received %d nodes for DHT lookup", len(nodes))
+        if nodes:
+            try:
+                await self.bandwidth_community.do_payout(nodes[0], total_bytes)
```
ichorid, line 40: I :heart: how concise this becomes.

src/tribler-core/tribler_core/modules/tests/bandwidth_accounting/test_database.py
```diff
@@ -0,0 +1,118 @@
...
+from tribler_core.modules.bandwidth_accounting.transaction import BandwidthTransaction
+
+
+@pytest.fixture
+def my_key():
+    return default_eccrypto.generate_key('curve25519')
+
+
+@pytest.fixture
+def bandwidth_db(tmpdir, my_key):
```
ichorid, line 20: use ":memory:" DB for testing?

src/tribler-core/tribler_core/modules/tunnel/community/triblertunnel_community.py
```diff
@@ -340,29 +304,61 @@ def update_torrent(self, peers, download):
...
-
-        payload = PayoutPayload.from_half_block(block, circuit_id, base_amount)
-        packet = self._ez_pack(self._prefix, 23, [payload], False)
+        tx = self.bandwidth_community.construct_signed_transaction(peer, amount)
+        self.bandwidth_community.database.BandwidthTransaction.insert(tx)
+        payload = BandwidthTransactionPayload.from_transaction(tx, circuit_id, base_amount)
+        packet = self._ez_pack(self._prefix, 30, [payload], False)
         self.send_packet(peer, packet)
 
+    def on_payout(self, source_address: AddressType, data: bytes) -> None:
```
ichorid, line 323: Does it have to be here, or is it better to move it to `BandwidthAccountingCommunity`? I see almost no usages of Tunnels-related properties and methods here.

ichorid (changes requested):
Also, could you please add a `misc` table to the database to identify the schema version? Alternatively, you can embed the schema ID in the filename, something like `bandwidth_v1.db`

ichorid (changes requested):
Otherwise, perfect PR :+1:

src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
```diff
@@ -0,0 +1,114 @@
...
+        return self.get_balance(self.my_pub_key)
+
+    @db_session
+    def get_num_peers_helped(self, public_key: bytes) -> int:
+        """
+        Return the number of unique peers that a peer with the provided public key has helped.
+        :param public_key: The public key of the peer of which we want to determine this number.
+        :return The unique number of peers helped by the specified peer.
+        """
+        result = list(select(count(g.public_key_b) for g in self.BandwidthTransaction if g.public_key_a == public_key))
```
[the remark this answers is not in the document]
kozlovsky: Actually, in PonyORM count is distinct by default, so it is not necessary to add `distinct=True`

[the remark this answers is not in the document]
ichorid: Well, that's not clear from the [documentation](https://docs.ponyorm.org/api_reference.html#count)...

src/tribler-core/tribler_core/modules/bandwidth_accounting/history.py
```diff
@@ -0,0 +1,17 @@
...
+    db = bandwidth_database.database
+
+    class BandwidthHistory(db.Entity):
+        """
+        This ORM class represents a mutation of ones bandwidth balance.
+        We store the last 100 mutations in ones bandwidth balance.
+        """
+
+        rowid = PrimaryKey(int, auto=True)
+        timestamp = Required(int, size=64)
```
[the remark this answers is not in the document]
kozlovsky: Looking at queries, I think the following indexes would be useful:
* composite_key(public_key_a, public_key_b), for searching by pair of (key_a, key_b) as well as just key_a;
* public_key_b
* timestamp (for ordering)

I don't see balance mentions in query conditions, so probably it is better not to create an index for it, as each unnecessary index slowdowns inserts a bit.

src/tribler-core/tribler_core/modules/bandwidth_accounting/transaction.py
```diff
@@ -0,0 +1,200 @@
+"""
+This file defines the required data structures for the bandwidth accounting mechanism.
+Note that we define two different class types of BandwidthTransaction: one for the object that resides outside the
+Pony database and another one that represents the class inside Pony.
+We make this separation to workaround the fact that Pony does not support objects that are created outside a database
```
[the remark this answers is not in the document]
kozlovsky: As I understand, the motivation for creating two different classes is because of two reasons:
1) Pony does not allow creating objects with `Required` fields which are not filled yet, and there were some places in the code where it is desirable to create a partially filled object (is it still so? @devos50, can you point to places in the code where it is desirable to have partially filled objects?)
2) Pony currently does not allow to have "detached" objects, which are created but not stored in the database as fast as possible. Like, if you want to have some long-lived in-memory object not stored in the database and not tied to a particular db_session.

If these reasons are still valid, I think it is normal to create such objects. The code repetition is actually minimal, and the methods are pretty different.

But it may be confusing that the two classes have the same name, `BandwidthTransaction`. How about name non-ORM class something like `BandwidthTransactionData`?

Also, it is possible to define the non-ORM class as a `@dataclass` to remove some boilerplate code:
https://docs.python.org/3.8/library/dataclasses.html

src/tribler-core/tribler_core/modules/bandwidth_accounting/transaction.py
kozlovsky: @ichorid, unique indexes for signature fields can be detrimental for insert performance.
Also, what about a very small chance for two different transactions to have the same signature? Do we want to receive an exception while adding the second transaction to the database?

src/tribler-core/tribler_core/modules/bandwidth_accounting/history.py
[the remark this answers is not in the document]
devos50: Since we only store the last 100 balance mutations (by default), the number of rows should be small. @kozlovsky do you think it would still be worth to add indices?

src/tribler-core/tribler_core/modules/tunnel/community/triblertunnel_community.py
devos50: Yeah this is not very pretty indeed, but I have good reasons to put it here. The main motivation is that a bandwidth payout in the tunnel community is slightly different from a direct payout (e.g., as the result of a libtorrent alert). Specifically, we want to include the circuit identifier when paying out to the next hop in a circuit so the subsequent hop knows which circuit the payment is associated with. Thus, we are using different IPv8 messages for these two forms of payouts.

I tried to think about a design where I just implemented `on_payout` once in `BandwidthAccountingCommunity` but that would result in code that was harder to read and maintain due to subtile differences between these two forms of payouts.

src/tribler-core/tribler_core/modules/bandwidth_accounting/transaction.py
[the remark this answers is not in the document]
devos50: Thanks for the `@dataclass` decorator! I didn't know about that :)

src/tribler-core/tribler_core/modules/bandwidth_accounting/transaction.py
devos50: Since our queries are only using the public keys currently, I have added indices on the public keys.

src/tribler-core/tribler_core/modules/bandwidth_accounting/history.py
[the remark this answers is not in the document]
kozlovsky: For such a small number of rows, indices are unnecessary, as the number of database pages we read will be very small anyway.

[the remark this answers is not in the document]
devos50: Ok thanks! 👍

devos50:
@ichorid @grimadas @kozlovsky processed the feedback 👍 

Note that I also have started to remove TrustChain constructs. There are, however, still a few things left that would require some more refactoring and changes. For example, the variable that references the (private) key that we are using in most communities is named `trustchain_keypair`. We should rename that in another PR.

I also updated the identifiers of the tunnel community, after communicating with @egbertbouman. I will deploy an exit node with the new code after this PR is merged.

Finally, the trust graph in the GUI is broken after this PR is merged. @xoriole already confirmed that he will fix this later.

kozlovsky (approved):

src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
```diff
@@ -0,0 +1,122 @@
...
+        self.db_path = db_path
+        self.my_pub_key = my_pub_key
+        create_db = str(db_path) == ":memory:" or not self.db_path.is_file()
+        self.database = Database()
+
+        self.MiscData = misc.define_binding(self.database)
+        self.BandwidthTransaction = transaction.define_binding(self)
+        self.BandwidthHistory = history.define_binding(self)
+
+        self.database.bind(provider='sqlite', filename=str(db_path), create_db=str(create_db), timeout=120.0)
```
kozlovsky, line 32: `create_db` option is a bool flag, conversion to `str` is probably error?

kozlovsky (changes requested):

src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
devos50: Good catch! I copied this particular line from the metadata store that was already there, so I guess it would also be a bug there?

ichorid (dismissed):
Fine, except for @kozlovsky 's remark.

src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
ichorid: Yep! Definitely a bug in my code :clap: 
That's the most prominent vector of bug proliferation - blind copy-paste.

devos50:
Win64 tests timeout... (we should really debug this one)

devos50:
retest this please

kozlovsky (approved):

ichorid (approved):

Issue #5262, opened 2020-04-01
DNS leak?


ichorid:
Some of our users showed concern about their system's resolver leaking DNS queries when checking trackers, etc. We should investigate it.

qstokkink:
Isn't this fixed in 7.5?

ichorid:
DHT leak is fixed in 7.5. This one is talking about DNS hostnames for trackers, which are served by system resolver.

qstokkink:
Oh my bad, I thought that was also anonymized now.

egbertbouman:
Torrent checking is not anonymous, only `DownloadManager.get_metainfo` is.

qstokkink:
Should we create a feature to only use anonymized DHT lookups?

egbertbouman:
DHT lookups all rely on `get_metainfo`, so they shouldn't leak. To fix any leakage in `TorrentChecker`, we need to allow it to use a SOCKS5 proxy, and also allow TCP scrape requests in the `TunnelCommunity`.

qstokkink:
> To fix any leakage in TorrentChecker, we need to allow it to use a SOCKS5 proxy, and also allow TCP scrape requests in the TunnelCommunity.

Seems like a reasonable feature for `7.6` to me.

egbertbouman:
I was thinking that maybe we should just allow all TCP requests that have a bencoded response. This would automatically allow both scrape and announce requests to TCP trackers while avoiding tracker specific logic/messages in the `TunnelCommunity`.

ichorid:
@egbertbouman , sounds like a (relatively) simple and robust solution.

PR #5628, opened 2020-10-12
Added anonymization support to TorrentChecker

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/5262.

This PR:
* Updates `Dispatcher` to allow multiple simultaneous connections to the tunnel SOCKS server. To achieve this, circuits are reserved for a particular SOCKS session.
* Adds `Socks5Client` that can be used for connecting to our `Socks5Server`. Both TCP and UDP are supported.
* Adds `Socks5Connector` that allows anonymous HTTP connections to be established using `aiohttp`.
* Adds anonymization support to `TorrentChecker`. The `TorrentChecker` now uses the default hop count for all traffic.
* Updates SOCKS5 client/server to use the IPv8 serializer.

### DIFF
--- a/src/tribler-core/tribler_core/modules/torrent_checker/test_torrentchecker.py
+++ b/src/tribler-core/tribler_core/modules/torrent_checker/test_torrentchecker.py
@@ -121,7 +121,7 @@ async def test_task_select_tracker(enable_chant, torrent_checker, session):
         tracker = session.mds.TrackerState(url="http://localhost/tracker")
         session.mds.TorrentState(infohash=b'a' * 20, seeders=5, leechers=10, trackers={tracker})
 
-    controlled_session = HttpTrackerSession("127.0.0.1", ("localhost", 8475), "/announce", 5)
+    controlled_session = HttpTrackerSession("127.0.0.1", ("localhost", 8475), "/announce", 5, None)
     controlled_session.connect_to_tracker = lambda: succeed(None)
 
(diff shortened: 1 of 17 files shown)


qstokkink:
Quick pytest fixture note. You can fix the `W0621` Pylint errors using this construction (this example exposes the name `ipv8` as a fixture):

https://github.com/Tribler/tribler/blob/9dfba003c74d004ba9452ef5e7d931ec00eb5497/src/tribler-core/tribler_core/modules/tests/test_ipv8_health_manager.py#L19-L20

egbertbouman:
@qstokkink Thanks, will do!

qstokkink (approved):
Cool 👍

egbertbouman:
I did some more testing. I've created a list of 28 HTTP trackers that are verified to be working. With some tweaking, I managed to successfully scrape 27 of them using the `TunnelCommunity`. The remaining one was redirecting to HTTPS (which we don't support).

ichorid:
So, we now have working HTTP tunnelling in Tribler?

egbertbouman:
@ichorid Yes, we do. That doesn't mean that it's possible to browse websites, though. HTTP tunneling is only allowed if the response is bencoded.

egbertbouman:
retest this please

qstokkink (approved):

egbertbouman:
retest this please

Issue #5625, opened 2020-10-11
[Popularity Community] Implement experiment 3


drew2a:
Extracted and adapted from https://github.com/Tribler/tribler/issues/3868

## Description of the experiment 

Given:

* Real network

Actions:

* Crawl random nodes
* Request from each crawled node top 100 torrents
* Calculate torrent infohash distribution among crawled nodes

Assumption: popular torrents (top 20?) must be represented on all nodes. 

The difference between the desired representation and the given representation should be a quality assessment of `Popularity Community`.

## Notes

We also can create a list of "trending" torrents, based on `infohash distribution`.
This experiment can be run on Jenkins every day.

The first part of the work (crawler) has been completed in https://github.com/Tribler/tribler/pull/5611

Jenkins job: https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_and_analyze_torrents/

drew2a:
Crawl speed, based on: [build #3](https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_torrents/3/)

After 1 hour, we polled 513 nodes.

![image](https://user-images.githubusercontent.com/13798583/95682546-6fc16780-0bee-11eb-91a3-70b5617764ba.png)

Daily statistics: https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_torrents/plot/

drew2a:
Example of analyzed data (310 nodes):
![image](https://user-images.githubusercontent.com/13798583/95752076-29c8da00-0ca8-11eb-8076-65df96b627aa.png)

Here we see torrent distribution among crawled nodes.

The `blue bar` here is a torrent. 
`Digit` is the count of nodes (in percent) in with this torrent has been represented.

drew2a:
The [job in Jenkins](https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_and_analyze_torrents/) has been scheduled to run daily at 7 am.

Artifacts:
* [peers.csv](https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_and_analyze_torrents/lastSuccessfulBuild/artifact/peers.csv) — count of crawled nodes and speed of crawling
* [distribution.csv](https://jenkins-ci.tribler.org/job/popularity_experiments/job/Crawl_and_analyze_torrents/lastSuccessfulBuild/artifact/distribution.csv) — top 100 torrents distribution among crawled nodes

PR #5632, opened 2020-10-13
Experiment 3

drew2a:
This PR contains:
1. Improvements for `crawl_torrents.py`:
   1. Crawler speed has been improved
   1. Console output informativeness has been increased
   1. Added `select limit per request = 20` for preventing overloading
1. Small improvements for `initial_filling_velocity.py` 
1. `analyze_crawled_data.py`

It solves #5625 

## analyze_crawled_data.py

Source: [analyze_crawled_data.py](https://github.com/drew2a/tribler/blob/experiment3/experiment/popularity_community/analyze_crawled_data.py)

### Description

This script analyzes the crawled data.

As an input, it takes `sqlite db`, produced by [crawl_torrents.py](https://github.com/drew2a/tribler/blob/experiment3/experiment/popularity_community/crawl_torrents.py).

The result will be stored in a `json` file:
```
[
  {
    "hash": "3f93ae6aabbba6d07176e778d9cbff088e9fbf1d",
    "title": "Ubuntu",
    "count": 224,
    "ratio": 72
  },
  {
    "hash": "95e8d04147f0d9cb8bf00ec8775c926126a65236",
    "title": "Fedora",
    "count": 139,
    "ratio": 44
  },
  {
    "hash": "ba693a610b43a1aa246d3382299775ebdf0fec34",
    "title": "Linux Mint",
    "count": 136,
    "ratio": 43
  }
]
```

and in `csv` file (short form):

```
position,node_count,torrent_count,ratio
0,55,39,70
1,55,34,61
2,55,30,54
```

`ratio` is a number, calculated as 

\Large l

Where:
 * \Large l  is a ratio for `i-th torrent`
 * \Large l  how many times `i-th torrent` encountered on crawled nodes
* \Large l  is a count of crawled nodes

### Usage

```
export PYTHONPATH=${PYTHONPATH}:`echo ../.. ../../src/{pyipv8,tribler-common,tribler-core} | tr " " :`

python3 analyze_crawled_data.py [-d <sqlite_db_path>] [-f <json_output_file_path>]
                                [-l <torrent_limit>] [-v]
```

Where:
* `sqlite_db_path` means the path to `sqlite` db file
* `json_output_file_path` means the path to result `json` file. 
Note, there is also a `csv` file that will be produced. It will be located on the same path, but with `.csv` extension
* `torrent_limit` is the maximum size of the result set
* `-v` means increase verbosity

 

### Example

```
python3 analyze_crawled_data.py -d crawled_data.sqlite 
python3 analyze_crawled_data.py -d crawled_data.sqlite -f result.json
python3 analyze_crawled_data.py -d crawled_data.sqlite -f result.json -l 100
python3 analyze_crawled_data.py -d crawled_data.sqlite -f result.json -l 100 -v
```

### DIFF
--- a/experiment/popularity_community/README.md
+++ b/experiment/popularity_community/README.md
@@ -81,7 +81,8 @@ Where:
 ```
 export PYTHONPATH=${PYTHONPATH}:`echo ../.. ../../src/{pyipv8,tribler-common,tribler-core} | tr " " :`
 
-python3 crawl_torrents.py [-t <timeout_in_sec>] [-f <db_file.sqlite>]
+python3 crawl_torrents.py [-t <timeout_in_sec>] [-f <db_file.sqlite>] [-v]
+                          [--peers_count_csv=<csv_file_with_peers_count>]
 ```
 
 Where:
@@ -89,6 +90,8 @@ Where:
 * `db_file.sqlite` means the path to `sqlite` db file. 
     If file doesn't exists, then new file will be created. 
     If file exists, then crawler will append it.
+* `csv_file_with_peers_count` means the path to `csv` file that contains
+    `(time, active_peers, crawled_peers)` tuples.
  
 
 ### Example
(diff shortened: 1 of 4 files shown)


kozlovsky:
README says that `ratio` value calculated as `100 * torrent_count / node_count`, but does not explain what is the actual meaning of `torrent_count` and `node_count` and which numbers will change in a series of experiment and which remains stable. Do I understand correctly that `torrent_count` is the count of nodes that have this torrent in their top-100 torrent list, and `node_count` is the total number of crawled nodes?

As I understand it, the goal of the experiment is to be able to compare distributions of torrents in a series of experiments. For that case, providing just a single number of `ratio` might be not enough. It may be better to keep in CSV file the raw values of `torrent_count` and `node_count` as well.

Also, for further calculations, it may be convenient to include a `position` value to each CSV row, which is basically a number of the row in the CSV file, or, equivalently,  the position of the torrent in the combined torrent's list.

drew2a:
@kozlovsky 

> As I understand it, the goal of the experiment is to be able to compare distributions of torrents in a series of experiments. For that case, providing just a single number of ratio might be not enough. It may be better to keep in CSV file the raw values of torrent_count and node_count as well.

Added.

drew2a:
@kozlovsky 
> Also, for further calculations, it may be convenient to include a position value to each CSV row, which is basically a number of the row in the CSV file, or, equivalently, the position of the torrent in the combined torrent's list.

The advantage is not clear to me. Please, give a particular example of using `position`.

kozlovsky:
> The advantage is not clear to me. Please, give a particular example of using position.

Like, if I want to take several experiment's data and compare the rating of 1st, 10th, and 20th most-popular torrents, I need to count rows in a file to find the 10th or 20th row of a file, and with the explicit position value, I can just search for a specific position. It would also be possible to combine information from multiple experiments in a single file without information loss. But I'm ok without explicit torrent position value, as it doesn't seem to be really important right now.

kozlovsky (dismissed):

drew2a:
@kozlovsky thank you for the example. I've added `position`.

xoriole (dismissed):

PR #5633, opened 2020-10-13
Add years and “Forever” (1c+) ETAs

da2x:
Added solar year precision to download ETAs, and labelled downloads that will take over a century as “Forever.” 

Users won’t see these labels except under exceptionally poor network conditions, or for stalled transfers. The following screenshot is from a stalled transfer:

ETA: 51601586521w 5d

I didn’t add months (m) to avoid confusion with minutes (m). Although, maybe that’s not a problem when the m is displayed after a y or in front of a w?


devos50:
ok to test

PR #5644, opened 2020-10-15
Show BEP33 support indicator in the Debug Pane

xoriole:
Adds an entry in the Debug Pane's General tab showing whether or not the build supports BEP33.
Screenshot from 2020-10-15 13-37-11

### DIFF
--- a/src/tribler-gui/tribler_gui/debug_window.py
+++ b/src/tribler-gui/tribler_gui/debug_window.py
@@ -15,6 +15,7 @@ import libtorrent
 import psutil
 
 import tribler_core.utilities.json_util as json
+from tribler_core.utilities.utilities import has_bep33_support
 
 from tribler_gui.defs import DEBUG_PANE_REFRESH_TIMEOUT, GB, MB
 from tribler_gui.dialogs.confirmationdialog import ConfirmationDialog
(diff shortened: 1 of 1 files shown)


xoriole:
retest this please

xoriole:
retest this please

src/tribler-gui/tribler_gui/debug_window.py
```diff
@@ -15,6 +15,7 @@
 import psutil
 
 import tribler_core.utilities.json_util as json
+from tribler_core.utilities.utilities import has_bep33_support
```
devos50, line 18: You are importing from the `tribler_core` in the `tribler_gui` module directly. Even though we have more core imports in the GUI code, this can have unintended side-effects, if `tribler_core.utilities.utilities` contains other imports.

Would there be an easy way to request this information from the core through the REST API?

xoriole: I checked utilities file which seems pretty independent from core so I did the import. Otherwise, I could simply replace the function call with `'dht_pkt_alert' in dir(libtorrent)`. What do you suggest?

devos50: On second thoughts, let's not try to overdo this for now. It's the debug panel, and a search through the GUI code base reveals _many_ architectural violations regarding imports from the core 👍

devos50 (approved):

PR #5650, opened 2020-10-16
Removed IRC mention from readme

devos50:
We usually do not hang out there...


devos50:
Seems that the Windows 64-bit tests are failing. I created an issue for that, see #5651 

I think this PR can be reviewed/merged though.

qstokkink (approved):

PR #5657, opened 2020-10-19
Request a torrent info after new torrent health has been received 

### LINKED ISSUE #5642: [PopularityCommunity] Unknown infohashes obtained through Popularity Community do not show up in Channels
Receiving a new `torrent health info` through Popularity Community does not affect the Channels interface.

When a new `torrent health info` received through Popularity Community, a new `TorrentState` entity is added:

```
self.metadata_store.TorrentState(infohash=infohash, seeders=seeders,
                                                        leechers=leechers, last_check=last_check)
```
https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/popularity/popularity_community.py#L66-L68

But it does not affect the Channels interface because there is no corresponding Torrent entry in `MetadataStore``.

@ichorid please, specify which data should be added to the Database.

drew2a:
Resolves #5642

The new behavior is the following: 
1. When `PopularityCommunity` receives an unknown torrent, it notifies `RemoteQueryCommunity` through the `Notifier` about it.
    ```
    self.notifier.notify(NTFY.POPULARITY_COMMUNITY_ADD_UNKNOWN_TORRENT, peer, infohash)
    ```
1. When `RemoteQueryCommunity` receives the notification, it will request the torrent's information by sending a remote query to the corresponding peer.
   ```
    def on_pc_add_unknown_torrent(self, peer, infohash):
        query = {'infohash': hexlify(infohash)}
        self.send_remote_select(peer, **query)
   ```


src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/metadata_node.py
```diff
@@ -109,6 +110,7 @@ def get_entries_query(
             pony_query = pony_query.where(lambda g: g.status != TODELETE) if exclude_deleted else pony_query
             pony_query = pony_query.where(lambda g: g.xxx == 0) if hide_xxx else pony_query
             pony_query = pony_query.where(lambda g: g.status != LEGACY_ENTRY) if exclude_legacy else pony_query
+            pony_query = pony_query.where(lambda g: g.infohash == infohash) if infohash else pony_query
```
ichorid, line 113: It is more correct to use `if infohash is not None` because otherwise, you will not be able to query on _empty string_ infohashes.

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -84,11 +87,24 @@ def __init__(self, my_peer, endpoint, network, metadata_store, settings=None, no
...
+        if self.notifier:
+            self.notifier.add_observer(NTFY.POPULARITY_COMMUNITY_ADD_UNKNOWN_TORRENT,
+                                       self.on_pc_add_unknown_torrent)
+
         self.add_message_handler(RemoteSelectPayload, self.on_remote_select)
         self.add_message_handler(SelectResponsePayload, self.on_remote_select_response)
 
         self.request_cache = RequestCache()
 
+    def __del__(self):
```
ichorid, line 99: `__del__` is unreliable in Python. It will only be called during garbage cleaning of the object. Python is not C unfortunately... We use a convention to add an explicit `shutdown` method to guarantee proper object deconstruction.

src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/metadata_node.py
drew2a: Agree, but what is wrong with querying an empty string infohashes?

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
drew2a: Thanks.

As I see the proper place to shutdown functionality for the `Community` class is 
```
async def unload(self)
```
Am I right?

drew2a:
@ichorid I've introduced a new query argument for remote select: `infohash` 

```
query = ({'infohash': 'f18461839b5e8bf227ac84c954cf1efbd039b0de'})
```

Should I change `RemoteQueryCommunity.community_id`?

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
ichorid: We use `unload` for communities and `shutdown` for everything else I guess.

src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/metadata_node.py
ichorid: There is nothing wrong on querying an empty string infohashes. The problem is, in Python both empty string and `None` is considered `False` in conditionals. So you should be very precise about it.

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
drew2a: Fixed.

src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/metadata_node.py
drew2a: Okay, that's a good point, thank you!

ichorid:
Regarding the `community_id` that's always a problem... Technically speaking, RemoteQueryCommunity does not change the format. It is the underlying MetadataStore that changed its behaviour. Typically, we prefer to change the ID in such cases, though doing it will cut the peers running the new version from the older one. So the *proper* way to do it is to go dual-stack. 

However, dual-stack is too much hassle in this case. So, just test what happens to older versions when they get a query with this new field. 

Also, there is another way to do it: you can try exploiting `attribute_ranges` to query for infohashes. AFAIK, SQLite allows range [comparisons on blobs](https://stackoverflow.com/questions/24011247/fast-search-on-a-blob-starting-bytes-in-sqlite).

src/tribler-core/tribler_core/tests/test_notifier.py
```diff
@@ -1,14 +1,38 @@
...
+
 from tribler_common.simpledefs import NTFY
 
 from tribler_core.notifier import Notifier
 
 
-def test_notifier():
-    notifier = Notifier()
+@pytest.fixture
+def notifier():
```
qstokkink, line 10:
```suggestion
@pytest.fixture(name="notifier")
def fixture_notifier():
```

Should avoid `# pylint: disable=redefined-outer-name`.

drew2a: @qstokkink 👍

drew2a:
retest please

drew2a:
retest this please

src/tribler-core/tribler_core/modules/popularity/popularity_community.py
```diff
@@ -49,23 +50,28 @@ def gossip_torrents_health(self):
...
-                        torrent_state.last_check = last_check
-                    elif not torrent_state:
-                        _ = self.metadata_store.TorrentState(infohash=infohash, seeders=seeders,
-                                                             leechers=leechers, last_check=last_check)
-
-            self.metadata_store.disconnect_thread()
-        await get_event_loop().run_in_executor(None, _put_health_entries_in_db)
+        await self.put_health_entries_in_db(peer, all_torrents)
+
+    @db_session
```
xoriole, line 60: Does this work with async function? @kozlovsky

kozlovsky: No, it should not work. It is necessary to put db_session inside:

```python
    async def put_health_entries_in_db(self, peer, all_torrents):
        with db_session:
            for infohash, seeders, leechers, last_check in all_torrents:
```
(and never call functions with `await` inside db_session context manager body)

drew2a: @xoriole @kozlovsky 
Fixed. Thanks for pointing this out!

Please check the current implementation:
```
 @lazy_wrapper(TorrentsHealthPayload)
    async def on_torrents_health(self, peer, payload):
        self.logger.info("Received torrent health information for %d random torrents and %d checked torrents",
                         len(payload.random_torrents), len(payload.torrents_checked))
        all_torrents = payload.random_torrents + payload.torrents_checked

        with db_session:
            for infohash, seeders, leechers, last_check in all_torrents:
                torrent_state = self.metadata_store.TorrentState.get(infohash=infohash)
```

xoriole (approved):

kozlovsky (approved):

src/tribler-core/tribler_core/modules/popularity/popularity_community.py
ichorid: This should work. The idea is that `db_session` should never span asynchronous functions.

PR #5668, opened 2020-10-21
Gossip only healthy torrents

### LINKED ISSUE #5658: [PopularityCommunity] Gossip only healthy torrents
Based on [current data](https://jenkins-ci.tribler.org/job/popularity_experiments/job/Initial_filling_velocity/plot/) count of received alive torrents is relatively small.

To increase this value I propose to gossip through `PopularityCommunity`  only `healthy` torrents.

drew2a:
Resolves #5658 

New `PopularityCommunity` behavior: it gossips only healthy torrents. 
Here healthy means have `seeders>0`.
Note: gossiping is for checked torrents only.

##

Select torrent logic is isolated in `select_torrents_to_gossip` function:

```python
class PopularityCommunity(Community):
    @staticmethod
    def select_torrents_to_gossip(torrents):
        # select the torrents that have seeders
        alive = set((_, seeders, *rest) for (_, seeders, *rest) in torrents if seeders > 0)

        # select 5 most popular from alive torrents, using `seeders` as a key
        count = PopularityCommunity.GOSSIP_POPULAR_TORRENT_COUNT
        popular = set(heapq.nlargest(count, alive, key=lambda t: t[1]))

        # select 5 random torrents from the rest of the list
        rest = alive - popular
        count = min(PopularityCommunity.GOSSIP_RANDOM_TORRENT_COUNT, len(rest))
        rand = set(random.sample(rest, count))

        return popular, rand
```

##

As part of refactoring, `PopularityCommunity.process_torrents_health` has been extracted from `PopularityCommunity.on_torrents_health` to make DB work async.

### DIFF
--- a/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
+++ b/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
@@ -1,3 +1,5 @@
+import asyncio
+import heapq
 import random
 from binascii import unhexlify
 
@@ -11,13 +13,20 @@ from tribler_common.simpledefs import NTFY
 from tribler_core.modules.popularity.payload import TorrentsHealthPayload
 from tribler_core.utilities.unicode import hexlify
 
-PUBLISH_INTERVAL = 5
-
 
 class PopularityCommunity(Community):
     """
-    Community for disseminating the content across the network. Follows publish-subscribe model.
+    Community for disseminating the content across the network.
+
+    Every 5 seconds it gossips 5 the most popular torrents and 5 random torrents to
+    a random peer.
+
+    Gossiping is for checked torrents only.
     """
+    GOSSIP_INTERVAL = 5
+    GOSSIP_POPULAR_TORRENT_COUNT = 5
+    GOSSIP_RANDOM_TORRENT_COUNT = 5
+
     community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1648')
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


qstokkink:
Quick tip while this is still in WIP: the `TorrentsHealthPayload` `fix_pack_*` and `fix_unpack_*` stuff was a work-around to stay wire-format compatible with a previous version. If you're changing the wire format anyway, you could use the new `ListOf` with a new packer derived from `TorrentInfoFormat`. Something like this (@egbertbouman correct me if I'm wrong):

```python
# payload.py (excluding imports)
TorrentInfo = collections.namedtuple('TorrentInfo', ['infohash', 'seeders', 'leechers', 'timestamp'])

class TorrentInfoPacker:
    def pack(self, torrentinfo):
        return struct.pack('>20sIIQ', *torrentinfo)

    def unpack(self, data, offset, unpack_list):
        unpack_list.append(TorrentInfo(*struct.unpack_from('>20sIIQ', data, offset)))
        return offset + 36

@vp_compile
class TorrentsHealthPayload(VariablePayload):
    msg_id = 1
    format_list = ['torrents-list']
    names = ['torrents']

# community.py (insert into Community class)
def get_serializer(self):
    serializer = super().get_serializer()
    serializer.add_packer('torrents-list', ListOf(TorrentInfoPacker()))
    return serializer
```

drew2a:
> Quick tip while this is still in WIP: the `TorrentsHealthPayload` `fix_pack_*` and `fix_unpack_*` stuff was a work-around to stay wire-format compatible with a previous version. 

Ok, thanks, will change it 👍

kozlovsky (dismissed):

src/tribler-core/tribler_core/modules/popularity/popularity_community.py
```diff
@@ -29,36 +38,77 @@ def __init__(self, *args, **kwargs):
...
+
+        Args:
+            torrents: set of tuples (infohash, seeders, leechers, last_check)
+
+        Returns:
+            tuple (set(popular), set(random))
+
+        """
+        # select the torrents that have seeders
+        alive = set((_, seeders, *rest) for (_, seeders, *rest) in torrents
```
xoriole, line 60: I think here, if the `alive` set is empty then simply two empty set or None could be returned.

drew2a: Added

xoriole (dismissed):

drew2a:
retest this please

PR #5681, opened 2020-10-25
Update IPv8 pointer

egbertbouman:


devos50 (approved):
Looks good 👍 

Do you have more plans for breaking changes? Now's the time to do so 😀

egbertbouman:
@devos50 That's all for now.

At some point, I'd like to refactor the crypto, add support for different types of crypto, and add link encryption. However, I don't think it's possible to implement that in a timely fashion.

PR #5682, opened 2020-10-25
Disable #5642

### LINKED ISSUE #5677: [7.5.3] RemoteQueryCommunity - got an unexpected keyword argument 'infohash'
From error reporter [#36133](http://reporter.tribler.org/stacktrace/36133) 
```python
File "tribler_gui/event_request_manager.py", line 138, in <lambda>
File "tribler_gui/event_request_manager.py", line 80, in on_error
tribler_gui.event_request_manager.CoreConnectTimeoutError: Could not connect with the Tribler Core within 60 seconds
[PID:2608] 2020-10-23 14:44:34,910 - ERROR <taskmanager:135> RemoteQueryCommunity.done_cb(): Task resulted in error: get_entries_query() got an unexpected keyword argument 'infohash'
File "ipv8/taskmanager.py", line 131, in done_cb
File "tribler_core/modules/metadata_store/community/remote_query_community.py", line 124, in on_remote_select
File "tribler_core/modules/metadata_store/orm_bindings/metadata_node.py", line 140, in get_entries_threaded
File "concurrent/futures/thread.py", line 57, in run
File "tribler_core/modules/metadata_store/orm_bindings/metadata_node.py", line 135, in _get_results
File "<string>", line 2, in get_entries
File "pony/orm/core.py", line 528, in new_func
File "tribler_core/modules/metadata_store/orm_bindings/metadata_node.py", line 150, in get_entries
TypeError: get_entries_query() got an unexpected keyword argument 'infohash'
```
Platform: Linux
Tribler uptime : 22818 seconds ~ 6.3hrs

@ichorid @drew2a can you have a look at this?

drew2a:
Resolves #5677 by disabling #5642

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
@@ -72,7 +72,8 @@ class RemoteQueryCommunity(Community):
 
     community_id = unhexlify('dc43e3465cbd83948f30d3d3e8336d71cce33aa7')
 
-    def __init__(self, my_peer, endpoint, network, metadata_store, settings=None, notifier=None):
+    def __init__(self, my_peer, endpoint, network, metadata_store, settings=None, notifier=None,
+                 enable_resolve_unknown_torrents_feature=False):
         super(RemoteQueryCommunity, self).__init__(my_peer, endpoint, network)
 
         self.notifier = notifier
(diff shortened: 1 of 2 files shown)


devos50:
As you most likely already know, this will break compatibility with older versions of Tribler (e.g., Tribler 7.5.x users will not be able to search the content of Tribler 7.6 users since they are in different IPv8 overlays now). Are you sure that this is intended?

ichorid:
"The best way to make a mistake is to hastily try to fix a just discovered mistake." (C)

Indeed, that's basically "packet of death" problem we introduced in our system :skull: . However, updating the community id will tear the network apart... The proper way to fix it is to go "double-stack" by subclassing RQC and starting two instances simultaneously. Both subclassed instances must be protected from the issue.

A much simpler way to fix this is just to put `try-excep` block around the query line (actually, I thought I did it before... hence the error)

drew2a:
I see no other good option here.

To solve https://github.com/Tribler/tribler/issues/5642 we need to change the protocol of `PopularityCommunity` or protocol of `RemoteQueryCommunity`.

The second option looked more elegant and less invasive. This approach has been implemented in https://github.com/Tribler/tribler/pull/5657.

But https://github.com/Tribler/tribler/issues/5677 shows that we have two problems: https://github.com/Tribler/tribler/issues/5683 and https://github.com/Tribler/tribler/issues/5684.

So, the options to solve https://github.com/Tribler/tribler/issues/5677:
1. Revert back https://github.com/Tribler/tribler/issues/5642
2. Change community id

drew2a:
@ichorid the need for implementation "double-stack" shows another unresolved problem: community migration.

devos50:
> community migration

Fortunately, we did not have _that_ many occurrences where we would have to break compatibility with older Tribler versions. I think the most invasive one was the migration to IPv8, which was still designed to remain compatible with Dispersy messages. Still, community migration is something we should always think about and try to plan as good as possible like we are doing now with the tunnel community 👍

drew2a:
After discussion with @ichorid, I've decided to disable #5642

ichorid (approved):

PR #5689, opened 2020-10-27
Added basic query functionality to bandwidth accounting

### LINKED ISSUE #5676: [devel] Implement a crawling mechanism for our new bandwidth accounting mechanism
Now that #5626 has been merged, we require a mechanism for peers to construct the global work graph. Specifically, each peer should slowly explore the (latest) transactions of other peers, which acts as input for our upcoming reputation mechanism. #5626 does not implement the required primitives for requesting transactions from other peers. We also need this to deploy a crawler that monitors network health.

### The approach

The mechanism relies on simple pull-based gossip. Peers send out `TransactionQuery` message to other peers of which it would like to request the latest, pairwise transactions. Peers send back individual `Transaction` messages. Queries do not have an associated request cache, to keep the mechanism simple. In fact, our mechanism can deal with missing messages or peers that go offline quickly after receiving a query.

For now, we will query the immediate neighbours in a round-robin fashion. We can later extend our approach to also query the transactions of two-hop/three-hop peers.

devos50:
This PR extends the bandwidth accounting mechanism with a basic query mechanism. Specifically, this PR:
- Adds a new query message.
- Adds a separate settings object where we can quickly change the different parameters of the mechanism. Right now, we do not know what the preferred values of many parameters are, e.g., the speed at which we send out queries. Deployment results will help us to answer this question but for now, I picked a value that I see reasonable.
- Adds functionality where the transactions of a random peer are periodically queried.
- Added a setting to store _all_ incoming transactions in the database, even when they belong to the same pair of users. This setting is used by our network crawler.
- Added a crawler and an accompanying `systemd` file.

We have verified whether our crawling mechanism works by deploying it on a separate server. So far, it has successfully collected over 700 transactions 👍 

Fixes #5676

### DIFF
--- a/src/tribler-core/run_bandwidth_crawler.py
+++ b/src/tribler-core/run_bandwidth_crawler.py
@@ -1,0 +1,71 @@
+"""
+This executable script starts a Tribler instance and joins the BandwidthAccountingCommunity.
+"""
+import argparse
+import sys
+from asyncio import ensure_future, get_event_loop
+from pathlib import Path
+
+from ipv8.loader import IPv8CommunityLoader
+
+from tribler_core.config.tribler_config import TriblerConfig
+from tribler_core.modules.bandwidth_accounting.database import BandwidthDatabase
+from tribler_core.modules.bandwidth_accounting.settings import BandwidthAccountingSettings
+from tribler_core.modules.ipv8_module_catalog import BandwidthCommunityLauncher
+from tribler_core.session import Session
+
+
+class PortAction(argparse.Action):
+    def __call__(self, _, namespace, values, option_string=None):
+        if not 0 < values < 2**16:
+            raise argparse.ArgumentError(self, "Invalid port number")
+        setattr(namespace, self.dest, values)
+
+
+class BandwidthCommunityCrawlerLauncher(BandwidthCommunityLauncher):
+
+    def get_kwargs(self, session):
+        settings = BandwidthAccountingSettings()
+        settings.outgoing_query_interval = 5
+        database = BandwidthDatabase(session.config.get_state_dir() / "sqlite" / "bandwidth.db",
+                                     session.trustchain_keypair.pub().key_to_bin(), store_all_transactions=True)
+
+        return {
+            "database": database,
+            "settings": settings,
+            "max_peers": -1
+        }
+
+
+async def start_crawler(tribler_config):
+    session = Session(tribler_config)
+
+    # We use our own community loader
+    loader = IPv8CommunityLoader()
+    session.ipv8_community_loader = loader
+    loader.set_launcher(BandwidthCommunityCrawlerLauncher())
+
+    await session.start()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=('Start a crawler in the bandwidth accounting community'))
+    parser.add_argument('--statedir', '-s', default='bw_crawler', type=str, help='Use an alternate statedir')
+    parser.add_argument('--restapi', '-p', default=8085, type=str, help='Use an alternate port for the REST API',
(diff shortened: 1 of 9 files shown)


src/tribler-core/tribler_core/modules/bandwidth_accounting/community.py
```diff
@@ -103,38 +110,73 @@ async def received_transaction(self, source_address: Address, data: bytes) -> No
...
+        else:
+            # This transaction involves two unknown peers. We can add it to our database.
+            self.database.BandwidthTransaction.insert(tx)
+
+    def query_random_peer(self) -> None:
+        """
+        Query a random peer neighbouring peer and ask their bandwidth transactions.
+        """
+        peers = list(self.network.verified_peers)
+        random_peer = choice(peers)
```
grimadas, line 156: For reproducibility consider also a seed for random

synctext:
- its a network crawler run by every Tribler peer
- the term "query" could imply a local SQL cmd
- "Adds functionality where the transactions of a random peer are periodically queried"
  - why collect random info?
  - we need to optimise collecting useful info
  - maximize "cache hit"
- This is a full network crawler, not just a remote query

devos50:
retest this please

src/tribler-core/run_bandwidth_crawler.py
```diff
@@ -0,0 +1,71 @@
...
+from ipv8.loader import IPv8CommunityLoader
+
+from tribler_core.config.tribler_config import TriblerConfig
+from tribler_core.modules.bandwidth_accounting.database import BandwidthDatabase
+from tribler_core.modules.bandwidth_accounting.settings import BandwidthAccountingSettings
+from tribler_core.modules.ipv8_module_catalog import BandwidthCommunityLauncher
+from tribler_core.session import Session
+
+
+class PortAction(argparse.Action):
```
ichorid, line 18: :+1:  for `argparse` usage!

src/tribler-core/tribler_core/modules/bandwidth_accounting/community.py
```diff
@@ -75,22 +83,22 @@ def do_payout(self, peer: Peer, amount: int) -> Future:
...
+        :param peer: The IP address and port of the peer.
         :param request_id: The identifier of the message, is usually provided by a request cache.
         """
         payload = BandwidthTransactionPayload.from_transaction(transaction, request_id)
         packet = self._ez_pack(self._prefix, 1, [payload], False)
-        self.endpoint.send(peer.address, packet)
+        self.endpoint.send(address, packet)
 
-    async def received_transaction(self, source_address: Address, data: bytes) -> None:
+    def received_transaction(self, source_address: Address, data: bytes) -> None:
```
ichorid, line 101: Well, I'm a little bit worried that the remote peer can blow up our database by generating an endless stream of fake transactions in response to our query. Do you handle that in some way?

ichorid (approved):
Perfect PR! :+1:

src/tribler-core/tribler_core/modules/bandwidth_accounting/community.py
devos50: Currently no but I acknowledge that this is a concern. I will address this in a subsequent PR, before the release of 7.6 👍

synctext:
To document now:
- **Ledger-Zero**
- more radical design then trustchain!
  - even simpler
  - nearly as simple as tit-for-tat (as light-as-possible)
  - near-zero overhead: no mining, scripting, smart contracts
- pair-wise t4t focus
- not useful for Euro, Dollar or Bitcoins natively
  - open question: how efficient is a credits-on-top
  - certain limit and ledger-approach

Issue #5678, opened 2020-10-24
[7.5.3] Version checker - local variable 'response_dict' referenced before assignment


xoriole:
From error reporter [#36093](http://reporter.tribler.org/stacktrace/36093) 
```python
Traceback (most recent call last):
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 116, in on_read_data
RuntimeError: local variable 'response_dict' referenced before assignment
--LONG TEXT--
Traceback (most recent call last):
File "src\tribler-core\tribler_core\modules\versioncheck_manager.py", line 60, in check_new_version_api
UnboundLocalError: local variable 'response_dict' referenced before assignment

--CONTEXT--
{'message': 'Task exception was never retrieved', 'exception': UnboundLocalError("local variable 'response_dict' referenced before assignment"), 'future': <Task finished name='Task-171' coro=<VersionCheckManager.check_new_version_api() done, defined at src\tribler-core\tribler_core\modules\versioncheck_manager.py:42> exception=UnboundLocalError("local variable 'response_dict' referenced before assignment")>}
```
Platform: Windows-10
Uptime: 981.57 seconds

xoriole:
Fixed by https://github.com/Tribler/tribler/pull/5695

Issue #5670, opened 2020-10-22
Soft-crash of all 7.5.2 instances (Mac,Linux)


synctext:
With a historic bug we might have managed to kill off all online instances of Tribler 7.5.2

Every online Tribler peers running 7.5.2 (reproduced on Ubuntu and Mac) who conducts a ```New_Tribler_Version()``` check has a popup window which overwrites the default Tribler GUI from which there is no escape:
![Screenshot from 2020-10-22 17-34-34](https://user-images.githubusercontent.com/325224/96895731-feef3a80-148c-11eb-85f1-d03ec69116ef.png)
Both the "Ignore" and the "Later" button do not make the blocking popup go away. The "OK" button opens a new browser window on both Ubuntu and Mac successfully, but does not close the popup. Open questions: is this also happening on Windows?

devos50:
I am unable to reproduce this on both my Mac and the Win10-64bit machine in proxmox.

synctext:
On Linux I just discovered I was running: ```tribler_7.5.2-2-g33fa321bb_all.deb```.

xoriole:
I think it is because there are multiple popup stacked. Not sure how that happened since the version check duration is 6 hours unless it is running for a long time.

To reproduce the issue, I tried reducing the version check duration to a few seconds and I had a similar screen quickly. The popup should be a singleton and dispose any previous popus.

ichorid:
Installed 7.5.2 on Ubuntu 20.04, everything works fine :shrug:

synctext:
> I think it is because there are multiple popup stacked. Not sure how that happened since the version check duration is 6 hours unless it is running for a long time.

Great to hear this! That could explain a lot. Tribler usually runs for at least a week. For past 15 years its been often in the background on one or two laptops.

Capturing these cornercases has consistently been hard. For many years we needed an easier way for our users to report bug and capture state in detail.

xoriole:
> Can we have a unit test reproducing the multiple popups?

https://github.com/Tribler/tribler/pull/5700 includes a test for it

PR #5695, opened 2020-10-27
Fix version checker issues

xoriole:
- Fixes https://github.com/Tribler/tribler/issues/5678 : Referenced before assignment
- Fixes https://github.com/Tribler/tribler/issues/5670 : Multiple version checker popups

### DIFF
--- a/src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
+++ b/src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
@@ -25,6 +25,8 @@ class TestVersionCheck(TestAsServer):
 
         self.session.notifier.notify = Mock()
 
+        self.response_lag = 0  # lag in seconds to send a response
+
     def notifier_callback(self, subject, *args):
         self.new_version_called = True
 
@@ -45,7 +47,9 @@ class TestVersionCheck(TestAsServer):
         site = web.TCPSite(self.runner, 'localhost', self.port)
         await site.start()
 
-    def handle_version_request(self, request):
+    async def handle_version_request(self, _):
+        if self.response_lag > 0:
+            await sleep(self.response_lag)
         return RESTResponse(self.response, status=self.response_code)
 
     async def assert_new_version_called(self):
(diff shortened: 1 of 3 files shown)


devos50:
Shouldn’t this go to the 7.5 release branch?

xoriole:
retest this please

xoriole:
retest this please

devos50:
Ok, this is getting a bit problematic. We cannot easily run the "Clone Tribler" job since there is this annoying file with a unicode filename. However, the "Clone Tribler" job seems to be required for several of our checkers.

Given all the activity on the `release-7.5` branch, this is getting a priority. I will look into it.

xoriole:
>Ok, this is getting a bit problematic. We cannot easily run the "Clone Tribler" job since there is this annoying file with a unicode filename. However, the "Clone Tribler" job seems to be required for several of our checkers.
>Given all the recent activity on the release-7.5 branch, this is getting a priority. I will look into it.

@devos50 Thanks

devos50:
Please rebase this PR 👍

src/tribler-core/tribler_core/modules/tests/test_versioncheck.py
```diff
@@ -116,6 +120,19 @@ async def test_version_check_timeout(self):
         self.should_call_new_version_callback = False
         await self.check_version()
 
+    @timeout(10)
+    async def test_version_check_api_timeout(self):
+        await self.setup_version_server(json.dumps({'name': 'v1337.0'}))
+
+        # Setting the API response lag to 2 seconds and
+        # client's version checkout timeout of 1ms, version checks should fail returning False
```
devos50, line 128: Is the checkout timeout in seconds or milliseconds (you are using both units)?

xoriole: Thanks for pointing out. It is in seconds. I'll update the comment.

devos50 (approved):

PR #5697, opened 2020-10-27
Removed file with unicode character

devos50:
Instead, this file will be generated on demand by the test. This should make the Jenkins clone job functional again.


devos50:
retest this please

ichorid (approved):

PR #5698, opened 2020-10-27
Fix GUI crashing on race condition in addtochannel dialog

### LINKED ISSUE #5673: [7.5.3]: tribler_gui/dialogs/addtopersonalchanneldialog.py, in on_channel_contents: KeyError: 'results'
In the following code block, the `response` may somehow miss the "results" key, which leads to KeyError:

```python
    # tribler_gui/dialogs/addtopersonalchanneldialog.py
    def on_channel_contents(self, response, channel_id):
        if not response:
            return

        # No results means this node is a leaf
        self.channels_tree[channel_id] = set() if response["results"] else None
```

The last event was `NTFY.LOW_SPACE`, maybe it is related somehow.

ichorid:
Fix [reporter bug 36066](http://reporter.tribler.org/stacktrace/36066)

Fixes #5673

### DIFF
--- a/src/tribler-gui/tribler_gui/dialogs/addtopersonalchanneldialog.py
+++ b/src/tribler-gui/tribler_gui/dialogs/addtopersonalchanneldialog.py
@@ -133,9 +133,9 @@ class AddToChannelDialog(DialogContainer):
             return
 
         # No results means this node is a leaf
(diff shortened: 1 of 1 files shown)


devos50 (approved):

devos50:
Seems that this PR is tied to an issue: #5673.

PR #5701, opened 2020-10-28
READY: Isolated RemoteQueryCommunity

### LINKED ISSUE #5699: [devel] Peer explosion
oh no!

We have a new peer explosion bug. The RemoteQueryCommunity sets a custom `max_peers` count and custom `RandomWalk` target peer count, but **uses the same Network instance** as all other communities. This causes the peer balancing to break (note that some Communities get LESS peers due to this).

To fix this, the RemoteQueryCommunity should either:

 1. Stick to the defaults.
 2. Use its own `Network()` to not kill off the other communities.

qstokkink:
Fixes #5699

Changes:

 - Isolated the walking of the `RemoteQueryCommunity` in its own `Network` instance.
 - Added a `RemovePeers` strategy to actively cycle peers in the `RemoteQueryCommunity`.
 - Fixed the `SyncChannels` strategy only running when there are less than 20 peers.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
@@ -6,6 +6,7 @@ from ipv8.community import Community
 from ipv8.lazy_community import lazy_wrapper
 from ipv8.messaging.lazy_payload import VariablePayload, vp_compile
 from ipv8.peer import Peer
+from ipv8.peerdiscovery.network import Network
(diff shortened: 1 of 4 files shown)


qstokkink:
retest this please 🤔 freak Windows GUI failure

devos50:
I think it's fine to ignore that Windows GUI test. It should be fixed on the `devel` branch.

qstokkink:
Alright, in that case this is done 👍

ichorid (approved):

PR #5705, opened 2020-10-29
Move Download button to the left menu

ichorid:
This PR moves the :heavy_plus_sign:  download button to the left menu. Also, it makes the "Add torrent" menu to only be created once, as recommended by QT guidelines.

There are several reasons to move the button:
 * the original position in the right top corner was too close to "Close X" button on some OSes and to "Settings" button. It was easy to misclick.
 * the original button was too small and not intuitive, lacking a textual description.
 * the position on the left menu follows the style of known services (e.g. Google Drive)

изображение

### DIFF
--- a/src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
+++ b/src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
@@ -495,57 +495,6 @@ background: none;</string>
          </property>
         </spacer>
        </item>
-       <item>
-        <widget class="CircleButton" name="add_torrent_button">
-         <property name="minimumSize">
-          <size>
-           <width>32</width>
-           <height>32</height>
-          </size>
-         </property>
-         <property name="maximumSize">
-          <size>
-           <width>32</width>
-           <height>32</height>
-          </size>
-         </property>
-         <property name="cursor">
-          <cursorShape>PointingHandCursor</cursorShape>
-         </property>
-         <property name="styleSheet">
-          <string notr="true"/>
-         </property>
-         <property name="text">
-          <string/>
-         </property>
-         <property name="icon">
-          <iconset>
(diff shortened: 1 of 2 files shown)


devos50:
Interesting!

I agree that that button there is not the most intuitive location. I think @synctext also pointed this out already some time ago, but no one actually made a proposal to move it to somewhere else. Until now 👍 

Comments:
- I'm not sure if we even have to make it a button if we simply use a horizontal divider in the left menu. This would be similar to how Spotify does it (see below). When hovering with the mouse over this menu item, it will light up, communicating to the user that the element is clickable.
- I would rename the 'Add torrent' to 'Start download'. I think this is a more generic and intuitive description of the buttons' purpose.

Schermafbeelding 2020-10-29 om 16 23 53

ichorid:
> I would rename the 'Add torrent' to 'Start download'. I think this is a more generic and intuitive description of the buttons' purpose.

I deliberately decided to use the "_torrent_" word instead of the "_download_" word in the button description to avoid mixing it up with the "Downloads" button that is already there. Also, there is a function to create torrents in that menu. So, "Add torrent" actually _describes_ the action better. However, I agree that "Start download" is more instructive and clear in regards to the button's function. On the other hand, when I changed it to "Start download" I constantly find myself trying to drag'n'drop the torrents from channels interface to it :dizzy:

Let's just merge it as it is now, and then ask people around about how to better name it.


Regarding the style and hover stuff, I suggest you can play with that on your own :wink: Also, I'm planning to use the "Spotify-style" position on the bottom of the left menu for "Create channel" function as a part of moving subscribed and personal channels to the left menu (Spotify+Drive style).

ichorid:
Also, note that Spotify _does not want you to create many playlists_. While Drive _wants you to upload files_. Hence the button position/style.

devos50 (approved):

PR #5708, opened 2020-10-30
Fix log rendering in the debug pane

xoriole:
This PR fixes the log rendering in the Debug Pane. Some tests are refactored as well.

fix-logs

### DIFF
--- a/src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
+++ b/src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
@@ -1,10 +1,12 @@
 import os
 import sys
+from pathlib import Path
 from unittest.mock import Mock
 
 import pytest
 
 from tribler_core.restapi.base_api_test import do_request
+from tribler_core.utilities.osutils import get_root_state_directory
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
```diff
@@ -82,61 +83,104 @@ async def test_dump_memory(enable_api, session):
...
+
+
+@pytest.mark.asyncio
+async def test_debug_pane_no_logs(tribler_root_dir, enable_api, session):
+    """
+    Test whether the API returns the default response when no log files are found.
+    """
+    # To make sure that the test is not taking the OS dependent root state directory, we
+    # explicitly set the state directory here but we don't create any logs files there.
+    os.environ['TSTATEDIR'] = str(tribler_root_dir)
```
devos50, line 177: Would this environment variable persist and leak to other tests?

xoriole: I don't think it does. I have set that in multiple tests which sets the value to `tribler_root_dir` which itself is different for different tests and it seems to work fine.

devos50: I'm not entirely sure if this would also hold when running the tests with `pytest-xunit`, as we do in the PR testers. Can you double-check this and make sure it does not cause nasty, hidden issues?

xoriole: to be safe, now state directory environment variable is reset at the end

src/tribler-core/tribler_core/utilities/osutils.py
```diff
@@ -235,3 +235,16 @@ def get_root_state_directory(home_dir_postfix=u'.Tribler'):
         path = os.environ['TSTATEDIR']
         return Path(path) if os.path.isabs(path) else Path(os.getcwd(), path)
     return Path(get_appstate_dir(), home_dir_postfix)
+
+
+def reset_env_state_directory(state_dir: str):
```
devos50, line 240: This method is only used by the debug endpoint tests. Can you move it to `test_debug_endpoint.py`, instead of having it in this file?

xoriole: sure

src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
```diff
@@ -82,61 +84,124 @@ async def test_dump_memory(enable_api, session):
...
+
+    create_dummy_logs(log_dir, process='core', num_logs=num_logs)
+
+    json_response = await do_request(session, f'debug/log?process={process}&max_lines={num_logs}', expected_code=200)
+    logs = json_response['content'].strip().split("\n")
+
+    # Check number of logs returned is correct
+    assert len(logs) == num_logs
+
+    reset_env_state_directory(old_env_state_dir)
```
ichorid, line 167: And if the test fails, will this line be executed? The correct and safe way to do this is to do it through pytest fixtures teardown (e.g. save the var state in the fixture startup and restore it after `yield` in the fixture).

Also, I don't understand why you really need to use `get_root_state_directory()` here. The intention of the test is not to test `get_root_state_directory()`, right? You're not testing the real chain of calls `run_tribler->get_root_...->session->logger` either. Then why bother? Just use a fake tmp dir.

ichorid (changes requested):

src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
ichorid: @devos50 , thanks for catching this. I almost missed it.

src/tribler-core/tribler_core/restapi/tests/test_debug_endpoint.py
xoriole: Good point. I have updated the PR with fixtures.

Regarding the intention of the test, it is to test that log files in the root directory is taken as fallback to display the logs in the debug pane. This is because the logs are actually written in the root directory as per the current implementation (https://github.com/Tribler/tribler/blob/devel/src/run_tribler.py) but the tests were testing the preset state directory.

ichorid (approved):

devos50 (approved):

PR #5712, opened 2020-11-01
READY: Add Subscribed Channels menu

ichorid:
This PR remove the "Subscribed" and "My channel" buttons in favour of a more ergonomic and fancy menu showing subscribed channels list directly.

Also, it adds a confirmation dialog to "Unsubscribe" and "Delete channel" actions.

изображение

p.s.
Synchronizing "Subscribe" toggle buttons everywhere took about 8 hours of work :facepalm:


devos50 (commented):
I checked out the PR, and I have the following comments:
- Thanks for doing these GUI improvements! 👏 You managed to get rid of another entire page! 
- Overall, the new box feels disconnected from the rest of the menu items. I have a few suggestions below on how we might be able to reduce that.
    - There is a small black box around the entries. That seems to unnecessarily clutter the left menu. I suggest to remove it.
    - I think it's fine to have a zero padding to the right of the box with subscribed channels. This will also align the scroll bar with the edge of the left menu, which should look a bit better.
    - Judging from your screenshot, there is an icon in front of some entries in that menu. I don't see this icon when I'm running Tribler. Is this icon needed?
    - Even though I think it should be clear, adding some header text above the box with subscribed channels (e.g., "SUBSCRIBED CHANNELS") would better convey what the items in the left menu are.
- It was not entirely sure to me why I couldn't click the items in the left menu, even though I can browse the channel content from the 'discovered' page. I also didn't check the tooltip. Why would we want to prevent users from clicking these channel items, even if the download has not completed yet?

ichorid:
@devos50 

>     * Thanks for doing these GUI improvements! clap You managed to get rid of another entire page!

Two pages, actually :monocle_face: 

>     * Overall, the new box feels disconnected from the rest of the menu items. I have a few suggestions below on how we might be able to reduce that.
>       
>       * There is a small black box around the entries. That seems to unnecessarily clutter the left menu. I suggest to remove it.

Good point, going to remove that

>       * I think it's fine to have a zero padding to the right of the box with subscribed channels. This will also align the scroll bar with the edge of the left menu, which should look a bit better.

Good point. Though I hope people are not going to add so many channels that the scrollbar will be needed...

>       * Judging from your screenshot, there is an icon in front of some entries in that menu. I don't see this icon when I'm running Tribler. Is this icon needed?

The icon marks personal channels, which are editable.

>       * Even though I think it should be clear, adding some header text above the box with subscribed channels (e.g., "SUBSCRIBED CHANNELS") would better convey what the items in the left menu are.

Actually, I do not want to clutter the interface with that kind of stuff at all. If the user does not use the Channels system, they should not be bothered with Channels-related labels. Conversely, when the user subscribes to a channel, the channel appears immediately and from its name it is obvious that's a channel. So, no need to for additional labels, IMHO. Though I will add a small margin just to space that out from other buttons.

>     * It was not entirely sure to me why I couldn't click the items in the left menu, even though I can browse the channel content from the 'discovered' page. I also didn't check the tooltip. Why would we want to prevent users from clicking these channel items, even if the download has not completed yet?

Yes, that's intended. The channels that have not been downloaded yet are greyed-out and non-clickable. The tooltip notifies on that. Maybe later I will add a placeholder screen for "Downloading" channels, but that will be done later, in a separate PR.

xoriole:
I tried running the PR locally to get the feel of it. I experienced the following:
1. When I subscribe a channel it gets added to the left sidebar, I liked this instant feedback.
2. Clicking on the new channel link on the bottom left, opens a dialog box to create a new channel. Once the channel is created, it gets added to the top of the channels list, above subscribed channels. Nice.
3. I subscribed to a bunch of big channels and tried creating a new channel. No instant feedback of adding the channel in the list. Even though the dialog box seemed to work fine, I couldn't see any new channels created anymore. On further checking, the request was successful but it was only not updated in the sidebar and there is no way to refresh the sidebar in the UI.
4. Going inside one of the created channels, the channel page is cleaner. Liked the subscribe button and popularity indicators. But I found a few inconsistencies. Popup menu on clicking the ellipsis appears below the search input field. The number of items in the channel representation is sometimes incorrect. See attached screenshot.
5. As Martijn also pointed out, nothing happens when you click any subscribed channel until it is complete. You only find that out by clicking on the channel name (multiple times). Maybe changing mouse pointer could be an option to give feedback to the user that it is not enabled yet.
6. Hovering over the channel name, it shows a tool tip with the name on the first line and status (eg. `search for metainfo`, `complete`) on the next line. Maybe the status could be indicated by an icon on the left.
6. The opacity between the incomplete channels and complete channels is not clear on the first glance. You mostly find out the difference by observing the response on a click. An icon on the left could improve to distinguish the contrast.

ksnip_20201104-174606

Screenshot from 2020-11-04 18-12-23

ichorid:
@xoriole 
Thanks for trying out the PR!

>I subscribed to a bunch of big channels and tried creating a new channel. No instant feedback of adding the channel in the list. Even though the dialog box seemed to work fine, I couldn't see any new channels created anymore. On further checking, the request was successful but it was only not updated in the sidebar and there is no way to refresh the sidebar in the UI.

It is nice that you noticed it. Some signal was lost somewhere during all the refactorings... I'll fix this.

> Popup menu on clicking the ellipsis appears below the search input field.

This is a known problem. It was there for a long time. I'll fix it either in this or another PR.


>The number of items in the channel representation is sometimes incorrect. See attached screenshot.

This is harder to fix as it is not GUI-related. It will require changing the logic of the Channels system. I'll fix it in another PR.

> As Martijn also pointed out, nothing happens when you click any subscribed channel until it is complete. You only find that out by clicking on the channel name (multiple times). Maybe changing mouse pointer could be an option to give feedback to the user that it is not enabled yet.

I changed that in my latest fixes here. The user can open the channel now, but the name is still grayed out.


> Hovering over the channel name, it shows a tool tip with the name on the first line and status (eg. search for metainfo, complete) on the next line. Maybe the status could be indicated by an icon on the left.

I was thinking about it too. Though, I'm hesitant of cluttering the interface with icons too much. I think that graying out incomplete channels is good enough for now. We can try different approaches later.

devos50 (commented):
Much better! I really like how it's starting to look. Now it feels like a natural element in the left menu. 👍 

One small thing that I noticed is that the scroll bar is not entirely aligned to the right of the left menu yet.

Also, now that I've played a bit more with it, I'm not in favour of the icon on the left to indicate that it is your channel. I'm not sure what a sensible solution for this would be actually.

Another thing I noticed: the channel items are currently (text) aligned with the 'big functionality' buttons (e.g., discovered/downloads) but they are less important and "minor" elements in the left menu. As such, I wouldn't place them on the same level as these buttons.

I have a few ideas about some layout changes. One of them includes a more dark background color of the left menu. I think the text will look a bit better if the background is more of a black color.

ichorid:
@devos50 , well, that stuff could be easily adjusted through CSS. It is hard to discuss GUI changes through text. If you provide a mock-up, I'll try implementing it. Otherwise, let's just merge it as it is, so the team can evaluate it and add proceed with cosmetic changes in other PRs?

xoriole:
Exporting channel (from ellipsis popup menu) returns 404. Other than that, I'm okay with merging it.
export-channel-nov5-2020

ichorid:
@xoriole , I removed the feature. No one ever used it anyway...

ichorid:
retest this please

xoriole:
@ichorid could you also check remove all from ellipsis popup menu?

ichorid:
@xoriole , good find. The button was dead for a long time. I removed it.

ichorid:
retest this please

devos50:
Seems that there is a conflict. Can you rebase the PR first?

Also, I recommend fixing the following pylint error (the new super style is much more readable and less confusing):

```
43: R1725: (super-with-arguments), ChannelListItem.setData: Consider using Python 3 style super() without arguments
```

xoriole (approved):

Issue #4320, opened 2019-03-06
Add a separate queue for threaded asynchronous Channels DB access


ichorid:
I've just done an improvised benchmark of anonymous downloading Ubuntu torrent while simultaneously processing a large GigaChannel from disk. DB i/o **definitely** affects tunnels performance. Badly.

If the database is allowed to take as much time as it wants, even on a separate thread, and even if I break up the transaction into small batches that take approx 100ms each, tunnels performance drops from 600-700 kb/s to 150-350 kb/s and GigaChannel-related GUI becomes unresponsive. If I add `sleep(0.05)` after each batch, the responsiveness and performance are restored to 80% of their original levels (at least it feels so).

Apparently, we've hit both the [thread starvation problem](https://stackoverflow.com/questions/37123486/thread-starvation-while-locking-in-a-loop-in-python) and the limits of Twisted reactor scheduling. The only way to fix this problem is to **move GigaChannel and TorrentChecker into a separate process with a separate REST endpoint**.

devos50:
Nice that we have some numbers on the impact of database I/O now! I must say that the drop in tunnel performance is not that bad and better than I expected (I expected it to drop significantly, to only a few kb/s).

A suggestion made by @synctext is to notify the user by displaying the status bar when a (big) channel is being processed. We could make a new event for this, and transmit it to the GUI over the events socket.

For further optimization, I would suggest to first focus on a (more formal) experiment (e.g. written in Gumby) which we can quickly repeat with different parameters.

qstokkink:
[I have some master thesis work on fast platform independent inter-process communication](https://github.com/qstokkink/tribler/tree/tunnelprocess_rc2/Tribler/community/tunnel/processes) (almost most of the logic on top of the communication logic will be out-of-date by now). This should give a nice copy-paste basis for testing some stuff.

Do we want this for 7.3 already though?

ichorid:
On my workplace PC, when I download Ubuntu torrent over the tunnels, it gives me 2.1 MB/s. If GigaChannel database just **contains** a channel of 1 million entries, the speed drops to 600-800 KB/s.
If I start to upload the channel to the database, the connection dies almost completely, dropping to 50-150 KB/s. Now that's in the case of unoptimized upload routine (500ms batches).

When I set the batch size to 100ms and add 50ms `sleep` periods, the download is able to proceed with 600-800 KB/s.

[...]

PR #5720, opened 2020-11-05
READY: Use WAL journal mode for sqlite connections

ichorid:
Well, this _should_ help with DB performance...

### DIFF
--- a/src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
+++ b/src/tribler-core/tribler_core/modules/bandwidth_accounting/database.py
@@ -25,6 +25,17 @@ class BandwidthDatabase:
         create_db = str(db_path) == ":memory:" or not self.db_path.is_file()
         self.database = Database()
 
+        # This attribute is internally called by Pony on startup, though pylint cannot detect it
+        # with the static analysis.
+        # pylint: disable=unused-variable
+        @self.database.on_connect(provider='sqlite')
+        def sqlite_disable_sync(_, connection):
+            cursor = connection.cursor()
+            cursor.execute("PRAGMA journal_mode = WAL")
+            cursor.execute("PRAGMA synchronous = 1")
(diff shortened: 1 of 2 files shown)


devos50:
Can you do a basic benchmark to verify that the performance has indeed increased?

ichorid:
Yes, I've tested it. When processing channels, patched Tribler is 20% faster. Also, it is important to note that WAL is most useful by enabling reads concurrent with writes. So, it is good for e.g. improving the responsiveness of the GUI, especially on slower media.

Let's merge it and see how it fares on our PCs, shall we?

kozlovsky:
Note that the SQLite documentation [says](https://www.sqlite.org/wal.html):

> The default strategy is to allow successive write transactions to grow the WAL until the WAL becomes about 1000 pages in size, then to run a checkpoint operation for each subsequent COMMIT until the WAL is reset to be smaller than 1000 pages. By default, the checkpoint will be run automatically by the same thread that does the COMMIT that pushes the WAL over its size limit. **This has the effect of causing most COMMIT operations to be very fast but an occasional COMMIT (those that trigger a checkpoint) to be much slower**. If that effect is undesirable, then the application can disable automatic checkpointing and run the periodic checkpoints in a separate thread, or separate process.

So, while in general WAL journaling mode should be faster than DELETE journaling mode, sometimes COMMIT might be very slow. If SQLite updates are performed directly from the event loop, it sometimes may lead to long pauses with the event loop processing.

I'm for using WAL mode, but ideally, it should be combined with performing all transactions which modify the database from a separate thread to execute them outside of the event loop. Or, at least, the should be a separate thread for periodic checkpointing.

devos50 (approved):

ichorid:
@kozlovsky , good point! This PR is very far from a perfect solution, but at least it is a step in the right direction. Of course, we'll still have to implement #4320 to solve the problem for good.

PR #5725, opened 2020-11-09
READY: Channels 2.0 docs

ichorid:


kozlovsky (dismissed):
We discussed with @ichorid  this PR in person, and he has already taken into account my comments

PR #5726, opened 2020-11-10
Combine subscribed column with state

ichorid:
This PR combines the state column with the subscription control column. The subscription toggle switch is now shown only for hovered rows. This reduces "*orangefication*" (:copyright:  @xoriole) of the interface and reduces its complexity as @qstokkink  suggested.
lkj

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
+++ b/src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
@@ -264,9 +264,16 @@ class SubscribedControlMixin(object):
         if data_item[u'state'] == u'Personal':
             return True
 
-        self.subscribe_control.paint(
-            painter, option.rect, index, toggled=data_item.get('subscribed'), hover=index == self.hover_index
-        )
+        # When cursor leaves the table, we must "forget" about the button_box
+        if self.hoverrow == -1:
+            self.button_box = QRect()
+
+        if index.row() == self.hoverrow:
+            self.subscribe_control.paint(
+                painter, option.rect, index, toggled=data_item.get('subscribed'), hover=index == self.hover_index
+            )
+        else:
+            self.draw_channel_state(painter, option, index, data_item)
 
         return True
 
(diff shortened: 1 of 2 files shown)


src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
```diff
@@ -264,9 +264,16 @@ def draw_subscribed_control(self, painter, option, index, data_item):
         if data_item[u'state'] == u'Personal':
             return True
 
-        self.subscribe_control.paint(
-            painter, option.rect, index, toggled=data_item.get('subscribed'), hover=index == self.hover_index
-        )
+        # When cursor leaves the table, we must "forget" about the button_box
+        if self.hoverrow == -1:
```
drew2a, line 268: Nit: I always prefer to keep magic numbers in constants and can only recommend you do the same ;)

drew2a:
It is not clear to me now, which states do we have in the `Subscribed` row?

* "Subscribed" (check)
* "Not Subscribed" (toggle)

if so, how you will represent "Subscribing" then?

devos50:
Even though I understand the reasoning of combining these rows, having the checkbox only pop up when hovering the row is quite confusing. I actually think it was more user-friendly before. Also, I don't have a toolbox anymore to see what the red question mark means.

If "orangeficiation" is a problem, why not use a more neutral color?

ichorid:
>  If "orangeficiation" is a problem, why not use a more neutral color?

See the previous discussion on Slack. I tried to use green, but there was a rightful objection that we should not add another color to the GUI without reason.

Personally, I think orange is too aggressive. However, I'm in no mood for redesigning the whole Tribler color scheme...

ichorid:
> I don't have a toolbox anymore to see what the red question mark means

That's strange, because the tooltip on the subscribed column quite literally repeats the tooltip on the former "state" column. Works for me in Ubuntu. Is it a Mac only problem?

ichorid:
I understand it is less intuitive now, but design is about tradeoffs... That's why I ask you, guys.

drew2a:
I still don't understand how it works (see the question above), but whatever...

Less intuitive means objectively worse, right? So, the question is "what we get then?".

disclaimer: I see nothing wrong with "orangefication".

ichorid:
> I see nothing wrong with "orangefication".

@drew2a , subscribe to the top 10 channels and try reading through the channels list for a minute. Then you will see.

ichorid:
No consensus on this, so closing this one.

Issue #3971, opened 2018-10-18
True GigaChannels with 1 billion magnet links


synctext:
Tribler aims to be a realistic alternative to Youtube. Goal: support an unbounded wealth of user-generated content. Solution: create a Bittorrent swarm of any magnet link collection.

_1 channel may contain up to 1 billion magnet links_

A Youtube-alternative requires seamless scalability of a server-free video platform; beyond anything scientists have been able to create so far. Fast growth needs to be accommodated, for instance, Netflix has a growth rate 200.000 new users per day.
![image](https://user-images.githubusercontent.com/325224/47154004-693ed480-d2e1-11e8-9074-78490495cee8.png)
Current Tribler channels do not support envisioned growth rate. For instance, Youtube channels obtains lot of new subscribers each day. Taking some [stats](https://socialblade.com/blog/youtube-milestones-four-channels-hit-one-million-subscribers-every-day/) of the Internets, we see that:
- More than 300 channels are hitting the 10,000 subscriber mark every day.
- More than 40 channels are hitting the 100,000 subscriber mark every day.
- Around 4 channels are hitting the 1 million subscriber mark every day.

Youtube channels with a gigantic amount of videos:
1. [1013672 videos](https://www.youtube.com/watch?v=8nW_cPi_Sbs&list=UUtykkZYAT4lbeG3CBhSfLrw)
1. [727931 videos](https://www.youtube.com/watch?v=e8NWj6cMMps&list=UUdf7KqPteLls_gmI7kjntfw)
1. [624745 videos](https://www.youtube.com/watch?v=btPqcoECBeM&list=UUsLiV4WJfkTEHH0b9PmRklw)

The 10 Biggest YouTube Channels Right Now
1. PewDiePie. Subscribers: 66.1 million.
1. Dude Perfect. Subscribers: 34.54 million.
1. HolaSoyGerman. Subscribers: 34.48 million.
1. whinderssonnunes. Subscribers: 31.4 million.
1. elrubiusOMG. Subscribers: 30.83 million.
1. Fernanfloo. Subscribers: 29.34 million.
1. JuegaGerman. Subscribers: 28.95 million.
1. VEGETTA777. Subscribers: 23.85 million.

### Technical Details

[...]

PR #5729, opened 2020-11-13
Fix batch size handling during metadata during channel content upload

kozlovsky:
I spend some time profiling the database access code and noticed that the Tribler application spends most of the time inside `db_session` wrapped around the `process_payload` method of MetadataStore.

As it turns out, MetadataStore split objects into batches of very small size, and on my machine typical batch size was just two metadata objects.

MetadataStore determines batch size dynamically depending on the duration of the previous batch execution, but the calculation includes sleep time between batches as well, which skewed results.

I fixed the formula and also restricted the minimum and maximum batch size.

After the changes presented in this pull request, the rate of metadata objects insertion on my machine increased by 23%, from 60.63 objects/seconds to 74.84 objects/seconds for the first 4000 objects loaded after the start of the application.

Over the course of the time the application is running, the speed of inserts execution drops significantly, and this does not depend on the size of the database, but on the time the application is running. This should be the next important topic for research since the drop is quite significant.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/store.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/store.py
@@ -52,6 +52,11 @@ UNKNOWN_TORRENT = 5
 DELETED_METADATA = 6
 UNKNOWN_COLLECTION = 7
 
+
+MIN_BATCH_SIZE = 10
+MAX_BATCH_SIZE = 1000
+
+
 # This table should never be used from ORM directly.
 # It is created as a VIRTUAL table by raw SQL and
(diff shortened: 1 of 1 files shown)


ichorid (approved):

synctext:
> After the changes presented in this pull request, the rate of metadata objects insertion on my machine increased by 23%, from 60.63 objects/seconds to 74.84 objects/seconds for the first 4000 objects loaded after the start of the application.

nice! Can we please have a performance graph on Jenkins with the speed channels? Related to huge item requested 2 years ago, #3971

PR #5734, opened 2020-11-17
Channels-related workarounds for another 7.5 release

ichorid:
Ok, this is going to be tricky to explain...

Apparently, when one specifies a `UNIQUE` index in SQLite, the index is created automatically as a special kind of `autoindex`. These indexes have two peculiar properties:
 1. autoindexes are only shown with a SQLite-specific `PRAGMA` (i.e. SQLitebrowser does not show them)
 2. unique autoindexes **cannot be deleted**.

Well, I happily left managing indexes stuff to Pony and never saw the results due to (1). However, the first versions of Channels included a unique `public_key+infohash` index to be compatible with the older GUI and REST logic. Later we completely change channels identification model in 7.4-7.5 by moving to the more robust `public_key+id_` approach. The logic for serving the corner case of unique public_key+infohash combination was still there, though, causing extreme slowdowns on channel processing *for people who started using Tribler since 7.5*! For those who used older, "upgraded" DBs, the index was still in place, speeding up the obsolete checks.

Now, we removed those checks in #5732 . But the index still remains there! Thus, user DB versions have split into two types: "with unique index" and "without unique index".

Note that if we turn the checks off for everyone, some people will experience crashes due to UNIQUE constraint fails. If we keep the checks on for everyone, some people (new users) will experience *extremely* slow Channels performance.

I'm going to do a full DB upgrade for 7.6, re-creating the DB completely. Til' that moment, I propose keeping the check ON for those users who have the "with unique index" DB, but turn the check OFF for those users who got no index in their DB.

This PR does just that - turns off the check for those DBs that must have it off. (It is ugly, but I checked it by hand - it works).

Also, this PR adds support for `infohash` and `public_key` -based queries in RemoteQueryCommunity. This will help smooth the transition to 7.6, where all Channels- related functions will be served by RemoteQueryCommunity.

We should do another 7.5 series release after merging this.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/store.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/store.py
@@ -149,6 +149,17 @@ class MetadataStore(object):
                 default_vsids = self.Vsids.create_default_vsids()
             self.ChannelMetadata.votes_scaling = default_vsids.max_val
 
+        with db_session:
+            self.pk_infohash_uniq_index_exists = False
+            for ind in range(0, 10):
+                result = list(self._db.execute(f"pragma index_info(sqlite_autoindex_ChannelNode_{ind});"))
+                if result:
+                    indexed_columns = []
+                    for col in result:
+                        indexed_columns.append(col[2])
+                    if len(indexed_columns) == 2 and 'public_key' in indexed_columns and 'infohash' in indexed_columns:
+                        self.pk_infohash_uniq_index_exists = True
+
     @db_session
     def upsert_vote(self, channel, peer_pk):
         voter = self.ChannelPeer.get(public_key=peer_pk)
@@ -440,19 +451,20 @@ class MetadataStore(object):
             ):
                 return check_update_opportunity()
 
-            # Check for a node with the same infohash
-            node = self.TorrentMetadata.get_for_update(
-                public_key=database_blob(payload.public_key), infohash=database_blob(payload.infohash)
-            )
-            if node:
-                if node.timestamp < payload.timestamp:
-                    node.delete()
-                    result.append((None, DELETED_METADATA))
-                elif node.timestamp > payload.timestamp:
-                    result.append((node, GOT_NEWER_VERSION))
-                    return result
-                else:
-                    return result
+            if self.pk_infohash_uniq_index_exists:
+                # Check for a node with the same infohash
+                node = self.TorrentMetadata.get_for_update(
+                    public_key=database_blob(payload.public_key), infohash=database_blob(payload.infohash)
+                )
+                if node:
+                    if node.timestamp < payload.timestamp:
+                        node.delete()
+                        result.append((None, DELETED_METADATA))
+                    elif node.timestamp > payload.timestamp:
+                        result.append((node, GOT_NEWER_VERSION))
+                        return result
+                    else:
+                        return result
                 # Otherwise, we got the same version locally and do nothing.
 
         # Check for the older version of the same node
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
@@ -25,6 +25,16 @@ def sanitize_query(query_dict, cap=100):
     first = first or 0
     last = last if (last is not None and last <= (first + cap)) else (first + cap)
     query_dict.update({"first": first, "last": last})
+
+    # convert hex infohash to binary
+    infohash = query_dict.get('infohash', None)
+    if infohash:
+        query_dict['infohash'] = unhexlify(infohash)
+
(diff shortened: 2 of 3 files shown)


src/tribler-core/tribler_core/modules/metadata_store/store.py
```diff
@@ -149,6 +149,17 @@ def sqlite_disable_sync(_, connection):
                 default_vsids = self.Vsids.create_default_vsids()
             self.ChannelMetadata.votes_scaling = default_vsids.max_val
 
+        with db_session:
+            self.pk_infohash_uniq_index_exists = False
+            for ind in range(0, 10):
```
devos50, line 154: Why 10?

devos50 (commented):
Thanks for the explanation!

src/tribler-core/tribler_core/modules/metadata_store/store.py
ichorid: just random number that is more than the possible number of indexes that could have been created by SQLite automatically?

devos50: Ok. Is there any simple way to get the number of indices from sqlite? If not, at the very least please add a comment that this value should be higher than the number of existing indices 👍

qstokkink: > Is there any simple way to get the number of indices from sqlite?

If you just want to list the indices you can run this:

```sql
SELECT name FROM sqlite_master WHERE type='index'
```

ichorid: I knew that at least *one* of the team will suggest this :facepalm:  Professional programmers' response :fist_raised: 

Yes, querying the number of indices will limit the number of checks to *about 20* checks, that is far *more* than 10. So it will not optimize anything (mind you, it is hard to imagine that the thing must be optimized). I doubt there is, or was, or ever will be (except for Quinten-factor, but where would we be if we always think about that?) any Tribler installation of 7.5.x featuring more than 10 autoindexes.

So, the only option to solve this properly is to add a query that will count the number of autoindexes in the list.
Personally, I am almost ashamed to spend even *these* letters, yours and my time to discuss and *solve* such a trivial "problem".

But if you guys insist, I will solve this for good :shrug:   :wink:

qstokkink: > except for Quinten-factor

Lol 😄, even I don't try to bend over backwards to get sqlite to make extra indices (I only have 3 as you can see). It is indeed quite unlikely anyone would have more than 10 autoindices.

That said, I agree with @devos50: if 2~3 easy extra lines of code makes this code work 100% of the time instead of 99% of the time, I'd vote to spend an extra 2 minutes on this PR and solve it for good.

devos50: > Personally, I am almost ashamed to spend even these letters, yours and my time to discuss and "solve" such a trivial "problem".

Isn't that the whole point of PR reviews? When I review a PR, I point out all issues that can arise, both major and minor. In this situation, the issue is very minor but it is an issue nonetheless and something that can trivially be fixed. If you have a different opinion on how we should treat PR reviews, please let me know and I can take it into consideration.

You indicated that you already knew that one of us would identify this issue, which (in my opinion) should be reason enough to have addressed this "trivial" issue before opening a PR.

ichorid: > If you have a different opinion on how we should treat PR reviews, please let me know and I can take it into consideration.

Whenever I review a PR, when I notice a risk or potential problem, or another kind of opportunity for improvement, before even *notifying* the author about the thing, I consider the tradeoff between:
 1. the potential impact of the change (e.g. how big it the risk that )
 2. the amount of energy the author and the team as a whole will have to spend on fixing that.

That's called [leverage](https://gist.github.com/rondy/af1dee1d28c02e9a225ae55da2674a6f). It is the most important concept in the Effective Engineer book.

If the tradeoff does not look good, I don't notify the author. Thus, I conserve their time, mental energy, and provide as little obstructions to their thinking process as possible. That's why I, e.g. never correct grammar mistakes in other's PRs: *it is just not worth it*. 



Changing even a single letter requires much more than changing a letter in the text editor these days. It requires:
1. committing the change
2. squashing the changes
3. requesting another re-review
4. rebasing the PR (because while you were doing steps 1-3 someone else already committed their stuff before you)
5. requesting another re-review because of the previous step

Pipelining multiple PRs does not help either, because it forces one to multitask, which is very inefficient and taxing on mental energy, and you guarantee to force step 4 on yourself (rebasing your work on top of you own work that was just accepted).

So, that's my position on both writing *and* reviewing code: maximize *leverage*, skip stuff with low *leverage* completely.

ichorid:
Dev meeting decided to skip 7.5.5 and go straight to 7.6. Closing this one

PR #5735, opened 2020-11-18
Add sentry

### LINKED ISSUE #5727: Add Sentry error reporting to the Tribler Client
This issue is linked to https://github.com/Tribler/tribler/issues/5715

Updated version:

As a developer, I want to increase the quality of receiving debug information and decrease the count of supported entities (Reporter).

So let's add Sentry to Tribler Client as a "backup system" that will accomplish the same work as the current Reporter do. Keep all the UX unchanged.

Will release it in `7.6.0` and evaluate how it works.

Old version:
```
Now we use Sentry as a mirror for the actual reporter. This has greatly increased the convenience of working with error reports. But the way we use Sentry is unnatural for Sentry itself and leaves an intermediate entity (reporter) that is used as a proxy.

The idea is to send crash reports directly to Sentry.

Changes:
1. Add to the installer a checkbox "Automatically sends crash reports to the Tribler team?"
2. Add to the crash reporter dialog the same checkbox.
3. If the checkbox is set, then let's sentry to be initialized and be able to send crash reports.

Note. Before sending a crash report we should cut all sensitive information.

In the first month, we will use both the old reporter and Sentry. 
After the first month, we have to review how it works.
```

drew2a:
Resolves #5727

This PR adds Sentry support to Tribler Client.
The UX keeps unchanged.

The main concept: suppress (but store) all Sentry events until a user clicked on the "send report" button.

There are two new entities:

1. `SentryReporter`(singleton) for managing work with Sentry
1. `SentryScrubber` for scrubbing sensitive and redundant information.

# How it works
Basically, Sentry-based reporting works the same way the old Reporter did:

* catch the error in the Core (or in the GUI)
* send it to the GUI over the Events endpoint (GUI errors are handled directly)
* present the "report error" dialog to the user
* send the report when the user clicks the "send report" button

# Protecting user identities
A user will now be assigned a unique pseudonym (e.g.Jonh Doe) based on the `hash` of their `public_key`.

Note. The current implementation specifies user only for errors raised in the Core process, but it can be extended in the future.

# Simplified python-friendly explanation

```python
# `SentryReporter` will be inited at the start of a Tribler Client run
SentryReporter.init(sentry_url=sentry_url, scrubber=SentryScrubber())

# Then all Sentry events will be suppressed (`session.py`):
SentryReporter.allow_sending_globally(False)

# In a case of an error in  the `Core process`, the Sentry event will be obtained and sent to `GUI` (`session.py`):
sentry_event = SentryReporter.last_event
self.api_manager.get_endpoint('events').on_tribler_exception(text_long, sentry_event)
```

In the case of an error inside the GUI process, the event will be obtained directly in the GUI process.

After the user's click on the "sent report" button, the Sentry event will be sent to the server (`feedbackdialog.py`):

```python
SentryReporter.send(post_data, sentry_event, sys_info_dict)
```

# Removing sensitive and redundant information

`SentryScrubber` is the class responsible for scrubbing.
It is calling on every Sentry's attempt to sent an event.
By using regex it scans all the necessary event fields and replaces real information with placeholders.

What is sensitive information:
* IPs
* User Name
* 40-symbols-long hashes

What is redundant information:
* Information on installed python modules

Which fields will be affected by scrubbing:
* Extra
* Logentry
* Breadcrumbds
* Os.environ
* Stacktrace
* Sysinfo

## Tests
This PR introduced the first tests for the `tribler-common` module. 
All the corresponding Jenkin's jobs have to be updated:
- [x] PR Linux
- [x] PR macOS
- [x] PR Windows

Thanks to @ichorid for contributing to the description.

### DIFF
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -1,0 +1,330 @@
+import logging
+from contextvars import ContextVar
+from hashlib import md5
+
+from faker import Faker
+
+import sentry_sdk
+from sentry_sdk.integrations.logging import LoggingIntegration
+from sentry_sdk.integrations.threading import ThreadingIntegration
+
+from tribler_common.sentry_reporter.sentry_tools import (
+    delete_item,
+    get_first_item,
+    get_value,
+    parse_os_environ,
+    parse_stacktrace,
+)
+
+PLATFORM_DETAILS = 'platform.details'
+STACKTRACE = '_stacktrace'
+SYSINFO = 'sysinfo'
+OS_ENVIRON = 'os.environ'
+SYS_ARGV = 'sys.argv'
+TAGS = 'tags'
+CONTEXTS = 'contexts'
+EXTRA = 'extra'
+BREADCRUMBS = 'breadcrumbs'
+LOGENTRY = 'logentry'
+REPORTER = 'reporter'
+
+
+class SentryReporter:
+    """SentryReporter designed for sending reports to the Sentry server from
+    a Tribler Client.
+
+    # Main concept
+
+    It aims to work in a Tribler Client with the following specifics:
+
+    1. An exception can be raised in two separate processes: GUI and Core.
+    2. A report will be sent after a user presses the "send report" button.
+
+    The main concept behind this class is the following: let's add controls that
+    can allow or disallow sending Sentry reports by-default.
+
+    If we have these controls, then it is easy to use Sentry with a Tribler
+    Client.
+
+    Algorithm:
+    1. Initialise Sentry.
+    2. Disallow sending messages (but storing the last event).
+    3. Waiting for user action: the "sent report" button is pressed.
+    4. Allow Sentry sending messages
+    5. Send the last event.
+
+    SentryReporter is thread-safe.
+   """
+
+    last_event = None
+
+    _allow_sending_global = False
+    _allow_sending_in_thread = ContextVar('Sentry')
+
+    _scrubber = None
+    _logger = logging.getLogger('SentryReporter')
+
+    @staticmethod
+    def init(sentry_url='', scrubber=None):
+        """ Initialization.
+
+        This method should be called in each process that uses SentryReporter.
+
+        Args:
+            sentry_url: URL for Sentry server. If it is empty then Sentry's
+                sending mechanism will not be initialized.
+
+            scrubber: a class that will be used for scrubbing sending events.
+                Only a single method should be implemented in the class:
+                ```
+                    def scrub_event(self, event):
+                        pass
+                ```
+        Returns:
+            Sentry Guard.
+        """
+        SentryReporter._logger.info(f"Init: {sentry_url}")
+
+        SentryReporter._scrubber = scrubber
+        return sentry_sdk.init(
+            sentry_url,
+            release=None,
+            # https://docs.sentry.io/platforms/python/configuration/integrations/
+            integrations=[
+                LoggingIntegration(
+                    level=logging.INFO,  # Capture info and above as breadcrumbs
+                    event_level=logging.ERROR,  # Send errors as events
+                ),
+                ThreadingIntegration(propagate_hub=True),
+            ],
+            before_send=SentryReporter._before_send,
+        )
+
+    @staticmethod
+    def send(event, post_data, sys_info):
+        """Send the event to the Sentry server
+
+        This method
+            1. Enable Sentry's sending mechanism.
+            2. Extend sending event by the information from post_data.
+            3. Send the event.
+            4. Disables Sentry's sending mechanism.
+
+        Scrubbing the information will be performed in the `_before_send` method.
+
+        During the execution of this method, all unhandled exceptions that
+        will be raised, will be sent to Sentry automatically.
+
+        Args:
+            event: event to send. It should be taken from SentryReporter at
+            post_data: dictionary made by the feedbackdialog.py
+                previous stages of executing.
+            sys_info: dictionary made by the feedbackdialog.py
+
+        Returns:
+            Event that was sent to Sentry server
+        """
+        SentryReporter._logger.info(f"Send: {post_data}, {event}")
+
+        if event is None:
+            return event
+
+        with AllowSentryReports(value=True, description='SentryReporter.send()'):
+            # prepare event
+            if CONTEXTS not in event:
+                event[CONTEXTS] = {}
+
+            if TAGS not in event:
+                event[TAGS] = {}
+
+            event[CONTEXTS][REPORTER] = {}
+
+            # tags
+            tags = event[TAGS]
+            tags['version'] = get_value(post_data, 'version')
+            tags['machine'] = get_value(post_data, 'machine')
+            tags['os'] = get_value(post_data, 'os')
+            tags['platform'] = get_first_item(get_value(sys_info, 'platform'))
+            tags[('%s' % PLATFORM_DETAILS)] = get_first_item(get_value(sys_info, PLATFORM_DETAILS))
+
+            # context
+            context = event[CONTEXTS]
+            reporter = context[REPORTER]
+            version = get_value(post_data, 'version')
+
+            context['browser'] = {'version': version, 'name': 'Tribler'}
+
+            reporter[STACKTRACE] = parse_stacktrace(get_value(post_data, 'stack'))
+            reporter['comments'] = get_value(post_data, 'comments')
+
+            reporter[OS_ENVIRON] = parse_os_environ(get_value(sys_info, OS_ENVIRON))
+            delete_item(sys_info, OS_ENVIRON)
+            reporter[SYSINFO] = sys_info
+
+            sentry_sdk.capture_event(event)
+
+            return event
+
+    @staticmethod
+    def set_user(user_id):
+        """ Set the user to identify the event on a Sentry server
+
+        The algorithm is the following:
+        1. Calculate hash from `user_id`.
+        2. Generate fake user, based on the hash.
+
+        No real `user_id` will be used in Sentry.
+
+        Args:
+            user_id: Real user id.
+
+        Returns:
+            Generated user (dictionary: {id, username}).
+        """
+        # calculate hash to keep real `user_id` in secret
+        user_id_hash = md5(user_id).hexdigest()
+
+        SentryReporter._logger.info(f"Set user: {user_id_hash}")
+
+        Faker.seed(user_id_hash)
+        user_name = Faker().name()
+        user = {'id': user_id_hash, 'username': user_name}
+
+        sentry_sdk.set_user(user)
+        return user
+
+    @staticmethod
+    def get_allow_sending():
+        """ Indicate whether Sentry allowed or disallowed to sent events.
+
+        Returns:
+            Bool
+        """
+        allow_sending_in_thread = SentryReporter._allow_sending_in_thread.get(None)
+        if allow_sending_in_thread is not None:
+            return allow_sending_in_thread
+
+        return SentryReporter._allow_sending_global
+
+    @staticmethod
+    def allow_sending_globally(value, info=None):
+        """ Setter for `_allow_sending_global` variable.
+
+        It globally allows or disallows Sentry to send events.
+        If `_allow_sending_in_thread` is not set, then `_allow_sending_global`
+        will be used.
+
+        Args:
+            value: Bool
+            info: String that will be used as an indicator of allowing or
+            disallowing reason (or the place from which this method has been
+            invoked).
+
+        Returns:
+            None
+        """
+        SentryReporter._logger.info(f"Allow sending globally: {value}. Info: {info}")
+        SentryReporter._allow_sending_global = value
+
+    @staticmethod
+    def allow_sending_in_thread(value, info=None):
+        """ Setter for `_allow_sending` variable.
+
+        It allows or disallows Sentry to send events in the current thread.
+        If `_allow_sending_in_thread` is not set, then `_allow_sending_global`
+        will be used.
+
+        Args:
+            value: Bool
+            info: String that will be used as an indicator of allowing or
+            disallowing reason (or the place from which this method has been
+            invoked).
+
+        Returns:
+            None
+        """
+        SentryReporter._logger.info(f"Allow sending in thread: {value}. Info: {info}")
+        SentryReporter._allow_sending_in_thread.set(value)
+
+    @staticmethod
+    def _before_send(event, _):
+        """The method that is called before each send. Both allowed and
+        disallowed.
+
+        The algorithm:
+        1. If sending is allowed, then scrub the event and send.
+        2. If sending is disallowed, then store the event in
+            `SentryReporter.last_event`
+
+        Args:
+            event: event that generated by Sentry
+            hint: root exception (can be used in some cases)
+
+        Returns:
+            The event, prepared for sending, or `None`, if sending is suppressed.
+        """
+        if not event:
+            return event
+
+        SentryReporter._logger.info(f"Before send event: {event}")
+        SentryReporter._logger.info(f"Is allow sending: {SentryReporter._allow_sending_global}")
+        # to synchronise error reporter and sentry, we should suppress all events
+        # until user clicked on "send crash report"
+        if not SentryReporter.get_allow_sending():
+            SentryReporter._logger.info("Suppress sending. Storing the event.")
+            SentryReporter.last_event = event
+            return None
+
+        # clean up the event
+        SentryReporter._logger.info(f"Clean up the event with scrubber: {SentryReporter._scrubber}")
+        if SentryReporter._scrubber:
+            event = SentryReporter._scrubber.scrub_event(event)
+
+        return event
+
+
+class AllowSentryReports:
+    """ This class designed for simplifying allowing and disallowing
+    Sentry's sending mechanism for particular blocks of code.
+
+    It is thread-safe, and use `SentryReporter.allow_sending_in_thread` method
+    for setting corresponding variable.
+
+    Example of use:
+    ```
+        with AllowSentryReports(value=True):
+            do_some_work()
+    ```
+    """
+
+    def __init__(self, value=True, description='', reporter=None):
+        """ Initialising a value and a reporter
+
+        Args:
+            value: Value that will be used for passing in
+                `SentryReporter.allow_sending`.
+            description: Will be used while logging.
+            reporter: Instance of a reporter. This argument mostly use for
+                testing purposes.
+        """
+        self._logger = logging.getLogger(self.__class__.__name__)
+        self._logger.info(f'Value: {value}, description: {description}')
+
+        self._value = value
+        self._saved_state = None
+        self._reporter = reporter or SentryReporter()
+
+    def __enter__(self):
+        """Set SentryReporter.allow_sending(value)
+        """
+        self._logger.info('Enter')
+        self._saved_state = self._reporter.get_allow_sending()
+
+        self._reporter.allow_sending_in_thread(self._value, 'AllowSentryReports.__enter__()')
+        return self._reporter
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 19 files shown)


drew2a:
retest it please

drew2a:
retest this please

drew2a:
retest this please

drew2a:
retest this please

ichorid (changes requested):
Would you kindly run `pre-commit` hooks to format the stuff using `black`?

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
```diff
@@ -0,0 +1,188 @@
...
+        "os": 'posix',
+        "timestamp": 42,
+        "sysinfo": '',
+        "comments": 'comment',
+        "stack": 'some\nstack',
+    }
+
+    assert reporter.send({'a': 'b'}, post_data, None) == {
+        'a': 'b',
+        'contexts': {'browser': {'name': 'Tribler', 'version': '0.0.0'},
```
ichorid, line 116: This stuff could really use some `black` formatting.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_scrubber.py
```diff
@@ -0,0 +1,253 @@
...
+    SYS_ARGV
+from tribler_common.sentry_reporter.sentry_scrubber import SentryScrubber
+
+
+@pytest.fixture(name="scrubber")  # this workaround implemented only for pylint
+def fixture_scrubber():
+    return SentryScrubber()
+
+
+def test_patterns(scrubber):
```
ichorid, line 16: :+1: for really thorough tests!

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
```diff
@@ -0,0 +1,125 @@
...
+        l: a list
+        default: a default value to return in case no items in list.
+
+    Returns:
+        The first item in list.
+    """
+    return l[0] if l else default
+
+
+def last(l, default=''):
```
ichorid, line 63: Personally, I think adding 9 lines of documentation for a trivial one-liner is clear overkill, and actually *decreases* readability. I found it much easier to understand what the function does just by looking at the code. Having to parse through the comments required some effort from me.

Also, the variable is named `l`. People don't use `l` as a variable name for a reason (there is even a pylint check for that).

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_scrubber.py
```diff
@@ -0,0 +1,253 @@
...
+    assert any(regex.search('/usr/local/path/') for regex in scrubber.re_folders)
+    assert any(regex.search('/Users/username/some/') for regex in scrubber.re_folders)
+    assert any(regex.search('/users/username/some/long_path') for regex in scrubber.re_folders)
+    assert any(regex.search('/home/username/some/') for regex in scrubber.re_folders)
+    assert any(regex.search('/data/media/username3/some/') for regex in scrubber.re_folders)
+    assert any(regex.search('WINNT\\Profiles\\username\\some') for regex in scrubber.re_folders)
+    assert any(regex.search('Documents and Settings\\username\\some') for regex in scrubber.re_folders)
+
+    # ip negative
+    assert not scrubber.re_ip.search('0.0.0')
```
ichorid, line 33: Maybe it is better to implement these tests as a list of arguments?

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_scrubber.py
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+
+def test_entities_recursively(scrubber):
+    # positive
+    assert scrubber.scrub_entity_recursively(None) is None
+    assert scrubber.scrub_entity_recursively({}) == {}
+    assert scrubber.scrub_entity_recursively([]) == []
+    assert scrubber.scrub_entity_recursively('') == ''
+    assert scrubber.scrub_entity_recursively(42) == 42
```
ichorid, line 197: :dolphin: :+1: :fish:

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
drew2a: Why should we use black formatting?

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
drew2a: What can you offer instead?

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_scrubber.py
drew2a: Yes, maybe :)

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_scrubber.py
drew2a: 🌸

src/tribler-core/tribler_core/session.py
```diff
@@ -249,11 +256,20 @@ def unhandled_error_observer(self, loop, context):
...
+        with AllowSentryReports(value=False, description=description):
+            self._logger.error("Unhandled exception occurred! %s", text_long)
+
+        sentry_event = SentryReporter.last_event
+
+        if not self.api_manager:
+            return
+
+        events = self.api_manager.get_endpoint('events')
+        events.on_tribler_exception(text_long, sentry_event)
```
drew2a, line 269: @ichorid is it the proper way to transfer exceptions from `core` to `GUI`?

ichorid: Actually, there are two mechanisms for transferring exceptions to the GUI:
* through the `state` endpoint
* through the `events` endpoint

The `events` thing is asynchronous and is the primary way of indicating errors. The `state` thing is synchronous (it is called by `check_core_ready` in the GUI) and it is a redundant way of passing the error to the GUI even in case the `events` endpoint has not connected yet.

So, you should keep both, at least until we come up with a better system.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
ichorid: https://github.com/Tribler/tribler/issues/4104

Short answer: because we had a thorough discussion about code autoformatting and decided that we will use it in the GUI and in some parts of the codebase.

Actually, I should have already added `pre-commit` checks to Jenkins jobs to check the conformance. I'll probably do it this weekend.

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
ichorid: `x`?

src/tribler-core/tribler_core/session.py
drew2a: Ok, thank you! Will implement it.

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
drew2a: Nice! `iterable` looks ok, will change the name.

drew2a: Fixed.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
drew2a: @ichorid no, please! 

Now we have:
1. pylint check
2. flake8 check
3. sonarcloud
4. deepcode-ci-bot

And you want to add a mandatory `pre-commit` check 🤦 

We need to decrease the count of checking tools, not increase.

Your citate:
> After that, every developer's work will become much easier, as instead of fixing pylint errors by hand the developer will only have to hit Ctl-Alt-L once in their PyCharm

That's not true at all! I spent a lot of my time in an effort to please the numerous checkers *after* Ctl-Alt-L. 
It is boring and annoying.

PS In the issue, you gave me I see no consensus on `Black`

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
```diff
@@ -47,30 +47,30 @@ def parse_stacktrace(stacktrace):
...
     Args:
-        l: a list
+        iterable: an iterable sequence
         default: a default value to return in case no items in list.
 
     Returns:
         The first item in list.
     """
-    return l[0] if l else default
+    return iterable[0] if iterable else default
```
ichorid, line 60: Actually, I expected the description to be removed too, to increase readability.
Contrast:

```python

def first(iterable, default=''):
    """ Safe return the first item in a list.

    Args:

        l: a list
        iterable: an iterable sequence
        default: a default value to return in case no items in list.

    Returns:

        The first item in list.

    """
    return iterable[0] if iterable else default
```

to

```python

def first_item(iterable, default=''):
    return iterable[0] if iterable else default
```

of even

```python

def first_item(iterable, default=''):
    # Return the first item in the iterable or substitute it with a default value
    return iterable[0] if iterable else default
```



The second one is much easier to parse visually. [The code *is* the documentation in this case](https://en.wikipedia.org/wiki/Self-documenting_code).

When reading the 9-lines-comment version, it took me a disproportionate amount of mental energy to skip through all the comments and find the single line of code under all those comments. The thing is just a non-method, simple function (:+1:  for functional style!) that "returns the first item in the list or substitutes it with a default value". That's a human-readable description that has fewer words then there are lines of code in the function!). Python was *designed* to be readable w/o lengthy comments in such trivial cases.

@drew2a , your code is *beautiful* and *disciplined* but, you know, **sometimes less is more**.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_tools.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_last():
+    assert last(None, '') == ''
+    assert last([], '') == ''
+    assert last(['some'], '') == 'some'
+    assert last(['some', 'value'], '') == 'value'
+
+    assert last((), '') == ''
+    assert last(('some', 'value'), '') == 'value'
+    assert last(range(5)) == 4
```
ichorid, line 26: Do we really have to test Python itself :wink:

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
drew2a: `Self-documenting code` is a beautiful concept. 

I often hear this sentence here in Tribler.
Like "we have no documentation, but our code is the documentation". 
It doesn't work.

That's why sometimes I write *more* comments than needed. 
Because to skip comment is simpler than imagine absence documentation.

But agree, we should keep all in balance.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -0,0 +1,319 @@
...
+            Sentry Guard.
+        """
+        SentryReporter._logger.info(f"Init: {sentry_url}")
+
+        SentryReporter._scrubber = scrubber
+        return sentry_sdk.init(
+            sentry_url,
+            release=None,
+            # https://docs.sentry.io/platforms/python/configuration/integrations/
+            integrations=[
```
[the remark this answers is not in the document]
drew2a: Yes, you a right, But now it is easy to configure levels of capturing information.
Otherwise, you need to spend some time reading documentation.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_tools.py
ichorid: I mean, these are trivial helper one-liners that are intended to make the code *shorter* and *more readable*. What is the probability of someone *changing* them in a way that could break these test? Why would anyone want to do that?

There is literally no complex logic in these one-liners, there is *nothing* to break. Python is a mature language, we trust its developers not to break the basic conventions of the language (e.g. iteration). And as long as those conventions hold these tests *cannot* fail.

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        array: strings that represents tuples delimited by `:`
+            Example: ["KEY:VALUE", "PATH:~/"]
+
+    Returns:
+        Dictionary that made from given array.
+            Example: {"KEY": "VALUE", "PATH": "~/"}
+
+    """
+    result = defaultdict(lambda: '')
```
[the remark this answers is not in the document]
drew2a: Yes, thank you, `defaultdict` can be removed now.

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
kozlovsky: It is probably a bikeshedding, but `iterable` is not a good name here because it means something specific in Python: it is an object with `__iter__()` method for which you can do `iter(obj)` to get an iterator. Something like `items` looks less ambiguous.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -0,0 +1,319 @@
...
+                event[TAGS] = {}
+
+            event[CONTEXTS][REPORTER] = {}
+
+            # tags
+            tags = event[TAGS]
+            tags['version'] = safe_get(post_data, 'version', None)
+            tags['machine'] = safe_get(post_data, 'machine', None)
+            tags['os'] = safe_get(post_data, 'os', None)
+            tags['platform'] = first(safe_get(sys_info, 'platform'
```
ichorid, line 159: You know, it is actually very mysterious why one would write a one-liner with a default value, and then use this default value explicitly as an argument? :thinking: 

Also, I can't understand why `pylint` did not catch those... There is a check for "using the default values in arguments".
Having to parse/count through nested braces is already a nasty exercise(hello, Lisp!). Finding a *default* values after all that effort is... disappointing.

Contrast 
```python

            tags['platform.details'] = first(safe_get(sys_info, 'platform.details',
                                                      None), None)
```
to 
```python

            tags['platform.details'] = first(safe_get(sys_info, 'platform.details'))
```

Which one is easier to understand?

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
ichorid: >That's not true at all! I spent a lot of my time in an effort to please the numerous checkers after Ctl-Alt-L.
It is boring and annoying.

I had the same problem before I embraced `pre-commit`, never had to appease import checkers and linters since that moment: `black` and other `pre-commit`-invoked tools do all the job for me. Try it out, you'll like it! :wink:

ichorid: In the issue, the consensus was: for those areas of code that are marked as `black`-formatted, we use `black`. Developers may add *the code they own* to `black` areas if they wish so. I own the GUI code (together with @devos50) and the MetadataStore code. `tribler-common` was created by me during refactoring, so it is owned by me now. If you wish, you 
can exclude the `sentry`-related code from `black`-area in `tribler-common`- that's your code after all.

Regardless of that, `pre-commit` also **performs automatic import sorting for the whole codebase** and runs lint-checks locally (for the programmer's convenience).

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
drew2a: @kozlovsky @ichorid renamed.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -0,0 +1,319 @@
...
+    server.
+
+    # SentryReporter and multithreading.
+
+    SentryReporter doesn't know in which thread it is running. That means all
+    the "threading" orchestration should be done outside this class.
+
+    SentryReporter should be initialized in each thread separately.
+
+    Note. SentryReporter is a completely static class, and it is store state in
```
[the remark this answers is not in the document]
drew2a: 🙏

xoriole:
@drew2a Could you update `sentry_url` from environment variable to `version.py` via this script below? It will make sure that builder works as well.
https://github.com/Tribler/tribler/blob/d89f5ca0f2b9f58632b9d73b60c52862af62aab0/build/update_version_from_git.py#L35-L37

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
drew2a: Yes, thank you to pointing this out. 
Changed.

>You know, it is actually very mysterious why one would write a one-liner with a default value, and then use this default value explicitly as an argument? 🤔

Because defaults in functions can be changed. 
My intention was to keep the behavior of this particular function unchanged and independent of function defaults.

drew2a:
@xoriole yes, sure! 👌

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -0,0 +1,319 @@
...
+    class variables to share them between all instances.
+    """
+    last_event = None
+
+    _allow_sending = False
```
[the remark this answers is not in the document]
drew2a: Have discussed it p2p.

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_tools.py
drew2a: >Do we really have to test Python itself

It has been added after renaming `l`->`iterable`. Not needed now.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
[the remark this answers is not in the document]
drew2a: Fixed

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_tools.py
drew2a: Fixed

src/tribler-core/tribler_core/session.py
drew2a: Implemented

src/tribler-common/tribler_common/sentry_reporter/tests/test_sentry_reporter.py
drew2a: `pre-commit` formating has been applied

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
```diff
@@ -0,0 +1,124 @@
...
+        items: a sequence
+        default: a default value to return in case no items in list.
+
+    Returns:
+        The first item in list.
+    """
+    return items[0] if items else default
+
+
+def last(items, default=None):
```
ichorid, line 62: изображение

ichorid (approved):
Personally, I still perceive 9-line comments and tests for trivial one-liners a waste of both the writer's and the reader's mental energy, exactly the thing Python was designed to avoid.

But we need this PR in 7.6. 

p.s. 
Be warned that I will remove excessive comments if I ever touch the said code :smiling_imp:

ichorid (approved):
Thanks! :+1:

drew2a:
@ichorid you are welcome :)

kozlovsky (dismissed):

xoriole (approved):
After the merge, please make sure to update the build jobs with the right sentry URL.
https://jenkins-ci.tribler.org/job/Build-Tribler_release/job/Build/

PR #5743, opened 2020-11-20
Added constant for infinite amount of walk strategy target peers

qstokkink:
You can now use `target_peers=INFINITE`, which is equivalent to `target_peers=-1`.


drew2a (approved):
Beautiful!

PR #5745, opened 2020-11-22
Removed encoding from json loads call

### LINKED ISSUE #5744: Encoding argument removed in Python 3.9
In Python 3.9, the `encoding` keyword argument got removed from the call to `json.loads`. This results in the following error:

```
[PID:55108] 2020-11-21 16:58:23,992 - ERROR <tribler_window:108> root.on_exception(): Traceback (most recent call last):
  File "/Users/martijndevos/Documents/tribler/src/tribler-gui/tribler_gui/tribler_request_manager.py", line 120, in <lambda>
    request.reply.finished.connect(lambda: request.on_finished(request))
  File "/Users/martijndevos/Documents/tribler/src/tribler-gui/tribler_gui/tribler_request_manager.py", line 203, in on_finished
    json_result = json.loads(bytes(data), encoding='latin_1')
  File "/Users/martijndevos/Documents/tribler/src/tribler-core/tribler_core/utilities/json_util.py", line 112, in loads
    return json.loads(s, *args, **kwargs)
  File "/usr/local/Cellar/python@3.9/3.9.0_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 359, in loads
    return cls(**kw).decode(s)
TypeError: __init__() got an unexpected keyword argument 'encoding'
```

I'm not entirely sure why we have chosen for that particular encoding, but removing the keyword argument seems to correctly start Tribler.

devos50:
This fixes compatibility with Python 3.9.

Fixes #5744

### DIFF
--- a/src/tribler-gui/tribler_gui/tribler_request_manager.py
+++ b/src/tribler-gui/tribler_gui/tribler_request_manager.py
@@ -200,7 +200,7 @@ class TriblerNetworkRequest(QObject):
             if not self.decode_json_response:
                 self.received_json.emit(data)
                 return
-            json_result = json.loads(bytes(data), encoding='latin_1')
+            json_result = json.loads(bytes(data))
             if (
                 'error' in json_result
                 and self.capture_core_errors


qstokkink (approved):
We should perhaps also look at if the `bytes(data)` cast is really necessary, in a future PR.

devos50:
@qstokkink indeed, I didn't even notice that cast 😅

PR #5746, opened 2020-11-23
ipv8 module catalog refactoring

drew2a:
This PR replaces some strings from `ipv8_module_catalog.py` which represents `Class.__module__` and `Class.__name__` by corresponding special attributes.

It makes navigation in IDE easy and will save us from changing them manually in case of class names changes or location changes.

What do you think?

### DIFF
--- a/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py
+++ b/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py
@@ -1,3 +1,4 @@
+# pylint: disable=import-outside-toplevel
 import inspect
 import sys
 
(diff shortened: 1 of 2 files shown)


qstokkink:
This PR _unmodularizes the modules_ by killing the lazy imports and using top-level imports.
Top-level imports are exactly what this file is supposed to avoid.

If you absolutely want IDE friendly import statements, I would suggest going back to the old explicit lazy imports. See this link for how it used to look: https://github.com/Tribler/tribler/blob/32eb399990d5d183464fa84ad91be5a2f0c144a4/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py

**Trivia**. Next to my fundamental problem with this PR, these wrappers can also take an actual class argument instead of deriving modules and names:

```python
@overlay(PopularityCommunity.__module__, PopularityCommunity.__name__)
 |
 v
@overlay(PopularityCommunity)
```

ichorid:
I would prefer to have the thing in YAML for readability.

qstokkink:
> I would prefer to have the thing in YAML for readability.

You would miss out on these beautiful docstrings describing the functionality of each of these decorators when you press `CTRL+Q` in PyCharm: https://github.com/Tribler/py-ipv8/blob/master/ipv8/loader.py#L103-L123

Also, not all Communities can be loaded completely declaratively, for example: https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/modules/ipv8_module_catalog.py#L77-L86
In other words, 50% would be in the config script and 50% in the code. This would only give the illusion of being completely configurable.

> That would enable the user to change it, BTW. Could help with debugging on Windows, you never know.

These decorators allow programmers to execute arbitrary code. Exposing this and giving users the power to do so from a config script seems like a debugging nightmare to me.

Imagine having to figure out that some user decided to remove the `metadata_store` argument from the `GigaChannelCommunity` `kwargs` 😃

qstokkink (changes requested):
_Must fix_: this PR breaks the lazy importing functionality.

qstokkink:
I've been giving this some thought. This is another possible middle-of-the-road solution (using both explicit imports and lazy loading):

```python
def GigaChannelCommunity():
    from tribler_core.modules.metadata_store.community.gigachannel_community import GigaChannelCommunity
    return GigaChannelCommunity

 ...

@overlay(GigaChannelCommunity)
class GigaChannelCommunityLauncher(IPv8CommunityLauncher):
    pass
```

This would require some changes to IPv8 to allow it to take a `callable` as the decorator argument.

Personally, I don't like this as much as the current form. But, I understand why you would want explicit imports.

drew2a:
@qstokkink indeed, string form is shorter and looks nicer, but your last proposition looks more...  solid? 
I would definitely prefer the last.

I also can offer any help with the implementation :)

qstokkink:
You're welcome to author the PR on IPv8.

The edits would involve changing:

1. These two lines:

https://github.com/Tribler/py-ipv8/blob/0504e61d8a8f7b14171a6256327ac5a3d417943b/ipv8/loader.py#L134
https://github.com/Tribler/py-ipv8/blob/0504e61d8a8f7b14171a6256327ac5a3d417943b/ipv8/loader.py#L177

into

```python
return str_module_or_class() if isinstance(str_module_or_class, types.FunctionType) else str_module_or_class
```

and

```python
strategy_class = str_module_or_class() if isinstance(str_module_or_class, types.FunctionType) else str_module_or_class
```

 2. Updating the docstrings.
 3. Adding two unit tests to https://github.com/Tribler/py-ipv8/blob/master/ipv8/test/test_loader.py for `test_overlay_class_from_function` and `test_walk_strategy_from_function`.

drew2a:
Another option: to roll back my PR and use:

```python
def GigaChannelCommunity():
    from tribler_core.modules.metadata_store.community.gigachannel_community import GigaChannelCommunity
    return GigaChannelCommunity

 ...

@overlay(GigaChannelCommunity())
class GigaChannelCommunityLauncher(IPv8CommunityLauncher):
    pass
```

:)

drew2a:
retest this please

qstokkink:
👍 implemented properly. However, to run this you need the latest version of IPv8. The latest version is incompatible with the current Tribler version. I'm resolving this in #5753.

qstokkink:
Alright, merged. Please rebase onto the latest devel and everything should be solved.

qstokkink:
The `hiddenimports` error is expected, there are no more hidden imports (so the set is empty). The test can be changed.

qstokkink (approved):

Issue #5715, opened 2020-11-02
[7.5.4] Exception: Tribler is shutting down


drew2a:
Stack:
```
[
Traceback (most recent call last):, 
  File "src\tribler-gui\tribler_gui\event_request_manager.py", line 138, in <lambda>, 
  File "src\tribler-gui\tribler_gui\event_request_manager.py", line 80, in on_error, 
tribler_gui.event_request_manager.CoreConnectTimeoutError: Could not connect with the Tribler Core within 60 seconds, 
[PID:8592] 2020-11-01 18:45:18,015 - ERROR <rest_manager:38> tribler_core.restapi.rest_manager.error_middleware(): Tribler is shutting down, 
Traceback (most recent call last):, 
  File "src\tribler-core\tribler_core\restapi\rest_manager.py", line 35, in error_middleware, 
Exception: Tribler is shutting down
]
```

devos50:
Shouldn't we attach this issue to a milestone?

drew2a:
This issue looks like an issue about `lack of information`.

Probably I can fix it by adding sentry on the customer side.

Issue #3567, opened 2018-04-06
python-tk is missing from deb dependencies


Dmole:
python-tk is missing from deb dependencies.

devos50:
Thank you for your report. It seems that some Python distributions are missing Tkinter. We will add it to the recommended section of the Debian configuration files.

PR #5747, opened 2020-11-24
Better error handling

drew2a:
Resolves https://github.com/Tribler/tribler/issues/5715

This PR:
1. Adds unhandled error handling on startup (before "core-gui" interaction is established)
1. Adds error handling to the core error handler 🕵️ 

Two important places to review:
1. run_tribler.py
1. session.py

If there is an error occurs during startup, then two message boxes will appear:

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -88,13 +88,18 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_m
     get_event_loop().run_forever()
 
 
+def excepthook(exctype, value, traceback): # pylint: disable=unused-argument
+    SentryReporter.send_exception_with_confirmation(value)
+
+
 if __name__ == "__main__":
     SentryReporter.init(sentry_url=sentry_url, scrubber=SentryScrubber())
     SentryReporter.allow_sending_globally(False, 'run_tribler.__main__()')
 
+    sys.excepthook = excepthook
+
     # Get root state directory (e.g. from environment variable or from system default)
     root_state_dir = get_root_state_directory()
-
     # Check whether we need to start the core or the user interface
     if 'CORE_PROCESS' in os.environ:
         # Check for missing Core dependencies
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -1,6 +1,9 @@
 import logging
+import sys
 from contextvars import ContextVar
 from hashlib import md5
+
+from PyQt5.QtWidgets import QApplication, QMessageBox
 
 from faker import Faker
 
@@ -83,8 +86,7 @@ class SentryReporter:
         Returns:
             Sentry Guard.
         """
-        SentryReporter._logger.info(f"Init: {sentry_url}")
-
+        SentryReporter._logger.debug(f"Init: {sentry_url}")
         SentryReporter._scrubber = scrubber
         return sentry_sdk.init(
             sentry_url,
@@ -101,7 +103,7 @@ class SentryReporter:
         )
 
     @staticmethod
-    def send(event, post_data, sys_info):
+    def send_event(event, post_data=None, sys_info=None):
         """Send the event to the Sentry server
 
         This method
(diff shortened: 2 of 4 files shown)


src/run_tribler.py
```diff
@@ -3,7 +3,9 @@
 import os
 import signal
 import sys
+import tkinter
```
qstokkink, line 6: [Quick note while this is in WIP] `tkinter` is no longer a default package in Python, see:

https://github.com/Tribler/tribler/issues/3567
https://github.com/Tribler/tribler/issues/4799

The recommended approach is to use Qt for popups.

drew2a: It is a pity... Well, thank you for pointing this out.

drew2a: Fixed

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -165,6 +167,31 @@ def send(event, post_data, sys_info):
 
             return event
 
+    @staticmethod
+    def send_exception_with_confirmation(exception):
```
drew2a, line 171: I'm not sure that SentryReporter class is the proper place for this function.
If you have an idea of a better place, please, let me know.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -165,6 +167,31 @@ def send(event, post_data, sys_info):
...
+        SentryReporter._logger.debug(f"Send exception with confirmation: {exception}")
+
+        _ = QApplication(sys.argv)
+        messagebox = QMessageBox(icon=QMessageBox.Critical, text=f'{exception}.')
+        messagebox.setWindowTitle("Error")
+        messagebox.exec()
+
+        messagebox = QMessageBox(
+            icon=QMessageBox.Question,
+            text='Do you want to sent the crash report to the Tribler team? ' 'All the data will be anonymized.',
```
drew2a, line 188: Please, verify the correctness of the string.

qstokkink: - "sent" -> "send"

Nits:

- "the crash report" -> "this crash report"

 - Perhaps you could make the second sentence a bit less sterile by stating something like:

"We anonymize all your data, who you are and what you downloaded."

drew2a: @qstokkink thank you!

src/run_tribler.py
```diff
@@ -88,7 +88,13 @@ async def start_tribler():
     get_event_loop().run_forever()
 
 
+def excepthook(exctype, value, _):
+    SentryReporter.send_exception_with_confirmation(f'{exctype}: {value}.')
```
kozlovsky, line 92: It is possible that `SentryReporter.send_exception_with_confirmation()` will be called before `SentryReporter.init`, if the exception happens inside `SentryReporter.init()`. I think it should be handled. At least it is possible to set up some flag at the end of `SentryReporter.init()` that the reporter is properly initialized and check this flag inside `SentryReporter.send_exception_with_confirmation()`

xoriole (approved):

src/tribler-core/tribler_core/session.py
```diff
@@ -220,48 +220,51 @@ def unhandled_error_observer(self, loop, context):
...
+            if not self.api_manager:
+                return
+
+            events = self.api_manager.get_endpoint('events')
+            events.on_tribler_exception(text_long, sentry_event)
+
+            state = self.api_manager.get_endpoint('state')
+            state.on_tribler_exception(text_long, sentry_event)
+        except Exception as e:
+            SentryReporter.send_exception_with_confirmation(e)
```
kozlovsky, line 266: This code looks like if some exception `e` were watched, it would be sent to Sentry. But, actually, `SentryReporter. send_exception_with_confirmation` uses `e` exception only for displaying it to the user. The exception that is actually sent is `SentryReporter.last_event`, which is not set here or inside `SentryReporter. send_exception_with_confirmation`.

Is it really true that `SentryReporter.last_event` is properly initialized here, does not equal to None, and actually corresponds to the `e` exception? The logic does not look straightforward to me.

As I understand it, `SentryReporter.last_event` becomes initialized as a side effect of the `before_send` hook that should be called when Sentry is trying to send some event. But does Sentry really sees the `e` exception? It was generated in our code, we catch it successfully, so if I understand correctly, Sentry will not have a chance to look at the exception and will not call the `before_send` hook as well.

drew2a: >  But does Sentry really sees the e exception?

Yes, it does.

src/run_tribler.py
drew2a: Yes, it is possible that `SentryReporter.init` will fail.
But if `SentryReporter.init` fails it will not be possible to send us the event.
Then no reason to handle this situation.

src/tribler-core/tribler_core/session.py
drew2a: To make the function purpose more clear, its name has been changed.

drew2a: I will take the time to test it once more.

drew2a: Yes, you are right. I've changed  `SentryReporter.send_exception_with_confirmation(e)` to send this particular exception with no magic.

kozlovsky (dismissed):

drew2a:
retest this please

PR #5758, opened 2020-11-30
Add upgrade procedure for Channels DB v8 to v10

ichorid:
The upgrade procedure creates a temporary Channels DB and directly dumps the old Channels tables into the corresponding tables in the new DB.
If the upgrade finishes correctly, the new DB replaces the old one.
If the upgrade is skipped by the user, the old DB is removed
If the upgrade crashes, at the next Tribler startup it will be tried anew.

The main reason to upgrade is the removal of unused unique indexes that mess up the new DB logic. Unfortunately, it is impossible to remove them without recreating the whole database.

5 GB database converts in 5 minutes on my PC.

### DIFF
--- a/src/tribler-core/tribler_core/upgrade/db8_to_db10.py
+++ b/src/tribler-core/tribler_core/upgrade/db8_to_db10.py
@@ -1,0 +1,124 @@
+import contextlib
+import datetime
+import logging
+import sqlite3
+from asyncio import sleep
+
+TABLE_NAMES = (
+    "ChannelNode", "TorrentState", "TorrentState_TrackerState", "ChannelPeer", "ChannelVote", "TrackerState", "Vsids")
+
+
+class PonyToPonyMigration(object):
+
+    def __init__(self, old_db_path, new_db_path, notifier_callback=None, logger=None):
+        self._logger = logger or logging.getLogger(self.__class__.__name__)
+        self.notifier_callback = notifier_callback
+        self.old_db_path = old_db_path
+        self.new_db_path = new_db_path
+        self.shutting_down = False
+
+    async def update_convert_progress(self, amount, total, elapsed, message=""):
+        if self.notifier_callback:
+            elapsed = 0.0001 if elapsed == 0.0 else elapsed
+            amount = amount or 1
+            est_speed = amount / elapsed
+            eta = str(datetime.timedelta(seconds=int((total - amount) / est_speed)))
+            self.notifier_callback(
+                f"{message}\nConverted: {amount}/{total} ({(amount * 100) // total}%).\nTime remaining: {eta}"
+            )
+            await sleep(0.001)
+
+    async def convert_async(self, convert_command, total_to_convert, offset=0, message=""):
+        """
+        This method copies entries from one Pony db into another one splitting the process into chunks dynamically.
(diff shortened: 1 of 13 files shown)


ichorid:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

src/tribler-core/tribler_core/upgrade/db8_to_db10.py
```diff
@@ -0,0 +1,116 @@
+import contextlib
+import datetime
+import logging
+import sqlite3
+from asyncio import sleep
+
+TABLE_NAMES = (
+    "ChannelNode", "TorrentState", "TorrentState_TrackerState", "ChannelPeer", "ChannelVote", "TrackerState", "Vsids")
```
kozlovsky, line 8: Should the `MiscData` table also be on this list?

ichorid: No. MiscData is currently only used for upgrader marks and database version.

ichorid:
retest this please

src/tribler-core/tribler_core/upgrade/db8_to_db10.py
```diff
@@ -0,0 +1,116 @@
...
+    def get_table_entries_count(self, cursor, table_name):
+        return cursor.execute(f"SELECT COUNT(*) FROM {table_name};").fetchone()[0]
+
+    async def convert_table(self, cursor, table_name, column_names):
+        def convert_command(offset, batch_size):
+            sql_command = None
+            try:
+                sql_command = f"INSERT OR IGNORE INTO {table_name} ({column_names}) " + \
+                              f"SELECT {column_names} FROM old_db.{table_name} " + \
+                              f"LIMIT {batch_size} {('OFFSET ' + str(offset)) if offset else ''} ;"
```
kozlovsky, line 73: I'd add `ORDER BY  ROWID` just to be sure that SELECT returns rows in a stable order

kozlovsky:
Screenshot 2020-11-30 at 19 45 56

Currently, when I press the "Skip" button during the upgrade I see the window with the following message: "Abort the conversion of old channels. The upgrade procedure is now converting torrents in channels collected by the previous installation of Tribler. Are you sure you want to abort the conversion process?".

If I understand correctly, choosing "Abort" will delete the previous database without the possibility to recover it. The message does not warn about it. It can be understood as "the upgrade is optional, you can abort conversion procedure and then continue to use your old database". We need to warn the user that all data will be lost if pressing the "Abort" button.

ichorid:
изображение


Is it better now? :wink:

kozlovsky:
image

When I started the upgrade the speed of copying metadata info was about 1000 rows/seconds, and now it is about 200 rows/seconds. It seems that all upgrade works as a single transaction, and big transactions are not always efficient. Probably it should be split into several transactions of smaller size.

kozlovsky:
> Is it better now? 😉

Definitely. But the user will lose not only the personal channel but all other loaded channels as well.

kozlovsky (approved):

kozlovsky (approved):

PR #5760, opened 2020-11-30
Add unhandled exceptions catching 

drew2a:
This PR restores https://github.com/Tribler/tribler/pull/5747 by capturing all unhandled exceptions with mandatory confirmation before sending.

## The issue:
https://github.com/Tribler/tribler/pull/5747 handles all types of exceptions, which caused false reports. Because of that, it has been temporarily disabled.

## The solution:
Сapturing all unhandled exceptions with mandatory confirmation before sending has been added.
Now Sentry can operate in three modes:

1. Send reports are allowed
1. Send reports are allowed with a confirmation dialog
1. Send reports are suppressed (disallowed, but the last event will be stored)

By default, it works in the second mode.

Exception filtering for unhandled exceptions has been added as well: 
```python
ignored_exceptions = [KeyboardInterrupt, SystemExit]
```

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -89,16 +89,14 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_m
 
 
 if __name__ == "__main__":
-    SentryReporter.init(sentry_url=sentry_url, scrubber=SentryScrubber())
-    SentryReporter.allow_sending_globally(False, 'run_tribler.__main__()')
-
+    SentryReporter.init(sentry_url=sentry_url, scrubber=SentryScrubber(),
+                        strategy=SentryReporter.Strategy.SEND_ALLOWED_WITH_CONFIRMATION)
     # Get root state directory (e.g. from environment variable or from system default)
     root_state_dir = get_root_state_directory()
     # Check whether we need to start the core or the user interface
     if 'CORE_PROCESS' in os.environ:
         # Check for missing Core dependencies
         check_for_missing_dependencies(scope='core')
-
         base_path = os.environ['CORE_BASE_PATH']
         api_port = os.environ['CORE_API_PORT']
         api_key = os.environ['CORE_API_KEY']
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -1,6 +1,7 @@
 import logging
 import sys
 from contextvars import ContextVar
+from enum import Enum, auto
 from hashlib import md5
 
 from PyQt5.QtWidgets import QApplication, QMessageBox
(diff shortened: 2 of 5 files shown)


src/run_tribler.py
```diff
@@ -88,10 +89,21 @@ async def start_tribler():
     get_event_loop().run_forever()
 
 
+def excepthook(exctype, value, traceback):  # pylint: disable=unused-argument
+    ignored_exceptions = [KeyboardInterrupt, ]
+    ignored_exceptions.extend(IGNORED_ERRORS)
+    if exctype in ignored_exceptions:
+        return
```
qstokkink, line 96: `IGNORED_ERRORS` contains both exception classes and tuples of exception classes and error codes:

https://github.com/Tribler/tribler/blob/63dc402cb4def5743099e1789fefd07942f9d5ea/src/tribler-core/tribler_core/session.py#L63-L76

Your code would ignore all tuple entries (essentially only checking for `gaierror`). If this is intended please add a line comment. If it is unintended.. there is probably some way to get the error code from the `traceback` 🤷.

I would also suggest adding [SystemExit](https://docs.python.org/3/library/exceptions.html#SystemExit) to the ignored exceptions - it is the second "special" type of Python exception.

qstokkink (changes requested):

src/run_tribler.py
drew2a: 🤦 thanks

drew2a: At first, I want to add all errors (without codes) from `IGNORED_ERRORS', but it is not necessary. 
`session.unhandled_error_observer(self, loop, context)` should eat them anyway.

I've added `SystemExit` and remove `ignored_exceptions.extend(IGNORED_ERRORS)`.

Thank you for pointing this out.

drew2a:
@qstokkink @xoriole the PR is ready to review.

The first PR's version was short and straightforward, but it added a hidden layer of complexity (three levels of exception handlers).

I rethought and rewrite SentryReporter.
Now we have only two levels of crashes handlers:
1. All unhandled exceptions will go to the SentryReporter.
1. All handled exceptions will go to the GUI.

qstokkink (approved):
The PR history is a bit messy for my taste, but the content seems fine 👍

xoriole (dismissed):

xoriole:
Tried out the PR, the issue with terminating the process is no more. It now shows the KeyboardInterrupt logs in the console.
```
  File "/home/jupiter/tudelft/tribler/src/run_tribler.py", line 105, in <module>
start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_mode=core_test_mode)
  File "/home/jupiter/tudelft/tribler/src/run_tribler.py", line 88, in start_tribler_core
    get_event_loop().run_forever()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 1823, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/lib/python3.8/selectors.py", line 468, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
KeyboardInterrupt
```

drew2a:
> The PR history is a bit messy for my taste
I'll squash it, don't worry :)

qstokkink (approved):

PR #5773, opened 2020-12-02
Fixes #5762: unset 'auto_managed' flag to prevent torrents resuming automatically

### LINKED ISSUE #5762: Cannot stop torrent downloading using devel branch of Tribler
If I start downloading some torrent using devel branch (let's say [Ubuntu 20.10 torrent](https://releases.ubuntu.com/20.10/ubuntu-20.10-desktop-amd64.iso.torrent) for specificity), and then press the "stop" button (looked as a square in a circle), or select the "Stop" command of the drop-down menu for this torrent file, I see the following:

In the next few seconds, the torrent status switches several times between "Downloading" and "Stopped" back and force, and then stabilizes on "Stopped". But then, half a minute later, the torrent again become "Downloading". It was not possible for me to stop downloading completely, it always started again after some time.

When I try the same on the 7.5.3 release, everything works as expected. The torrent switches from "Downloading" to "Stopped" (well, there is actually a very short switch back to "Downloading" and then finally to "Stopped"), and then it remains in the "Stopped" state without a spontaneous switch to "Downloading" state again.

kozlovsky:
In LibTorrent auto_managed flag is now on by default, and as a result, any torrent's state can change from Stopped to Downloading at any time. Here I unset this flag to prevent auto-resuming of stopped torrents.

### DIFF
--- a/src/tribler-core/tribler_core/modules/libtorrent/download.py
+++ b/src/tribler-core/tribler_core/modules/libtorrent/download.py
@@ -182,6 +182,11 @@ class Download(TaskManager):
 
         self.handle = alert.handle
         self._logger.debug("Added torrent %s", str(self.handle.info_hash()))
+        # In LibTorrent auto_managed flag is now on by default, and as a result
+        # any torrent's state can change from Stopped to Downloading at any time.
+        # Here we unset this flag to prevent auto-resuming of stopped torrents.
+        if hasattr(self.handle, 'unset_flags'):
+            self.handle.unset_flags(lt.add_torrent_params_flags_t.flag_auto_managed)
 
         self.set_selected_files()
 


ichorid:
retest this please

egbertbouman:
Since we're adding torrents without the `flag_auto_managed` set, this probably means that the flags field in `atp` isn't being respected. If that's the case, I guess this is a libtorrent bug?

kozlovsky:
retest this please

kozlovsky:
> Since we're adding torrents without the flag_auto_managed set, this probably means that the flags field in atp isn't being respected. If that's the case, I guess this is a libtorrent bug?

It is an interesting question, to which I don't have a definitive answer yet. Probably need to check libtorrent sources.

ichorid (approved):

ichorid:
@kozlovsky , are you going to report this to Arvid, or should I do that myself? You'll have to provide a minimum reproducible test though.

kozlovsky:
> are you going to report this to Arvid, or should I do that myself?

At this moment it is unclear to me yet, is it a bug in libtorrent or our misunderstanding how to use it

kozlovsky:
I see this example use another way to set atp flags, using some C++ object `libtorrent.add_torrent_params()` instead of dict. Maybe we need to use something similar?
https://github.com/arvidn/libtorrent/blob/773f0f7eef085800617b8800303304b097ec8252/bindings/python/client.py#L157

ichorid:
@egbertbouman , your opinion?

> I see this example use another way to set atp flags, using some C++ object `libtorrent.add_torrent_params()` instead of dict. Maybe we need to use something similar?
> https://github.com/arvidn/libtorrent/blob/773f0f7eef085800617b8800303304b097ec8252/bindings/python/client.py#L157

egbertbouman:
I don't really see a reason to prefer `libtorrent.add_torrent_params()` over a `dict`, unless it fixes some issue. Most Python code I've seen uses a `dict`.

PR #5797, opened 2020-12-04
Change PopularityCommunity id to prevent crashing legacy clients

ichorid:


ichorid:
Useless to fix the problem with crashing 7.5.4, as pointed out by @egbertbouman ...

PR #5798, opened 2020-12-04
Fix GUI crashing on negative votes

ichorid:


src/tribler-gui/tribler_gui/utilities.py
```diff
@@ -325,7 +325,8 @@ def votes_count(votes=0.0):
     votes = float(votes)
     # FIXME: this is a temp fix to cap the normalized value to 1.
     #  The votes should already be normalized before formatting it.
-    votes = 1.0 if votes > 1 else votes
+    votes = 1.0 if votes > 1.0 else votes
+    votes = 0.0 if votes < 0.0 else votes
```
drew2a, line 329: You can use `max(0,min(votes, 1))` here.

drew2a (approved):

PR #5800, opened 2020-12-04
Added "compensation payouts" to bandwidth crawler

devos50:
In this PR, I extended the bandwidth crawler to support "compensation payouts". These crawlers can payout some bandwidth to users, which helps us in reducing the effects of updating to Tribler 7.6 (which resets the token balance). Only users with a positive token balance are eligible for this compensation.

I also fixed two other, minor issues:
- Set the `target_peers` for the random walk strategy used by the bandwidth crawler to unlimited.
- Fixed an issue with fetching the latest transaction from the database.

### DIFF
--- a/src/tribler-core/run_bandwidth_crawler.py
+++ b/src/tribler-core/run_bandwidth_crawler.py
@@ -1,14 +1,21 @@
 """
 This executable script starts a Tribler instance and joins the BandwidthAccountingCommunity.
+
+This bandwidth crawler optionally support a payout file. This is used to payout the old balances of users during
+the migration from Tribler 7.5 to 7.6. We will keep track of who we did payout already in a separate file to avoid
+payout out twice.
 """
 import argparse
+import os
 import sys
 from asyncio import ensure_future, get_event_loop
+from binascii import hexlify, unhexlify
 from pathlib import Path
 
-from ipv8.loader import IPv8CommunityLoader
+from ipv8.loader import IPv8CommunityLoader, overlay
 
 from tribler_core.config.tribler_config import TriblerConfig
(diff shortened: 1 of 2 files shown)


grimadas (approved):
L G T M

ichorid:
This is a crawler, so we're supposed to run it ourselves? Why do we have to put it on `release-7.6` then? I would argue that we indeed should not even merge it in devel, but just keep it in a separate branch and run it manually.

devos50:
I deployed this crawler on a machine on our infrastructure.

Regarding code organization, we might choose to move all crawler code to a separate branch or repo. I'm fine with that but then we should move _all_ scripts (including the tunnel helper scripts) for consistency. For now, I've chosen to do it like this.

Should we create a separate issue for this?

devos50:
> We should at least discuss it. I don't want the stuff to be in the same repo, or at least in the same Python package as the main stuff.

That's fair. I would be fine with moving these scripts in a separate repo, but then I argue we should treat experimental code (e.g., in the `experiments` directory) the same.

synctext:
bit of a monorepo person. just browsing through all the fragmentation we have. So many Tribler/repoX fragments, I find it more messy..

devos50:
> I don't want the stuff to be in the same repo

> bit of a monorepo person.

I think we already had this discussion a while ago and we decided to keep everything within one repository, right?

synctext:
yes please

ichorid:
I propose to create a separate `tribler-tools` package/dir then in our `src` directory for such stuff. We want to start doing proper package management eventually.

devos50:
This is fixed by #5995.

PR #5835, opened 2020-12-10
Add try-except block to dependency message box

### LINKED ISSUE #5832: [7.6.0] FileNotFoundError: [Errno 2] No such file or directory: 'xmessage'
Sentry Issue: [SENTRY-FOR-760-Q](https://sentry.tribler.org/organizations/tribler/issues/271/?referrer=github_integration)

```
FileNotFoundError: [Errno 2] No such file or directory: 'xmessage'
  File "run_tribler.py", line 115, in <module>
    check_for_missing_dependencies(scope='both')
  File "tribler_core/dependencies.py", line 84, in check_for_missing_dependencies
    _show_system_popup("Dependencies missing!",
  File "tribler_core/dependencies.py", line 52, in _show_system_popup
    subprocess.Popen(['xmessage', '-center', text])
  File "subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
```

drew2a:
Fixes #5832 by adding try-except block.

Fix uses broad exception handling as a savest way to catch all exceptions. 
It can't cause any harm in an unobvious way, to further Tribler's execution, because, at the step, when the dependency message box is shown, Tribler stops his work.

### DIFF
--- a/src/tribler-core/tribler_core/dependencies.py
+++ b/src/tribler-core/tribler_core/dependencies.py
@@ -41,20 +41,28 @@ def _show_system_popup(title, text):
     :param text: the pop-up body
     """
     sep = "*" * 80
-    print('\n'.join([sep, title, sep, text, sep]), file=sys.stderr)
+
+    # pylint: disable=import-outside-toplevel, import-error, broad-except
+    print('\n'.join([sep, title, sep, text, sep]), file=sys.stderr)  # noqa: T001
 
     system = platform.system()
-    if system == 'Windows':
-        import win32api
-        win32api.MessageBox(0, text, title)
-    elif system == 'Linux':
-        import subprocess
-        subprocess.Popen(['xmessage', '-center', text])
-    elif system == 'Darwin':
-        import subprocess
-        subprocess.Popen(['/usr/bin/osascript', '-e', text])
-    else:
-        print(f'cannot create native pop-up for system {system}')
+    try:
+        if system == 'Windows':
+            import win32api
+            win32api.MessageBox(0, text, title)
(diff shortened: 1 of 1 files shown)


src/tribler-core/tribler_core/dependencies.py
```diff
@@ -41,20 +41,28 @@ def _show_system_popup(title, text):
...
+        elif system == 'Darwin':
+            import subprocess
+            subprocess.Popen(['/usr/bin/osascript', '-e', text])
+        else:
+            print(f'cannot create native pop-up for system {system}')  # noqa: T001
+    except Exception as exception:
+        # Use base Exception, because code above can raise many
+        # non-obvious types of exceptions:
+        # (SubprocessError, ImportError, win32api.error, FileNotFoundError)
+        print(f'Error while showing a message box: {exception}')  # noqa: T001
```
kozlovsky, line 65: The intention of this `_show_system_popup` function is to show some important message to the user.
When the function is falling back to using print calls, it completely ignores the original message and shows secondary error only.
I think it is better to add a second print call with the title and text passed to this function, to not silently skip it (here and in previous print call)

drew2a: So, do you mean that printing the title and the text on line 46 is not enough?

kozlovsky: Sorry, I misread it

kozlovsky (approved):

PR #5848, opened 2020-12-14
Rollback to "Tribler starts on events start received" logic

### LINKED ISSUE #5845: GUI stuck on "Loading" screen
Latest `devel`, Ubuntu 20.04

90% of attempts to start Tribler result in endless "Loading" screen with spinning gears, while the Core starts normally. The problem disappears if I bring back triggering `tribler_started` signal on `EVENTS_START` notification from the Core.

Apparently, #5821 left our fragile startup mechanism broken on some machines...

drew2a:
This PR fixes #5845 by rolling back to the old version "Tribler starts on event mechanism".

I don't like this version, because it looks like a tricky way to solve some unknown problem.
Probably we can spend some time on "unwinding the clew of Tribler's starting logic" in the future.

### DIFF
--- a/src/tribler-gui/tribler_gui/event_request_manager.py
+++ b/src/tribler-gui/tribler_gui/event_request_manager.py
@@ -28,7 +28,7 @@ class EventRequestManager(QNetworkAccessManager):
 
     node_info_updated = pyqtSignal(object)
     received_remote_query_results = pyqtSignal(object)
-    tribler_started = pyqtSignal(object, str)  # arguments are version and public_key
+    tribler_started = pyqtSignal(object)
     upgrader_tick = pyqtSignal(str)
     upgrader_finished = pyqtSignal()
     new_version_available = pyqtSignal(str)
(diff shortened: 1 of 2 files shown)


src/tribler-gui/tribler_gui/event_request_manager.py
```diff
@@ -59,12 +59,22 @@ def __init__(self, api_port, api_key):
             NTFY.LOW_SPACE.value: self.low_storage_signal.emit,
             NTFY.REMOTE_QUERY_RESULTS.value: self.received_remote_query_results.emit,
             NTFY.TRIBLER_SHUTDOWN_STATE.value: self.tribler_shutdown_signal.emit,
-            NTFY.EVENTS_START.value: lambda _: None,
```
ichorid, line 62: Well, you don't have to rollback your key-related stuff. You only have to call `events_start_received` on `EVENTS_START`. I checked that, that was enough to solve the problem. So, it is better to keep the key thing in - we really want it.

ichorid (approved):

PR #5852, opened 2020-12-14
Convert PyQt5 to local import in tribler common

xoriole:
Tribler common is required by the core where it is best not to include PyQt dependency. I had an issue finding & resolving all required dependencies of PyQt5 while running in headless mode.

### DIFF
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -212,7 +212,11 @@ class SentryReporter:
         """
         # Prevent importing PyQt globally in tribler-common module.
         # pylint: disable=import-outside-toplevel
-        from PyQt5.QtWidgets import QApplication, QMessageBox
+        try:
+            from PyQt5.QtWidgets import QApplication, QMessageBox
+        except ImportError:
+            SentryReporter._logger.debug("PyQt5 is not available. User confirmation is not possible.")
+            return False
 
         SentryReporter._logger.debug(f"Get confirmation: {exception}")
 
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -4,8 +4,6 @@ from contextvars import ContextVar
 from enum import Enum, auto
 from hashlib import md5
 
-from PyQt5.QtWidgets import QApplication, QMessageBox
-
 from faker import Faker
 
 import sentry_sdk
@@ -212,6 +210,10 @@ class SentryReporter:
         Args:
             exception: exception to be sent.
         """
+        # Prevent importing PyQt globally in tribler-common module.
+        # pylint: disable=import-outside-toplevel
+        from PyQt5.QtWidgets import QApplication, QMessageBox
+
         SentryReporter._logger.debug(f"Get confirmation: {exception}")
 
         _ = QApplication(sys.argv)


devos50 (dismissed):

ichorid (commented):
Maybe add `try..except` here on imports?

xoriole:
@ichorid thought about using `try..except` but normally the function and the import would not be called if there is no exception so local import should be ok.

ichorid:
> @ichorid thought about using `try..except` but normally the function and the import would not be called if there is no exception so local import should be ok.

So, there is gonna be a crash on exception if local import will not work?

xoriole:
> So, there is gonna be a crash on exception if local import will not work?

Yes, it should. Note that in the enclosing function call, we're trying to trying to ask user consent to send the exception/crash report to sentry. If there is no PyQt5, we cannot (should not) send error report silently without user consent so in my opinion it is best to just crash there if that happens.

https://github.com/Tribler/tribler/blob/c43bc76cd6566681b35a17cf71ea05da354d99f6/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py#L196-L227

ichorid:
> If there is no PyQt5, we cannot (should not) send error report silently without user consent so in my opinion it is best to just crash there if that happens.

OK. The Core would gonna crash anyway, so why not?

xoriole:
Some observation based on the current state of the code and error reporting
1. If there is an unhandled exception in the core, the strategy used is `Strategy.SEND_SUPPRESSED`. That means it will not ask for confirmation, rather the event is stored and dropped. Therefore the above code is never executed.

2. Even if the strategy is set to `Strategy.SEND_ALLOWED_WITH_CONFIRMATION` where a user confirmation is expected. I still do not see import error in the logs. For example, introducing a runtime error in torrent checker produces the following logs:

```
[PID:12223] 2020-12-16 13:33:16,475 - ERROR <session:250> Session.unhandled_error_observer(): Unhandled exception occurred! Raising an error intentionally
--LONG TEXT--
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/jupiter/tudelft/release/tribler/src/pyipv8/ipv8/taskmanager.py", line 131, in done_cb
    future.result()
  File "/home/jupiter/tudelft/release/tribler/src/pyipv8/ipv8/taskmanager.py", line 18, in interval_runner
    await task(*args)
  File "/home/jupiter/tudelft/release/tribler/src/pyipv8/ipv8/util.py", line 35, in call_async
    return func(*args, **kwargs)
  File "<string>", line 2, in check_random_torrent
  File "/home/jupiter/.local/lib/python3.8/site-packages/pony/orm/core.py", line 528, in new_func
    result = func(*args, **kwargs)
  File "/home/jupiter/tudelft/release/tribler/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py", line 162, in check_random_torrent
    raise RuntimeError("Raising an error intentionally")
RuntimeError: Raising an error intentionally

--CONTEXT--
{'message': "Exception in callback TaskManager.register_task.<locals>.done_cb(<Task finishe...tentionally')>) at /home/jupiter/tudelft/release/tribler/src/pyipv8/ipv8/taskmanager.py:128", 'exception': RuntimeError('Raising an error intentionally'), 'handle': <Handle TaskManager.register_task.<locals>.done_cb(<Task finishe...tentionally')>) at /home/jupiter/tudelft/release/tribler/src/pyipv8/ipv8/taskmanager.py:128>}
[PID:12223] 2020-12-16 13:33:16,478 - INFO - SentryReporter(252) - Event from exception: Raising an error intentionally
[PID:12223] 2020-12-16 13:33:16,492 - INFO - SentryReporter(314) - Before send strategy: Strategy.SEND_ALLOWED_WITH_CONFIRMATION
[PID:12223] 2020-12-16 13:33:16,732 - INFO - TriblerTunnelCommunity(312) - Sending payout of 0 (base: 0) to Peer<81.171.8.17:35195, W8pxSwFeatcJq63LE1uXsAaxC8g=> (cid: 3912062244)
[PID:12223] 2020-12-16 13:33:16,747 - INFO - TriblerTunnelCommunity(319) - Removing DATA circuit 3912062244 timeout, 5 tries left
```
It was expected that import error would mask the actual error but not really (strange).

drew2a (approved):

PR #5854, opened 2020-12-14
Add popular torrents in torrent checker periodic loop

xoriole:
To improve the sharing of the popular torrents via popularity community, this PR adds local popular torrents in the list to be checked periodically and be eventually shared via the community. 

Note, no changes in the popularity community is needed (yet) to include popular torrents for sharing. It is because the community only makes use of `torrents_checked` set which is updated when torrent health check is processed.

This will follow another PR to store the information about who did the torrent check so the recent torrent checks data can be reused to populate the popularity community message.

### DIFF
--- a/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
+++ b/src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
@@ -168,11 +168,11 @@ class TorrentChecker(TaskManager):
         2. Old torrents (50%)
         By old torrents, we refer to those checked quite farther in the past, sorted by the last_check value.
         """
-        four_hours_ago = time.time() - HEALTH_FRESHNESS_SECONDS
-        popular_torrents = list(self.tribler_session.mds.TorrentState.select(lambda g: g.last_check < four_hours_ago).
+        last_fresh_time = time.time() - HEALTH_FRESHNESS_SECONDS
+        popular_torrents = list(self.tribler_session.mds.TorrentState.select(lambda g: g.last_check < last_fresh_time).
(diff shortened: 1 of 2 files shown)


qstokkink:
retest this please

src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
```diff
@@ -152,30 +155,42 @@ async def connect_to_tracker(self, session):
...
+        excluded from the selection considering the health information is still fresh.
+
+        1. Popular torrents (50%)
+        The indicator for popularity here is considered as the seeder count with direct proportionality
+        assuming more seeders -> more popular. There could be other indicators to be introduced later.
+
+        2. Old torrents (50%)
+        By old torrents, we refer to those checked quite farther in the past, sorted by the last_check value.
+        """
+        four_hours_ago = time.time() - HEALTH_FRESHNESS_SECONDS
```
drew2a, line 171: It is 4 hours by default, yes. But it depends on `HEALTH_FRESHNESS_SECONDS` if I understand it correctly. Therefore, maybe we need another name here instead of `four_hours_ago`

drew2a (changes requested):

drew2a:
The change of the torrent_checker looks correct and clear for me.

The tests looks a bit complicated, but it is fine.

👍

src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
xoriole: Yes, I missed that. Thank you for pointing out.

drew2a (approved):

PR #5855, opened 2020-12-15
Fixed bug in relay payouts

### LINKED ISSUE #5789: [7.6.0-exp1] Token balance bug
For some reason my token balance just shot up by like 10 GB, and later 3. I've only maybe gotten like a gigabyte and a half in 7.5.4 in a day, and this has only been running for a few hours. 
When I updated from 7.5.4, my token balance got reset. It wasn't exactly very much, but still annoying.

image

devos50:
When a relay initiates a payout to the next hop in the circuit, the values were computed incorrectly. This PR fixes the behaviour and extends the payout tests to capture such bugs.

At this point, the damage is already done but we should at least fix it to reduce the impact of this bug.

Fixes #5789


qstokkink (approved):

drew2a:
Probably, the target branch should be 'devel' as far we will not release `7.6.2`.

ichorid (approved):

devos50:
GitHub got a bit confused because of the base branch changes. Note that [all tests are actually passing](https://jenkins-ci.tribler.org/job/GH_Tribler_PR/1685/).

PR #5860, opened 2020-12-15
Fast channel discovery

### LINKED ISSUE #5828: Fast channel discovery after first startup
Goal: accelerate the genesis discovery of channels.

The experience we give to our user is that when Tribler starts for the first time your screen is just bombarded with newly discovered channels. It goes very fast, barely able to read all the information Tribler is reading of The Internet. After some seconds the most popular channels start to float to the top. Re-sorting based on popularity is done every second or so. Within 30 seconds the most popular channels on Tribler are stable and fill the page. The user is able to scroll to all these new channels.

Out of scope: fast pre-preview of when clocking on a new channel. Also out of scope: fast start of download of channel after subscribe.

By walking multiple times within IPv8 we can attempt to discover 25 new channels per second for first 30 seconds of Tribler startup. The introduction of the first peer is usually 600ms, after that we aim to discover channel entries at full speed.
https://jenkins-ci.tribler.org/job/ipv8/job/performance_peer_discovery/plot/

drew2a:
This PR resolves #5828 

### The behavior

At a startup, a user will be "bombarded with newly discovered channels" for 30 seconds.
The new channels' appearance speed: 1-3 channels per second.

The user will receive only "subscribed" channels to prevent "propagation of junk" (@ichorid)

#### Note
This code will be invoked on each run.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/discovery_booster.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/discovery_booster.py
@@ -1,0 +1,86 @@
+import logging
+
+from ipv8.peerdiscovery.discovery import EdgeWalk
+
+
+class DiscoveryBooster:
+    """This class is designed for increasing the speed of peers' discovery during a limited time.
+
+    It can be applied to any community.
+    """
+    def __init__(self, timeout_in_sec=10.0, max_peers=200, take_step_interval_in_sec=0.05, walker=None):
+        """
+
+        Args:
+            timeout_in_sec: DiscoveryBooster work timeout. When this timeout will be reached,
+                `finish` function will be called.
+            max_peers: Temporary value of max peers that will be set to the community during DiscoveryBooster work.
+            take_step_interval_in_sec: Сall frequency of walker's `take_step` function.
+            walker: walker that will be used during boost period.
+        """
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+        self.timeout_in_sec = timeout_in_sec
+        self.max_peers = max_peers
+        self.take_step_interval_in_sec = take_step_interval_in_sec
+        self.walker = walker
+
+        self.community = None
+        self.saved_max_peers = None
+
+        self._take_step_task_name = 'take step'
+
+    def apply(self, community):
+        """Apply DiscoveryBooster to the community
+
+        Args:
+            community: community to implement DiscoveryBooster
+
+        Returns: None
+        """
+        if not community:
+            return
+
+        self.logger.info(
+            f'Apply. Timeout: {self.timeout_in_sec}s, '
+            f'Max peers: {self.max_peers}, '
+            f'Take step interval: {self.take_step_interval_in_sec}s'
+        )
+
+        self.community = community
+        self.saved_max_peers = community.max_peers
+
+        community.max_peers = self.max_peers
+
+        if not self.walker:
+            # values for neighborhood_size and edge_length were found empirically to
+            # maximize peer count at the end of a 30 seconds period
+            self.walker = EdgeWalk(community, neighborhood_size=25, edge_length=25)
+
+        community.register_task(self._take_step_task_name, self.take_step, interval=self.take_step_interval_in_sec)
+        community.register_task('finish', self.finish, delay=self.timeout_in_sec)
+
+    def finish(self):
+        """Finish DiscoveryBooster work.
+
+        This function returns defaults max_peers to the community.
+
+        Will be called automatically from community's task manager.
+
+        Returns: None
+        """
+        self.logger.info(
+            f'Finish. Set self.max_peers={self.saved_max_peers}. Cancel pending task: {self._take_step_task_name}'
+        )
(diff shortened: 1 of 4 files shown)


qstokkink:
Your PR is making our unit tests connect to the wild Internet: please don't do that (you're calling `self.walker.take_step()` manually, which is causing this).

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
```diff
@@ -131,13 +131,18 @@ def notify_gui(_, processing_results):
...
+        self.components = kwargs.pop('components', None)
         super().__init__(*args, **kwargs)
 
         # Register legacy payload
         self.add_message_handler(LegacySelectResponsePayload, self.legacy_on_remote_select_response)
-
         self.new_style_peers = set()
 
+        # apply components
+        if self.components:
```
ichorid, line 142: It is better to move it into `NonLegacyGigaChannelCommunity`. Otherwise, we may accidentally delete it when removing the legacy stuff.

src/tribler-core/tribler_core/modules/metadata_store/community/discovery_booster.py
```diff
@@ -0,0 +1,52 @@
+import logging
+
+from ipv8.peerdiscovery.discovery import EdgeWalk
+
+
+class DiscoveryBooster:
```
ichorid, line 6: :+1: for generic solution!

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
drew2a: Yes, sure, thank you :)

src/tribler-core/tribler_core/modules/ipv8_module_catalog.py
```diff
@@ -185,7 +185,9 @@ class PopularityCommunityLauncher(IPv8CommunityLauncher):
 @walk_strategy(random_walk, target_peers=30)
 @walk_strategy(remove_peers, target_peers=INFINITE)
 class GigaChannelCommunityLauncher(IPv8CommunityLauncher):
-    pass
+    def get_kwargs(self, session):
+        from tribler_core.modules.metadata_store.community.discovery_booster import DiscoveryBooster
+        return {'components': [DiscoveryBooster()]}
```
ichorid, line 190: This way of running subcomponents is not standard to IPv8. If there would be more than a single subsystem running like this, this would be the correct way to do it (however, it would be even more correct to add the "components" thing into IPv8 `Community` class). 

But you're running only a single component like this. Inventing a separate subsystem to run just a single component is a form of overengineering. Of course, this sends the other programmers a signal: "do it like this next time". However, that's not the point of this PR. Putting the thing here evoked a "WTF?" moment from me. So, please just put the thing into the GigaChannelCommunity `_init_`.

drew2a: @ichorid proposed approach will break our test system because I calling 

```python
    def take_step(self):
        self.logger.debug('Take a step')
        self.walker.take_step()
```
https://github.com/drew2a/tribler/blob/e54e3f080dae3278e9ca29695774959fef3378ad/src/tribler-core/tribler_core/modules/metadata_store/community/discovery_booster.py#L52

ichorid: Why not `patch` it then?

drew2a: What do you mean?

ichorid: Like this:
https://github.com/drew2a/tribler/compare/feature/5828...ichorid:feature/5828

drew2a: 👍 thank you

src/tribler-core/tribler_core/modules/metadata_store/community/discovery_booster.py
```diff
@@ -0,0 +1,52 @@
...
+
+        community.max_peers = self.max_peers
+
+        # values for neighborhood_size and edge_length were found empirically to
+        # maximize peer count at the end of a 30 seconds period
+        self.walker = EdgeWalk(community, neighborhood_size=25, edge_length=25)
+
+        community.register_task(self._take_step_task_name, self.take_step, interval=self.take_step_interval_in_sec)
+
+        community.register_anonymous_task('', self.finish, delay=self.timeout_in_sec)
```
qstokkink, line 41: It's good practice to give your anonymous tasks a basename that is not the empty string (in your case perhaps "DiscoveryBooster timeout").

This makes any potentially failing tasks be logged as "\<basename> \<random number>" instead of " \<random number>", allowing for easier debugging.

drew2a: 👌 thank you for the explanation.

ichorid (approved):
:+1:

Issue #5858, opened 2020-12-15
Trustgraph crashes on node selection


xoriole:
Trustgraph crashes on clicking on any node. Seems like the latest changes in connecting pyqtsignal mechanism has some side effects.

```
[PID:1017469] 2020-12-15 13:33:32,223 - ERROR <tribler_window:124> root.on_exception(): Traceback (most recent call last):
  File "/home/jupiter/tudelft/release/tribler/src/run_tribler.py", line 150, in <module>
    window = TriblerWindow()
  File "/home/jupiter/tudelft/release/tribler/src/tribler-gui/tribler_gui/tribler_window.py", line 240, in __init__
    self.trust_graph_page.initialize_trust_graph()
  File "/home/jupiter/tudelft/release/tribler/src/tribler-gui/tribler_gui/widgets/trustgraphpage.py", line 117, in initialize_trust_graph
    self.trust_graph.set_node_selection_listener(self.on_node_clicked)
  File "/home/jupiter/tudelft/release/tribler/src/tribler-gui/tribler_gui/widgets/trustgraphpage.py", line 38, in set_node_selection_listener
    connect(self.scatter.sigClicked, listener)
tribler_gui.utilities.CreationTraceback

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/jupiter/tudelft/release/tribler/src/tribler-gui/tribler_gui/utilities.py", line 385, in trackback_wrapper
    raise exc from source_exc
  File "/home/jupiter/tudelft/release/tribler/src/tribler-gui/tribler_gui/utilities.py", line 382, in trackback_wrapper
    callback(*args, **kwargs)
TypeError: on_node_clicked() takes 2 positional arguments but 3 were given
```

PR #5863, opened 2020-12-16
Fix trust graph node click and max node issue

xoriole:
- It fixes the crash on clicking a node in the Trust graph. The issue was because of reorganization of the arguments in pyqtsignal. The list of items where user clicked was splited to a single top point and a list of rest points. (Qt signal is magic).
Fixes https://github.com/Tribler/tribler/issues/5858

- It fixes the crash occurring when there are nodes more than the max value. There was a test for that but the failure was suppressed as failure. The removal inheritance of unnecessary `TaskManager` fixes that apparently. Additionally, added separate tests for max nodes and max transactions.

- The query parameter was removed in the last refactor. It is fixed now. A test is added as well. The refresh from the GUI should work nicely now.

- Increased the size of nodes slightly.

**Screenshot**
trust-graph

### DIFF
--- a/src/tribler-core/tribler_core/restapi/tests/test_trustview_endpoint.py
+++ b/src/tribler-core/tribler_core/restapi/tests/test_trustview_endpoint.py
@@ -222,3 +222,42 @@ async def test_trustview_max_response(enable_api, mock_ipv8, session, mock_boots
     assert response_json['graph']
     assert response_json['num_tx'] <= max_tx
     assert len(response_json['graph']['node']) <= max_peers
+
+
+@pytest.mark.asyncio
+async def test_trustview_with_refresh(enable_api, mock_ipv8, session, mock_bootstrap):
+    """
+    Test whether refresh query parameters works as expected.
+    If refresh parameter is not set, the cached graph is returned otherwise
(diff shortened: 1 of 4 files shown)


xoriole:
retest this please

src/tribler-core/tribler_core/restapi/tests/test_trustview_endpoint.py
```diff
@@ -199,26 +199,91 @@ def verify_response(response_json):
     verify_response(response)
 
 
+def insert_node_transactions(root_key, session, node_public_key=None, count=1):
+    for idx in range(count):
+        counterparty = unhexlify(node_public_key if node_public_key else get_random_node_public_key())
+        # seq_num = random.randint(1, 100)
```
drew2a, line 205: nit: this code comment can be safely removed :)

drew2a (changes requested):

xoriole:
GUI test on macOS is failing because of https://github.com/Tribler/tribler/issues/5864 so keeping this PR on hold until the blocking issue is fixed.

drew2a (approved):

PR #5866, opened 2020-12-18
Fix database upgrader & tests

kozlovsky:
There were some bugs in the database upgrade process.
1. When it is necessary to perform a series of updates (versionA -> versionB -> versionC), first steps of this process may fail, as Python classes used on step versionA -> versionB are actually from the current version (versionC) of application code. It leads to two types of errors:
- Automatic check of columns may fail, as it currently checks for the actual version of code (versionC), whereas the database may be on version B instead.
- some selects may fail, as ORM expects all columns specified in a class definition are presented in the database.
2. When it is necessary to perform a series of updates (versionA -> versionB -> versionC), first step (versionA -> versionB) after completion updated database version number to current version (versionC) instead of versionB, which lead to skipped upgrade versionB -> versonC.

This PR fixes above mentioned problems:
- database structure checks are postponed to the moment when database structure is corresponding to Python classes
- each upgrade function from version A to version B specifies hardcoded version B as the resulted database version instead of the current version specified in the application code.
- some queries which may be affected by the latest changes specified using SQL, to not be dependent on Python classes.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/store.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/store.py
@@ -85,7 +85,9 @@ sql_add_fts_trigger_update = """
 
 
 class MetadataStore(object):
-    def __init__(self, db_filename, channels_dir, my_key, disable_sync=False, notifier=None):
+    def __init__(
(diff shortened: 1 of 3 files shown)


xoriole:
retest this please

xoriole (approved):

PR #5869, opened 2020-12-18
Bootstrap torrents checked from database

xoriole:
Changes:
- Previously checked torrents are used to bootstrap the `torrent_checked` set which is used by popularity community to compose health message.
- A `self_checked` column is added to `TorrentState` table which indicates whether the health information of a torrent is actually checked by the user. The default is False.
- A working upgrader for the database is also included, thanks to @kozlovsky


src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
```diff
@@ -225,7 +246,7 @@ def update_torrents_checked(self, new_result):
         new_result_tuple = (new_result['infohash'], new_result['seeders'],
                             new_result['leechers'], new_result['last_check'])
         if new_result['seeders'] > 0:
-            self.torrents_checked.add(new_result_tuple)
+            self._torrents_checked.add(new_result_tuple)
```
kozlovsky, line 249: self._torrents_checked should probably be dict infohash -> (seeders, leechers, last_check) instead of a set, otherwise, multiple calls of update_torrents_checked can potentially create multiple records with the same infohash value and different seeders/leechers count

xoriole: Good point

xoriole:
retest this please

src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/torrent_state.py
```diff
@@ -14,6 +14,7 @@ class TorrentState(db.Entity):
         seeders = orm.Optional(int, default=0)
         leechers = orm.Optional(int, default=0)
         last_check = orm.Optional(int, size=64, default=0)
+        self_checked = orm.Optional(bool, default=False)
```
kozlovsky, line 17: Currently, PonyORM applies defaults in Python and does not add `default` clauses to columns in the database automatically (because in PonyORM default may be arbitrary Python function).
It is better to add a second option `sql_default="0"` to this attribute. Otherwise, upgrade from 7.5 to 7.6 will put NULL instead of 0 to this column for previously existed records.
It is not critical for the current logic that processes this attribute, but this way the data will be more consistent.

xoriole: Good to know, thanks

kozlovsky (dismissed):

src/tribler-core/tribler_core/modules/torrent_checker/test_torrentchecker.py
```diff
@@ -15,15 +17,15 @@
 from tribler_core.utilities.unicode import hexlify
 
 
-@pytest.fixture
-async def torrent_checker(session):
+@pytest.fixture(name="torrent_checker")
+async def fixture_torrent_checker(session):
```
drew2a, line 21: Pylint....

src/tribler-core/tribler_core/modules/torrent_checker/test_torrentchecker.py
```diff
@@ -81,6 +83,55 @@ async def test_health_check_cached(enable_chant, torrent_checker, session):
     assert result['db']['leechers'] == 10
 
 
+@pytest.mark.asyncio
+async def test_load_torrents_check_from_db(enable_chant, torrent_checker, session):  # pylint: disable=unused-argument
```
drew2a, line 87: 👍 the test is really illustrative and clear.

src/tribler-core/tribler_core/modules/torrent_checker/torrent_checker.py
```diff
@@ -225,7 +244,7 @@ def update_torrents_checked(self, new_result):
         new_result_tuple = (new_result['infohash'], new_result['seeders'],
                             new_result['leechers'], new_result['last_check'])
         if new_result['seeders'] > 0:
-            self.torrents_checked.add(new_result_tuple)
+            self._torrents_checked[new_result['infohash']] = (new_result_tuple)
```
drew2a, line 247:
```suggestion
            self._torrents_checked[new_result['infohash']] = new_result_tuple
```

Probably, parentheses are redundant here.

drew2a: NIT
`new_result['infohash']` are used twice. So, we can extract it to
```
infohash = new_result['infohash']
```

drew2a (changes requested):

drew2a (dismissed):

xoriole:
retest this please

xoriole:
retest this please

kozlovsky (approved):

PR #5870, opened 2020-12-18
Add an option to automatically send exceptions

### LINKED ISSUE #5799: [devel] Core is importing from PyQt5
Minor issue, but I found that the core is importing Qt stuff through the packages in `tribler-common`:

```
  File "/Users/martijndevos/Documents/tribler/src/tribler-core/run_bandwidth_crawler.py", line 20, in <module>
    from tribler_core.session import Session
  File "/Users/martijndevos/Documents/tribler/src/tribler-core/tribler_core/session.py", line 23, in <module>
    from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
  File "/Users/martijndevos/Documents/tribler/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py", line 7, in <module>
    from PyQt5.QtWidgets import QApplication, QMessageBox
```

When running the core, I would like to avoid importing from `PyQt5` as much as possible.

drew2a:
Partially resolves #5799 by adding an option to send unhandled exceptions automatically.

To enable or disable this option you need to change "tribler.conf":
```
[error_handling]
error_reporting_requires_user_consent=boolean(default=True)
```

It is turn on or turn off mechanism for sending unhandled exceptions **only**. 
Ordinary crash reports will not be affected.

This change has entailed SentryReporter refactoring:
* `this_sentry_strategy` context manager has been extracted 
* ContextVar's strategy has been extended by global strategy

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -5,7 +5,7 @@ import signal
 import sys
 from asyncio import ensure_future, get_event_loop
 
-from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
+from tribler_common.sentry_reporter.sentry_reporter import SentryReporter, SentryStrategy
 from tribler_common.sentry_reporter.sentry_scrubber import SentryScrubber
 
 import tribler_core
@@ -64,6 +64,10 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_m
         state_dir = get_versioned_state_directory(root_state_dir)
 
         config = TriblerConfig(state_dir, config_file=state_dir / CONFIG_FILENAME)
+
+        if not config.get_error_reporting_requires_user_consent():
+            SentryReporter.global_strategy = SentryStrategy.SEND_ALLOWED
+
         config.set_api_http_port(int(api_port))
         # If the API key is set to an empty string, it will remain disabled
         if config.get_api_key() not in ('', api_key):
@@ -90,7 +94,7 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_m
 
 if __name__ == "__main__":
     SentryReporter.init(sentry_url=sentry_url, release_version=version_id, scrubber=SentryScrubber(),
-                        strategy=SentryReporter.Strategy.SEND_ALLOWED_WITH_CONFIRMATION)
+                        strategy=SentryStrategy.SEND_ALLOWED_WITH_CONFIRMATION)
     # Get root state directory (e.g. from environment variable or from system default)
     root_state_dir = get_root_state_directory()
     # Check whether we need to start the core or the user interface
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -1,5 +1,6 @@
 import logging
 import sys
+from contextlib import contextmanager
 from contextvars import ContextVar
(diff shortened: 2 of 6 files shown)


src/run_tribler.py
```diff
@@ -64,6 +64,10 @@ async def start_tribler():
         state_dir = get_versioned_state_directory(root_state_dir)
 
         config = TriblerConfig(state_dir, config_file=state_dir / CONFIG_FILENAME)
+
+        if config.get_automatically_send_unhandled_exceptions():
+            SentryReporter.strategy = SentryStrategy.SEND_ALLOWED
```
drew2a, line 69: This is the main change for the PR.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -33,39 +33,60 @@
...
-    last_event = None
+    def __enter__(self):
+        SentryReporter.lock.acquire()
+        self.saved_strategy = SentryReporter.strategy
+        SentryReporter.strategy = self.value
 
-    ignored_exceptions = [KeyboardInterrupt, SystemExit]
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        SentryReporter.strategy = self.saved_strategy
+        SentryReporter.lock.release()
```
drew2a, line 71: @kozlovsky please, confirm that lock is used properly.

src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -33,39 +33,58 @@
...
+    3. Send reports are suppressed (but the last event will be stored)
+    """
+
+    SEND_ALLOWED = auto()
+    SEND_ALLOWED_WITH_CONFIRMATION = auto()
+    SEND_SUPPRESSED = auto()
+
+
+class ThisSentryStrategy:
+    """Class simplify execution of a particular block of code with given SentryStrategy
```
kozlovsky, line 51: You can write this context manager as a generator, this way it is a bit easier to understand how it handles exceptions:

```python
from contextlib import contextmanager


@contextmanager
def this_sentry_strategy(strategy: SentryStrategy):
    saved_strategy = SentryReporter._context_strategy.get()
    try:
        SentryReporter._context_strategy.set(strategy)
        yield
    finally:
        SentryReporter._context_strategy.set(saved_strategy)
```

drew2a: Wow, cool, thank you!

kozlovsky (dismissed):

src/tribler-core/tribler_core/config/tribler_config.spec
```diff
@@ -106,3 +106,6 @@ history_size = integer(min=1, default=20)
 [popularity_community]
 enabled = boolean(default=True)
 cache_dir = string(default=health_cache)
+
+[error_handling]
+automatically_send_unhandled_exceptions=boolean(default=True)
```
xoriole, line 111: - Suggestion: it is basically representing user consent to send the reports so maybe could be renamed as `error_reporting_consent` or `user_consent`.
- I think the default should be false

xoriole (commented):
Maybe a small checkbox in GUI settings could be a nice addition

xoriole (approved):

drew2a:
retest this please

Issue #5857, opened 2020-12-15
Setup test nodes to run Tribler 24/7


xoriole:
Setup test nodes to run Tribler 24/7

Why?
- To find out if there are issues on normal Tribler network.
- To find out if there are crashes on the release candidate or past stable versions and report them.

Where to setup?
- Proxmox or Zulu

Tribler version:
- last stable (eg. 7.6.1)
- new candidate (eg. 7.7/ devel)

Requirements:
- Sentry should be configured to report crashes
- Alert developers if a crash is found

drew2a:
There is one more improvement to Sentry, that should be done:
1. We should add a value to the config to allow sending crash reports without confirmations.

xoriole:
Update: Jenkins jobs running the old and newer versions are setup here:
1. https://jenkins-ci.tribler.org/job/Deployment-Tests/job/long-runner-new-release/
2. https://jenkins-ci.tribler.org/job/Deployment-Tests/job/long-runner-old-release/

The newer version installer is obtained via jenkins build and the last stable installer is obtained via Github release URL.

xoriole:
With the above jenkins jobs setup and running, I consider this issue as resolved.

PR #5873, opened 2020-12-21
Apply `error_reporting_requires_user_consent` to all core's errors

### LINKED ISSUE #5799: [devel] Core is importing from PyQt5

drew2a:
This PR partially resolves #5799, #5857 by extending https://github.com/Tribler/tribler/pull/5870

Now, setting `error_reporting_requires_user_consent` covers all "core" errors. It will not be used by a regular Tribler user and will set to 'True' by default:

```
[error_handling]
error_reporting_requires_user_consent=boolean(default=True)
```

Note: this value does not affect errors that occurred in GUI.

### DIFF
--- a/src/tribler-core/tribler_core/restapi/events_endpoint.py
+++ b/src/tribler-core/tribler_core/restapi/events_endpoint.py
@@ -115,11 +115,12 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
 
     # An exception has occurred in Tribler. The event includes a readable
     # string of the error and a Sentry event.
-    def on_tribler_exception(self, exception_text, sentry_event):
+    def on_tribler_exception(self, exception_text, sentry_event, error_reporting_requires_user_consent):
         self.write_data({
             "type": NTFY.TRIBLER_EXCEPTION.value,
             "event": {"text": exception_text},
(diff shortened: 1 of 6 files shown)


kozlovsky (dismissed):

xoriole (commented):
When require user consent is False, does the Feedback dialog appear briefly and close like a flicker?

drew2a:
@xoriole yes

xoriole (approved):

PR #5874, opened 2020-12-22
Add an option to run Tribler with sentry URL for tests

drew2a:
This PR is referred to #5857 

It adds an option to run a Tribler client in test mode and send all the crash reports automatically.

To enable it you have to specify TEST_SENTRY_URL:

```bash
export TEST_SENTRY_URL=<some_url>
```

And enable crash reports auto-send in "triblerd.conf":

```
[error_handling]
error_reporting_requires_user_consent=False
```

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -92,9 +92,30 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, core_test_m
     get_event_loop().run_forever()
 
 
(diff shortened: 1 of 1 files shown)


xoriole (commented):
Might be better to enclose it into a function, the main would look cleaner

xoriole (approved):

kozlovsky (approved):

PR #5876, opened 2020-12-22
Fix #5867

### LINKED ISSUE #5867: [devel] Request error on clicking breadcrumb of the search results page
devel

- Search for 'ubuntu'
- Results appear
- Click on breadcrumb 'search results for Ubuntu'

- Nothing happens on clicking the base breadcrumb, better even click is not enabled.

##### Actual behavior: #####
- - Request error dialog appears

##### Screenshot #####
breadcrumb-clicked

drew2a:
This PR fixes #5867 by removing `self.reset_view()` calling after clicking on a "search result breadcrumb".

This call was the cause of sending `TriblerNetworkRequest` with unexpected arguments.

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
+++ b/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
@@ -268,8 +268,6 @@ class ChannelContentsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class)
     def on_breadcrumb_clicked(self, tgt_level):
         if int(tgt_level) + 1 != len(self.channels_stack):
             self.go_back_to_level(tgt_level)
-        else:
-            self.reset_view()
 
     def go_back_to_level(self, level):
         level = int(level)


xoriole (commented):
This should fix the mentioned issue but I'm not sure about the impact of removing the else clause on other pages. ChannelContentsWidget is widely used (search_results_page, channel_contents_page, discovered_page). Maybe @ichorid can comment.

ichorid (changes requested):
@xoriole is right.

Indeed, resetting the view on clicking the top/last breadcrumb is an important feature that allows to e.g. reset the list contents in case the user is lost in the list, or just wants to get the default sorting options.

The proper solution to this problem is to either fill the `filter` property from the "Search" lineEdit, or overload the `reset_view` for the `SearchModel` (IIRC).

drew2a:
@ichorid @xoriole you can check the PR. The logic to return to the previous page is kept. Only the ability to click on the rightmost (the last breadcrumb) item is removed.

Why we should reset the view on click at the rightmost item? It is counterintuitive. 

Anyway, @ichorid maybe it is better to reassign the issue to you, because it is your code, and you can provide a more correct solution.

ichorid:
> Why should we reset the view on click at the rightmost item? It is counterintuitive.

Because we don't have a "Refresh" :recycle:  button.

You're right, I'll better fix it myself.

drew2a:
@ichorid reset and refresh are not the same things.

ichorid:
Yep. And we're not a browser. That's why we don't have a separate button to reset the page contents - because it is useless most of the time. But it is still sometimes necessary to reset the page, especially during development :wink: 

If you have a better idea of where to put it - welcome!

drew2a:
Not everything that is good for development is good for the user.

We can (and should) add some tools that will make our lives easier as developers. But these tools should not affect our users.

If you want to refresh the page it is ok to add this logic to the rightmost item in breadcrumbs ("current page" in the example below).

```
root page / child page / current page
```

If you want to reset the page it is ok to add this logic to the leftmost item in breadcrumbs (root page in the example above)

PR #5879, opened 2020-12-23
Fix/5867

### LINKED ISSUE #5867: [devel] Request error on clicking breadcrumb of the search results page

ichorid:
Fix #5867 

Also, enable querying for non-torrent preview entries in channels.

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
+++ b/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
@@ -225,8 +225,8 @@ class ChannelContentsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class)
         self.update_labels()
         self.channel_torrents_filter_input.setText("")
 
-    def reset_view(self):
-        self.model.text_filter = ''
+    def reset_view(self, text_filter=None):
+        self.model.text_filter = text_filter or ''
         self.model.category_filter = None
 
         with self.freeze_controls():
@@ -270,7 +270,13 @@ class ChannelContentsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class)
     def on_breadcrumb_clicked(self, tgt_level):
         if int(tgt_level) + 1 != len(self.channels_stack):
             self.go_back_to_level(tgt_level)
+        elif isinstance(self.model, SearchResultsModel) and len(self.channels_stack) == 1:
+            # In case of remote search, when only the search results are on the stack,
+            # we must keep the txt_filter (which contains the search term) before resetting the view
+            text_filter = self.model.text_filter
+            self.reset_view(text_filter=text_filter)
         else:
+            # Reset the view if the user clicks on the last part of the breadcrumb
             self.reset_view()
 
     def go_back_to_level(self, level):
--- a/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
+++ b/src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
@@ -176,8 +176,9 @@ class ChannelContentsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class)
         self.channels_stack.append(model)
         connect(self.model.info_changed, self.on_model_info_changed)
 
-        connect(self.window().core_manager.events_manager.received_remote_query_results,
-                self.model.on_new_entry_received)
+        connect(
+            self.window().core_manager.events_manager.received_remote_query_results, self.model.on_new_entry_received
+        )
         connect(self.window().core_manager.events_manager.node_info_updated, self.model.update_node_info)
 
         with self.freeze_controls():
@@ -239,8 +240,9 @@ class ChannelContentsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class)
     def disconnect_current_model(self):
         self.model.info_changed.disconnect()
         disconnect(self.window().core_manager.events_manager.node_info_updated, self.model.update_node_info)
-        disconnect(self.window().core_manager.events_manager.received_remote_query_results,
-                   self.model.on_new_entry_received)
+        disconnect(
+            self.window().core_manager.events_manager.received_remote_query_results, self.model.on_new_entry_received
+        )
         self.controller.unset_model()  # Disconnect the selectionChanged signal
 
     def go_back(self, checked=False):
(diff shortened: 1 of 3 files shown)


ichorid:
retest this please

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
```diff
@@ -268,7 +270,13 @@ def go_back(self, checked=False):
     def on_breadcrumb_clicked(self, tgt_level):
         if int(tgt_level) + 1 != len(self.channels_stack):
             self.go_back_to_level(tgt_level)
+        elif isinstance(self.model, SearchResultsModel) and len(self.channels_stack) == 1:
+            # In case of remote search, when only the search results are on the stack,
```
drew2a, line 274: **praise:** you add an explanation to the code blocks, it is great!

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
```diff
@@ -268,7 +270,13 @@ def go_back(self, checked=False):
...
+            # we must keep the txt_filter (which contains the search term) before resetting the view
+            text_filter = self.model.text_filter
+            self.reset_view(text_filter=text_filter)
         else:
+            # Reset the view if the user clicks on the last part of the breadcrumb
```
drew2a, line 279: **issue:** I still disagree with resetting the view by clicking on the rightmost breadcrumb 
More information: https://github.com/Tribler/tribler/pull/5876#issuecomment-750223950
**suggestion:** remove reset logic from click on rightmost breadcrumb or implement the more straightforward general solution.

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
```diff
@@ -268,7 +270,13 @@ def go_back(self, checked=False):
     def on_breadcrumb_clicked(self, tgt_level):
         if int(tgt_level) + 1 != len(self.channels_stack):
             self.go_back_to_level(tgt_level)
+        elif isinstance(self.model, SearchResultsModel) and len(self.channels_stack) == 1:
+            # In case of remote search, when only the search results are on the stack,
+            # we must keep the txt_filter (which contains the search term) before resetting the view
+            text_filter = self.model.text_filter
+            self.reset_view(text_filter=text_filter)
```
drew2a, line 277: **issue:** I could be wrong, but it seems that the base class (ChannelContentsWidget) adopted his behavior for 3 different views (search, discovery, channel content). It is unclear, which parts of the class are used by which views.
For those who are not the author of the class, It is hard to understand the logic and hard to make proper changes to the class.

**suggestion:** use different implementations of ChannelContentsWidget to different views (search, discovery, channel content)

drew2a (changes requested):
This PR solves the issue, but it decreases overall code health, that's why changes are requested.

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
ichorid: > use different implementations of ChannelContentsWidget to different views (search, discovery, channel content)

pre-7.5 versions used this approach. This resulted in *massive* code duplication because Qt does not allow for inheritance in `.ui`-defined classes. So, we were left with two options (besides the copy-paste):
 * stop using QtDesigner for Channels stuff and lose the ability to see and manipulate the layout visually
 * create a single uber-class and disable its features as necessary

The current state of affairs it the result of choosing the second option...

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
ichorid: > suggestion: remove reset logic from click on rightmost breadcrumb or implement the more straightforward general solution.

IMHO, adding a special "Reset" button to the GUI - that would be overkill. The fewer buttons - the better. Also, the current solution is general enough (but, sadly, not too straightforward.)

drew2a: You can add reset logic to the leftmost breadcrumb.

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
drew2a: Can you give me a link to this "QT feature"?

drew2a: > create a single uber-class and disable its features as necessary

If it is the only possible solution (if we assume that QtDesigner is mandatory), then probably, we should set boundaries inside this class with more accuracy.

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
ichorid: If something was pointed out by a colleague as unobvious, that should be commented :wink:

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
ichorid: >we should set boundaries inside this class with more accuracy.

Indeed, we should. ChannelConentsWidget was never *designed*, but instead born as a merge of several classes. I'm not going to refactor it until we plan some big new features for it. It works fine for now.

src/tribler-gui/tribler_gui/widgets/channelcontentswidget.py
ichorid: The leftmost breadcrumb is for navigation to the root of the channels stack...

drew2a: > The leftmost breadcrumb is for navigation to the root of the channels stack...

From my understanding, it is the same as "reset". Am I wrong?

ichorid: Clicking on each part of the path in the breadcrumb will get you to the corresponding part of the channels stack.

When you're looking at the contents of folder `D` the breadcrumb looks like this:
`A / B / C / D`

Clickin on "B" will pop `D` and `C` from the stack, gettin you to look at the contents of `B`. That is straightforward and intuitive behaviour. Clicking on `A` will show you the contents of `A` (A could be either a search request root, a channel or a list of discovered channels). Clicking on `D` (the name of the current folder in the breadcrumb) will reset the contents of `D`.

PR #5880, opened 2020-12-23
Fix sorting search result by health

kozlovsky:
This PR fixes a problem when folders are not displayed in the channel content view after sorting by "health".

The essence of the fix is in replacing INNER JOIN to LEFT JOIN in SQL query, so ChannelNode objects without TorrentState corresponding objects are still returned by the query. 

Also, in this PR I add default ordering by `rowid` to all queries which return ChannelNode objects, to have guaranteed stable sorting order in tests.

Some tests are marked with FIXME as they are not returning totally correct results. Namely, full-text search queries don't return more than one folder. This is a separate bug in full-text search queries and will be fixed in a separate PR.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/tests/test_channel_metadata.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/tests/test_channel_metadata.py
@@ -452,50 +452,53 @@ def test_get_channels(metadata_store):
 
 @db_session
 def test_sort_by_health(metadata_store):
+    def save():
+        metadata_store._db.flush()  # pylint: disable=W0212
+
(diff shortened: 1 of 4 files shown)


kozlovsky:
retest this please

src/tribler-core/tribler_core/modules/metadata_store/tests/test_channel_metadata.py
```diff
@@ -450,6 +450,132 @@ def test_get_channels(metadata_store):
     assert len(channels) == 5
 
 
+@db_session
```
ichorid, line 453: :+1: Wow! Quite a test :clap:

ichorid (approved):

PR #5897, opened 2021-01-04
Add ability to send exceptions automatically in case `TEST_SENTRY_URL` is defined

drew2a:
This PR adds an ability to send handled exceptions to the reporter automatically in case `TEST_SENTRY_URL` is defined.

Requested by @xoriole for using in 24/7 Tribler instance tester.

### DIFF
--- a/src/tribler-gui/tribler_gui/dialogs/feedbackdialog.py
+++ b/src/tribler-gui/tribler_gui/dialogs/feedbackdialog.py
@@ -94,9 +94,14 @@ class FeedbackDialog(AddBreadcrumbOnShowMixin, QDialog):
         # Users can remove specific lines in the report
         connect(self.env_variables_list.customContextMenuRequested, self.on_right_click_item)
 
-        self.error_reporting_requires_user_consent = error_reporting_requires_user_consent
-        if not error_reporting_requires_user_consent:
+        self.send_automatically = FeedbackDialog.can_send_automatically(error_reporting_requires_user_consent)
+        if self.send_automatically:
             self.on_send_clicked(True)
+
+    @staticmethod
+    def can_send_automatically(error_reporting_requires_user_consent):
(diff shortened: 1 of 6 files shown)


src/tribler-gui/tribler_gui/dialogs/feedbackdialog.py
```diff
@@ -94,10 +94,15 @@ def add_item_to_info_widget(key, value):
...
-        self.error_reporting_requires_user_consent = error_reporting_requires_user_consent
-        if not error_reporting_requires_user_consent:
+        self.send_automatically = FeedbackDialog.can_send_automatically(error_reporting_requires_user_consent)
+        if self.send_automatically:
             self.on_send_clicked(True)
 
+    @staticmethod
+    def can_send_automatically(error_reporting_requires_user_consent):
+        test_sentry_url_is_defined = os.environ.get('TEST_SENTRY_URL', None) is not None
+        return test_sentry_url_is_defined or not error_reporting_requires_user_consent
```
xoriole, line 104: On second thought, maybe `and` instead of `or`.

drew2a: It looks safer for a user (errors will not be sent in case of defined-by-mistake `TEST_SENTRY_URL`) but adds more complexity for the understanding of `TEST_SENTRY_URL` and  `core_error_reporting_requires_user_consent "collaborative work". Am I right?

BTW I'm ok with `and`.

xoriole: Yes, I'd go for safer option.

drew2a: Done

xoriole (approved):

Issue #4930, opened 2019-11-05
Debugging user issues better


xoriole:
Some bugs on Tribler are difficult to fix because they are difficult to reproduce. Knowing the internal state of Tribler during the faulty execution would be immensely useful for debugging such issue. 
Some directions discussed at today's meeting for that:
- Full state recording, extreme logging in Tribler/python. User can then upload the logs along with the error report. 
- Debug option in PyInstaller build

Seems there is a related old ticket [#2910](https://github.com/Tribler/tribler/issues/2910)

qstokkink:
We already have the `--debug` flag. What do you propose to extend this with?

ichorid:
The problem is, if you need to turn on the "debugging mode" to capture the complete Tribler state, you'll have to know *in advance* that Tribler is going to crash. This will only help the most advanced and determined user.

An alternative route is to do it the way all major OSes do it: in case of crash dump the whole process memory state on disk, *replacing the previous dump*. Then, the reported can have a checkbox to attach the dump. Note, though, that this kind of stuff will invariably leak private user info... Even if we try to filter the dump for things like, e.g. private key, before sending.

xoriole:
@qstokkink Yes, we already have the flag but it may not be sufficient to reproduce the issues. Besides, we don't get the user logs in the error report. So, I think we should extend the error reporter to add a choice (disabled by default) whether to upload the logs along with the report.
@ichorid crash dump and attaching to the error report is a good idea and I'm in favor of it. Regarding privacy, it is a user consent to attach the files. Also, we could try some basic anonymisation on data.

qstokkink:
Two notes on crash dumps:
 - They are big (our server is already filling up) and can take in the order of minutes to properly capture.
 - Someone has to plow through them to find out what was wrong (which might be OS specific).

Regarding logging:
 - Three years ago I performed a clean performance experiment on Dispersy throughput. By stripping out all logging I gained 400 MB/s in download speed. Be careful with overzealous logging, especially in Python - which is basically single thread execution.

ichorid:
relate to _the plan_ #5348

PR #5910, opened 2021-01-07
Add check on infohash key existing

### LINKED ISSUE #5909: [7.7.0] KeyError: 'infohash' (in triblertablecontrollers.py:135)
```
  File "src\run_tribler.py", line 179, in <module>, 
  File "src\tribler-gui\tribler_gui\utilities.py", line 382, in trackback_wrapper, 
  File "src\tribler-gui\tribler_gui\event_request_manager.py", line 116, in on_read_data, 
  File "src\tribler-gui\tribler_gui\event_request_manager.py", line 77, in tribler_started_event, 
  File "src\tribler-gui\tribler_gui\utilities.py", line 382, in trackback_wrapper, 
  File "src\tribler-gui\tribler_gui\tribler_window.py", line 488, in on_tribler_started, 
  File "src\tribler-gui\tribler_gui\widgets\channelcontentswidget.py", line 124, in initialize_content_page, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 297, in __init__, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 93, in __init__, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 195, in __init__, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 53, in __init__, 
tribler_gui.utilities.CreationTraceback, 
  File "src\tribler-gui\tribler_gui\utilities.py", line 385, in trackback_wrapper, 
  File "src\tribler-gui\tribler_gui\utilities.py", line 382, in trackback_wrapper, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 54, in <lambda>, 
  File "src\tribler-gui\tribler_gui\widgets\triblertablecontrollers.py", line 135, in check_torrent_health, 
KeyError: 'infohash'
```
[Here](https://github.com/Tribler/tribler/blob/abd610cda9efc01972ce59ed66d7de9d3434472e/src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py#L135):
```python
class HealthCheckerMixin:
    def check_torrent_health(self, data_item, forced=False):
        # TODO: <not relevant for this bug>
        infohash = data_item[u'infohash']
        ...
```
I think the fix should be similar to [this](https://github.com/Tribler/tribler/pull/5292/files) one. I mean:
```python
if data_item['type'] != REGULAR_TORRENT:
    return
```

drew2a:
This PR resolves #5909 by adding a check on the existing of `infohash` key.

The root cause of the issue is too general an inaccurate design of `ContentTableViewController`.
My proposal: to schedule a refactoring for `ContentTableViewController`.

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
+++ b/src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
@@ -17,7 +17,7 @@ from tribler_core.utilities.json_util import dumps
 from tribler_gui.defs import HEALTH_CHECKING, HEALTH_UNCHECKED
 from tribler_gui.tribler_action_menu import TriblerActionMenu
 from tribler_gui.tribler_request_manager import TriblerNetworkRequest
-from tribler_gui.utilities import connect, get_health
+from tribler_gui.utilities import connect, dict_item_is_any_of, get_health
 
 HEALTHCHECK_DELAY_MS = 500
 
@@ -115,7 +115,7 @@ class TableSelectionMixin(object):
             return
 
         data_item = selected_indices[-1].model().data_items[selected_indices[-1].row()]
-        if 'type' in data_item and data_item['type'] != REGULAR_TORRENT:
+        if not dict_item_is_any_of(data_item, 'type', [REGULAR_TORRENT]):
             return
 
         # Trigger health check if necessary
@@ -132,7 +132,7 @@ class TableSelectionMixin(object):
 class HealthCheckerMixin:
     def check_torrent_health(self, data_item, forced=False):
         # TODO: stop triggering multiple checks over a single infohash by e.g. selection and click signals
-        if 'infohash' not in data_item:
+        if not dict_item_is_any_of(data_item, 'type', [REGULAR_TORRENT]):
             return
 
         infohash = data_item[u'infohash']
@@ -290,7 +290,7 @@ class ContextMenuMixin:
     def selection_can_be_added_to_channel(self):
         for row in self.table_view.selectionModel().selectedRows():
             data_item = row.model().data_items[row.row()]
-            if 'type' in data_item and data_item['type'] in (REGULAR_TORRENT, CHANNEL_TORRENT, COLLECTION_NODE):
+            if dict_item_is_any_of(data_item, 'type', [REGULAR_TORRENT, CHANNEL_TORRENT, COLLECTION_NODE]):
                 return True
         return False
 
--- a/src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
+++ b/src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
@@ -132,6 +132,9 @@ class TableSelectionMixin(object):
 class HealthCheckerMixin:
     def check_torrent_health(self, data_item, forced=False):
         # TODO: stop triggering multiple checks over a single infohash by e.g. selection and click signals
+        if 'infohash' not in data_item:
+            return
+
         infohash = data_item[u'infohash']
 
         if u'health' not in self.model.column_position:
--- a/src/tribler-gui/tribler_gui/tests/test_util.py
+++ b/src/tribler-gui/tribler_gui/tests/test_util.py
@@ -1,6 +1,6 @@
 from urllib.parse import unquote_plus
 
-from tribler_gui.utilities import compose_magnetlink, quote_plus_unicode, unicode_quoter
+from tribler_gui.utilities import compose_magnetlink, dict_item_is_any_of, quote_plus_unicode, unicode_quoter
 
 
 def test_quoter_char():
@@ -105,3 +105,22 @@ def test_compose_magnetlink():
     assert composed_link1 == expected_link1
     assert composed_link2 == expected_link2
     assert composed_link3 == expected_link3
+
+
+def test_is_dict_has():
+    assert not dict_item_is_any_of(None, None, None)
+    assert not dict_item_is_any_of({}, None, None)
+
+    d = {
+        'k': 'v',
+        'k1': 'v1'
+    }
+
+    assert not dict_item_is_any_of(d, 'missed_key', None)
+    assert not dict_item_is_any_of(d, 'missed_key', ['any_value'])
+    assert not dict_item_is_any_of(d, 'k', ['missed_value'])
+    assert not dict_item_is_any_of(d, 'k', ['missed_value', 'missed_value1'])
+
+    assert dict_item_is_any_of(d, 'k', ['v'])
+    assert dict_item_is_any_of(d, 'k', ['v', 'a'])
+    assert dict_item_is_any_of(d, 'k', ['a', 'v'])
--- a/src/tribler-gui/tribler_gui/utilities.py
+++ b/src/tribler-gui/tribler_gui/utilities.py
@@ -422,3 +422,9 @@ def disconnect(signal: pyqtSignal, callback: Callable):
 
 class CreationTraceback(Exception):
     pass
+
+
+def dict_item_is_any_of(d, key, values):
+    if not d or not key or not values:
+        return False
+    return key in d and d[key] in values


src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
```diff
@@ -132,6 +132,9 @@ def _on_selection_changed(self, selected, deselected):
 class HealthCheckerMixin:
     def check_torrent_health(self, data_item, forced=False):
         # TODO: stop triggering multiple checks over a single infohash by e.g. selection and click signals
+        if 'infohash' not in data_item:
```
kozlovsky, line 135: As I understand it, channels have infohashes as well as regular torrents, but we probably don't want to show health information for channels in this view.

In other similar places, there is a check `if data_item['type'] != REGULAR_TORRENT`. I think we need to use a similar check here as well.

drew2a: Checking infohash is a more reliable approach if we look at it as a "bugfix".

But your suggestion is the more in `triblertablecontrollers' style`, which is more correct if we look at the change as a "triblertablecontrollers fix".

I've implemented your suggestion (still not sure that this is the best version of bugfix).

ichorid: It is indeed more correct to check for the entry type, not for its attributes. In this way, we catch the errors in the application *logic*, though at the cost of the thing crashing for the user.

src/tribler-gui/tribler_gui/widgets/triblertablecontrollers.py
```diff
@@ -287,7 +293,7 @@ def add_menu_item(self, menu, name, item_index, callback):
     def selection_can_be_added_to_channel(self):
         for row in self.table_view.selectionModel().selectedRows():
             data_item = row.model().data_items[row.row()]
-            if 'type' in data_item and data_item['type'] in (REGULAR_TORRENT, CHANNEL_TORRENT, COLLECTION_NODE):
+            if dict_has(data_item, 'type', REGULAR_TORRENT, CHANNEL_TORRENT, COLLECTION_NODE):
```
kozlovsky, line 293: Actually, I don't like this new `dict_has` function. In my opinion, it makes code much harder to understand.
1. Previously the code did not require additional context to understand what is going on. Now I need to look at the `dict_has` source to understand the details of its logic.
2. In the previous code it was clear to me that `dict_item` is a `dict` instance (or, at least, some objects with the Mapping interface). Now the code will not break if `dict_item` is, for example, `None`. So, if I need to extend this function in the future, I need to spend time to answer the question "can dict_item be anything other than dict", whereas in the previous version it was clear.
3. The new version is not significantly shorter. If the goal is to achieve brevity it was possible to write instead:
```python
if data_item.get('type') in (REGULAR_TORRENT, CHANNEL_TORRENT, COLLECTION_NODE):
    ...
```

So I propose to roll back this change and remove the `dict_has` function.

kozlovsky: Python is a dynamically typed language but it is strongly typed. The expression`foo[bar]` generates `KeyError` exception when it is incorrectly assumed that `foo` should be a mapping and `bar` is a required item of this mapping, and it is a good thing. It is like a programmer explicitly adds some assertions to its code which will guarantee that errors will raise fast and not be propagated to different places. When we add unnecessary checks `if foo` or `if bar in foo` we essentially convert strongly typed language to a weakly typed one.

drew2a: > Previously the code did not require additional context to understand what is going on. Now I need to look at the dict_has source to understand the details of its logic.

Yes, and this is ok. You need to understand once what this function does, and then relax for all occurences.
`dict_has` just extract a pattern that was used in this file three times

> Now the code will not break if data_item is, for example, None
Yes, and this is great. No excuses for crash on the user side.

> The new version is not significantly shorter.
Yes, but shortnesses is doubtful purpose. The new version reduces the number of repetitions (of a pattern and of key's value, inside the pattern). Therefore it decreases the chance to make a mistake.

kozlovsky: > dict_has just extract a pattern that was used in this file three times

Not all patterns probably should be extracted. `a + b * c` is better than `addmul(a, b, c)`

> Yes, and this is great. No excuses for crash on the user side.

If we a sure that, say, something can never be `None`, adding unnecessary handling for `None` leads to less clear code, as from now it says to the future reader: there are some obscure cases when this variable really can be `None`.

> Yes, and this is ok. You need to understand once what this function does, and then relax for all other occurrences.

Many developers need to do it, and even a single developer does not remember all details of all the small functions that he read in the entire codebase and needs to read it again.

Actually, this specific case is small enough to turn it into the bikeshedding, but the general pattern looks a bit like a code obfuscation to me.

ichorid: I agree with @kozlovsky . Python is not C++: Python programmers are used to some idiomatic ways to do simple stuff like checking `dict` attributes, etc. Python is not about inventing your own sub-language as C++ does. Python is about making the code immediately readable to anyone.


So, replacing an explicit one-liner with an implicit one-liner goes against [the Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python)

`import this` :wink:

drew2a: > Not all patterns probably should be extracted. a + b * c is better than

Yes, not all. But some -- should.

> there are some obscure cases when this variable really can be None

Yes, and if we take a look at the recent issues from the GUI side, we will be ensured that there are a lot of "obscure cases". So, here the question, what is better: to let the client crash at the next "obscure case" and fix it, or to assume that there are such cases, and carefully handle them?

ichorid: > what is better: to let the client crash at the next "obscure case" and fix it, or to assume that there are such cases, and carefully handle them?

To let the client crash at the next few obscure cases, fix 3-4 of these and then figure out the common pattern and refactor it for good. Alas, the problem with the GUI (especially with the PyQT one) is that there are far too many corner cases to devise a consistent solution upfront...

drew2a: @ichorid client's crash should be unacceptable for a client-oriented product. If we want a million users, we have to take more care of them.

ichorid: Silencing errors results in accumulation of unseen defects. What is worse: a button that crashes the application instantly, or a button that does not do anything?

drew2a: I'll give you an answer from my POV, assuming that we talking about Tribler (not in general):
"a button that does not do anything" is better than "a button that crashes the application instantly"

ichorid: How do you know that your buttons work for your users then?

drew2a: I assume that the author of the button checked the button's correctness when he added the corresponding feature. 
Plus all new features should be covered by tests.

But even if no. It is better to receive a user's report that "the button doesn't work" than receive the Sentry event "application is crashed".

In the first case, we will get a loyal user, in the second case, we will lose him.

ichorid: >But even if no. It is better to receive a user's report that "the button doesn't work" than receive the Sentry event "application is crashed". In the first case, we will get a loyal user, in the second case, we will lose him.

What's the probability of the first case vs the second? How many users write reports about broken buttons? More important, *what kind of user* is going to produce these reports? Is it new users, or experienced/loyal ones? What's the price of a broken button in terms of silently disappointed users?

We [already had](https://github.com/Tribler/tribler/issues/4930) this kind of discussions before. The consensus was: "We are going to make Tribler work in two modes: either *debug* or *production*. Debug should always crash, production should never crash. Production release should only be done after a couple of weeks of RCs in Debug mode." No one ever bothered to implement it properly, though.

Well, we don't do extensive RCs now, since we are on the monthly release schedule. So, every release must **look** as stable as possible, even if it means hiding defects, right?

kozlovsky: The number of safety checks which can be added to the code "just for a case" is very big. If we start adding such safety checks to every possible place then it would be very hard to understand, which of these safety checks are real checks added to prevent some possible situations. As in the current code we don't have situations when `data_item` can be None or when it might do not contain "type" key, in my opinion, it is better not to include such checks. Otherwise, if we start applying consistently the same approach to all possible places in code, the codebase will be much harder to read.

drew2a: >  As in the current code we don't have situations when data_item can be None or when it might do not contain "type"

How can you prove that?

kozlovsky: > How can you prove that?

By the absence of the corresponding exceptions in reports

drew2a: Dijkstra would say that exceptions in reports can be used to show the presence of bugs, but never to show their absence :)

drew2a: @ichorid 

> So, every release must look as stable as possible, even if it means hiding defects, right?

Probably it is better to just answer "yes" on your question.

But I would say, if release doesn't look stable, you should publish it. Taking into account, that "obscure cases" always will be presented... we need to make release stable with the margin of safety.

(in my humble view)

ichorid: > But I would say, if release doesn't look stable, you should publish it.

Maybe you meant "you **shouldn't** publish it"? :wink:

drew2a: Yes 🤦 
Fixed.

ichorid: > Dijkstra would say that exceptions in reports can be used to show the presence of bugs, but never to show their absence :)

Of course. Yet exceptions is almost the only kind of feedback and support that we get from our userbase. Because, you know, we can't do telemetry.

xoriole (approved):

kozlovsky (approved):

PR #5933, opened 2021-01-11
Ignore patch version in upgrade scenarios

### LINKED ISSUE #5927: No old state directory for rollback
Tribler crashes if the state directory for the rollback to the old version does not exist. Instead of crashing, it would be better to start a fresh.

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/<user>/.Tribler/7.6'
```

Sentry Issue: [TRIBLER-41](https://sentry.tribler.org/organizations/tribler/issues/289/?referrer=github_integration)

ichorid:
*hopefully* fixes #5927 

There is a double check that should have prevented that scenario:
* the directory existence check
* the version check

This PR fixes the version check. Indeed, there was a problem (we did not ignore the patch number when considering if we should  upgrade the version.) Well, I still can't understand how the user pulled that off, 'cause the directory existence check seems to work (at least according to automated and manual tests). They could still trigger that if they manually deleted the directory during the upgrade procedure or something like that.

### DIFF
--- a/src/tribler-core/tribler_core/upgrade/version_manager.py
+++ b/src/tribler-core/tribler_core/upgrade/version_manager.py
@@ -54,6 +54,23 @@ and "stashes" the state dirs with conflicting names by renaming them.
 Note that due to failures in design pre-7.4 series and 7.4.x series get special treatment.
 """
 
+def must_upgrade(old_ver:LooseVersion, new_ver:LooseVersion):
+    """
+    This function compares two LooseVersions by combination of major/minor components only, omitting the patch version.
+    By convention, in Tribler we only fork version directories on major/minor version changes, and never touch
+    the directory format in patch versions.
+    e.g 1.2.3 -> 1.3.0 must upgrade (return True)
+    but 1.2.3 -> 1.2.4 must not upgrade (return False)
+
+    :param old_ver: the version to upgrade from
+    :param new_ver:  the version to upgrade to
+    :return: True, if new_ver's major/minor version is higher than old_ver's major/minor version, otherwise False
+    """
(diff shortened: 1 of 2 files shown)


drew2a:
@ichorid release has been already published. Therefore, the target branch has to be changed.

src/tribler-core/tribler_core/upgrade/version_manager.py
```diff
@@ -54,6 +54,20 @@
...
+    the directory format in patch versions.
+    e.g 1.2.3 -> 1.3.0 must upgrade (return True)
+    but 1.2.3 -> 1.2.4 must not upgrade (return False)
+
+    :param old_ver: the version to upgrade from
+    :param new_ver:  the version to upgrade to
+    :return: True, if new_ver's major/minor version is higher than old_ver's major/minor version, otherwise False
+    """
+    return (new_ver.version[0] > old_ver.version[0]) or \
+           ((new_ver.version[0] == old_ver.version[0]) and (new_ver.version[1] > old_ver.version[1]))
```
drew2a, line 70: **suggestion:** this version probably will be more reader-friendly:

```python
    major_version_is_greater = new_ver.version[0] > old_ver.version[0]
    major_version_is_equal = new_ver.version[0] == old_ver.version[0]
    minor_version_is_greater = new_ver.version[1] > old_ver.version[1]

    return major_version_is_greater or (major_version_is_equal and minor_version_is_greater)
```

drew2a (changes requested):

xoriole (approved):

ichorid:
> @ichorid release has been already published. Therefore, the target branch has to be changed.

I thought we were probably going to do a patch release?

drew2a (approved):

PR #5940, opened 2021-01-14
Fix memory leak in trackback_wrapper

kozlovsky:
In this PR I preprocess creation traceback to avoid storing all frames with locals inside a signal handler.


xoriole (approved):

xoriole:
Before fix
Screenshot from 2021-01-19 10-40-30
After this fix
Screenshot from 2021-01-19 10-41-11
This does not fully fix the issue but slows down

xoriole (approved):

PR #5945, opened 2021-01-18
Skip all the versions produced by the deployment tester

drew2a:
This PR fixes the issue of unnecessary multiplication of "release version" during a deployment test:

https://sentry.tribler.org/organizations/tribler/releases/?project=7

### DIFF
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
@@ -112,21 +112,23 @@ def distinct_by(list_of_dict, key):
     return result
 
 
-def skip_dev_version(version):
-    """
-    For the release version let's ignore all "developers" versions
-    to keep the meaning of the `latest` keyword:
-    See Also:https://docs.sentry.io/product/sentry-basics/search/
-    Args:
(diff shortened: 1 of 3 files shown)


src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
```diff
@@ -112,21 +112,23 @@ def distinct_by(list_of_dict, key):
...
     if 'GIT' in version:
         return None
 
-    return version
+    parts = version.split('-', maxsplit=2)
+    if len(parts) < 2:
+        return version
+    
+    # in case of a release candidate, leave "<version>-RC<number>"
+    if parts[1].lower().startswith('rc'):
```
xoriole, line 130: There could be other suffixes besides RC eg. exp for experimental so it might be better to use regular expression to check for the version here allowing any alphanumeric suffix

xoriole (dismissed):

src/tribler-common/tribler_common/sentry_reporter/sentry_tools.py
drew2a: fixed

PR #5948, opened 2021-01-18
Stop crash an application when GUI raise an exception

drew2a:
This PR addresses _problem1_ from https://github.com/Tribler/tribler/discussions/5938 and starts to solve it.

In this PR refactoring was made to explicitly split GUI errors and Core errors.

* In the case of Core errors, we always kill the application.
* In the case of GUI errors, we only kill the application when `CoreConnectTimeoutError` will occur.

Tribler will report only the first occurred GUI exception (filtered by Type).
Repetitive exceptions will be ignored.

### DIFF
--- a/src/tribler-gui/tribler_gui/error_handler.py
+++ b/src/tribler-gui/tribler_gui/error_handler.py
@@ -1,0 +1,87 @@
+import logging
+import os
+import traceback
+
+from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
+
+from tribler_gui.dialogs.feedbackdialog import FeedbackDialog
+from tribler_gui.event_request_manager import CoreConnectTimeoutError
+
+
+class ErrorHandler:
+    def __init__(self, tribler_windows):
+        logger_name = self.__class__.__name__
+        self._logger = logging.getLogger(logger_name)
+        SentryReporter.ignore_logger(logger_name)
+
+        self.tribler_windows = tribler_windows
+
+        self._handled_exceptions = set()
+        self._tribler_stopped = False
+
+    def gui_error(self, *exc_info):
+        if self._tribler_stopped:
+            return
+
+        info_type, info_error, tb = exc_info
+        if info_type in self._handled_exceptions:
+            return
+        self._handled_exceptions.add(info_type)
+
+        text = "".join(traceback.format_exception(info_type, info_error, tb))
+
+        if info_type is CoreConnectTimeoutError:
+            text = text + self.tribler_windows.core_manager.core_traceback
+            self._stop_tribler(text)
+
+        self._logger.error(text)
+
+        FeedbackDialog(
+            self.tribler_windows,
+            text,
+            self.tribler_windows.tribler_version,
+            self.tribler_windows.start_time,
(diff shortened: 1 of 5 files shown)


src/tribler-gui/tribler_gui/error_handler.py
```diff
@@ -0,0 +1,87 @@
...
+import traceback
+
+from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
+
+from tribler_gui.dialogs.feedbackdialog import FeedbackDialog
+from tribler_gui.event_request_manager import CoreConnectTimeoutError
+
+
+class ErrorHandler:
+    def __init__(self, tribler_windows):
```
kozlovsky, line 12: Probably singular `tribler_window` should be used here, as this is the reference to the Tribler main window

src/tribler-gui/tribler_gui/error_handler.py
```diff
@@ -0,0 +1,87 @@
...
+        self._tribler_stopped = True
+
+        self.tribler_windows.tribler_crashed.emit(text)
+        self.tribler_windows.delete_tray_icon()
+
+        # Stop the download loop
+        self.tribler_windows.downloads_page.stop_loading_downloads()
+
+        # Add info about whether we are stopping Tribler or not
+        os.environ['TRIBLER_SHUTTING_DOWN'] = str(self.tribler_windows.core_manager.shutting_down)
```
kozlovsky, line 80: I understand that this PR just move this code from some other place, but it probably should be fixed anyway.
When Tribler code checks the value of this environment variable, it expects to get "TRUE" or something else:
```python
@web.middleware
async def error_middleware(request, handler):
    try:
        if os.environ.get('TRIBLER_SHUTTING_DOWN', "FALSE") == "TRUE":
            raise Exception('Tribler is shutting down')
```
The values of environment variables are case-sensitive, at least on non-Windows systems. So the check above should not work when the actual value is `"True"` instead of `"TRUE"`

So I suspect you need to add `upper`:
```python
os.environ['TRIBLER_SHUTTING_DOWN'] = str(self.tribler_windows.core_manager.shutting_down).upper()
```

kozlovsky: Actually, it is not obvious to me, why we use this `TRIBLER_SHUTTING_DOWN` environment variable at all, instead of just setting some flag in memory. It may be useful when two processes are communicating using an environment variable value, but in our case, GUI does not read this environment variable value in its code.

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/error_handler.py
drew2a: @devos50 seems to be an author of this line :)

drew2a: @kozlovsky changed to
```
os.environ['TRIBLER_SHUTTING_DOWN'] = str(self.tribler_windows.core_manager.shutting_down).upper()
```

kozlovsky (approved):

xoriole (approved):

PR #5949, opened 2021-01-19
Fix memory leak in Downloads page

### LINKED ISSUE #5934: ubuntu 18.04 LTS, ram filling up completely
Hi,
I installed tribler 7.7.0 last friday (downloaded deb file, installation via Ubuntu sw centre) on a ubuntu 18.04 LTS PC as alternative to transmission. Every day I had to reboot the PC as it was completely stuck. Yesterday I started following RAM: after starting tribler I get 4 entries in the list (I have a couple of active torrents), of which 1 starts at ~100 MB RAM, and this continues to grow until the RAM (5 GB in my case) is completely filled up. This morning I stopped tribler, and restarted, but the same behaviour is still there. After installing 7.7.1 and starting today I still see RAM-usage increasing (from 150 M at 14:00 today, to 1.8 G at 18:00).

I am running remmina as remote terminal for the ubuntu 18.04 LTS PC, on a laptop running ubuntu 20.04 LTS, but I cannot imagine that that causes the problem.

Is this expected behaviour? What maximum RAM-usage do you expect from tribler?

I saw a similar problem (issue 2123, 22-4-2016), but I would expect this to be solved in the mean time.

Let me know if you need more information.

ichorid:
Fixes #5934 

The moral of the story: if you connect a QT signal to a QT object, don't expect it to be GCed in PyQT until you disconnect the signal manually (or better yet, just reuse the object).

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/downloadspage.py
+++ b/src/tribler-gui/tribler_gui/widgets/downloadspage.py
@@ -96,6 +96,11 @@ class DownloadsPage(AddBreadcrumbOnShowMixin, QWidget):
         self.window().downloads_list.header().setSortIndicator(12, Qt.AscendingOrder)
         self.window().downloads_list.header().resizeSection(12, 146)
 
+        self.downloads_timeout_timer.setSingleShot(True)
+        self.downloads_timer.setSingleShot(True)
+        connect(self.downloads_timer.timeout, self.load_downloads)
+        connect(self.downloads_timeout_timer.timeout, self.on_downloads_request_timeout)
+
     def on_filter_text_changed(self, text):
         self.window().downloads_list.clearSelection()
         self.window().download_details_widget.hide()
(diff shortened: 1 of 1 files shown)


ichorid:
retest this please

drew2a:
@ichorid congrats! 🎊 

But I have a question. Are you sure about " don't expect it to be GCed in PyQT until you disconnect the signal manually"?

Because:

>To connect slots to it, pass callbacks into connect. The connections are maintained through weakref, thus you don't need to search for them and disconnect whenever you're up to destroy some object.
https://pypi.org/project/qsignal/

BTW yes, we always should help GC: https://github.com/spesmilo/electrum/issues/4905

drew2a (approved):

ichorid:
> @ichorid congrats! confetti_ball
> 
> But I have a question. Are you sure about " don't expect it to be GCed in PyQT until you disconnect the signal manually"?

I learned the hard way never to trust PyQT promises... :crying_cat_face:

drew2a:
@ichorid what do you think, do we have to use something like:

```python
class QAutoDisconnectableObject(QObject):
    def __init__(self):
        super().__init__()

        self.connected_signals = set()

        connect(self.destroyed, self.on_destroy)

    def connect(self, signal, callback):
        connect(signal, callback)
        self.connected_signals.add(signal)

    def on_destroy(self, *args):
        for signal in self.connected_signals:
            signal.disconnect()


class Table(QAutoDisconnectableObject):
    def __init__(self):
        super().__init__()
        self.connect(self.button.clicked, self.on_button_clicked)
        self.connect(self.button1.clicked, self.on_button1_clicked)
```

qstokkink:
@drew2a It seems useful to have a construction like that: I don't see any downsides to your suggestion.

ichorid:
https://www.sep.com/blog/prevent-signal-slot-memory-leaks-in-python/

PR #5951, opened 2021-01-22
Add auto-disconnecting mixin for QObject

drew2a:
This PR adds a Mixin that can disconnect QR signals automatically.

Based on the discussoin https://github.com/Tribler/tribler/pull/5949 and regarding @ichorid experience, it is necessary to manually disconnect a QT signal. This should help for avoiding memory leaks like https://github.com/Tribler/tribler/issues/5934

The current version of the Mixin is working with no errors (at least for me).

### The pitfalls

During developing this Mixin I saw a lot of "strange" bugs when python crashed with the following error: "Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)". 

They occurred, for example, when trying to log some of the signals:

```python
                logger.info(f'Signal: {signal}')
```

And when trying to use (some) of the signals as a key in a dictionary:

```python
                self.connected_signals[signal] += 1
```

Another case. Some signals lead to a python crash if you try to disconnect all callbacks at once:
```python
                signal.disconenct()
```

Therefore:
* `id(signal)` is using for logging and as a key for a dictionary
* a full disconnect is performed by calling callback-by-callback disconnect.

### Conclusion

The Mixing is working for me with no errors, but I consider the QT disconnect mechanism is fragile.

Unfortunately, I see more risks for merging this PR rather than benefits.

What do you think about that?

### DIFF
--- a/src/tribler-gui/tribler_gui/dialogs/auto_disconnecting_mixin.py
+++ b/src/tribler-gui/tribler_gui/dialogs/auto_disconnecting_mixin.py
@@ -1,0 +1,78 @@
+import logging
+from collections import defaultdict
+from threading import RLock
+
+from PyQt5.QtCore import pyqtBoundSignal
+
+from tribler_gui.utilities import connect, disconnect
+
+lock = RLock()
+logger = logging.getLogger('QAutoDisconnectingMixin')
+
+
+class QAutoDisconnectingMixin:
+    enabled = False
+
+    def lazy_init(self):
+        if not QAutoDisconnectingMixin.enabled:
+            return
+
+        with lock:
+            # abbreviation "adm" here is the attempt to avoid naming conflicts
+            if hasattr(self, 'adm_inited'):
+                return
+
+            self.adm_inited = True
+            self.adm_destroyed_callback_connected = False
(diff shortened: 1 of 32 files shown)


src/tribler-gui/tribler_gui/dialogs/auto_disconnecting_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+
+from PyQt5.QtCore import pyqtBoundSignal
+
+from tribler_gui.utilities import connect, disconnect
+
+lock = RLock()
+logger = logging.getLogger('QAutoDisconnectingMixin')
+
+
+class QAutoDisconnectingMixin:
```
drew2a, line 13: This is the main improvement for the PR.

ichorid:
Indeed, connecting/disconnecting signals is a fragile topic in PyQT. I'm afraid it can bring more trouble because of potential interference with different QT version on different platforms, etc. I would prefer adding a separate warning ("caveat") chapter to our developer's guidelines instead. And of course, point out the PR reviewers to pay special attention to signals.

PR #5961, opened 2021-01-28
Remove max_peers from DiscoveryBooster

drew2a:
This PR removes the `max_peers` argument from `DiscoveryBooster` because it has no effect on discovery boosting.

Linked to the dev morning meeting.


egbertbouman (approved):
Looks good to me (assuming that you've verified that `max_peers` is indeed not needed).

drew2a:
I've verified it empirically :)

PR #5955, opened 2021-01-27
Updated ipv8 pointer

qstokkink:
This PR:

 - Updates the IPv8 pointer.
 - Updates the `IPv8CommunityLauncher` base class to load the usual bootstrapping servers for each overlay.
 - Refactors the `get_ipv8_bootstrap_override` check in `Session` to the `IPv8CommunityLauncher` to reduce clutter.

Gumby will have to be changed after this PR.

PR #5964, opened 2021-01-28
Remove hard-coded peer limits from DebugWindow

egbertbouman:
Note that this needs to go to devel since we didn't want to move the IPv8 pointer too far (https://github.com/Tribler/tribler/pull/5955#issuecomment-768202925)


drew2a (approved):
👍

qstokkink (approved):
Thanks 😃

qstokkink (approved):

PR #5992, opened 2021-02-10
Allow exitnodes to run without libtorrent installed

kozlovsky:
This PR allows running the `run_tunnel_helper.py` script without libtorrent installing.

Currently, many modules import `libtorrent` even if it is not necessary. `run_tunnel_helper.py` script even [sets `libtorrent_enabled` option to `False`](https://github.com/Tribler/tribler/blob/devel/src/tribler-core/run_tunnel_helper.py#L124), but still imports `libtorrent` library. As installing libtorrent from sources may be a non-trivial task, it is better to allow running `run_tunnel_helper.py` without `libtorrent` installed.

This PR does two things:

1. Refactor imports to import libtorrent not directly, but from `tribler_core.utils.libtorrent_helper`. When `libtorrent` is not installed, the `libtorrent_helper` module returns a fallback object instead, which raises `ImportError` on any attempt to access `libtorrent` module members. The `ImportError` exception was chosen as it provides a clear explanation for the reason for the error. It is possible to use another exception instead. Still, it should not be `RuntimeError`, as the `libtorrent` module itself uses `RuntimeError` (in the `bdecode` function, for example). Tribler code may erroneously catch the exception and assume that the reason for the error is incorrect data for the `bdecode` function.

2. Add `is_bencoded(x)` pure-python function, which checks that the data appears to be a valid bencoded string. This function is used in `TriblerTunnelCommunity` instead of `libtorrent.bdecode` call. It is not necessary to actually perform `bdecode` in `TriblerTunnelCommunity`, as the purpose is just to check that the data is not something unrelated to libtorrrent.

Later, it will be easy to add a pure-python implementation of the `bencode` and `bdecode` functions to the `LibtorrentFallback` class if the need arises, but right now, there are no such use-cases where it may be helpful.

### DIFF
--- a/src/tribler-core/tribler_core/utilities/tests/test_bencodecheck.py
+++ b/src/tribler-core/tribler_core/utilities/tests/test_bencodecheck.py
@@ -3,109 +3,83 @@ import pytest
 from tribler_core.utilities.bencodecheck import is_bencoded
 
 
-def test_bcheck_nobytes():
+def test_bencode_checker():
     # only bytes is valid argument type
     with pytest.raises(ValueError, match='^Value should be of bytes type. Got: str$'):
         is_bencoded('3:abc')
 
-def test_bcheck_empty():
+    # empty encoded string
     assert not is_bencoded(b'')
 
-def test_bcheck_excess_data1():
-    # excess data after the end of encoded string
+    # excess data after the end of encoded data
     assert not is_bencoded(b'3:abc3:abc')
 
-def test_bcheck_excess_data2():
-    # excess end marker after the end of encoded string
+    # excess end marker after the end of the encoded data
     assert not is_bencoded(b'3:abce')
 
-def test_bcheck_str1():
     # empty string
     assert is_bencoded(b'0:')
 
-def test_bcheck_str2():
     # normal string with 'abc' value
     assert is_bencoded(b'3:abc')
 
-def test_bcheck_str3():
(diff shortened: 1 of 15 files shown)


src/tribler-core/tribler_core/utilities/tests/test_bencodecheck.py
```diff
@@ -0,0 +1,111 @@
...
+
+from tribler_core.utilities.bencodecheck import is_bencoded
+
+
+def test_bcheck_nobytes():
+    # only bytes is valid argument type
+    with pytest.raises(ValueError, match='^Value should be of bytes type. Got: str$'):
+        is_bencoded('3:abc')
+
+def test_bcheck_empty():
```
ichorid, line 11: For the sake of readability and test speed, I would put all these tests into a single long test instead.

kozlovsky:
retest this please

src/tribler-core/tribler_core/utilities/tests/test_bencodecheck.py
kozlovsky: Changed to a single test

egbertbouman (dismissed):
Looks good to me. As far as I can tell the dependency checker will still produce an error message in case someone tries to run Tribler normally without libtorrent.

ichorid (approved):

PR #5804, opened 2020-12-04
Hide token balance

devos50:
This PR hides the token balance from the GUI, at least until 7.6.1 has been released.

PR #5995, opened 2021-02-11
Wiping bandwidth transactions + upgrading community IDs

devos50:
This PR addresses our payout issue (see #5789) where the token balances of peers became disproportional high. To fix this issue, I have made the following changes:
- Updated the bandwidth transaction database to version 9. When updating to this version, I wipe all transactions + history items from this database.
- Updated the community ID of the `BandwidthAccountingCommunity`. This should prevent peers acting in the 'infected' community from performing payouts to peers running Tribler 7.8.
- Updated the community ID of the `TriblerTunnelCommunity`.
- Showing the token balance again, effectively reverting #5804.


xoriole (commented):
Should the token graph be enabled as well?

devos50:
The graph seems to be enabled, as is the charts with the evolution of ones balance 👍

xoriole (approved):

PR #6004, opened 2021-02-19
Python random is discouraged, move to secrets

NoNameForMee:
Some of the CI/CD jobs for Tribler already warns about this:
https://github.com/Tribler/tribler/runs/1913957805
https://sonarcloud.io/project/security_hotspots?id=Tribler_tribler&hotspots=AW_DWqEXH4jiAyAXJjGK

Official python doc:
https://docs.python.org/3/library/secrets.html
https://docs.python.org/3/library/random.html

Please note: I used `secrets.choise(range(min,max))` as this is somewhat
clearer than `secrets.randbelow(n)` which treats edge different from random..
(I.e. `secrets.randbelow` would have to have "`n=max-min+1`" and then add "`min`").

And also note: I did not remove the "`import random`",
as some lines could not be translated 1:1 with the new secrets.

Feel free to edit this merge request, discard it or merge it at your will...

### DIFF
--- a/src/tribler-common/tribler_common/network_utils.py
+++ b/src/tribler-common/tribler_common/network_utils.py
@@ -3,6 +3,7 @@ This module contains some utility functions for networking.
 """
 import logging
 import random
+import secrets
 import socket
 import struct
 import sys
@@ -45,7 +46,7 @@ def get_random_port(socket_type="all", min_port=5000, max_port=60000):
     assert 0 < min_port <= max_port <= MAX_PORT, f"Invalid min_port and mac_port values {min_port}, {max_port}"
 
     working_port = None
-    try_port = random.randint(min_port, max_port)
+    try_port = secrets.choice(range(min_port, max_port))
     while try_port <= MAX_PORT:
         if check_random_port(try_port, socket_type):
             working_port = try_port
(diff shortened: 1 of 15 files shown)


ichorid:
ok to test

egbertbouman:
I'm not really seeing any issues with using `random`. Of course, `random` is only pseudo-random, but it's not like we're using it for anything security-related. If we decide to merge this, it would mostly be to silence the CI/CD jobs.

qstokkink:
As I was requested for review: completely agree with @egbertbouman.
You pay the price in performance (roughly 1.5 times slower) if you use secure random numbers unnecessarily:

```python
>>> timeit.timeit("secrets.choice(range(255))", "import secrets")
0.9644661999999755
>>> timeit.timeit("random.choice(range(255))", "import random")
0.5771826999999803

>>> timeit.timeit("secrets.choice(range(65535))", "import secrets")
1.0464367000000152
>>> timeit.timeit("random.choice(range(65535))", "import random")
0.6713555999999983
```

_Disclaimer: if you use random numbers for security you absolutely need to use `secrets` or `os.urandom`._

ichorid:
@qstokkink , @egbertbouman  do we use secure random generators for tunnels and private key generation?

qstokkink:
@ichorid yes, always.

ichorid:
> @ichorid yes, always.

Then I see no reasons for moving to `secrets`. 

Thanks for all the work done though, @NoNameForMee . We prefer to only change stuff if it makes a positive impact. Better luck next time :wink:

Issue #5907, opened 2021-01-06
No space to upgrade


xoriole:
What is the best course of action if there is not enough storage to copy the previous version state directory and continue with the upgrade? (Reported in error reporter)
```python
Traceback (most recent call last):
File "src\run_tribler.py", line 179, in <module>
File "src\tribler-gui\tribler_gui\utilities.py", line 382, in trackback_wrapper
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 150, in connect
tribler_gui.utilities.CreationTraceback

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "src\tribler-gui\tribler_gui\utilities.py", line 385, in trackback_wrapper
File "src\tribler-gui\tribler_gui\utilities.py", line 382, in trackback_wrapper
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 150, in <lambda>
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 84, in on_error
tribler_gui.event_request_manager.CoreConnectTimeoutError: Could not connect with the Tribler Core within 60 seconds
[PID:18788] 2021-01-06 11:05:53,877 - ERROR <base_events:1707> asyncio.default_exception_handler(): Task exception was never retrieved
future: <Task finished name='Task-1' coro=<start_tribler_core.<locals>.start_tribler() done, defined at src\run_tribler.py:53> exception=OSError(28, 'No space left on device')>
Traceback (most recent call last):
File "src\run_tribler.py", line 62, in start_tribler
File "src\tribler-core\tribler_core\upgrade\version_manager.py", line 171, in fork_state_directory_if_necessary
File "src\tribler-core\tribler_core\upgrade\version_manager.py", line 133, in copy_state_directory
File "lib\shutil.py", line 415, in copy
File "lib\shutil.py", line 261, in copyfile
OSError: [Errno 28] No space left on device: 'C:\\Users\\<User>\\AppData\\Roaming\\.Tribler\\7.7\\ecpub_multichain.pem'
```

With monthly releases scheduled, this could be an issue for many. We should start removing old state directories.

qstokkink:
+1 on this. My disk also quickly ran out of space with 8 GB of state per version. Tribler has become too fat to run normally for me and I need to manually remove databases to run it 😞.

drew2a:
I see only one reason to keep the old state directory: in case of migration failure, we can rollback to the previous version manually.

[...]

PR #6017, opened 2021-03-01
Add support for removing old state directories on upgrade

xoriole:
This PR
1. Asks user whether to clean up old state directories on upgrade
settings-data-0

2. Adds a `DATA` tab in settings which allows users to selectively select and delete the old state directories.
settings-data-2


xoriole:
retest this please

ichorid (changes requested):
When the "should remove old dirs?" dialog window is shown, if the user waits (thinks) too long and then clicks "No", the GUI never switches from the "Loading" screen. This is reproducible and is probably a race condition between the GUI receiving the "Core started" signal and the dialog window. Also, showing the dialog window _before_ Tribler GUI finished loading falsely indicates to the user that something is wrong.
Please, only show the window **after** the GUI completely loaded.

Also, please squash the commits :wink:

src/tribler-gui/tribler_gui/widgets/settingspage.py
```diff
@@ -223,6 +233,78 @@ def initialize_with_settings(self, settings):
...
+
+    def on_confirm_remove_version_dirs(self, selected_versions):
+        message_box = QMessageBox()
+        message_box.setIcon(QMessageBox.Question)
+
+        if selected_versions:
+            version_dirs_str = "\n- ".join(selected_versions)
+            versions_info = f"Versions: \n- {version_dirs_str}"
+
+            title = "Confirm delete older versions?"
```
ichorid, line 291: Please, change these to
"Really delete old state directories?"

src/tribler-gui/tribler_gui/widgets/settingspage.py
```diff
@@ -223,6 +233,78 @@ def initialize_with_settings(self, settings):
...
+            message_body = f"Are you sure to remove the selected versions? " \
```
ichorid, line 292: Better to
"Are you sure you want to remove the state directories for selected versions?"

src/tribler-core/tribler_core/upgrade/tests/test_version_manager.py
```diff
@@ -207,3 +212,98 @@ def test_copy_state_directory(tmpdir):
...
+
+                # Create an empty version directory if does not exist
+                (root_state_dir / version_dir).mkdir(exist_ok=True)
+
+    # Write the version history file before checking disposable directories
+    (root_state_dir / VERSION_HISTORY_FILE).write_text(json.dumps(version_history))
+
+    # Case 0: If the code version and last version is the same, no disposable directories shown at startup
+    code_version = last_version  # 8.9.2
+    disposable_dirs = get_disposable_state_directories(root_state_dir, code_version, skip_last_version=True)
```
ichorid, line 255: the output of `get_disposable_state_directories` should be always the same, regardless of whether the code version is there or not. Otherwise, you're mixing _decision_ logic with _analysis_ logic.

Examples: (code version is 8.9.2)
```
8.7
8.8
8.9
disposable_dirs -> [8.7]
```

```
8.7
8.8
disposable_dirs -> []
```

```
8.3
8.7
8.8
disposable_dirs -> []
```

src/tribler-core/tribler_core/upgrade/tests/test_version_manager.py
```diff
@@ -207,3 +212,98 @@ def test_copy_state_directory(tmpdir):
...
+    assert disposable_dirs is None
+
+    # Now, assuming the new code version is a newer major version
+    code_version = f"{major_versions[0] + 1}.{minor_versions[0]}.{patch_versions[0]}"  # 9.0.0
+
+    # Case 1: Skip last version is True, then those two last directories will not returned as disposable dirs.
+    disposable_dirs = get_disposable_state_directories(root_state_dir, code_version, skip_last_version=True)
```
ichorid, line 262: Again, please don't mix *decision* logic with *analysis* logic

src/tribler-core/tribler_core/upgrade/tests/test_version_manager.py
```diff
@@ -207,3 +212,98 @@ def test_copy_state_directory(tmpdir):
...
+    assert last_version_dir not in disposable_dirs
+    assert second_last_version_dir not in disposable_dirs
+
+    # Case 2: Skip last version is False, then only the upgrade version is kept and not the one before that.
+    disposable_dirs = get_disposable_state_directories(root_state_dir, code_version, skip_last_version=False)
+    assert last_version_dir not in disposable_dirs
+    assert second_last_version_dir in disposable_dirs
+
+
+def test_installed_versions_and_removal(tmpdir_factory):
```
ichorid, line 272: Sorry, I don't get what this test is intended to test from the name :face_with_head_bandage:

src/tribler-core/tribler_core/upgrade/tests/test_version_manager.py
```diff
@@ -207,3 +212,98 @@ def test_copy_state_directory(tmpdir):
...
+    assert len(installed_versions) == len(major_versions) * len(minor_versions) + 1  # including the current version
+
+    # 2. exclude current version
+    installed_versions = get_installed_versions(root_state_dir, current_version=False)
+    assert current_version_dir not in installed_versions
+    assert len(installed_versions) == len(major_versions) * len(minor_versions)  # the current version not included
+
+    # 3. Skip a few other versions
+    skip_versions = ['7.5', '7.6']
+    installed_versions = get_installed_versions(root_state_dir, current_version=False, skip_versions=skip_versions)
```
ichorid, line 300: Again, mixing decision logic with analytical logic

src/tribler-core/tribler_core/upgrade/version_manager.py
```diff
@@ -189,3 +190,52 @@ def fork_state_directory_if_necessary(root_state_dir, code_version):
...
+    version_history, _, src_dir, tgt_dir, last_version = should_fork_state_directory(root_state_dir, code_version)
+
+    if src_dir is not None:
+        # If versioned state directory exists
+        if src_dir != root_state_dir:
+            skip_dirs = [src_dir, tgt_dir]
+            skip_versions = [last_version]
+
+            # Get the second last version if available and add it to the dirs to save
+            second_last_version = version_history.get_last_upgradable_version(root_state_dir, last_version)
```
ichorid, line 205: Why do this `skip_versions` hack instead of just forming a full list of versions and using Python slicing to select every one but the last two?

ichorid (changes requested):

src/tribler-core/tribler_core/upgrade/version_manager.py
xoriole: 1. `skip_versions` is used to give flexibility to choose different version from Settings/Data page.
2. If only last two version are selected, it does not clean up non-version directories eg. pre 7.4 directories
3. Full list of version should be sorted separately in the order of installation based on version history, additional logic

xoriole:
@ichorid 
>When the "should remove old dirs?" dialog window is shown, if the user waits (thinks) too long and then clicks "No", the GUI never switches from the "Loading" screen. This is reproducible and is probably a race condition between the GUI receiving the "Core started" signal and the dialog window. Also, showing the dialog window before Tribler GUI finished loading falsely indicates to the user that something is wrong.
Please, only show the window after the GUI completely loaded.

It is a good point you pointed out that the GUI could get stuck in the loading screen if user waits too long since the core can timeouts by then. That can be solved by increasing the timeout or pausing it when the dialog is shown. But, showing the dialog only after fully loaded defeats the purpose of the PR to solve the issue with not sufficient storage to upgrade. https://github.com/Tribler/tribler/issues/5907

> Also, please squash the commits wink

Yes, was planning after the review

src/tribler-core/tribler_core/upgrade/version_manager.py
ichorid: I would expect _choosing_ the version to delete from the Data page to be completely independent from the process _listing_ the versions to choose from

>Full list of version should be sorted separately in the order of installation based on version history, additional logic

I was not clear enough in my request, my fault. By "the full list of versions" I mean "the ordered list of versions". By "ordered" I mean "ordered by last usage time, last usage goes last".

So, the expected logic is:
1. Get the list of state directories from the directory listing A)
2. Get the ordered list of versions from the version history file (B) 
3. Filter list B to only contain entries from A
4. Get `B[:-2]` and show it to the user. Let the user form the list C of versions to delete.
5. Delete every version in list C.

ichorid:
> It is a good point you pointed out that the GUI could get stuck in the loading screen if user waits too long since the core can timeouts by then. That can be solved by increasing the timeout or pausing it when the dialog is shown. But, showing the dialog only after fully loaded defeats the purpose of the PR to solve the issue with not sufficient storage to upgrade. #5907

Increasing the timeout is a bad idea. Pausing will work, though. The Core should not be even run before the dialog finishes.

src/tribler-gui/tribler_gui/widgets/settingspage.py
xoriole: I would like to avoid mentioning `state directories`. I think `version` is more understandable to users than `state directories`.

src/tribler-gui/tribler_gui/widgets/settingspage.py
```diff
@@ -1,16 +1,20 @@
-from PyQt5.QtWidgets import QFileDialog, QSizePolicy, QWidget
+from PyQt5.QtWidgets import QCheckBox, QFileDialog, QMessageBox, QSizePolicy, QWidget
 
 from tribler_common.sentry_reporter.sentry_mixin import AddBreadcrumbOnShowMixin
 from tribler_common.simpledefs import MAX_LIBTORRENT_RATE_LIMIT
 
 import tribler_core.utilities.json_util as json
+from tribler_core.upgrade.version_manager import get_installed_versions, get_versioned_state_directory, \
```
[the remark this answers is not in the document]
xoriole: I fully agree, but I consider that to be a separate PR and not mix with the original reason for this PR.

src/tribler-gui/tribler_gui/core_manager.py
```diff
@@ -87,20 +89,65 @@ def start(self, core_args=None, core_env=None):
...
+        storage_info = ""
+        claimable_storage = 0
+        for old_state_dir in disposable_dirs:
+            dir_size = get_dir_size(old_state_dir)
+            claimable_storage += dir_size
+            storage_info += f"{format_size(dir_size)} \t {relpath(old_state_dir, root_state_dir)}\n"
+
+        # Show a question to the user asking if the user wants to remove the old data.
+        title = "Delete older version?"
+        message_body = f"Press 'Yes' to remove data of older versions " \
```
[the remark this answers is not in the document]
xoriole: This is better :+1:

src/tribler-gui/tribler_gui/core_manager.py
```diff
@@ -87,20 +89,65 @@ def start(self, core_args=None, core_env=None):
...
             self.use_existing_core = False
             self.start_tribler_core(core_args=core_args, core_env=core_env)
 
+        do_cleanup, old_version_dirs = self.should_cleanup_old_versions(root_state_dir, version_id)
+        if do_cleanup:
+            for version_dir in old_version_dirs:
+                shutil.rmtree(str(version_dir), ignore_errors=True)
+
+        # Connect to the events manager only after the cleanup is done
         self.events_manager.connect()
```
xoriole, line 104: Trying to connect the core after the cleanup is done should address the issue about GUI being stuck if user waits long. There is no need to increase the timeout or pause the connection attempts with this.

src/tribler-gui/tribler_gui/widgets/settingspage.py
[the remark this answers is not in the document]
ichorid: Ok

src/tribler-gui/tribler_gui/widgets/settingspage.py
ichorid: "Version" and "state directories" are two different things. I propose asking the guys in the dev chat :wink:

src/tribler-core/tribler_core/upgrade/version_manager.py
kozlovsky: I'm not sure that the `B[:-2]` logic ("we should suggest to delete everything except two last directories") is correct. If we currently have versions:
- 8.3
- 8.4
- **8.5 (currently running version)**
- 8.6
- 8.7
what should we delete? In my opinion, definitely not `[8.3, 8.4, 8.5]`. I think the correct answer is `[8.3, 8.6, 8.7]`.

src/tribler-gui/tribler_gui/widgets/settingspage.py
kozlovsky: To me "version" means binary files of a program, and a suggestion to delete old version I'd probably interpreted as a suggestion to delete old exe/dmg files.

As an alternative to "state directories for selected version" we can probably use "data files for selected version"

kozlovsky:
Screenshot 2021-03-08 at 10 02 19

1) Currently, the code suggests deletion of every unknown folder within the root directory. You can see on the screenshot some folders with underscores. These are my "backups" of old versions. Also, you can see three directories for old versions of Tribler, in which the `.Tribler` folder itself was the data folder: `sqlite`, `channels`, and `dlcheckpoints`. Also, there is a folder `"tunnel--1"` for tunnel helpers. I don't want to remove any of them except three folders from the ancient version. So, currently, the code suggests deleting every unknown folder inside `.Tribler` except several known folders (for the current version of Tribler and two previous versions). Probably instead it should only suggest deletion of known folders of previous versions. That is, instead of searching the root directory for all folders, it should construct folder names that are candidates for deletion from the version history.

2) I think it is better to swap buttons in the dialog to have the default (the safest) button on the right.

3) As choosing "Yes" lead to data deletion, it is probably better to ask for a second confirmation, like "Press Continue if you really want to delete all these old directories".

Regarding the code itself, it looks good enough for me. I think the code logic may be significantly simplified by introducing TriblerVersion class which loaded from the version history and combines version name, version number, version directory, and some methods like `directory_exists` and `delete`. But this refactoring can be done in a separate PR.

kozlovsky:
retest this please

kozlovsky:
retest this please

kozlovsky:
Retest this please

ichorid (approved):

kozlovsky (approved):

PR #6018, opened 2021-03-03
Enriched log entry with peer info

devos50:
Hopefully this will help to debug a bug where there is a ping-pong of bandwidth transactions between two peers.


devos50:
Seems that the deepcode analyser has a day off but this PR should be ready for review 👍

ichorid (approved):

Issue #6003, opened 2021-02-19
Transport protocol over IPv8


ichorid:
Further development of Tribler Channels features requires adding a way to quickly and reliably transport small files over the network. Libtorrent does not fit this role for too many reasons.
The biggest problem is that we're forced to build the thing atop IPv8 addressing system, which denies us reusing every other non-pure-Python protocol library in existence.

* Sans protocol QUIC implementation is ruled out as too complex.
* No proper TCP implementations in pure Python.

One alternative way is to use IPv8 to coordinate NAT puncture for something like [UDT](https://en.wikipedia.org/wiki/UDP-based_Data_Transfer_Protocol)

My current attempt at solving the problem: #6002 

Some useful links:
* https://github.com/vmarkovtsev/udt4py
* https://github.com/delattj/py_udt

qstokkink:
Short recap of "offline" discussion on using UDT and hooking in external transport libraries into IPv8:

 - Using UDT puncturing is unnecessary, as IPv8 can give you an already punctured port (which it gives you the WAN address estimate for to connect to it).
 - Ideally, just use the IPV8 port with your transport protocol.
 - Less ideally, use IPv8 to puncture a port with `socket.so_reuseaddr` and hand it off to the transport protocol library.
 - Equally less ideally, wrap the external socket in an IPv8 `Endpoint` object and make it available in the `DispatcherEndpoint` mapping.
 - In the worst case: use local socket loopback to feed the data back into your own process.

ichorid:
>Ideally, just use the IPV8 port with your transport protocol.

I still don't get it, what will process the packets in this case: IPv8 or my protocol library? How would they know which packet should be processed by which socket listener? :shrug:

qstokkink:
> IPv8 or my protocol library? 

Both. For receiving, IPv8 first strips its header information from the packet and then the library unserializes the remainder. For sending, IPv8 would add its extra header info to the serialized packet from the external library.

> How would they know which packet should be processed by which socket listener?

If properly layered this should be completely opaque: whether the packet comes from the Internet or not _should_ be a matter of hooking up the streams correctly. For example, if your library of choice supports asyncio transports, you can easily reroute this by overwriting the default transport:

![rerouteexample](https://user-images.githubusercontent.com/3630389/108618698-7378bc80-7420-11eb-8418-dd511cad1721.jpg)

PR #6021, opened 2021-03-03
Binary data transfer protocol over ipv8.

drew2a:
This PR makes it possible to transfer big binary data over ipv8.
Closes https://github.com/Tribler/tribler/issues/6003

The protocol based on TFTP with windowsize (RFC 7440).

Features:

 * timeout
 * retransmit
 * dynamic window size

The maximum data size that can be transferred through the protocol can be calculated as "block_size * 4294967295" where 4294967295 is the max segment number (4B unsigned int). But the protocol is slow by design, so try to send less rather than more.

### How to use:

```python
class MyCommunity(EVAProtocolMixin, Community):
    community_id = b'0' * 20

        self.eva_init()

        self.eva_register_receive_callback(self.on_receive)
        self.eva_register_send_complete_callback(self.on_send_complete)
        self.eva_register_error_callback(self.on_error)

    def my_function(self, peer):
        self.eva_send_binary(peer, b'info1', b'data1', 42)
        self.eva_send_binary(peer, b'info2', b'data2', 42)
        self.eva_send_binary(peer, b'info3', b'data3', 42)

   def on_receive(self, peer, binary_info, binary_data, nonce):
        logger.info(f'Data has been received: {binary_info}')

   def on_send_complete(self, peer, binary_info, binary_data, nonce):
        logger.info(f'Transfer has been completed: {binary_info}')

   def on_error(self, peer, exception, nonce):
        logger.error(f'Error has been occurred: {exception}')
```

Due to ipv8 specifics, we can use only one socket port per one peer. Therefore, at one point in time, the protocol can only transmit one particular piece of data for one particular peer.

In case the `eva_send_binary` method is invoked multiply times for a single peer, the data transfer will be scheduled and performed when the current transfer session is finished.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/eva_protocol.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/eva_protocol.py
@@ -1,0 +1,564 @@
+# EVA protocol: a protocol for transferring big binary data over ipv8.
+#
+# Limitations and other useful information described in the corresponding class.
+# Example of use:
+#
+# class MyCommunity(EVAProtocolMixin, Community):
+#     community_id = os.urandom(20)
+#
+#     def __init__(self, *args, **kwargs):
+#         super().__init__(*args, **kwargs)
+#         self.eva_init()
+#
+#         self.eva_register_receive_callback(self.on_receive)
+#         self.eva_register_send_complete_callback(self.on_send_complete)
+#         self.eva_register_error_callback(self.on_error)
+#
+#     def my_function(self, peer):
+#         self.eva_send_message(peer, b'info1', b'data1')
+#         self.eva_send_message(peer, b'info2', b'data2')
+#         self.eva_send_message(peer, b'info3', b'data3')
+#
+#     def on_receive(self, peer, binary_info, binary_data, nonce):
+#         logger.info(f'Data has been received: {binary_info}')
+#
+#     def on_send_complete(self, peer, binary_info, binary_data, nonce):
+#         logger.info(f'Transfer has been completed: {binary_info}')
+#
+#     def on_error(self, peer, exception):
+#         logger.error(f'Error has been occurred: {exception}')
+
+import logging
+import math
+import time
+from collections import defaultdict, deque
+from enum import Enum, auto
+from types import SimpleNamespace
+
+from ipv8.lazy_community import lazy_wrapper
+from ipv8.messaging.lazy_payload import VariablePayload, vp_compile
+
+logger = logging.getLogger('EVA')
+
+
+# fmt: off
+
+@vp_compile
+class WriteRequest(VariablePayload):
+    format_list = ['I', 'I', 'raw']
+    names = ['data_size', 'nonce', 'info_binary']
+
+
+@vp_compile
+class Acknowledgement(VariablePayload):
+    format_list = ['I', 'I', 'I']
+    names = ['number', 'window_size', 'nonce']
+
+
+@vp_compile
+class Data(VariablePayload):
+    format_list = ['I', 'I', 'raw']
+    names = ['block_number', 'nonce', 'data_binary']
+
+
+@vp_compile
+class Error(VariablePayload):
+    format_list = ['raw']
+    names = ['message']
+
+
+class TransferException(Exception):
+    def __init__(self, transfer_type, info_binary, nonce, message):
+        super().__init__(message)
+        self.transfer_type = transfer_type
+        self.info_binary = info_binary
+        self.nonce = nonce
+
+
+class SizeLimitException(TransferException):
+    pass
+
+
+class TimeoutException(TransferException):
+    pass
+
+
+class EVAProtocolMixin:
+    """This mixin makes it possible to transfer big binary data over ipv8.
+
+    The protocol based on TFTP with windowsize (RFC 7440).
+    Features:
+        * timeout
+        * retransmit
+        * dynamic window size
+
+    The maximum data size that can be transferred through the protocol can be
+    calculated as "block_size * 4294967295" where 4294967295 is the max segment
+    number (4B unsigned int).
+    """
+
+    def eva_init(  # pylint: disable=too-many-arguments
+            self,
+            block_size=1000,
+            window_size_in_blocks=16,
+            start_message_id=186,
+            retransmit_interval_in_sec=3,
+            retransmit_attempt_count=3,
+            timeout_interval_in_sec=10,
+            binary_size_limit=1024 * 1024 * 1024,
+    ):
+        """Init should be called manually within his parent class.
+
+        Args:
+            block_size: a single block size in bytes. Please keep in mind that
+                ipv8 adds approx. 177 bytes to each packet.
+            window_size_in_blocks: size of consecutive blocks to send
+            start_message_id: a started id that will be used to assigning
+                protocol's messages ids
+            retransmit_interval_in_sec: an interval until the next attempt
+                to retransmit will be made
+            retransmit_attempt_count: a limit for retransmit attempts
+            timeout_interval_in_sec: an interval after which the transfer will
+                be considered as "dead" and will be terminated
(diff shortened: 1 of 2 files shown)


src/tribler-core/tribler_core/modules/metadata_store/community/eva_protocol.py
```diff
@@ -0,0 +1,502 @@
...
+        )
+
+        # specify messages' ids
+        _assign_id_to_messages(start_message_id, [WriteRequest, Acknowledgement, Data, Error])
+
+        # add messages' handlers
+        self.add_message_handler(WriteRequest, self.on_eva_write_request)
+        self.add_message_handler(Acknowledgement, self.on_eva_acknowledgement)
+        self.add_message_handler(Data, self.on_eva_data)
+        self.add_message_handler(Error, self.on_eva_error)
```
qstokkink, line 203: Quick note: you don't need to change the `msg_id` of the class instances (which could screw with other community instances doing the same). You can specify the handlers by message number instead:

```suggestion
        # add messages' handlers
        self.add_message_handler(start_message_id, self.on_eva_write_request)
        self.add_message_handler(start_message_id + 1, self.on_eva_acknowledgement)
        self.add_message_handler(start_message_id + 2, self.on_eva_data)
        self.add_message_handler(start_message_id + 3, self.on_eva_error)
```

qstokkink: My bad, forgot to mention that you're not allowed to use `self.ez_send()` anymore then.

```python
self.ez_send(peer, Acknowledgement(0, transfer.nonce))
```
⬇️ (with `self.ack_msg_id = start_message_id + 1`)
```python
self.endpoint.send(peer.address, self.ezr_pack(self.ack_msg_id, Acknowledgement(0, transfer.nonce)))
```

drew2a: I see. 
I'll rewrite all the occurrences because probably it seems like a single correct way to implement a dynamic message-id assignation (in case this mixin can be used by multiple classes in the future).

qstokkink:
Some further high-level comments while this is still in WIP:

 - Using `asyncio.create_task` is dangerous. This (a) schedules a task outside of the IPv8 error handler and consequently (b) doesn't cancel tasks when the Community is unloaded (calling dead objects usually leads to weird stuff on shut down). You're encouraged to use `self.register_anonymous_task` instead to bind the task to the Community.

 - Binding to `self.community*` and `self.peer*` in the tests is unnecessary. For this the `self.overlay(*)` and `self.peer(*)` exist. On another note, if you use `my_peer` from a community you're using the `Peer` instance that has a private key instead of a public key, which can lead to issues (`self.peer(*)` provides the public key view of a peer).

 - TestBase instantiates a new Community instance per test run. There is no need to remember the overwritten instance methods and restore them in a `finally` clause (for example the callbacks in `decode_map`). Also there doesn't seem to be a need to call `clear()` for the same reason.

 - Storing states in `dict`s almost always leads to memory leaks. You're encouraged to use [RequestCache](https://py-ipv8.readthedocs.io/en/latest/basics/requestcache_tutorial/#the-easier-way)s to keep track of states. These automatically time out and IPv8 can retrieve them in automatically in your message flow using the decorator. This also downsizes your code and makes it less complex.

 - I'd encourage the use of `asyncio` `Future`s for your calls. This allows external code to, for example, `await eva_send_binary()`. This will also downsize your tests. The `RequestCache` `Caches` have the `register_future()` to make sure `Future`s are also properly canceled when the overlay is destroyed.

 - (minor) `drain_loop` seems functionally equivalent to `self.deliver_messages(0.0)`.

drew2a:
@synctext yes, sure. Tests already there: https://github.com/Tribler/tribler/pull/6021/files#diff-0e1f354d589644baf35d810ec6117405b591e2428200678ac38fd8abab1b23a6R319-R377

drew2a:
@qstokkink thank you for the comments. 
I will work on them today.

>I'd encourage the use of asyncio Futures for your calls. This allows external code to, for example, await eva_send_binary(). This will also downsize your tests. The RequestCache Caches have the register_future() to make sure Futures are also properly canceled when the overlay is destroyed.

My original version of the protocol's interface was with `async` functionality. But @ichorid strongly asked me to rewrite it with normal calls.

I'm more in favor of the `async` version, BTW.

drew2a:
@qstokkink thanks for the comments.

> Using asyncio.create_task is dangerous. This (a) schedules a task outside of the IPv8 error handler and consequently (b) doesn't cancel tasks when the Community is unloaded (calling dead objects usually leads to weird stuff on shut down). You're encouraged to use self.register_anonymous_task instead to bind the task to the Community.

Implemented.

> Binding to self.community* and self.peer* in the tests is unnecessary. For this the self.overlay(*) and self.peer(*) exist. On another note, if you use my_peer from a community you're using the Peer instance that has a private key instead of a public key, which can lead to issues (self.peer(*) provides the public key view of a peer).

Implemented

> TestBase instantiates a new Community instance per test run. There is no need to remember the overwritten instance methods and restore them in a finally clause (for example the callbacks in decode_map). Also there doesn't seem to be a need to call clear() for the same reason.

Implemented

> Storing states in dicts almost always leads to memory leaks. You're encouraged to use RequestCaches to keep track of states. These automatically time out and IPv8 can retrieve them automatically in your message flow using the decorator. This also downsizes your code and makes it less complex.

About this. At the moment I continue to use dicts because I don't see the real need to change them to RequestCaches.
But it is still an open question.

qstokkink:
Cool 👍 

> At the moment I continue to use dicts because I don't see the real need to change them to RequestCaches.

Well you don't strictly _need_ them. They are just an easier way to store states of message flows and deal with all of the possible unhappy flows that can cause data to stick around in your dicts.

Without RequestCaches you'll have to manually make sure your dicts are cleaned up if any response message does not arrive (or crashes mid-handler or whatever). Historically, we've not been very good at this. We've spent quite some time digging through memory dumps (which is not fun).

qstokkink (approved):
I already gave my high-level comments and apart from that I only have some minor comments left that you could consider for future work/PRs:

 - The order in which `_eva_register_message_handler` is called defines the message wire format. I'd add a comment here so a future developer doesn't order the handlers alphabetically.
 - The docstring examples don't follow RST styling and won't be rendered correctly.
 - 10~20 second unit tests are not ideal.
 - I'd prefer `logging.info()` over `print()` in the tests (which is the recommended style approach by most code style tools). This also gets rid of the `# noqa: T001`.

I don't see any game-breakers and in my opinion this PR is therefore good to merge.

drew2a:
@qstokkink thank you 🙏 

> The order in which _eva_register_message_handler is called defines the message wire format. I'd add a comment here so a future developer doesn't order the handlers alphabetically.

Implemented

> The docstring examples don't follow RST styling and won't be rendered correctly.

Correct. But there is no need to render them (am I wrong?). 

> 10~20 second unit tests are not ideal.

Correct, but all tests here are not unit tests (it is also correct for almost all descendants of TestBase). 
Anyway, the tests should last as little time as possible.
I decrease the default time for timeouts and retransmit. It should decrease the execution time of the tests.

> I'd prefer logging.info() over print() in the tests (which is the recommended style approach by most code style tools). This also gets rid of the # noqa: T001.

Implemented.

Once again, thank you for the feedback.

qstokkink:
Nice, I'll re-approve momentarily.

>> The docstring examples don't follow RST styling and won't be rendered correctly.
>
> Correct. But there is no need to render them (am I wrong?).

If you press Ctrl+Q in PyCharm you get the RST rendering. Also, we used to automatically generate the documentation from the classes for ReadTheDocs.. however: https://tribler.readthedocs.io/en/latest/py-modindex.html kinda broken apparently 😃

qstokkink (approved):

ichorid (approved):
After three iterations, almost perfect :+1:

PR #6030, opened 2021-03-18
Fix popularity endpoint to limit selection by the last 24hrs

ichorid:
Make popular endpoint return torrent checked in the last 24 hrs.


drew2a (dismissed):

drew2a:
Don't forget to solve all issues from the checker (pylint, coverage, etc)

drew2a (approved):

PR #6034, opened 2021-03-26
Add tracking of peers subscribed to channels

ichorid:
This PR adds tracking of the last 10 peers who introduced us to a channel, for each channel. If a peer does not respond to our queries, it is dropped. Also, it adds a convenient debug endpoint and a corresponding Channels debug tab.

This is another step towards live previews in Channels 2.5

изображение

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
@@ -1,9 +1,12 @@
+import time as unixtime
 import uuid
 from binascii import unhexlify
+from collections import defaultdict, deque
 from dataclasses import dataclass
 from random import sample
 
 from ipv8.peerdiscovery.network import Network
+from ipv8.types import Peer
 
 from pony.orm import db_session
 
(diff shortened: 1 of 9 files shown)


ichorid:
retest this please

ichorid:
retest this please

drew2a (changes requested):
Can't approve: tests are falling.

ichorid:
> Can't approve: tests are falling.

Yep. We have to first upgrade our Linux testers to 3.8.

ichorid:
retest this please

ichorid:
retest this please

ichorid:
retest this please

ichorid:
retest this please

drew2a (changes requested):

.flake8
```diff
@@ -1,5 +1,6 @@
 [flake8]
 jobs = auto 
 max-line-length = 120
-select=E901,E999,F821,F822,F823,I100,I101,I201,I202
+select=E901,E999,F821,F822,F823,I100,I101,I201,I202,T001,T002,T003,T004
```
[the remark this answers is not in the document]
ichorid: Done. I'll have to rebase the thing again though \*sigh\*

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
```diff
@@ -117,11 +164,32 @@ def notify_gui(_, processing_results):
...
+            )
+        else:
+            peers_to_query = self.get_random_peers(self.settings.max_query_peers)
+
+        for p in peers_to_query:
             self.send_remote_select(p, **kwargs, processing_callback=notify_gui)
 
         return request_uuid
 
+    def get_known_subscribed_peers_for_node(self, node_pk, node_id, limit=None):
```
drew2a, line 180: **improvement:** seems that you go to the database for each request.

What do you think about caching this information to prevent frequent access to the database?

drew2a (changes requested):

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
ichorid: Caching these results can break the logic in some cases, such as when the channel author moves some intermediary folders to another channel. Or when our instance fails to find the root node correctly the first time (e.g. after a search), but then receives the missing ancestor nodes. Unless we build a subsystem for caching the stuff into the MetadataStore itself, along with a way to **invalidate** the caching results, but that's worth a different PR, with an extensive test suite. Another way to solve it would be to add transitive closures to node entries.

That said, it should be harmless in 99.99% of cases to cache the thing. My only real concern is how to indicate to the programmers (basically, *us*) in the future, that the thing is unreliable in some corner cases? Argument flags? Big warning signs?

I would like to hear your suggestions.

ichorid:
retest this please

drew2a (approved):

PR #6045, opened 2021-03-31
Fix for DHTHealthManager returning incorrect results

egbertbouman:
This PR fixes an issue with increased peer counts being returned by the `DHTHealthManager`. The problem is that when we send BEP33 responses for active downloads, they are picked up by the `DHTHealthManager` and added to the nearest infohash. A similar issue could present when we receive BEP33 responses for active downloads.

To fix the issue we keep track of transaction_id -> infohash. Unfortunately, this means that we have to process all incoming and outgoing DHT messages. Note that I'm assuming that transaction_ids for in-flight requests are unique.


egbertbouman:
retest this please

ichorid (approved):

PR #6046, opened 2021-04-01
Fix test_dynamically_changed_window_size

drew2a:
This PR fixes sporadic test fails caused by not necessary assertion.
This assertion doesn't work properly for some cases for the last packet. 
Seems like a "naive test design issue".


xoriole (approved):

xoriole:
Seems there are false negatives on some of the test jobs

PR #6048, opened 2021-04-06
Fix test_get_first_free_port

### LINKED ISSUE #6042: test_get_first_free_port fails randomly
On `main`. See [this test run](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux_pytest/1159/testReport/junit/src.tribler-core.tribler_core.tests/test_network_utils/test_get_first_free_port/), for instance

```python
Error Message

assert 50001 == 50000   +50001   -50000

Stacktrace

def test_get_first_free_port():
        # target port is free
>       assert get_first_free_port(start=50000) == 50000
E       assert 50001 == 50000
E         +50001
E         -50000
```

drew2a:
This PR fixes #6042.

I was not able to reproduce the actual issue.
My guess is the cause of the issue was simultaneously running tests. 

The fix is simple: to use a randomly picked port instead of a fixed one.

### DIFF
--- a/src/tribler-core/tribler_core/tests/test_network_utils.py
+++ b/src/tribler-core/tribler_core/tests/test_network_utils.py
@@ -52,13 +52,14 @@ def test_autodetect_socket_style():
 
 def test_get_first_free_port():
     # target port is free
-    assert get_first_free_port(start=50000) == 50000
+    port = random.randint(50000, 60000)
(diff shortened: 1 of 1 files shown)


ichorid (approved):
Should nicely decrease the chance for random fail :+1:  :four_leaf_clover:

Issue #5972, opened 2021-02-01
Suggestions for Tribler UI


TriblerFan:
Here are a few options that should be added to Tribler which are common to almost all other torrent clients.

    1. Option to start Tribler on Windows login.
    2. Option to start Tribler minimized.
    3. Option to minimize to taskbar OR minimize to tray. (currently Tribler minimizes to both locations at the same time)
    4. Option when clicking X to close Tribler OR minimize instead. (To prevent from closing Tribler by accident.)
    5. Under the “Files”, “Trackers”, “Peers” tabs we should be able to sort by the column of our choosing.
    6. Under the “Trackers” tab we should be able to add or remove trackers from any torrent.
    
Of course there are many more UI tweaks possible but as I mentioned these 6 options are already available in all torrent clients I have tested. Thanks for reading.

synctext:
thnx!
Great list of Quality of Life improvements in Tribler, very handy.

kotenok2000:
Can you also add an option to choose database directory?
I don't want to fill my ssd

PR #6086, opened 2021-04-29
READY: Add GUI options to minimize to tray

### LINKED ISSUE #1054: option to close to tray
As a user I expect Tribler to close to tray (just like uTorrent). I would like to have this option available in the GUI. Currently it is only possible to minimize to tray, which feels unnatural to me (after all these years). I don't ever want to close the application of course. I just want it to work in the background.

### LINKED ISSUE #3631: Small UI Requests: Close to Tray
- [Y ] I have read CONTRIBUTING.rst

Unknown (Windows binary download)

Windows 10

##### REQUEST #####

Could we add two things to the settings menu (gear icon):

1. 'About' tab with version number, info etc
1. Checkbox to change "Close to Tray" vs "Close to Exit/End Process" behavior

Thanks

ichorid:
This adds an option in Settings to minimize Tribler GUI to system tray. Users asked for it repeatedly over the years:
fixes #3631
fixes #1054 

related to #5972 

recent [forum post](https://forum.tribler.org/t/minimize-to-system-tray/5047/7)

p.s.
for some reason the newly adde checkbox looks different than the older checkboxes. I looked basically everywhere, but AFAICT there is no difference in definition of the new and old checkboxes :face_with_head_bandage: 

@devos50 , @xoriole , maybe you guys know something about this checkbox mystery?

Also, dear Mac :apple: users, could you please test it out? I only tested it on Windows and on Linux, works as expected.

[Qt reference for TrayIcon](https://doc.qt.io/qt-5/qsystemtrayicon.html#details)

EDIT:
I guess it [makes not sense](https://stackoverflow.com/questions/44286102/minimize-to-tray-on-macosx) to talk about minimizing to tray on Mac.

EDIT:
Checkboxes mystery solved (after about an hour of experiments with .ui files)! The problem was that https://github.com/Tribler/tribler/commit/f5a8474a2de6d0ef54b36b9bd2c2803580a60f94 was setting checkboxes' style sheets _individually_ by a list of widgets.

Guys, please! There are three "don'ts" in QT design:
1. **Don't set CSS in the code**, set it in the widget `.ui`-file
2. **Don't set it on individual widgets**, set it on the highest parent widget, which is the main Tribler window in our case. CSS will do the rest
3.  Don't copy-paste the stylesheets, try to move as much stuff as possible to the highest possible parent widget and subclass the widgets in Qt Creator.

### DIFF
--- a/src/tribler-gui/tribler_gui/defs.py
+++ b/src/tribler-gui/tribler_gui/defs.py
@@ -1,6 +1,7 @@
 """
 This file contains various definitions used by the Tribler GUI.
 """
+import sys
 from collections import namedtuple
 
 DEFAULT_API_PROTOCOL = "http"
@@ -222,3 +223,6 @@ MB = 1024 * KB
 GB = 1024 * MB
 TB = 1024 * GB
 PB = 1024 * TB
+
+DARWIN = sys.platform == 'darwin'
+WINDOWS = sys.platform == 'win32'
--- a/src/tribler-gui/tribler_gui/dialogs/startdownloaddialog.py
+++ b/src/tribler-gui/tribler_gui/dialogs/startdownloaddialog.py
@@ -16,7 +16,6 @@ from tribler_gui.tribler_request_manager import TriblerNetworkRequest
 from tribler_gui.utilities import (
     connect,
     format_size,
-    get_checkbox_style,
     get_gui_setting,
     get_image_path,
     get_ui_file_path,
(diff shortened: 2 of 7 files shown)


devos50:
Doesn't work on Mac (11.0.1). Even with the 'miminize to tray' checkbox added, Tribler closes directly.

> for some reason the newly adde checkbox looks different than the older checkboxes.

I remember that there are visual differences, even between Tribler in frozen environment and when running it from source. I suspect that Pyinstaller is doing some magic and stripping out stuff that affects how checkboxes are being displayed. We might want to investigate this at some point, but in general, Mac makes it very hard to customise checkboxes (even in webkit).

ichorid:
retest this please

kozlovsky (approved):

devos50:
@ichorid It seems that this PR is already merged, but I looked a bit further into the applications on Mac and how I'm interacting with them. For most applications that I'm regularly using, clicking the close button (the left-most button in the upper-left side of the window) closes the current window but the app itself remains open (as evident by a small dot below the app icon in my dock). I can stop any application by right-clicking the app icon and by selecting 'stop' in the context menu. In Tribler, however, pressing the close button completely stops the application.

But that Tribler exhibits this behaviour when clicking the close button is actually something I never realised. Changing this could help with user retention though since users still run Tribler when they close the window (while their downloads progress etc).

Also, could you please add these "Qt don't" to the contribution guidelines?

PR #6094, opened 2021-05-03
Ignore PackErrors in Socks5Client

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/6082.


ichorid (dismissed):

egbertbouman:
@ichorid I've removed the `except` from the `sendto` function because it's probably not a good idea to ignore exceptions that occur during encoding. If you agree, could you please re-approve?

ichorid (approved):

Issue #5582, opened 2020-09-25
test_hidden_services still failing occasionally


devos50:
```
=================================== FAILURES ===================================
_____________________________ test_hidden_services _____________________________
[gw5] linux -- Python 3.7.3 /usr/bin/python3

enable_ipv8 = None
proxy_factory = <test_full_session.test_tunnel_community.ProxyFactory object at 0x7f103aec6a90>
session = <tribler_core.session.Session object at 0x7f103aec69b0>
hidden_seeder_session = <tribler_core.session.Session object at 0x7f103b168390>
video_tdef = <tribler_core.modules.libtorrent.torrentdef.TorrentDef object at 0x7f103b1715f8>

    @pytest.mark.asyncio
    @pytest.mark.timeout(30)
    async def test_hidden_services(enable_ipv8, proxy_factory, session, hidden_seeder_session, video_tdef):
        """
        Test the hidden services overlay by constructing an end-to-end circuit and downloading a torrent over it
        """
        await load_tunnel_community_in_session(session, exitnode=False, start_lt=True)
    
        hidden_seeder_session.tunnel_community.build_tunnels(1)
    
        relays, exit_nodes = await create_nodes(proxy_factory, num_relays=3, num_exitnodes=2)
        await introduce_peers([session, hidden_seeder_session] + relays + exit_nodes)
        await deliver_messages(timeout=1)
    
        for ses in [session, hidden_seeder_session] + relays + exit_nodes:
>           assert len(ses.tunnel_community.get_peers()) == 6
E           assert 7 == 6
E             +7
E             -6

src/tribler-core/tribler_core/modules/tunnel/test_full_session/test_tunnel_community.py:233: AssertionError
---------------------------- Captured stdout setup -----------------------------
seeder: b'video.avi' DLSTATUS_EXIT_NODES 0
seeder: b'video.avi' DLSTATUS_HASHCHECKING 0.0
------------------------------ Captured log setup ------------------------------
DEBUG    asyncio:selector_events.py:53 Using selector: EpollSelector
INFO     Session:session.py:388 Session is using state directory: /tmp/pytest-of-jenkins/pytest-75/popen-gw5/test_hidden_services0/dot.Tribler
DEBUG    tribler_core.utilities.network_utils:network_utils.py:38 Got a working random port 24269
DEBUG    TriblerConfig:tribler_config.py:111 Get random port 24269 for [libtorrent~port]
[...]

PR #6098, opened 2021-05-04
READY: Add live channel preview feature

ichorid:
This change affects how non-subscribed channels are shown to the user. Now _all_ the results are shown from a remote peer that is known to host the channel.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
@@ -18,6 +18,7 @@ from tribler_core.modules.metadata_store.community.remote_query_community import
 )
 from tribler_core.modules.metadata_store.serialization import CHANNEL_TORRENT
 from tribler_core.modules.metadata_store.store import ObjState
+from tribler_core.modules.metadata_store.utils import NoChannelSourcesException
 from tribler_core.utilities.unicode import hexlify
 
 minimal_blob_size = 200
@@ -159,6 +160,17 @@ class GigaChannelCommunity(RemoteQueryCommunity):
         }
         self.send_remote_select(peer, **request_dict, processing_callback=on_packet_callback)
 
+    async def remote_select_channel_contents(self, **kwargs):
+
+        peers_to_query = self.get_known_subscribed_peers_for_node(kwargs["channel_pk"], kwargs["origin_id"], 1)
+        if not peers_to_query:
+            raise NoChannelSourcesException()
+
+        result = await self.send_remote_select(peers_to_query[0], force_eva_response=True, **kwargs).processing_results
+        request_results = [r.md_obj.to_simple_dict() for r in result]
+
+        return request_results
+
     def send_search_request(self, **kwargs):
         # Send a remote query request to multiple random peers to search for some terms
         request_uuid = uuid.uuid4()
@@ -178,7 +190,7 @@ class GigaChannelCommunity(RemoteQueryCommunity):
         # Try sending the request to at least some peers that we know have it
         if "channel_pk" in kwargs and "origin_id" in kwargs:
             peers_to_query = self.get_known_subscribed_peers_for_node(
-                unhexlify(kwargs["channel_pk"]), kwargs["origin_id"], self.settings.max_mapped_query_peers
+                kwargs["channel_pk"], kwargs["origin_id"], self.settings.max_mapped_query_peers
             )
         else:
             peers_to_query = self.get_random_peers(self.settings.max_query_peers)
--- a/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
@@ -1,5 +1,6 @@
 import json
 import struct
+from asyncio import Future
 from binascii import unhexlify
 from dataclasses import dataclass
 
@@ -14,6 +15,7 @@ from tribler_core.modules.metadata_store.community.eva_protocol import EVAProtoc
 from tribler_core.modules.metadata_store.orm_bindings.channel_metadata import LZ4_EMPTY_ARCHIVE, entries_to_chunk
 from tribler_core.modules.metadata_store.serialization import CHANNEL_TORRENT, COLLECTION_NODE, REGULAR_TORRENT
 from tribler_core.modules.metadata_store.store import MetadataStore, ObjState
+from tribler_core.modules.metadata_store.utils import RequestTimeoutException
 from tribler_core.utilities.unicode import hexlify
 
 BINARY_FIELDS = ("infohash", "channel_pk")
@@ -37,6 +39,23 @@ def sanitize_query(query_dict, cap=100):
             sanitized_dict[field] = unhexlify(value)
 
     return sanitized_dict
+
+
+def convert_to_json(parameters):
+    sanitized = dict(parameters)
+    # Convert frozenset to string
+    if "metadata_type" in sanitized:
+        sanitized["metadata_type"] = [int(mt) for mt in sanitized["metadata_type"] if mt]
+
+    for field in BINARY_FIELDS:
+        value = parameters.get(field)
+        if value is not None:
+            sanitized[field] = hexlify(value)
+
+    if "origin_id" in parameters:
+        sanitized["origin_id"] = int(parameters["origin_id"])
+
+    return json.dumps(sanitized)
 
 
 @vp_compile
@@ -77,6 +96,17 @@ class SelectRequest(RandomNumberCache):
     def on_timeout(self):
         if self.timeout_callback is not None:
             self.timeout_callback(self)
+
+
+class EvaSelectRequest(SelectRequest):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # For EVA transfer it is meaningless to send more than one message
+        self.packets_limit = 1
+
+        self.processing_results = Future()
+        self.register_future(self.processing_results, on_timeout=RequestTimeoutException())
 
 
 class PushbackWindow(NumberCache):
@@ -144,7 +174,8 @@ class RemoteQueryCommunity(Community, EVAProtocolMixin):
         self.logger.warning(f"EVA transfer error: peer {hexlify(peer.mid)}, exception: {exception}")
 
     def send_remote_select(self, peer, processing_callback=None, force_eva_response=False, **kwargs):
-        request = SelectRequest(
+        request_class = EvaSelectRequest if force_eva_response else SelectRequest
+        request = request_class(
             self.request_cache,
             hexlify(peer.mid),
             kwargs,
@@ -155,11 +186,12 @@ class RemoteQueryCommunity(Community, EVAProtocolMixin):
         self.request_cache.add(request)
 
         self.logger.info(f"Select to {hexlify(peer.mid)} with ({kwargs})")
-        args = (request.number, json.dumps(kwargs).encode('utf8'))
+        args = (request.number, convert_to_json(kwargs).encode('utf8'))
         if force_eva_response:
             self.ez_send(peer, RemoteSelectPayloadEva(*args))
         else:
             self.ez_send(peer, RemoteSelectPayload(*args))
+        return request
 
     async def process_rpc_query(self, json_bytes: bytes):
         """
@@ -227,10 +259,6 @@ class RemoteQueryCommunity(Community, EVAProtocolMixin):
         if request is None:
             return
 
-        # Remember that at least a single packet was received was received from the queried peer.
-        if isinstance(request, SelectRequest):
-            request.peer_responded = True
-
         # Check for limit on the number of packets per request
         if request.packets_limit > 1:
             request.packets_limit -= 1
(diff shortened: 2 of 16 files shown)


ichorid:
This thing features some `Future`s programming. @qstokkink , @egbertbouman , could you please look at the stuff, so I don't :boom: myself in the :foot: ?

ichorid:
retest this please

src/tribler-core/tribler_core/modules/metadata_store/restapi/channels_endpoint.py
```diff
@@ -156,10 +157,22 @@ async def get_channel_contents(self, request):
...
-            total = self.session.mds.get_total_count(**sanitized) if include_total else None
+        remote = sanitized.pop("remote", None)
+
+        total = None
+
+        remote_failed = False
+        if remote:
+            try:
+                contents_list = await self.session.gigachannel_community.remote_select_channel_contents(**sanitized)
+            except (RequestTimeoutException, NoChannelSourcesException):
```
qstokkink, line 168: You may also get a `CancelledError` here too.

src/tribler-core/tribler_core/modules/metadata_store/community/tests/test_gigachannel_community.py
```diff
@@ -289,3 +290,51 @@ async def test_drop_silent_peer_from_channels_map(self):
             self.nodes[1].overlay.send_remote_select(self.nodes[0].my_peer, **kwargs_dict)
             await self.deliver_messages(timeout=1)
             self.nodes[1].overlay.channels_peers.remove_peer.assert_not_called()
+
+    @pytest.mark.timeout(0)
+    async def test_remote_select_channel_contents(self):
```
qstokkink, line 295: I noticed this test in passing: this test does not follow the [AAA pattern](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) and does not make use of the available test case code.

The pattern here is _Arrange-Act-Assert-Act-Assert-Act-Assert-Act-Assert_. Now I'm not for anally following all software patterns to a T, but your pattern puts a lot of stress on a reviewer and makes it harder to interpret the unit test suite's failing tests. Splitting this test into four separate tests would adhere to the Arrange-Act-Assert standard and I see no technical reason to deviate from the standard in this case.

Of secondary concern is the code duplication in this test file. There is high duplication in the arranging of this test with `test_remote_search_mapped_peers` for example. This can be refactored into a common method of `TestGigaChannelUnits`.
The volume and complexity of this test can be further reduced by using the available shortcuts. For example, `self.nodes[0].overlay.mds.TorrentMetadata` can be replaced with `TestGigaChannelUnits.torrent_metadata(0)`. Another commonly used shortcut you can employ is `self.overlay(i)` instead of `self.nodes[i].overlay`.

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
```diff
@@ -29,6 +30,9 @@
 MAGIC_GIGACHAN_VERSION_MARK = b'\x01'
 
 
+loop = asyncio.get_event_loop()
```
egbertbouman, line 33: Why is this here? As far as I can tell, you're not using it anywhere.

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -240,6 +271,9 @@ async def on_remote_select_response(self, peer, response_payload):
         processing_results = await self.mds.process_compressed_mdblob_threaded(response_payload.raw_blob)
         self.logger.info(f"Response result: {processing_results}")
 
+        if isinstance(request, EvaSelectRequest):
+            request.processing_results.set_result(processing_results)
```
egbertbouman, line 275: Before calling `future.set_results`, I recommend checking if it's done first. If the exception gets set in the `on_timeout` before calling `set_results` it will throw an exception.

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
```diff
@@ -144,7 +177,8 @@ def on_error(self, peer, exception):
         self.logger.warning(f"EVA transfer error: peer {hexlify(peer.mid)}, exception: {exception}")
 
     def send_remote_select(self, peer, processing_callback=None, force_eva_response=False, **kwargs):
```
egbertbouman, line 179: This function seems to pass its results to either a callback or a feature, depending on the argument passed to `force_eva_response`. I'm not sure why this is, but I think it makes more sense to either stick with callbacks or futures (but not both).

src/tribler-core/tribler_core/modules/metadata_store/community/gigachannel_community.py
ichorid: :face_with_head_bandage:

src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py
ichorid: You make a very good point here. Indeed, I would prefer `Future`s+`await`s. The only reason callbacks are still there is for compatibility with multiple-UDP packets response for pre-EVA versions. In that case, it is impossible to `await` on `Future`, because we never know if the received packet was the last one.

I'm a bit afraid of moving *all* the transport messages to EVA for 7.10. After 7.10, I will refactor these methods to use `Futures` only :+1:

egbertbouman: OK, sounds good 👍

src/tribler-core/tribler_core/modules/metadata_store/community/tests/test_gigachannel_community.py
ichorid: I refactored the offending test in 3 separate tests. However, I still use the `TestBase` thing, and I don't understand how to combine that with `pytest` fixtures. I'm not aware of a way to apply `Patch` context managers to a whole test w/o pytest stuff. 

We should refactor the whole `TestBase` into a pytest fixture eventually.

egbertbouman (approved):

src/tribler-core/tribler_core/modules/metadata_store/community/tests/test_gigachannel_community.py
qstokkink: Nice 👍 this looks 100x better now.

> We should refactor the whole `TestBase` into a pytest fixture eventually.

Why? Fixtures are not some magical construct that solve.. well.. _any_ problems at all in this case. In fact, they make your life actively worse.

1. To stage communities you need to keep a state. This state is being kept explicitly in `TestBase` superclass now. It is also possible to keep this state in a fixture, but you would still need the a `TestBase` **fixture** to juggle the test state anyway. So, you're not lowering the number of objects in your stack, that's staying the same.

2. As a downside for keeping state in fixtures, tests no longer explicitly point to the `TestBase` fixture (e.g., `from ipv8.test.base import TestBase`) and programmers will have to go on an archeological adventure to find out what is providing the network mocking functionality. Trying to find where your internals are defined costs time. Trying to figure out what fixtures your fixtures are using takes even longer.

3. In the `TestBase` you can easily overwrite functionality in a subclass you fit your specific test case. Using fixtures you have to redefine the fixtures in a smaller scope (e.g., in the session scope instead of the global scope) and hope and pray you caught all the side effects. If you had used a normal `TestCase` subclass your IDE shows you exactly what objects are maintained where. All the code is grouped together and you can easily what is going on.

4. In a normal class you can dynamically change your data. You can, in the middle of a unit test, modify the attributes of `TestBase` (e.g., add nodes and whatnot). You cannot (easily) modify a fixture that is not a manager (adding to a fixture that is a list or a dict or something is fine, replacing an immutable fixture at runtime is not going to happen easily).

5. You're losing compatibility. Subclasses of `TestCase`, like `TestBase`, can be run on pretty much every test runner for example `nose`, `unittest` and `pytest`. On the other hand `pytest` syntax doesn't work without.. well.. `pytest`.

6. It costs time and effort to convert `TestBase` into a fixture and will most likely, like all refactorings, bring with it some bugs.

The only upside to converting `TestBase` to a fixture I can see is uniformity of all tests in Tribler. However if uniformity is what you want, for the reasons I presented above, I'd favor converting everything back to `TestBase` instead of converting everything to `pytest`. The original reason to switch to `pytest` was because it would magically fix a hiddenservices test bug, [which it didn't](https://github.com/Tribler/tribler/issues/5582).

To end with some conventional wisdom: if it ain't broke, don't fix it. `pytest` works perfectly fine with the existing `TestBase`. Or, to quote [C4](https://hintjens.gitbooks.io/social-architecture/content/chapter4.html):

> A patch SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.

> The user or Contributor SHOULD seek consensus on the accuracy of their observation, and the value of solving the problem.

I don't see a problem being solved nor do I see the added value of the solution.

qstokkink (approved):

src/tribler-core/tribler_core/modules/metadata_store/community/tests/test_gigachannel_community.py
ichorid: The point of fixtures is to arbitrarily combine external states for individual test cases, mixin-style. It is not easy to achieve the same level of code reuse by subclassing `TestBase`. Also, I do not agree with your point (2). It was very daunting for me to delve deep into the bowels of `TestBase` classes hierarchy with all the non-relevant stuff around. Conversely, it is very easy for me every time to follow the fixtures.

Anyways, the ability to `yield` from a fixture is amazing, especially in combination with context managers. You should try it.

qstokkink: > The point of fixtures is to arbitrarily combine external states for individual test cases, mixin-style. It is not easy to achieve the same level of code reuse by subclassing TestBase.

Sure. I agree that in some cases a mix-in pattern can be useful. However, you can add a mix-in to a class, you cannot add a superclass to a mix-in.

Also, "combining external states" easily is a minefield. Mix-ins make your code more and more complex as your software grows. Here's a breakdown by a React developer if you want that backed-up by developers of large frameworks: https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750

> It was very daunting for me to delve deep into the bowels of TestBase classes hierarchy with all the non-relevant stuff around. Conversely, it is very easy for me every time to follow the fixtures.

Is it though? Let's take an example, suppose we want to know where `seed_config` comes from in this file:

https://github.com/Tribler/tribler/blob/089fb6ab1c597b90a051f35350d0e3703d3a07a6/src/tribler-core/tribler_core/modules/tunnel/test_full_session/test_tunnel_community.py#L68

It's not defined in this file. It's not even imported in this file. **First full project namespace search.** Aha, it's coming from here:

https://github.com/Tribler/tribler/blob/041a26f3d8c0cdd483f369e753e6482036a905d0/src/tribler-core/tribler_core/conftest.py#L71

That is created through `tribler_config` (thankfully defined in the same file) and `tmpdir_factory`. **Second full project namespace search.** Ah, not defined in this project. **Pytest namespace search.** Aha! There it is, finally we know how this fixture is created.

This is not easy at all, I had to do three full searches manually in two different scopes. I was also lucky there were no name or scope collisions in this case. In my IDE and even with GitHub's Python integration, this would've taken three clicks. The problem is the same as with star-imports, you get a bucketload of stuff with side-effects with no clue what the source is. You say classes include more "non-relevant stuff", I've both shown and experienced the opposite.

> Anyways, the ability to yield from a fixture is amazing, especially in combination with context managers. You should try it.

I have tried it and it, once again, has its uses. However, sometimes I require fine-grained control of the setup and shutdown of components in different fixtures. In those cases, fixtures become a mess. The same point holds as before, you can do the exact same thing in a class, but you can't bring the fine-grained control into a fixture (without turning your code into fixture spaghetti).

ichorid: > I had to do three full searches manually in two different scopes. 

For me, it was a matter of two Ctrl-LMB clicks. Or two `gd` commands from the keyboard :shrug: (VIM emulation) What IDE do you use?

>you can't bring the fine-grained control into a fixture

Can't we just have the best of both worlds? Use `TestBase` and add a fixture here and there somehow? Or at least add `yield`-base pseudo-fixtures to `TestBase`?

qstokkink: > What IDE do you use?

I use PyCharm, like most of our contributors, and I use GitHub to review PRs.

> Can't we just have the best of both worlds? Use TestBase and add a fixture here and there somehow? Or at least add yield-base pseudo-fixtures to TestBase?

Yes and that is all I'm bargaining for. Just use `TestBase` subclasses where appropriate and use fixtures where appropriate. Don't force fixtures where they are not needed or they are actively harmful. It is perfectly reasonable to provide fixture shortcuts for `TestBase` for program flows that could benefit from it. However, as you first proposed, I consider it harmful to straight up refactor `TestBase` into a fixture.

PR #6101, opened 2021-05-06
Update Popularity community

xoriole:
Two changes:
- Popular and random torrents are gossiped separately in two different intervals (popular: 2 mins, random: 5 seconds)
- Popularity community now uses a separate Network instance

### DIFF
--- a/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
+++ b/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
@@ -2,6 +2,8 @@ import heapq
 import random
 from binascii import unhexlify
 
+from ipv8.peerdiscovery.network import Network
+
 from ipv8.lazy_community import lazy_wrapper
 
 from pony.orm import db_session
@@ -26,10 +28,11 @@ class PopularityCommunity(RemoteQueryCommunity):
 
     community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1648')
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, my_peer, endpoint, network, **kwargs):
         self.torrent_checker = kwargs.pop('torrent_checker', None)
 
-        super().__init__(*args, **kwargs)
+        # Creating a separate instance of Network for this community to find more peers
+        super().__init__(my_peer, endpoint, Network(), **kwargs)
 
         self.add_message_handler(TorrentsHealthPayload, self.on_torrents_health)
 
--- a/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
+++ b/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
@@ -2,9 +2,8 @@ import heapq
 import random
 from binascii import unhexlify
 
-from ipv8.peerdiscovery.network import Network
-
 from ipv8.lazy_community import lazy_wrapper
+from ipv8.peerdiscovery.network import Network
 
 from pony.orm import db_session
 
@@ -17,14 +16,16 @@ class PopularityCommunity(RemoteQueryCommunity):
     """
     Community for disseminating the content across the network.
 
-    Every 5 seconds it gossips 5 the most popular torrents and 5 random torrents to
+    Every 2 minutes it gossips 10 popular torrents and
+    every 5 seconds it gossips 10 random torrents to
     a random peer.
 
     Gossiping is for checked torrents only.
     """
-    GOSSIP_INTERVAL = 5
-    GOSSIP_POPULAR_TORRENT_COUNT = 5
-    GOSSIP_RANDOM_TORRENT_COUNT = 5
+    GOSSIP_INTERVAL_FOR_POPULAR_TORRENTS = 120  # seconds
+    GOSSIP_INTERVAL_FOR_RANDOM_TORRENTS = 5  # seconds
+    GOSSIP_POPULAR_TORRENT_COUNT = 10
+    GOSSIP_RANDOM_TORRENT_COUNT = 10
 
     community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1648')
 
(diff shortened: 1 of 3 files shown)


src/tribler-core/tribler_core/modules/popularity/popularity_community.py
```diff
@@ -15,37 +16,44 @@ class PopularityCommunity(RemoteQueryCommunity):
     """
     Community for disseminating the content across the network.
 
-    Every 5 seconds it gossips 5 the most popular torrents and 5 random torrents to
+    Every 2 minutes it gossips 10 most popular torrents and
+    every 5 seconds it gossips 10 most random torrents to
```
ichorid, line 20: "most random" is, like, :boom:  :game_die: :boom:

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
```diff
@@ -75,6 +75,51 @@ async def test_torrents_health_gossip(self):
...
+        for torrent_info in all_checked_torrents:
+            self.nodes[0].overlay.torrent_checker.torrents_checked.add(torrent_info)
+
+        await self.introduce_nodes()
+
+        # Gossip random torrents
+        self.nodes[0].overlay.gossip_random_torrents_health()
+        await self.deliver_messages(timeout=0.1)
+
+        # Check whether node 1 has received all random torrent health information
```
ichorid, line 108: db1 == node0 and db2 == node1 naming scheme is inconsistent and surprising. It is better to use the same index everywhere, eg db0, db1, or even node_sender/db_sender, node_receiver/db_receiver.

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
```diff
@@ -75,6 +75,51 @@ async def test_torrents_health_gossip(self):
...
+
+        # Gossip popular torrents
+        self.nodes[0].overlay.gossip_popular_torrents_health()
+        await self.deliver_messages(timeout=0.1)
+
+        # Check whether node 1 has received all popular torrent health information.
+        with db_session:
+            count = db2.select().count()
+            assert count >= PopularityCommunity.GOSSIP_POPULAR_TORRENT_COUNT
+            assert count <= PopularityCommunity.GOSSIP_RANDOM_TORRENT_COUNT \
```
ichorid, line 120: The count should be `==` here since you completely control the environment, and there is no reinsertion in sampling the torrents.

ichorid (changes requested):

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
xoriole: this test and the naming scheme is an extension to the existing test `test_torrents_health_gossip` (line 56). I can change the naming scheme in both of them if you prefer.

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
xoriole: Random torrents could include popular torrents as well so there is this check for range instead of `==`

synctext:
You made the spread of popular swarms 24 x slower. Good idea, this should fix matters or get really close.

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
ichorid: Yep, that would be superb! :+1:

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
ichorid: Hmm... :thinking: 
https://github.com/Tribler/tribler/blob/3c2abfbad404fb52fe756ec830d86335b6554009/src/tribler-core/tribler_core/modules/popularity/popularity_community.py#L77

This line suggests that you're sampling `random` from the `rest` which is, well, the set of  `all` with removed `popular`, right? That is basically sampling without reinsertion.

Even if you're doing reinsertion, putting the test like you did, with `a<=b<=c`, may still result in `a==b` always. E.g. the test will **still pass** even if you disable sampling random torrents, right? That is clearly a deficiency.

BTW, to avoid randomness you can use `random.seed(123)` or something like that. You can achieve it by, say, `Patch`ing the `random` module in the community namespace. For reference, see:
https://github.com/Tribler/tribler/blob/74f836c4075b592597529438c734d99d683eb3c1/src/tribler-core/tribler_core/modules/metadata_store/tests/test_channel_metadata.py#L506
https://github.com/Tribler/tribler/blob/74f836c4075b592597529438c734d99d683eb3c1/src/tribler-core/tribler_core/modules/metadata_store/community/tests/test_gigachannel_community.py#L49

xoriole:
retest this please

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
xoriole: Yes, I'm sampling from the `rest` but random sampling will still include popular torrents because in computing random torrents `include_popular=False` is passed so sampling happens in the whole instead of `rest`. This is still important because it introduces the random uniform probability of including popular torrent in the first 2 mins because an explicit popular torrent message is gossiped.

```python
   def gossip_random_torrents_health(self):
        """
        Gossip random torrent health information to another peer.
        """
        self._gossip_torrents_health(include_popular=False, include_random=True)

    def gossip_popular_torrents_health(self):
        """
        Gossip popular torrent health information to another peer.
        """
        self._gossip_torrents_health(include_popular=True, include_random=False)
```

Regarding the test, I've updated to check the overlap and the existence of the popular torrents.

xoriole:
retest this please

src/tribler-core/tribler_core/modules/popularity/test_popularity_community.py
ichorid: That's some confusing usage of names :confused:
Anyways, let's merge it as it is. I'll fix it later today.

ichorid (approved):

PR #6160, opened 2021-06-08
Refactoring: tribler_config.py

drew2a:
* Made `tribler_config.py` real-file independent (you can create config just in memory)
* Made `tribler_config.py` generic, without module-related-specifics.
* The config specification (`tribler_conf.spec`) has been split into module-specific `ini` files


qstokkink:
Quick tip: You can avoid all the backslashes in the method chaining by using parentheses, for example:

https://github.com/Tribler/tribler/blob/db6fe41dcec46f7c9f4832fdf4f186f2ac944992/src/tribler-core/tribler_core/session.py#L319-L325

drew2a:
@qstokkink cool, thank you. Will use it.

src/tribler-core/tribler_core/config/tribler_config.py
```diff
@@ -2,627 +2,152 @@
...
-        """
-        Create a new TriblerConfig instance.
 
-        :param config_file: path to existing config file
-        :param reset_config_on_error: Flag indicating whether to recover from corrupt config using default config.
-        :raises an InvalidConfigException if ConfigObj is invalid
-        """
-        self._logger = logging.getLogger(self.__class__.__name__)
-        self._state_dir = Path(state_dir)
+class TriblerConfig:
```
drew2a, line 18: The most significant changes are located in this class.
So, it is a good idea to start a review from this particular class.

ichorid (approved):

PR #6163, opened 2021-06-14
Refactoring: path_utils.py

drew2a:
This is a refactoring for `path_utils.py`.

### DIFF
--- a/src/tribler-core/tribler_core/utilities/torrent_utils.py
+++ b/src/tribler-core/tribler_core/utilities/torrent_utils.py
@@ -1,8 +1,8 @@
 import logging
 from hashlib import sha1
 
-from tribler_core.utilities import path_util
 from tribler_core.utilities.libtorrent_helper import libtorrent as lt
+from tribler_core.utilities.path_util import Path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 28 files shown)


src/tribler-core/tribler_core/utilities/torrent_utils.py
```diff
@@ -17,34 +17,27 @@ def commonprefix(paths_list):
...
-        for full_file_path in file_path_list_filtered:
-            #FIXME: there should be a better, cleaner way to define this
-            filename = path_util.join(*full_file_path.parts[len(base_path.parent.parts):])
-            fs.add_file(str(filename), path_util.getsize(full_file_path))
+    # get the directory where these files are in. If there are multiple files, take the common directory they are in
+    base_path = commonprefix(path_list).parent if len(path_list) > 1 else path_list[0].parent
+    base_path = base_path.absolute()
+    for path in path_list:
+        relative = path.relative_to(base_path)
+        fs.add_file(str(relative), path.size())
```
drew2a, line 40: Right now I keep the original logic.

If there is only one file, then the values will be the following:

```
path_list = ['/user/tmp/1.txt']
name='1.txt'
base_path='/user/tmp'
```

If two or more files, then

```python
path_list = ['/user/tmp/1.txt', '/user/tmp/2.txt']
names = ['tmp/1.txt', 'tmp/2.txt']
base_path='/user'
```

My question here: "Should we actually keep the origin logic?"
Or, in the case of two and more files, it is better to have:

```python
path_list = ['/user/tmp/1.txt', '/user/tmp/2.txt']
names = ['1.txt', '2.txt']
base_path='/user/tmp'
```

src/tribler-core/tribler_core/utilities/torrent_utils.py
```diff
@@ -17,34 +17,27 @@ def commonprefix(paths_list):
...
-    if len(file_path_list_filtered) == 1:
-        filename = path_util.basename(file_path_list_filtered[0])
-        fs.add_file(filename, path_util.getsize(file_path_list_filtered[0]))
-    else:
-        for full_file_path in file_path_list_filtered:
-            #FIXME: there should be a better, cleaner way to define this
-            filename = path_util.join(*full_file_path.parts[len(base_path.parent.parts):])
-            fs.add_file(str(filename), path_util.getsize(full_file_path))
+    # get the directory where these files are in. If there are multiple files, take the common directory they are in
+    base_path = commonprefix(path_list).parent if len(path_list) > 1 else path_list[0].parent
```
drew2a, line 36: Right now I keep the original logic.

If there is only one file, then the values will be the following:

```
path_list = ['/user/tmp/1.txt']
name='1.txt'
base_path='/user/tmp'
```

If two or more files, then

```python
path_list = ['/user/tmp/1.txt', '/user/tmp/2.txt']
names = ['tmp/1.txt', 'tmp/2.txt']
base_path='/user'
```

My question here: "Should we actually keep the origin logic?"
Or, in the case of two and more files, it is better to have:

```python
path_list = ['/user/tmp/1.txt', '/user/tmp/2.txt']
names = ['1.txt', '2.txt']
base_path='/user/tmp'
```

ichorid: The whole "assemble a torrent from multiple files" logic is a complete and utter mess. It is better to keep it as it is now, and rethink the UX associated with file sharing later.

ichorid (approved):

PR #6174, opened 2021-06-22
Add typings to DiscoveryBooster

drew2a:
This PR adds typings for `DiscoveryBooster`


drew2a:
This PR will be a part of further refactoring.

PR #6181, opened 2021-06-24
Add version community mixin

xoriole:
This community mixin add two messages to request and receive Tribler version and platform of the peer.
- VersionRequest: 101
- VersionResponse: 102

These community messages are not used for the normal operation of Tribler but will be useful for network experiments. Currently, Popularity community is implementing this mixin.

### DIFF
--- a/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
+++ b/src/tribler-core/tribler_core/modules/popularity/popularity_community.py
@@ -9,10 +9,11 @@ from pony.orm import db_session
 
 from tribler_core.modules.metadata_store.community.remote_query_community import RemoteQueryCommunity
 from tribler_core.modules.popularity.payload import TorrentsHealthPayload
+from tribler_core.modules.popularity.version_community_mixin import VersionCommunityMixin
 from tribler_core.utilities.unicode import hexlify
 
 
-class PopularityCommunity(RemoteQueryCommunity):
+class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
     """
     Community for disseminating the content across the network.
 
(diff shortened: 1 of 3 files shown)


ichorid:
Making this into a separate community instead of a mixin would be a cleaner design.

@qstokkink , what are the pros and cons of having a separate community for this? Could you name some in regards to NAT puncturing and how trackers work?

qstokkink:
> what are the pros and cons of having a separate community for this?

It really depends on what you want to use this for. In short:
1. If you favor isolation over network overhead and want a general idea of the deployed versions, use a Community. A typical use case would be a server that observes the network.
2. If you want minimal network overhead and the versions of the specific peers you're connected to, use a mixin-style. A typical use case would be a debug window to inspect connections.

In my opinion, as O.P. suggests that this will be used to inspect specific experimental Communities, a mixin style would be most applicable.

<details> <summary> [More details] </summary>

|                   | Community | Mixin |                                                                                                                                                                                                                                                                                                    |
|-------------------|-----------|-------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Network Overhead  | :(        | :)    | A Community is "fat": you add additional network overhead to establish connections to peers. You also have to claim a unique `community_id` in the wire protocol. In contrast, if you have more of a mixin-style, you don't need to run peer discovery protocols and send peer discovery messages. |
| Exact Peers       | :(        | :)    | A mixin-style allows you to query exactly the peers you are connected to in a particular Community. Regrettably, Tribler moved to separated pools of peers (using their own `Network` instances). So, you'll have to query a Community for each Network instance you use.                          |
| Isolated Messages | :)        | :(    | A mixin-style uses the message space of its attached Community. Usually Communities don't use all 255 available message ids though.                                                                                                                                                                |
| Isolated Peers    | :)        | :(    | A Community can be isolated from other Communities. Without running any wild Communities, you could still query versions.                                                                                                                                                                          |

</details> 

By the way, there is also a third way/hybrid approach: you can make an inert Community (this would have minimal network overhead, exact peers and isolated messages, but no peer isolation). This requires special loading into IPv8 though.

xoriole:
The motivation for this PR was to be able to observe the peers in a particular community under experiment (in this case Popularity Community) which is why mixin was created instead of a separate community. Making a separate community will not serve that purpose. 

An alternative would have been to put the mixin code on the base community level which would provide these protocol messages to all the communities. That will make experimental observation of all the communities much easier. Since these messages are not used in normal operation of tribler there should also not be additional traffic (assuming honest network peers). But this will require making changes to IPv8 and not in Tribler.

ichorid (approved):

PR #6182, opened 2021-06-24
Extract loaders from ipv8_module_catalog.py

drew2a:
One step closer to the atomic community: introduction of `launcher.py`

I've extracted `IPv8CommunityLauncher` for each Tribler's community from `ipv8_module_catalog.py` and put them into `launcher.py` next to corresponding communities. 

image

### Preconditions

`Preconditions` have been separated from `Launcher` definition to `Loader` implementation:

`community_loader.py`
```python
def create_default_loader(config: TriblerConfig, tunnel_testnet: bool = False,
                          bandwidth_testnet: bool = False, chant_testnet: bool = False):
    loader = IPv8CommunityLoader()

    loader.set_launcher(IPv8DiscoveryCommunityLauncher())
    loader.set_launcher(DHTCommunityLauncher())

    if bandwidth_testnet:
        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthTestnetCommunityLauncher
        loader.set_launcher(BandwidthTestnetCommunityLauncher())
    else:
        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
        loader.set_launcher(BandwidthCommunityLauncher())
   ...
```

This allowed the use of top imports inside each loader (which should lead to easier navigation and more convenient work inside IDE). 

`launcher.py`
```python
from ipv8.loader import overlay, set_in_session, walk_strategy
from ipv8.peerdiscovery.discovery import RandomWalk
from tribler_core.modules.community_loader import INFINITE, TriblerCommunityLauncher

from tribler_core.modules.metadata_store.community.sync_strategy import RemovePeers
from tribler_core.modules.popularity.community import PopularityCommunity


@set_in_session('popularity_community')
@overlay(PopularityCommunity)
@walk_strategy(RandomWalk, target_peers=30)
@walk_strategy(RemovePeers, target_peers=INFINITE)
class PopularityCommunityLauncher(TriblerCommunityLauncher):
    def get_kwargs(self, session):
        return {
            'settings': session.config.popularity_community,
            'rqc_settings': session.config.remote_query_community,

            'metadata_store': session.mds,
            'torrent_checker': session.torrent_checker,
        }
```

### Community loader

Community loader can be passed to `Session` constructor now:

`session.py`
```python
    def __init__(self, config: TriblerConfig, core_test_mode: bool = False,
                 community_loader: IPv8CommunityLoader = None):
        if not community_loader:
            community_loader = create_default_loader(config)
```

Therefore you don't need to cheat session to specify your own set of community launchers:

`run_bandwith_crawler.py`
```python
async def start_crawler(tribler_config):
    # We use our own community loader
    loader = IPv8CommunityLoader()
    loader.set_launcher(BandwidthCommunityCrawlerLauncher())
    session = Session(tribler_config, community_loader=loader)
```

### DIFF
--- a/src/tribler-core/tribler_core/modules/community_loader.py
+++ b/src/tribler-core/tribler_core/modules/community_loader.py
@@ -12,6 +12,7 @@ from ipv8.loader import (
     walk_strategy,
 )
 from ipv8.peer import Peer
+from tribler_core.config.tribler_config import TriblerConfig
 
 INFINITE = -1
 """
@@ -19,9 +20,9 @@ The amount of target_peers for a walk_strategy definition to never stop.
 """
 
 
-class IPv8CommunityLauncher(CommunityLauncher):
+class TriblerCommunityLauncher(CommunityLauncher):
     def get_my_peer(self, ipv8, session):
-        return (Peer(session.trustchain_testnet_keypair) if session.trustchain_testnet()
+        return (Peer(session.trustchain_testnet_keypair) if session.trustchain_testnet
                 else Peer(session.trustchain_keypair))
 
     def get_bootstrappers(self, session):
@@ -50,7 +51,6 @@ def dht_discovery_community():
     return DHTDiscoveryCommunity
 
 
-# strategies
 def random_churn():
     from ipv8.peerdiscovery.churn import RandomChurn
     return RandomChurn
@@ -77,7 +77,7 @@ def periodic_similarity():
 @walk_strategy(random_churn, target_peers=INFINITE)
 @walk_strategy(random_walk)
 @walk_strategy(periodic_similarity, target_peers=INFINITE)
-class IPv8DiscoveryCommunityLauncher(IPv8CommunityLauncher):
+class IPv8DiscoveryCommunityLauncher(TriblerCommunityLauncher):
     pass
 
 
@@ -87,8 +87,45 @@ class IPv8DiscoveryCommunityLauncher(IPv8CommunityLauncher):
 @kwargs(max_peers='60')
 @walk_strategy(ping_churn, target_peers=INFINITE)
 @walk_strategy(random_walk)
-class DHTCommunityLauncher(IPv8CommunityLauncher):
+class DHTCommunityLauncher(TriblerCommunityLauncher):
     pass
+
+
+def create_default_loader(config: TriblerConfig, tunnel_testnet: bool = False,
+                          bandwidth_testnet: bool = False, chant_testnet: bool = False):
+    loader = IPv8CommunityLoader()
+
+    loader.set_launcher(IPv8DiscoveryCommunityLauncher())
+    loader.set_launcher(DHTCommunityLauncher())
+
+    if bandwidth_testnet:
+        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthTestnetCommunityLauncher
+        loader.set_launcher(BandwidthTestnetCommunityLauncher())
+    else:
+        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
+        loader.set_launcher(BandwidthCommunityLauncher())
+
+    if config.tunnel_community.enabled and not tunnel_testnet:
+        from tribler_core.modules.tunnel.community.launcher import TriblerTunnelCommunityLauncher
+        loader.set_launcher(TriblerTunnelCommunityLauncher())
+
+    if config.tunnel_community.enabled and tunnel_testnet:
+        from tribler_core.modules.tunnel.community.launcher import TriblerTunnelTestnetCommunityLauncher
+        loader.set_launcher(TriblerTunnelTestnetCommunityLauncher())
+
+    if config.popularity_community.enabled:
+        from tribler_core.modules.popularity.launcher import PopularityCommunityLauncher
+        loader.set_launcher(PopularityCommunityLauncher())
+
+    if config.chant.enabled and not chant_testnet:
+        from tribler_core.modules.metadata_store.community.launcher import GigaChannelCommunityLauncher
+        loader.set_launcher(GigaChannelCommunityLauncher())
+
+    if config.chant.enabled and chant_testnet:
+        from tribler_core.modules.metadata_store.community.launcher import GigaChannelTestnetCommunityLauncher
+        loader.set_launcher(GigaChannelTestnetCommunityLauncher())
+
+    return loader
 
 
 def get_hiddenimports():
@@ -101,43 +138,3 @@ def get_hiddenimports():
         hiddenimports.update(getattr(obj, "hiddenimports", set()))
 
     return hiddenimports
-
-
-def register_launchers(session, loader):
-    """
-    Register the default CommunityLaunchers into the given CommunityLoader.
-    If you define a new default CommunityLauncher, add it here.
-
-    = Warning =
-     Do not perform any state changes in this function. All imports and state changes should be performed within
-     the CommunityLaunchers themselves to be properly scoped and lazy-loaded.
-    """
-    loader.set_launcher(IPv8DiscoveryCommunityLauncher())
-    loader.set_launcher(DHTCommunityLauncher())
-
-    if session.bandwidth_testnet():
-        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthTestnetCommunityLauncher
-        loader.set_launcher(BandwidthTestnetCommunityLauncher())
-    else:
-        from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
-        loader.set_launcher(BandwidthCommunityLauncher())
-
-    if session.config.tunnel_community.enabled and not session.tunnel_testnet():
-        from tribler_core.modules.tunnel.community.launcher import TriblerTunnelCommunityLauncher
-        loader.set_launcher(TriblerTunnelCommunityLauncher())
-
-    if session.config.tunnel_community.enabled and session.tunnel_testnet():
-        from tribler_core.modules.tunnel.community.launcher import TriblerTunnelTestnetCommunityLauncher
-        loader.set_launcher(TriblerTunnelTestnetCommunityLauncher())
-
-    if session.config.popularity_community.enabled:
-        from tribler_core.modules.popularity.launcher import PopularityCommunityLauncher
-        loader.set_launcher(PopularityCommunityLauncher())
-
-    if session.config.chant.enabled and not session.chant_testnet():
-        from tribler_core.modules.metadata_store.community.launcher import GigaChannelCommunityLauncher
-        loader.set_launcher(GigaChannelCommunityLauncher())
-
-    if session.config.chant.enabled and session.chant_testnet():
-        from tribler_core.modules.metadata_store.community.launcher import GigaChannelTestnetCommunityLauncher
-        loader.set_launcher(GigaChannelTestnetCommunityLauncher())
--- a/src/tribler-core/tribler_core/modules/community_loader.py
+++ b/src/tribler-core/tribler_core/modules/community_loader.py
@@ -1,7 +1,4 @@
 # pylint: disable=import-outside-toplevel
-import inspect
-import sys
-
 from ipv8.loader import (
     CommunityLauncher,
     IPv8CommunityLoader,
@@ -23,7 +20,8 @@ The amount of target_peers for a walk_strategy definition to never stop.
 
 class TriblerCommunityLauncher(CommunityLauncher):
     def get_my_peer(self, ipv8, session):
-        return (Peer(session.trustchain_testnet_keypair) if session.trustchain_testnet
+        trustchain_testnet = session.config.general.testnet or session.config.trustchain.testnet
+        return (Peer(session.trustchain_testnet_keypair) if trustchain_testnet
                 else Peer(session.trustchain_keypair))
 
     def get_bootstrappers(self, session):
@@ -92,20 +90,21 @@ class DHTCommunityLauncher(TriblerCommunityLauncher):
     pass
 
 
-def create_default_loader(config: TriblerConfig, tunnel_testnet: bool = False,
-                          bandwidth_testnet: bool = False, chant_testnet: bool = False):
+def create_default_loader(config: TriblerConfig):
     loader = IPv8CommunityLoader()
 
     loader.set_launcher(IPv8DiscoveryCommunityLauncher())
     loader.set_launcher(DHTCommunityLauncher())
 
-    if bandwidth_testnet:
+    bandwidth_accounting_testnet = config.general.testnet or config.bandwidth_accounting.testnet
+    if bandwidth_accounting_testnet:
         from tribler_core.modules.bandwidth_accounting.launcher import BandwidthTestnetCommunityLauncher
         loader.set_launcher(BandwidthTestnetCommunityLauncher())
     else:
         from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
         loader.set_launcher(BandwidthCommunityLauncher())
 
+    tunnel_testnet = config.general.testnet or config.tunnel_community.testnet
     if config.tunnel_community.enabled and not tunnel_testnet:
         from tribler_core.modules.tunnel.community.launcher import TriblerTunnelCommunityLauncher
         loader.set_launcher(TriblerTunnelCommunityLauncher())
@@ -118,6 +117,7 @@ def create_default_loader(config: TriblerConfig, tunnel_testnet: bool = False,
         from tribler_core.modules.popularity.launcher import PopularityCommunityLauncher
         loader.set_launcher(PopularityCommunityLauncher())
 
+    chant_testnet = config.general.testnet or config.chant.testnet
     if config.chant.enabled and not chant_testnet:
         from tribler_core.modules.metadata_store.community.launcher import GigaChannelCommunityLauncher
         loader.set_launcher(GigaChannelCommunityLauncher())
--- a/src/tribler-core/tribler_core/modules/community_loader.py
+++ b/src/tribler-core/tribler_core/modules/community_loader.py
@@ -12,6 +12,7 @@ from ipv8.loader import (
     walk_strategy,
 )
 from ipv8.peer import Peer
+
 from tribler_core.config.tribler_config import TriblerConfig
 
 INFINITE = -1
@@ -37,7 +38,7 @@ class TriblerCommunityLauncher(CommunityLauncher):
 
 
 class TestnetMixIn:
-    def should_launch(self, session):
+    def should_launch(self, _):
         return True
 
 
@@ -126,15 +127,3 @@ def create_default_loader(config: TriblerConfig, tunnel_testnet: bool = False,
         loader.set_launcher(GigaChannelTestnetCommunityLauncher())
 
     return loader
-
-
-def get_hiddenimports():
-    """
-    Return the set of all hidden imports defined by all CommunityLaunchers in this file.
-    """
-    hiddenimports = set()
-
-    for _, obj in inspect.getmembers(sys.modules[__name__]):
-        hiddenimports.update(getattr(obj, "hiddenimports", set()))
-
-    return hiddenimports
--- a/src/tribler-core/tribler_core/session.py
+++ b/src/tribler-core/tribler_core/session.py
@@ -8,18 +8,16 @@ import logging
 import os
 import sys
 import time as timemod
+from _socket import gaierror
 from asyncio import get_event_loop
 from io import StringIO
 from traceback import print_exception
 
-from _socket import gaierror
-
+import tribler_core.utilities.permid as permid_module
 from ipv8.loader import IPv8CommunityLoader
 from ipv8.taskmanager import TaskManager
-
 from ipv8_service import IPv8
 from tribler_common.network_utils import NetworkUtils
-
 from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
 from tribler_common.simpledefs import (
     NTFY,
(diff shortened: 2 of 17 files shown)


src/tribler-core/tribler_core/modules/community_loader.py
```diff
@@ -0,0 +1,141 @@
...
+        from tribler_core.modules.popularity.launcher import PopularityCommunityLauncher
+        loader.set_launcher(PopularityCommunityLauncher())
+
+    if config.chant.enabled and not chant_testnet:
+        from tribler_core.modules.metadata_store.community.launcher import GigaChannelCommunityLauncher
+        loader.set_launcher(GigaChannelCommunityLauncher())
+
+    if config.chant.enabled and chant_testnet:
+        from tribler_core.modules.metadata_store.community.launcher import GigaChannelTestnetCommunityLauncher
+        loader.set_launcher(GigaChannelTestnetCommunityLauncher())
```
qstokkink, line 127: I have two problems with this approach:

 1. It extracts the internal logic of a launcher (a module) outside of its scope. **Launchers are no longer self-contained.**
 2. This has now turned from managing abstract launchers, to actively setting up the launcher internals (checking launch conditions). **This file has turned into a god object.**

I agree that this makes it easier to work with an IDE, allowing top-level imports in the `launcher.py` definitions. However, ultimately, I don't think IDE integration for files we touch once every +- year is worth the **sacrifice of demodularizing the Tribler architecture**. Therefore, I consider this part of your PR an unacceptable regression in architectural integrity.

I'd be happy to help with a solution that allows top-level imports without the problems I stated above. For now, I'd suggest reverting these lines.

src/tribler-core/tribler_core/modules/community_loader.py
```diff
@@ -0,0 +1,141 @@
...
+
+    return loader
+
+
+def get_hiddenimports():
```
qstokkink, line 132: This file no longer contains all of the launchers, breaking this function.

src/tribler-core/tribler_core/session.py
```diff
@@ -105,14 +104,17 @@ def __init__(self, config, core_test_mode=False):
         self.bootstrap = None
 
         # modules
-        self.ipv8_community_loader = IPv8CommunityLoader()
-        register_default_launchers(self.ipv8_community_loader)
-
+        if not community_loader:
+            community_loader = create_default_loader(config, chant_testnet=chant_testnet)
```
qstokkink, line 108: Forwarding an entire loader object seems like overkill for the problem you want to solve. Why not just use a boolean like `use_default_launchers` instead of forwarding a `community_loader` object:

```python
self.ipv8_community_loader = IPv8CommunityLoader()
if use_default_launchers:
    register_default_launchers(self.ipv8_community_loader)
```

src/tribler-core/tribler_core/session.py
```diff
@@ -73,7 +70,9 @@ class Session(TaskManager):
     """
     __single = None
 
-    def __init__(self, config, core_test_mode=False):
+    def __init__(self, config: TriblerConfig, core_test_mode: bool = False,
```
qstokkink, line 73: There are now config arguments next to a config object. These should not be apart.

qstokkink (changes requested):
I believe that moving module (/community) definitions to their own files is a good idea. However, I observe that this PR creates more problems than it solves.

I'd strongly suggest closing this PR and drafting a more formal design in an issue first.

src/tribler-core/tribler_core/modules/community_loader.py
drew2a: Thank you for this comment. Will try to explain my reasoning.

> It extracts the internal logic of a launcher (a module) outside of its scope

It extracts conditions from a launcher to a loader. Yes. 
But I would disagree with the statement that conditions are an internal and integral part of the launcher. Conditions describe the high-level composition of launchers and then communities. It looks more natural to set this composition of launchers explicitly.
Launchers become more self-contained in this case. 

If I need a custom composition of launchers I will not edit the launchers themselves, but just define a different loader.
https://github.com/drew2a/tribler/blob/ada438d691d1e14b3da743f0c2f6e7a6cc8fa7fb/src/tribler-core/run_bandwidth_crawler.py#L40-L47

> This file has turned into a god object.

Yes, the file has turned to something different, but I doubt it is a god object now. 
It actually was a "god" not object, but file before. 
This single file consisted of launchers for all communities and loaders for all communities. 

Now `community_loader.py` responds only for a default set of launchers. (It contains also `TriblerCommunityLauncher` and `TestnetMixIn`, and yes we can extract them into separate files too)

>However, ultimately, I don't think IDE integration for files we touch once every +- year is worth the sacrifice of demodularizing the Tribler architecture. 

I would gladly agree with you, but I don't see "demodularisation" here. Moreover, I don't see modularization in Tribler. It is a kind of "some degree" of modularization. But the logic of modules, however highly intertwined and not isolated.

Below just one single example:
https://github.com/Tribler/tribler/blob/814a94ad5faaa915083c9c9b299a831684e28802/src/tribler-core/tribler_core/session.py#L306-L316

Short description: if chant enables, then create `mds`-instance, which is used in many places:
image

So, from my perspective, this PR is a step closer to modularisation. 
It extracts loader logic (that is an integral part of a community)  from one single big file next to the community itself.

src/tribler-core/tribler_core/session.py
drew2a: Because it is more flexible. 

With `community_loader` you can change loader for the session without changing `session` (or `ipv8_module_catalog.py`/`comunity_loader.py`) itself.

By introducing `community_loader` I've decreased the count of session's responsibilities.

Or you want something like:

```python
    def __init__(self, config: TriblerConfig, core_test_mode: bool = False,
                 community_loader: IPv8CommunityLoader = None, use_default_launchers: bool = False,
                 chant_testnet: bool = False, trustchain_testnet: bool = False):
        
        if not use_default_launchers:
            community_loader = create_default_loader(config, chant_testnet=chant_testnet)
```
?

src/tribler-core/tribler_core/session.py
drew2a: Agree, but `core_test_mode` comes from: https://github.com/drew2a/tribler/blob/ada438d691d1e14b3da743f0c2f6e7a6cc8fa7fb/src/run_tribler.py#L178

So, it is not a part of Tribler's config.

ichorid:
@qstokkink , I hoped to first discuss it with @drew2a , but you guys discussing the problem of modularisation here...
[Dependency Injector](https://python-dependency-injector.ets-labs.org/examples/application-multiple-containers.html) framework is just what you're looking for. I propose converting the IPv8 modules catalog to it first, then gradually convert all of the Tribler code to it, finishing with the Session itself.

src/tribler-core/tribler_core/modules/community_loader.py
qstokkink: > Conditions describe the high-level composition of launchers and then communities. It looks more natural to set this composition of launchers explicitly.
> Launchers become more self-contained in this case.
>
> If I need a custom composition of launchers I will not edit the launchers themselves, but just define a different loader.

Why does it look more natural to you?
This does not fit a module pattern at all.
This fits a mediator pattern.

A module should initialize and finalize itself.
In your approach the manager of launchers (i.e., the loader) actively inspects and checks launcher-specific logic.
This also means that to add a launcher you need to add a loader.
To give an example analogy, this is like recompiling your operating system every time you install a new application.
I don't think this is natural at all.

> Yes, the file has turned to something different, but I doubt it is a god object now.
> It actually was a "god" not object, but file before.
> This single file consisted of launchers for all communities and loaders for all communities.

You have converted a "manager of launchers" into "a list of if-statements that deal with the internal logic of launchers".
This kills the abstraction of modules.
God file or god object, whatever you want to call it.

Still, though, I am not against moving each launcher to a separate file.
I do believe that is indeed the next step in modularization.

> [...] I don't see "demodularisation" here. Moreover, I don't see modularization in Tribler. It is a kind of "some degree" of modularization. But the logic of modules, however highly intertwined and not isolated.

The demodularization is in the breaking of the module design pattern I discussed before, I won't repeat it again.
Two further points on this though:

 1. Modules can be intertwined, that doesn't make them less modular. Just like `pip` dependencies have dependencies themselves.
 2. Just because Tribler modularization is not complete yet, it doesn't mean we should just void all previous work towards it. An analogy: "Your child can walk, but it can't ride a bike. So, we're breaking its legs".

> It extracts loader logic (that is an integral part of a community) from one single big file next to the community itself.

A single file can specify modules, that doesn't make the modules less modular.
At the same time, you can split highly centralized logic into multiple files just the same.
I don't see how the amount of files relates to the degree of modularity.
That said (repeating), I still see the value in placing the individual launcher files next to the module directories.

src/tribler-core/tribler_core/session.py
qstokkink: `core_test_mode`, `chant_testnet`, `trustchain_testnet` and `community_loader` all serve to configure a `Session`. Therefore, they should be part of the configuration.

I know `core_test_mode` was already there, even though it shouldn't be. However, that doesn't justify adding more of a bad thing.

src/tribler-core/tribler_core/session.py
qstokkink: I fully agree that it is more flexible. However, extra flexibility is not always a good thing, as it makes the code more complex. 

A loader is simply a container that orchestrates launchers. Making a new loader is like subclassing a `list`. There are uses for subclassing the container itself, but I don't see any uses for this in our code base. Even in Gumby, which is arguably using launchers in the most modular fashion, we don't use a custom loader (only custom launchers). For now, at least, this still seems like overkill.

Looking at it again, perhaps even my boolean suggestion is overkill. We could even simply make a method `Session.register_launchers()` that registers the launchers. This could be overwritten by a `Session` subclass.

qstokkink:
@ichorid I agree that this warrants more in-depth discussion and a formal design. This is why I called to move this PR to an issue first, instead of turning this PR into a long discussion.

I have not checked out the Dependency Injector framework. However, from my experience with Python plug-ins, we need to make doubly-sure that whatever we use (e.g., Dependency Injector) is compatible with a frozen environment (as created by PyInstaller). Ideally, we should have a list of all plug-in frameworks and weigh the pro's and con's. We'll be stuck with this for years probably, so this is not something we should decide on a whim.

src/tribler-core/tribler_core/session.py
drew2a: All variables that are passed into a class constructor are some kind of configuration for this particular class.

There is no reason to always use only one variable for this purpose.

If there are too many variables, of course then you can consolidate them.

src/tribler-core/tribler_core/modules/community_loader.py
drew2a: > A module should initialize and finalize itself.
In your approach the manager of launchers (i.e., the loader) actively inspects and checks launcher-specific logic.
This also means that to add a launcher you need to add a loader.

Not sure. In my approach loader doesn't check any launcher-specific logic actually. 
In my approach, it specifies the composition of launchers (which can be unique for any use case of `Session`). 

If you like to use analogies it is like to specify `requirements.txt` for each specific application instead of using one single big complex `requirements.txt` that contains all possible modules and their conditions.

> You have converted a "manager of launchers" into "a list of if-statements that deal with the internal logic of launchers".

I would say, I've converted the file, that was responsible for:
1. Defining launchers to each community
1. Defining default set of launchers

into:

1. Four launchers located next to corresponding communities
1. Default loader that used in Tribler main application (Tribler itself)

That said I have split the logic of something named "module catalog"(which actually had nothing to do with modules, but managed the loading of communities instead) into `community_loader.py` and `launcher.py`  

> Just because Tribler modularization is not complete yet, it doesn't mean we should just void all previous work towards it. An analogy: "Your child can walk, but it can't ride a bike. So, we're breaking its legs".

First of all, emphasize that, `ipv8_module_catalog` is not a part of the Tribler "modules", but a part of Tribler communities, and in particular, it is a loader manager for communities. Let's call things by their proper names.

Next, again, If you like analogies, then "I'm trying to give your baby a bike, to give him a chance to learn how to ride"

src/tribler-core/tribler_core/session.py
drew2a: You are correct, fully agree.

My point here, we shouldn't specify a set of launchers in `Session`. We have left open the possibility of easily changing the list of loaded communities by specifying a custom loader, without changing `Session` itself.

It is an application of "Inversion of control" principle.

Probably the version:
```python
    def __init__(self, config: TriblerConfig, core_test_mode: bool = False,
                 community_loader: IPv8CommunityLoader = None,
                 chant_testnet: bool = False, trustchain_testnet: bool = False):
        self.community_loader = community_loader
```
is more strict and сreates fewer dependencies.

Well, it will require a bigger changeset in this PR as I should fix every creation of `Session` instance. 
But that is fine.

src/tribler-core/tribler_core/modules/community_loader.py
drew2a: 👌 thank you for this remark

src/tribler-core/tribler_core/session.py
qstokkink: There are always exceptions to a rule. However, in this case, you're passing additional boolean values to enable/disable functionality next to an object (`TriblerConfig`) that exists to capture the configuration of components, including enabling/disabling functionality.

`chant_testnet` and `trustchain_testnet` specify ways of loading respectively the channel and trustchain components, they belong next to the other settings to configure these components. These are not special configuration options, they require no obscure datatypes or initialization. These are just boolean configuration values.

src/tribler-core/tribler_core/session.py
qstokkink: > We have left open the possibility of easily changing the list of loaded communities by specifying a custom loader, without changing Session itself.

The possibility was already there. A loader is simply a container for launchers that has functionality to load launcher definitions. You can add and remove launchers from a loader. I believe you're missing the following information, causing our disconnect:

**A loader is nothing but a glorified list for launchers.**

You can freely modify this list without making a new list. The launchers are only loaded as a consequence of `Session.start()`, here:

https://github.com/Tribler/tribler/blob/6c25dac44e6c233afa029251e59de3ad8e2e015a/src/tribler-core/tribler_core/session.py#L136-L137

https://github.com/Tribler/tribler/blob/6c25dac44e6c233afa029251e59de3ad8e2e015a/src/tribler-core/tribler_core/session.py#L356

Why I object to your configuration proposal, is because it boils down to the following:

```python
def register_communities(container):
    container.append("1")
    container.append("2")
    container.append("3")

class Old:
    def __init__(self):
        self.community_loader = []
        register_communities(self.community_loader)

class New:
    def __init__(self, community_loader=["1", "2", "3"]):
        self.community_loader = community_loader
```

I hope that makes it more clear. The old implementation adds elements to a container, your proposal supplies a completely new container with prefilled elements. This is why I suggest a boolean option added to the old implementation.

src/tribler-core/tribler_core/modules/community_loader.py
qstokkink: > Not sure. In my approach loader doesn't check any launcher-specific logic actually.

`config.tunnel_community.enabled and not tunnel_testnet` checks two configuration options of the TunnelCommunity outside of the TunnelCommunity module.

`config.chant.enabled and not chant_testnet` checks two configuration options of the GigaChannelCommunity outside of the GigaChannelCommunity module.

etc.

This is very much module specific. There is no abstraction from the component internals.

> In my approach, it specifies the composition of launchers (which can be unique for any use case of Session).

Doing this outside of a module itself is a textbook mediator pattern.

> If you like to use analogies it is like to specify requirements.txt for each specific application instead of using one single big complex requirements.txt that contains all possible modules and their conditions.

No, you're doing exactly the opposite. You're taking the individual requirements and pulling them into a long list. This is why I request changes.

> > You have converted a "manager of launchers" into "a list of if-statements that deal with the internal logic of launchers".
>
> I would say, I've converted the file, that was responsible for:
>
> 1. Defining launchers to each community
> 2. Defining default set of launchers
> 
> into:
>
> 1. Four launchers located next to corresponding communities
> 2. Default loader that used in Tribler main application (Tribler itself)

I agree and I like the idea. However, one does not exclude the other. Through your changes you changed the responsiblities of loaders from simple managers to managers with embedded elements (see [my other answer](https://github.com/Tribler/tribler/pull/6182/files/ada438d691d1e14b3da743f0c2f6e7a6cc8fa7fb#r659699915) -- I hope that clarifies this more).

> That said I have split the logic of something named "module catalog"(which actually had nothing to do with modules, but managed the loading of communities instead) into community_loader.py and launcher.py

It has everything to do with modules. These launchers are simply elaborate configuration definitions, not the actual modules themselves. I believe you misunderstand the nature of launchers and the loader: modules are not actually loaded here (see also [my other answer](https://github.com/Tribler/tribler/pull/6182/files/ada438d691d1e14b3da743f0c2f6e7a6cc8fa7fb#r659699915)).

> Next, again, If you like analogies, then "I'm trying to give the baby a bike, to give him a chance to learn how to ride"

I like analogies and I think this is noble goal. I also like moving files next to modules. My analogy is what I inferred from your code and it does not match your analogy. This is why I object. I object because I do not see your goal align with your implementation.

drew2a: > config.tunnel_community.enabled and not tunnel_testnet checks two configuration options of the TunnelCommunity outside of the TunnelCommunity module.
config.chant.enabled and not chant_testnet checks two configuration options of the GigaChannelCommunity outside of the GigaChannelCommunity module.
etc.
This is very much module specific. There is no abstraction from the component internals.

Yes, you can look at them in this way. 

From my POV in your approach, launchers (by defining loading conditions that you described above) have an impact on the higher structure — composition of launchers. From my POV it is wrong. They should not care about their composition. But they can care about the order of load (you nicely implemented it in `after` annotation). 
(actually, the load order is another very discussable topic, but let's skip it for this PR).

I want a more flexible and explicit approach in loaders. 
I feel a lack of explicitly and of flexibility here. 

You can get it as feedback from a developer who tried to use the current system and was disappointed (many-many times). As I see it from code I'm not alone. 

But besides just the feedback, I spent time developing an option to present how the current system can be improved. 
You see it is as a degradation (that's fine). I see it is as an improvement.

Make note that this PR fits your limitations on dynamic loading (which you patiently explained to me during the consultation I took before I started PR).

Let's try to make an arrangement for this PR tomorrow. If we can't find a middle ground, then will take a second opinion from a team :) 🤝 

>Doing this outside of a module itself is a textbook mediator pattern.

I agree it is kind-a-mediator, but the mediator "encapsulates the interaction between a set of objects" and our loader just loads a set (or a composition) of communities via launchers.

src/tribler-core/tribler_core/session.py
drew2a: Just for note this is exactly what I want:

```python
class New:
    def __init__(self, community_loader=["1", "2", "3"]):
        self.community_loader = community_loader
```

qstokkink (approved):
With the shift in issue #1, dropping plugins, none of the points I raised matter anymore.

qstokkink (changes requested):
Oops, I was too fast. This one still holds: https://github.com/Tribler/tribler/pull/6182/files#r659146652
I've set the comments that were related to plug-ins to "resolved".

src/tribler-core/tribler_core/session.py
drew2a: Unfortunately `chant_testnet` and `trustchain_testnet` are not a part of Tribler config, but calculated as:

```python
('TESTNET' in os.environ or
            'CHANT_TESTNET' in os.environ or
            config.chant.testnet)
```

and 

```python
('TESTNET' in os.environ or
            'TRUSTCHAIN_TESTNET' in os.environ or
            config.trustchain.testnet)
```

So, they are special.

In that case, how would you like them to be transferred to the session?

qstokkink: Configuration should be set in the configuration object. A Session should not have to worry about fixing its configuration.

I consider this the perfect solution:

https://github.com/Tribler/tribler/blob/43a428d3b0bee3256c559d5cb0cbb1d4c6091c12/src/tribler-core/tribler_core/modules/metadata_store/config.py#L10

🔽 

```python
   testnet: bool = 'TESTNET' in os.environ or 'CHANT_TESTNET' in os.environ
```

drew2a: As far as I know, by doing `testnet: bool = 'TESTNET' in os.environ or 'CHANT_TESTNET' in os.environ` in  `MetadataStoreConfig` definition you specify a default value of `testnet` variable.

But `'TESTNET' in os.environ or 'CHANT_TESTNET' in os.environ` should be calculated in a runtime and be a disjunction with testnet value.

But wait.. probably I know the way. Wait a minute...

drew2a: I can do something like:

```python
class BandwidthAccountingSettings(TriblerConfigSection):
    testnet: bool = Field(..., env='CHANT_TESTNET')
```

https://pydantic-docs.helpmanual.io/usage/settings/#environment-variable-names

qstokkink: Perfect 👍

drew2a: @qstokkink done. Please, check it: 

https://github.com/Tribler/tribler/pull/6182/commits/5e99b28359e4d7721169fa2268b7e5dd1c87d672

qstokkink (approved):
Nice 👍 Thanks for sticking with us through an entire lab vision revision.

Some future work notes:

 - This breaks Gumby, specifically https://github.com/Tribler/gumby/blob/devel/gumby/modules/gumby_session.py
 - The launcher/loader classes are made for plug-ins. Now that we're not doing plug-ins anymore, they can be removed.

drew2a:
@qstokkink 🙏 

I may need to do another PR with Dependency Injection as a result of today's meeting, which will remove the use of `launcher/loader` from Tribler.

After that, the Gumby can be fixed.

ichorid:
Continuation of the DependencyInjector discussion: https://github.com/Tribler/tribler/issues/4953

qstokkink (approved):

PR #6185, opened 2021-06-27
EXAMPLE: Use DependencyInjector for modules

ichorid:
This PR shows how we can gradually start using the [DependencyInjector framework](https://github.com/ets-labs/python-dependency-injector). The example is a bit silly, because it puts MetadataStore and Trustchain keypair into two separate containers.
Note the usage of the `Singleton` provider.

While working on this example, I realised two things:
 * we probably don't really *need* DepencyInjector _at the Session level_, because Tribler consists of a bunch of singletons, and DI framework's most powerful feature is injecting the dependencies into factories (which we don't use). Nonetheless, the framework can be of great use _inside_ some communities, where we have to e.g. instance Tunnels, Caches, etc. Also, the future Channels design can put it to great use by instancing Communities per each Channel.
 * **we don' really need the Session god-object**. It will take only a limited amount of effort to stop passing the `Session` object around, thus ceasing the backreference antipattern. In turn that will allow us to start our components asynchronously, making each one into an independent `Task`.

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/container.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/container.py
@@ -5,8 +5,8 @@ from tribler_core.modules.metadata_store.store import MetadataStore
 
 class MetadataStoreContainer(containers.DeclarativeContainer):
     config = providers.Configuration()
-    trustchain_keypair = providers.Singleton()
-    notifier = providers.Singleton()
+    trustchain_keypair = providers.Object()
+    notifier = providers.Object()
 
     metadata_store = providers.Singleton(
         MetadataStore,
--- a/src/tribler-core/tribler_core/modules/metadata_store/container.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/container.py
@@ -1,0 +1,18 @@
+from dependency_injector import containers, providers
+
+from tribler_core.modules.metadata_store.store import MetadataStore
+
+
+class MetadataStoreContainer(containers.DeclarativeContainer):
+    config = providers.Configuration()
+    trustchain_keypair = providers.Singleton()
+    notifier = providers.Singleton()
+
+    metadata_store = providers.Singleton(
+        MetadataStore,
+        db_filename=config.db_filename,
+        channels_dir=config.channels_dir,
+        my_key=trustchain_keypair,
+        notifier=notifier,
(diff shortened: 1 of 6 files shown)


drew2a:
🔥 looks very attractive.

qstokkink:
Looks good so far. 👍 

Could you also port a unit test? Specifically: do pytest fixtures work with providers?

qstokkink:
(ref to documentation for my own convenience: https://python-dependency-injector.ets-labs.org/index.html)

Just finished plowing through all of the documentation. Two high-level highlights:

1.
> This makes easier to understand and change how application works.

2.
> **Common sense first.** Use a common sense when apply dependency injection. It is a good principle, but not a silver bullet. If you do it too much you will reveal too much of the implementation details. Experience comes with practice and time.

These are two things I find myself completely agreeing with:

 - Dependency Injector formalizes the interfaces between components, this provides abstraction between the implementation and the capabilities of code. This indeed helps to compose and change an application. However, this only formalizes the interfaces - it does not change or fix your application by itself. A bad architecture or implementation will stay the same, just with a formal interface.

 - Applying interfaces is a good fit to compose components in your software architecture. However when going too low level ("too much of the implementation details"), interfaces turn into overengineering. For example, you don't need an interface to get a value from a dictionary. Therefore, I disagree with your first realization @ichorid: for implementation details like cache lookups (which are already a wrapper around `dict.get()` -- not a system component), this doesn't seem to fit. Abstracting over a Community's capabilities with an interface does seem like a good fit though.

synctext:
This is a pull request without an attached issue that describes the problem that needs to be fixed. This has not been discussed in the meetings which are explicitly setup for this. Small overhauls are great, but this seems very disruptive.

I understand where this is coming from, architectural beauty is beautiful.

ichorid:
> This is a pull request without an attached issue that describes the problem that needs to be fixed. This has not been discussed in the meetings which are explicitly setup for this. Small overhauls are great, but this seems very disruptive.
> 
> I understand where this is coming from, architectural beauty is beautiful.

This PR is a solution to the old issue #4953 . This PR was created to just show/prove the solution. After getting initial reactions from developers, we discussed the idea with @drew2a  and @qstokkink in the office. Now @drew2a is producing another PR, which will actually solve #4953

PR #6196, opened 2021-07-01
Remove obsolete database_blob type

### LINKED ISSUE #6097: Remove usages of `database_blob`
Related to https://github.com/Tribler/py-ipv8/issues/982

The `database_blob` was used for Python 2/3 compatibility, now it is just equal to `bytes` and it can be removed. In many cases this also means the removal of additional casts, for example:

https://github.com/Tribler/tribler/blob/f14108a8d0e3f17ee0216ad461e93bb121d1736a/src/tribler-core/tribler_core/modules/metadata_store/store.py#L172

This example is casting a `bytes` into a `bytes` and then casts that to a `bytes`. This particular example can be replaced with:

```python
self.my_public_key_bin = self.my_key.pub().key_to_bin()[10:]
```

ichorid:
fixes #6097  with some `bowler` magic:

<details>

```python

import mmap
import re
from lib2to3.pgen2.parse import ParseError
from pathlib import Path

from bowler.query import Query

fun_name = "database_blob"

dbblob = re.compile(fun_name.encode())


def modify_file(filename):
    (
        Query(filename).
            select_function(fun_name).
            is_call().
            modify(remove_call).execute(interactive=False, write=True)
    )


def remove_call(ast_node, capture, _):
    args = capture.get('function_arguments')

    parent = ast_node.parent
    children_list = parent.children
    for ind, leaf in enumerate(children_list):
        if leaf is ast_node:
            children_list[ind] = args[0].clone()


if __name__ == "__main__":
    for file in Path("src/tribler-core").rglob("*.py"):
        try:
            with open(file, "r+") as f:
                data = mmap.mmap(f.fileno(), 0)
                if not re.search(dbblob, data):
                    continue
        except:
            continue
        try:
            modify_file(file)
        except ParseError:
            print("Can't handle file ", file)


```

</details>

### DIFF
--- a/src/tribler-core/tribler_core/modules/metadata_store/tests/test_gigachannel_manager.py
+++ b/src/tribler-core/tribler_core/modules/metadata_store/tests/test_gigachannel_manager.py
@@ -4,7 +4,6 @@ from datetime import datetime
 from pathlib import Path
 from unittest.mock import Mock, patch
 
-from ipv8.database import database_blob
 from ipv8.util import succeed
 
 from pony.orm import db_session
@@ -12,8 +11,8 @@ from pony.orm import db_session
 import pytest
 
 from tribler_common.simpledefs import DLSTATUS_SEEDING
-from tribler_core.config.tribler_config import TriblerConfig
 
+from tribler_core.config.tribler_config import TriblerConfig
 from tribler_core.modules.libtorrent.torrentdef import TorrentDef
 from tribler_core.modules.metadata_store.gigachannel_manager import GigaChannelManager
 from tribler_core.modules.metadata_store.orm_bindings.channel_node import NEW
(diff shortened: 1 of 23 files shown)


src/tribler-core/tribler_core/modules/metadata_store/tests/test_gigachannel_manager.py
```diff
@@ -290,16 +289,16 @@ def get_def(self):
...
         # Downloads for the updated external channel: "old ones" and "recent"
-        MockDownload(database_blob(b'12331244'), chan2.dirname),
-        MockDownload(database_blob(bytes(chan2.infohash)), chan2.dirname),
+        MockDownload(b'12331244', chan2.dirname),
+        MockDownload(bytes(chan2.infohash), chan2.dirname),
         # Downloads for the unsubscribed external channel
-        MockDownload(database_blob(b'1231551'), chan3.dirname),
-        MockDownload(database_blob(bytes(chan3.infohash)), chan3.dirname),
+        MockDownload(b'1231551', chan3.dirname),
+        MockDownload(bytes(chan3.infohash), chan3.dirname),
```
qstokkink, line 299: It seems likely that these `bytes` calls could also be removed. Could you double-check this is still needed?

ichorid: Nice find!

qstokkink (approved):
Thank you! This has been on my wishlist for quite a long time.

Issue #6177, opened 2021-06-23
Improvement Request: Restructure of `tribler-core`


drew2a:
The current structure of `core` is definitely better than [7.3.0](https://github.com/Tribler/tribler/tree/release-7.3.0/Tribler/Core) but it is still far from perfection.
I'm almost "one-year-tribler-developer" but still confused with the navigation and with the purpose of some folders.

I want to suggest one small step further.

The structure might look like this:

```python
tribler_core
|
├ communities
|  ├ accounting
|  ├ tunnel
|  ├ popularity
|  ├ gigachannel
|  └ remote_query
|
├ database
|  ├ mds
|  └ ledger_zero
|
├ rest
|  ├ libtorrent
|  ├ channels
|  └ tunnels
|
├ components
|  ├ libtorrent
|  ├ category_filter
|  ├ resource_monitor
|  ├ torrent_checker
|  ├ trust_calculation
|  ├ torrent_checker
|  └ upgrade
|
└ utilities
```

Why `rest` placed outside communities or components? Because it represents the whole Core-API and should be easy to find for those who want to use it.

Why `database` placed outside of `mds`? Because it is one of the core Tribler's parts that has been used from several components and communities and deserves its place in a hi-level of the structure.

Feel free to comments on a structure itself or on naming in particular.
When we will reach a consensus it will take one day of my time to implement the changes.

ichorid:
A couple of important notes:
 * First, there is more than a single set of ORM bindings: LedgerZero's ORM bindings are completely independent of MDS-related bindings. So, putting all the ORM bindings into a single dir will create a total mess.
 * Second, the [original idea behind the structure we're using now](https://github.com/Tribler/tribler/issues/4953) was to eventually support adding _modules_ to Tribler dynamically. 


In the end, it boils down to the classic question of "sort by type vs sort by function". Sort by type is the Unix's approach (`/bin`), while sort by function is the Windows's approach (`c:\Program Files`). The former better fits tight, cleanly designed systems, conservative systems with a limited scope of functionality. The latter is favoured by dynamic, highly extensible systems.
The Unix approach is about enforcing structure. The Windows approach is about isolating chaos.

Which one fits Tribler better at this point in development, I'm unable to tell :shrug:

drew2a:
@ichorid I've added "ledger zero" and "mds" ORM-bindings to the example above.

ichorid:
[...]

PR #6197, opened 2021-07-01
WIP Add dependency-injector to Tribler

drew2a:
This is an example of DI in Tribler.

I've tested it on `initial_filling.py` and `run_tribler.py`.

What can I say after a day of work: it works but the framework is surprisingly difficult to understand. 
I spent all day trying to make this example runnable.

Partially it is because of the `session.py` organization, but mainly because of the unintuitive framework.
(I am not stating that it is bad, but only that it is unintuitive).

Please disregard the writing style, I didn't work on it.
I want to show you roughly what the Tribler with DI will look like.

Framework: https://python-dependency-injector.ets-labs.org/
@ichorid example: https://github.com/Tribler/tribler/pull/6185/files

References:
https://github.com/Tribler/tribler/issues/4953
https://github.com/Tribler/tribler/issues/6177

### DIFF
--- a/experiment/popularity_community/initial_filling.py
+++ b/experiment/popularity_community/initial_filling.py
@@ -12,8 +12,10 @@ from pony.orm import count, db_session
 
 from ipv8.peer import Peer
 from ipv8.peerdiscovery.discovery import RandomWalk
+from tribler_core import containers
+from tribler_core.modules.popularity.community import PopularityCommunity
 from tribler_core.modules.remote_query_community.settings import RemoteQueryCommunitySettings
-from tribler_core.modules.popularity.popularity_community import PopularityCommunity
+from tribler_core.session import Session
 from tribler_core.utilities.tiny_tribler_service import TinyTriblerService
 
 _logger = logging.getLogger(__name__)
@@ -67,11 +69,16 @@ class ObservablePopularityCommunity(PopularityCommunity):
 
 class Service(TinyTriblerService):
     def __init__(self, interval_in_sec, output_file_path, timeout_in_sec, working_dir, config_path):
-        super().__init__(Service.create_config(working_dir, config_path), timeout_in_sec,
+        config = Service.create_config(working_dir, config_path)
+        super().__init__(config, timeout_in_sec,
                          working_dir, config_path)
 
         self._interval_in_sec = interval_in_sec
         self._output_file_path = output_file_path
+
+        self.application = containers.ApplicationContainer(state_dir=config.state_dir)
+        self.application.config.from_pydantic(config)
+        self.application.wire(modules=[Session])
 
     @staticmethod
     def create_config(working_dir, config_path):
@@ -85,7 +92,7 @@ class Service(TinyTriblerService):
         await super().on_tribler_started()
 
         session = self.session
-        peer = Peer(session.trustchain_keypair)
+        peer = Peer(session.trustchain_keys.keypair)
 
         session.popularity_community = ObservablePopularityCommunity(self._interval_in_sec,
                                                                      self._output_file_path,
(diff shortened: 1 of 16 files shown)


experiment/popularity_community/initial_filling.py
```diff
@@ -67,12 +69,17 @@ def check(self):
...
+        config = Service.create_config(working_dir, config_path)
+        super().__init__(config, timeout_in_sec,
                          working_dir, config_path)
 
         self._interval_in_sec = interval_in_sec
         self._output_file_path = output_file_path
 
+        self.application = containers.ApplicationContainer(state_dir=config.state_dir)
+        self.application.config.from_pydantic(config)
+        self.application.wire(modules=[Session])
```
drew2a, line 81: A good start point to discover PR.

src/tribler-core/tribler_core/containers.py
```diff
@@ -0,0 +1,36 @@
...
+class Ipv8Container(containers.DeclarativeContainer):
+    state_dir = providers.Dependency()
+    config = providers.Configuration()
+
+    ipv8_config = providers.Singleton(Ipv8Config, state_dir=state_dir, config=config.provider)
+
+    endpoint = providers.Singleton(
+        DispatcherEndpoint,
+        providers.List(providers.Object("UDPIPv4")),
+        UDPIPv4=providers.Dict(port=providers.Factory(config.port), ip=providers.Factory(config.address)),
```
ichorid, line 21: Yep, this line looks not too intuitive :confused:. I would expect it to be just `UDPIPv4={'port':config.port}`, etc. Why the `Factory` stuff is necessary here, what's the semantic reason?

src/tribler-core/tribler_core/session.py
```diff
@@ -254,21 +231,22 @@ def get_ipv8_statistics(self):
         """Return a dictionary with IPv8 statistics."""
         return TriblerStatistics(self).get_ipv8_statistics()
 
+    @inject
```
ichorid, line 234: I believe this is not necessary here, since there are not `Provide` statements in the method's arguments. Correct me if I'm wrong.

src/tribler-core/tribler_core/trustchain_keys.py
```diff
@@ -0,0 +1,30 @@
+import tribler_core.utilities.permid as permid_module
+from tribler_core.utilities.path_util import Path
+
+
+class TrustChainKeys:
```
ichorid, line 5: In the future, it is better to leave only the single keypair property in the code, loading either the real key or the testnet key based on the config option. AFAIK, there are no situations where *both* keys are used in the code simultaneously.

ichorid:
Looks promising! I see you're going to introduce DependencyInjector, but you're still keeping the Session in, right?

qstokkink:
Looks good 👍 All my implementation feedback is already covered by @ichorid. I have two meta questions about what you wrote in O.P.:

> What can I say after a day of work: it works but the framework is surprisingly difficult to understand.
I spent all day trying to make this example runnable.

1. Did you enjoy working with DI (is it intuitive, verbose, etc.)? In other words, did you feel like you were fighting the framework or did you feel like the framework was helping you?
2. Do you think DI is prohibitively difficult for new developers? It's not a good thing if this scares off new or external developers.

src/tribler-core/tribler_core/containers.py
drew2a: It can be replaced by (explanations later):

```python
    endpoint = providers.Singleton(
        DispatcherEndpoint, ["UDPIPv4"],
        UDPIPv4=providers.Dict(port=config.port, ip=config.address),
    )
```

qstokkink:
retest this please

src/tribler-core/tribler_core/containers.py
drew2a: If you use `UDPIPv4={'port':config.port}` you'll get:

```python
{'UDPIPv4': {'port': <dependency_injector.providers.ConfigurationOption('config.port') at 0x112961eb0>}}
```

src/tribler-core/tribler_core/session.py
drew2a: You are right! It is just an artifact.

Issue #1052, opened 2014-12-20
Redesign of the Family Filter


gjedeer:
It looks like anything with the word "girl" in it is marked as "Xxx" and blocked by the family filter. It's a word which generates a lot of false positives:

http://www.imdb.com/find?q=girl&&s=tt&ref_=fn_al_tt_mr

LipuFei:
This could be a good enhancement although it's for videos:
https://github.com/ryanjay0/miles-deep

ichorid:
State-of-the-art of the text classification today is [BERT](https://towardsdatascience.com/bert-text-classification-in-3-lines-of-code-using-keras-264db7e7a358).

To train it, we can use dataset from https://github.com/cental/stc described in [this](http://cental.fltr.ucl.ac.be/team/~panchenko/11_Camera_ready_paper.pdf) paper by https://github.com/alexanderpanchenko.

qstokkink:
In the upcoming version of Tribler, the family filter is removed in its entirety.

Some rationale. We no longer have random torrents pushed onto people's front page, nor do we have thumbnails. The most people will see is offensive torrent titles and, hopefully, only if they were looking for them.

Perhaps we will come crawling back from this decision, but for now this issue has been superseded.

PR #6259, opened 2021-08-13
Don’t filter any groups of people out of existence

da2x:
The keyword filter list is discriminating.

Some searches that aren’t allowed with the current filter:

* belgian waffles
* asian tv, asian cousine, asuan culture
* scandianvian nature, scandinavian architecture
* gay interest (a common label for media with LGBTQ characters)
* turkish movies
* danish children tv (watch [John Dillermand](https://www.youtube.com/watch?v=A51mJjFyG_w))

I’ve removed the groups of people (but left in slurs) I could identify from the list. I was surprised that there were so many given names are on the list. I understand these may be adult performers, but blocking every artist or creator named Alexis, Charlie, or one of the hundred other names on the list is crazy. These should be removed as well, but let’s address those in a separate pr.

### DIFF
--- a/src/tribler-core/tribler_core/modules/category_filter/filter_terms.filter
+++ b/src/tribler-core/tribler_core/modules/category_filter/filter_terms.filter
@@ -1,7 +1,5 @@
 *adult
 *fuck
-*gay
-*lesbian
 *porn
 *sex
 *shemale
@@ -52,7 +50,6 @@ agustina
 akiba
 asa akira
 alba
-albanian
 alektra
 alektra blue
 alex divine
@@ -126,7 +123,6 @@ amber rayne
 amor
 amore
 ampland
-amputee
 anaal
 anaallikken
 anaalneuken
@@ -167,7 +163,6 @@ animalporno
 animalsex
 animalsexcom
 animalsexverhalen
-anime
 anita crystal
 anja
 anjali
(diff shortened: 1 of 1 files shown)


qstokkink:
ok to test

ichorid:
Originally, the filter was a quick hack hastily assembled from some Dutch porn site by a couple of students. The last time I gave it a cleanup I've learned more than I asked for, despite considering myself a pretty open and educated person...

So, thanks for filtering the filter. 

I would still suggest leaving all the "teen"-related words in the filter. Losing TMNT and Teen Moms is a relatively small price for filtering out all the stuff that typically hides behind the term...

ichorid:
Also, please look at #1052

src/tribler-core/tribler_core/modules/category_filter/filter_terms.filter
```diff
@@ -3360,8 +3309,6 @@ teasing
 techniek
 teef
 teefjes
-teen
```
ichorid, line 3363: Please keep this word an other words with the same stem

ichorid (changes requested):
Please keep the `teen` words in

da2x:
I’ll put it back in if you insist, but I still believe it’s wrong to filter out Teenage Mutant Ninja Turtles and Teen Vouge just because people sexualize teenagers.

ichorid:
Don't worry, people would still be able to search for Teen Vogue - they'll
have to enable showing "unsafe" content though.

пн, 16 авг. 2021 г., 06:31 Daniel Aleksandersen ***@***.***>:

ichorid (approved):

PR #6275, opened 2021-08-20
Refactoring/session logic outside of session

### LINKED ISSUE #6254: Session logic outside of Session
The following statements operate on the `Session` class but are not bound to its namespace. This causes all of these methods and variables to have a `session` prefix or postfix.

https://github.com/Tribler/tribler/blob/5cb1ee094e5312ede323b4bb1166dea85c379768/src/tribler-core/tribler_core/components/base.py#L36

https://github.com/Tribler/tribler/blob/5cb1ee094e5312ede323b4bb1166dea85c379768/src/tribler-core/tribler_core/components/base.py#L93-L113

If these methods only operate on the `Session` class, they should be part of it. This also avoids using `global` statements and, more formally, gives coherence to the code (🎉).

For example:

```python
_default_session: Optional[Session] = None 

def set_default_session(session: Session): 
    global _default_session 
    _default_session = session 
```

Should become:

```python
class Session:
    _default: Optional[Session] = None

    def set_default(self): 
        Session._default = self
```

Edit: almost forgot, this also cleans up your imports (`from x import Session` instead of `from x import Session, set_default_session, get_session, session_counter, ...`).

kozlovsky:
Fixes #6254

In this pull request, I refactor session-based logic into class methods/attributes of Session.
It makes the code a bit cleaner.

- `get_session()` becomes `Session.current()`
- `set_default_session(session)` becomes `session.set_as_default()`

### DIFF
--- a/src/tribler-core/tribler_core/components/base.py
+++ b/src/tribler-core/tribler_core/components/base.py
@@ -1,17 +1,16 @@
 from __future__ import annotations
 
-from abc import abstractmethod
-from asyncio import Event, create_task, gather, get_event_loop
-from contextlib import contextmanager
-from itertools import count
 import logging
 import os
 import sys
-from typing import Dict, Iterable, List, Optional, Set, Type, TypeVar
-
+from abc import abstractmethod
+from asyncio import Event, create_task, gather
+from itertools import count
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


qstokkink (approved):
Nice! Thanks 👍

PR #6276, opened 2021-08-20
Fixes #6255: make `faifast` an option of the Session.start() method

### LINKED ISSUE #6255: Session has `failfast` configuration attribute
The `Session` now has a `failfast` boolean, which is used to configure the behavior of only the `Session.start()` method and not referenced externally or changed after the `Session` has initialized. Therefore, it should not be bound as a `Session` attribute.

https://github.com/Tribler/tribler/blob/5cb1ee094e5312ede323b4bb1166dea85c379768/src/tribler-core/tribler_core/components/base.py#L40-L49

If there is supposed to be a stable reference to this configuration for external reference, the `failfast` should be part of the `TriblerConfig`. If it is only to be used as it is now (just internally in `Session.start()`), it should be an argument to the `Session.start()` method.

kozlovsky:
Fixes #6255

### DIFF
--- a/src/tribler-core/tribler_core/components/base.py
+++ b/src/tribler-core/tribler_core/components/base.py
@@ -46,7 +46,6 @@ class Session:
         self.shutdown_event: Event = shutdown_event or Event()
         self.notifier: Notifier = notifier or Notifier()
         self.components: Dict[Type[Component], Component] = {}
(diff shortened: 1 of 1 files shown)


qstokkink (approved):
Thanks!

PR #6291, opened 2021-08-25
Allow optional dependencies between components

kozlovsky:
This PR adds an important possibility for components to have optional dependencies.

Basically, it means that a component B can declare: "I don't need to have component A enabled, but if it is enabled, it should be activated before me".

```python
# inside component B:
async def run(self):
    component_a = await self.use(ComponentA, required=False)
    if component_a.enabled:
         foo = component_a.foo
         bar = component_a.bar
     else:
         foo = bar = None
```

Optional component dependencies are necessary to fix various scripts like `run_tunnel_helper.py`, as not all components are used in such scripts.


qstokkink (dismissed):
Seems fine to merge as-is.

Some suggestions for improvement (not critical/nitpicks):
 - Refactor out duplicate code in tests into common functions to reduce duplication.
 - Get rid of the `loop` fixture in these tests. If the fixture is not used (as reported Pylint) it has some hidden side effects that are needed to configure this test (or alternatively: it does nothing). This implies it should've been set somewhere higher up in the fixture hierarchy.

kozlovsky:
@qstokkink Thanks! I'll refactor the tests code a bit later to remove duplication.
Regarding the `loop` fixture, you are right, and it is not necessary in these tests. I removed it.

qstokkink (approved):

PR #6292, opened 2021-08-25
Fix run tunnel helper

kozlovsky:
This PR fixes `run_tunnel_helper.py` scripts so it can work with component architecture.
It depends on #6291 and can be merged after #6291 is merged.

### DIFF
--- a/src/tribler-core/run_tunnel_helper.py
+++ b/src/tribler-core/run_tunnel_helper.py
@@ -23,39 +23,6 @@ from tribler_core.utilities.path_util import Path
 logger = logging.getLogger(__name__)
 
 
-class PortAction(argparse.Action):
-    def __call__(self, parser, namespace, values, option_string=None):
-        if not 0 < values < 2**16:
-            raise argparse.ArgumentError(self, "Invalid port number")
-        setattr(namespace, self.dest, values)
-
-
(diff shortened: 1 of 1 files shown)


src/tribler-core/run_tunnel_helper.py
```diff
@@ -90,85 +130,85 @@ async def signal_handler(sig):
...
+    def __call__(self, parser, namespace, values, option_string=None):
+        if not 0 < values < 2**16:
+            raise argparse.ArgumentError(self, "Invalid port number")
+        setattr(namespace, self.dest, values)
+
+
+class IPAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        try:
+            inet_aton(values)
```
qstokkink, line 191: Please consider using `ipaddress.IPv4Address()` instead of `inet_aton()`. This allows you to only catch `AddressValueError` exceptions (instead of general exceptions) and gives you a message that shows exactly what is wrong with the given IP address.

qstokkink (approved):
Looks good 👍

src/tribler-core/run_tunnel_helper.py
kozlovsky: Thanks, will made another PR with this fix

PR #6293, opened 2021-08-25
Fix tunnel helper arguments

kozlovsky:
This PR fixes missed the first argument when raising the `ArgumentError` exception, and also it creates an `ipaddress.IPv4Address` instance instead of calling the `inet_aton()` function.

Also, it reformats the code a bit to satisfy the linter.


qstokkink (approved):
Very nice 👍 Thanks!

PR #6316, opened 2021-09-16
Fix #6283

### LINKED ISSUE #6283: [main] The TunnelCommunity's `dht_provider` should be set to `None` without dht
Related to #6281

https://github.com/Tribler/tribler/blob/5cb1ee094e5312ede323b4bb1166dea85c379768/src/tribler-core/tribler_core/components/implementation/tunnels.py#L49

Given the config option `dht = False`, which is now properly respected, the ipv8 component may have no `dht_discovery_community`. In this case, `dht_provider` given to the `TunnelCommunity` should be set to `None` (instead of `DHTCommunityProvider(None, config.ipv8.port)`) or not given at all.

See also https://github.com/Tribler/py-ipv8/blob/master/ipv8/messaging/anonymization/community.py#L102

drew2a:
This PR fixes  #6283

### DIFF
--- a/src/tribler-core/tribler_core/components/implementation/tunnels.py
+++ b/src/tribler-core/tribler_core/components/implementation/tunnels.py
@@ -46,6 +46,8 @@ class TunnelsComponentImp(TunnelsComponent):
         else:
             tunnel_cls = TriblerTunnelCommunity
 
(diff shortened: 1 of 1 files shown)


drew2a:
retest it please

drew2a:
retest this please

qstokkink (dismissed):

PR #6318, opened 2021-09-16
Set exit node cache path on Tunnel community

xoriole:
Exit node cache path is not currently set on initialization of the Tunnel Community because of which the cache is not saved during shutdown


xoriole:
retest this please

src/tribler-core/tribler_core/components/implementation/tunnels.py
```diff
@@ -47,6 +49,7 @@ async def run(self):
             tunnel_cls = TriblerTunnelCommunity
 
         provider = DHTCommunityProvider(dht_discovery_community, config.ipv8.port) if dht_discovery_community else None
+        exitnode_cache = Path(config.state_dir) / "exitnode_cache.dat"
```
ichorid, line 52: Forcing `config.state_dir` to `Path` is unnecessary: `TriblerConfig` object ensures the thing is always a `Path`.

ichorid (changes requested):

ichorid (approved):

PR #6321, opened 2021-09-17
Fix #6237

### LINKED ISSUE #6237: [7.10.0] KeyError: 'health' in tribler_gui.widgets.tablecontentdelegate.CheckClickedMixin.check_clicked
[Here](https://github.com/Tribler/tribler/blob/4f8446c60a7c68fd8455c27b3fe845c5ecb2866c/src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py#L96):
```python
class CheckClickedMixin:
    def check_clicked(self, event, _, __, index):
        model = index.model()
        data_item = model.data_items[index.row()]
        if (
            event.type() == QEvent.MouseButtonRelease
            and model.column_position.get(self.column_name, -1) == index.column()
            and data_item[index.model().columns[index.model().column_position[self.column_name]].dict_key] != ''  # <-- here
```

The previous version of code used `column_position.get(self.column_name, -1)` instead of `column_position[self.column_name]`, probably we need to return this back

ichorid:
Fixes #6237

Fix GUI crash on CheckClickedMixin accessing non-existing column

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
+++ b/src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
@@ -90,10 +90,13 @@ class CheckClickedMixin:
     def check_clicked(self, event, _, __, index):
         model = index.model()
         data_item = model.data_items[index.row()]
+        if column_position := model.column_position.get(self.column_name) is None:
+            return False
+        attribute_name = model.columns[column_position].dict_key
         if (
             event.type() == QEvent.MouseButtonRelease
-            and model.column_position.get(self.column_name, -1) == index.column()
-            and data_item[index.model().columns[index.model().column_position[self.column_name]].dict_key] != ''
+            and column_position == index.column()
+            and data_item.get(attribute_name, '') != ''
         ):
             self.clicked.emit(index)
             return True


src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
```diff
@@ -90,10 +90,11 @@ class CheckClickedMixin:
     def check_clicked(self, event, _, __, index):
         model = index.model()
         data_item = model.data_items[index.row()]
+        column_position = model.column_position.get(self.column_name, -1)
         if (
             event.type() == QEvent.MouseButtonRelease
-            and model.column_position.get(self.column_name, -1) == index.column()
-            and data_item[index.model().columns[index.model().column_position[self.column_name]].dict_key] != ''
+            and column_position == index.column()
+            and data_item[model.columns[column_position].dict_key] != ''
```
kozlovsky, line 97: The current version of code (which contains a bug) looks like an attempt to do a logical short-circuiting: if clause A

```python
model.column_position.get(self.column_name, -1) == index.column()
```

is not `True`, then the next clause B

```python
data_item[index.model().columns[index.model().column_position[self.column_name]].dict_key] != ''
```

should not be executed.

Apparently, it fails to do the short-circuiting properly, and clause B is executed even if 

```python
model.column_position.get(self.column_name, -1)
```

is equals to `-1`. In my understanding that means that `index.column()` can be `-1` as well.

The new version of code simplifies the expression, but as it seems to me it still can execute updated clause B2:

```python
data_item[model.columns[column_position].dict_key] != ''
```
when `column_position` is equal to `-1`.

Maybe it is necessary to add another pre-condition `and column_position >= 0` before B2?

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
ichorid: Strictly speaking, you're right: if `index.column()` equals `-1` the next line will result in taking an incorrect column index. However, `index.column()` can never be `-1` in QT for properly constructed tables.
Anyway, your comment means this is unobvious to a non-QT programmer, so I should add at least a comment.

kozlovsky: I still think it is possible for `index.column()` to be equal to `-1` in some corner cases. But with your latest change it should be processed correctly 👍

src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
```diff
@@ -90,10 +90,13 @@ class CheckClickedMixin:
     def check_clicked(self, event, _, __, index):
         model = index.model()
         data_item = model.data_items[index.row()]
+        column_position = model.column_position.get(self.column_name, -1)
+        attribute_name = model.columns[column_position].dict_key
```
kozlovsky, line 94: It is probably necessary to use `.get` or `in` here

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/widgets/tablecontentdelegate.py
ichorid: `columns` is a tuple. I'll just add a check above...

kozlovsky (approved):

PR #6330, opened 2021-09-20
Change gaierror import

### LINKED ISSUE #6286: [main] Don't import from `_socket`
https://github.com/Tribler/tribler/blob/3343308ff7aacb184b84255a8eb7c6eb1714ebda/src/tribler-core/tribler_core/exception_handler.py#L8

This line is importing `gaierror` from `_socket` instead of from `socket`. Importing from `_` modules should generally be avoided and I see no reason to do so here.

drew2a:
This PR fixes #6286


qstokkink (dismissed):
Thanks 👍

PR #6337, opened 2021-09-21
Fix notifier not working from a thread

### LINKED ISSUE #6305: [main] "There is no current event loop in thread" error in logs
In the logs, I occasionally see the following error:

```
[PID:7320] 2021-09-07 11:29:20,651 - ERROR <gigachannel_manager:288> GigaChannelManager.process_channel_dir_threaded(): Error when processing channel dir download: There is no current event loop in thread 'asyncio_12'.
[PID:7320] 2021-09-07 11:29:20,822 - ERROR <gigachannel_manager:288> GigaChannelManager.process_channel_dir_threaded(): Error when processing channel dir download: There is no current event loop in thread 'asyncio_6'.
```

It also seems that one of my channels is stuck processing (at 47%), while the channel download has finished.

ichorid:
fixes #6305

### DIFF
--- a/src/tribler-core/tribler_core/notifier.py
+++ b/src/tribler-core/tribler_core/notifier.py
@@ -14,6 +14,10 @@ class Notifier:
     def __init__(self):
         self._logger = logging.getLogger(self.__class__.__name__)
         self.observers = {}
+        # We have to note the event loop reference, because when we call "notify" from an external thread,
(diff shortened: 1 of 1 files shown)


devos50:
Thanks! This seems to fix the issue for me (my channels are now making progress again) 👍

kozlovsky (approved):

PR #6338, opened 2021-09-21
Fix community component shutdown leaving stuff behind

### LINKED ISSUE #6248: [main] Community component shutdown leaves stuff behind
The Community components I've seen introduced in #6206 look something like this:

```python
class XComponentImp(XComponent):

    async def run(self):
        # other stuff omitted
        self.community = community

        ipv8.strategies.append((RandomWalk(community), 30))
        ipv8.strategies.append((RemovePeers(community), INFINITE))

        ipv8.overlays.append(community)

    async def shutdown(self):
        await self.community.unload()
```

This is filling IPv8's `strategies` and `overlays` lists manually and then not cleaning this up on shutdown. Only the `community` is being unloaded. The registered strategies in `strategies` should be removed at the very least or they will keep being called until IPv8 is unloaded, which also keeps a reference to the `community` instance and will lead to the `community` still being called even though it's unloaded.

I'd suggest doing this as follows (using the XComponent example):

```python
class XComponentImp(XComponent):

    async def run(self):
        # other stuff omitted
        self.community = community

        # The old way will also work, this is just a bit less work; no need to add to ipv8.overlays yourself.
        ipv8.add_strategy(community, RandomWalk(community), 30)
        ipv8.add_strategy(community, RemovePeers(community), INFINITE)

    async def shutdown(self):
        # This removes the overlay and all associated strategies.
        # You can also remove your stuff from the strategies and overlays manually if you really want to
        await ipv8.unload_overlay(community)
```

ichorid:
fixes #6248

### DIFF
--- a/src/tribler-core/tribler_core/components/implementation/bandwidth_accounting.py
+++ b/src/tribler-core/tribler_core/components/implementation/bandwidth_accounting.py
@@ -1,6 +1,7 @@
 from ipv8.peerdiscovery.discovery import RandomWalk
 
 from tribler_common.simpledefs import STATEDIR_DB_DIR
+
 from tribler_core.components.interfaces.bandwidth_accounting import BandwidthAccountingComponent
 from tribler_core.components.interfaces.ipv8 import Ipv8Component
 from tribler_core.components.interfaces.reporter import ReporterComponent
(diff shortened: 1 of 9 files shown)


qstokkink (approved):
Nice, fixes the core issue 👍 

Note: the switch from `strategies.append + overlays.append` to `add_strategy`, documented #6248, should be moved to a new "enhancement" issue after this PR is merged.

PR #6346, opened 2021-09-22
Switch to modern IPv8 Community addition

### LINKED ISSUE #6341: Switch to modern IPv8 Community addition
_Inherited from #6248_

Community loading now looks something like this:

```python
class XComponentImp(XComponent):

    async def run(self):
        # other stuff omitted
        self.community = community

        ipv8.strategies.append((RandomWalk(community), 20))
        ipv8.strategies.append((RemovePeers(community), INFINITE))

        ipv8.overlays.append(community)
```

The more modern slightly-shorter way to do this, is as follows (note that there is no need to access the `strategies` and `overlays` containers directly and that this is thread-safe):

```python
class XComponentImp(XComponent):

    async def run(self):
        # other stuff omitted
        self.community = community

        ipv8.add_strategy(community, RandomWalk(community), 20)
        ipv8.add_strategy(community, RemovePeers(community), INFINITE)
```

drew2a:
This PR closes #6341

### DIFF
--- a/src/tribler-core/tribler_core/components/implementation/bandwidth_accounting.py
+++ b/src/tribler-core/tribler_core/components/implementation/bandwidth_accounting.py
@@ -43,7 +43,6 @@ class BandwidthAccountingComponentImp(BandwidthAccountingComponent):
 
         community.bootstrappers.append(ipv8_component.make_bootstrapper())
 
-        ipv8.overlays.append(community)
         self.community = community
 
         rest_manager.get_endpoint('trustview').bandwidth_db = community.database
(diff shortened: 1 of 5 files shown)


qstokkink (changes requested):
The calls to `ipv8.overlays.append(community)` are covered by `ipv8.add_strategy()` and should be removed (the overlays are now being added twice).

<details><summary>Details/Trivia</summary>

The `add_strategy()` method adds to the `ipv8.overlays` and makes sure that there are no duplicates in the `ipv8.overlays` data structure. So, _technically_ you would only see an overlay appear twice if `ipv8.overlays.append()` is called after `ipv8.add_strategy()`. Even so, the call to `ipv8.overlays.append()` is unnecessary.

</details>

qstokkink (approved):
Thanks 😃

drew2a:
@qstokkink you are always welcome!

PR #6354, opened 2021-09-24
Fix #6263

### LINKED ISSUE #6263: [main] Tribler gets stuck on loading screen after component crash
After the crash of #6262 my Tribler window was stuck in the "Loading..." screen. Even worse, it was impossible to CTRL+C to death.

screenshot

ichorid:
fixes #6263 

Also, move `osutils.py` to `tribler-common`, since those functions are used by both GUI and Core

Also, fix some tests flaky (`"ValueError: set_wakeup_fd only works in main thread"`) from not using `pytest.mark.asyncio`.


ichorid:
retest this please

drew2a (dismissed):

qstokkink:
retest this please

PR #6376, opened 2021-09-28
Implement row deselection on mouse leave

devos50:
When the mouse leaves the table, it should deselect the currently selected row. This PR implements this behaviour. It also captures the event when moving the mouse through the (top) header of the table, which required the implementation of an event filter. Note that sometimes rows remain selected when moving the mouse quickly out of the viewport, but I believe this is acceptable for now.

I also removed the redundant `hoverrow` variable in the table delegate.


drew2a (dismissed):

devos50:
Is there anything needed from my side to merge this PR?

ichorid (dismissed):

devos50:
retest this please

devos50:
Encountered #6382, does not seem to be related to this PR since it only contains GUI changes.

PR #6381, opened 2021-09-28
Fixes #6360: Optional dependency on REST manager in components

### LINKED ISSUE #6360: [Components] Dependency on RESTComponent can be converted to "Optional"
During the refactoring (#6335), we found that in many Components, dependency on [RESTComponent](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/implementation/restapi.py#L10) is marked as "required".

But almost in all places, it can be converted to "optional".

We agreed that @kozlovsky will check if there is any "hidden" reason for "required" dependency. If not, we can easily convert all calls to "optional".

kozlovsky:
This PR makes the dependency of components on REST manager optional. It adds a base class `RestfulComponent`, which provides three methods:

The `init_endpoints` method simplifies endpoint initialization and can be used inside a component's `run` method to inject values to REST endpoints:

```python
async def run(self):
    ...
    await self.init_endpoints(
        ["endpoint_name1", "endpoint_name2"]
        [(attr_name1, value1), (attr_name2, value2)]
    )
```
The method sets specified attributes to all enlisted endpoints. If RESTComponent or some specific endpoint is not available, it just skipped silently.

It also is possible to initialize `ipv8` endpoints as well by specifying the `ipv8` instance.

```python
        await self.init_ipv8_endpoints(self._ipv8, ['tunnel'])
```
This way it becomes not necessary to explicitly access RESTComponent and REST endpoints in a component's code.

The `shutdown` method of RestfulComponent will release endpoints by replacing previously set values to `None`:

The `set_readable_status` abstracts out the setting Tribler user-readable status to a specific value.
```python
async def run(self):
    ...
    await self.set_readable_status(STATE_START_WATCH_FOLDER)
```

### DIFF
--- a/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
+++ b/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
@@ -47,8 +47,8 @@ class BandwidthAccountingComponent(RestfulComponent):
         community.bootstrappers.append(ipv8_component.make_bootstrapper())
 
         self.community = community
-        await self.init_endpoints(['trustview', 'bandwidth'],
-                                  [('bandwidth_db', community.database), ('bandwidth_community', community)])
+        await self.init_endpoints(endpoints=['trustview', 'bandwidth'],
+                                  values={'bandwidth_db': community.database, 'bandwidth_community': community})
 
     async def shutdown(self):
         await super().shutdown()
--- a/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
+++ b/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
@@ -51,5 +51,5 @@ class BandwidthAccountingComponent(RestfulComponent):
                                   [('bandwidth_db', community.database), ('bandwidth_community', community)])
 
     async def shutdown(self):
-        self.release_endpoints()
+        await super().shutdown()
         await self._ipv8.unload_overlay(self.community)
--- a/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
+++ b/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
@@ -23,7 +23,7 @@ class BandwidthAccountingComponent(RestfulComponent):
     _ipv8: IPv8
 
     async def run(self):
-        await self.get_component(ReporterComponent)
+        await super().run()
         await self.get_component(UpgradeComponent)
         config = self.session.config
 
--- a/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
+++ b/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
@@ -1,10 +1,13 @@
 from ipv8.peerdiscovery.discovery import RandomWalk
+
 from ipv8_service import IPv8
+
 from tribler_common.simpledefs import STATEDIR_DB_DIR
+
 from tribler_core.components.base import Component
 from tribler_core.components.ipv8 import Ipv8Component
 from tribler_core.components.reporter import ReporterComponent
-from tribler_core.components.restapi import RESTComponent
+from tribler_core.components.restapi import RestfulComponent
 from tribler_core.components.upgrade import UpgradeComponent
 from tribler_core.components.bandwidth_accounting.community.community import (
     BandwidthAccountingCommunity,
@@ -14,10 +17,9 @@ from tribler_core.components.bandwidth_accounting.db.database import BandwidthDa
 from tribler_core.restapi.rest_manager import RESTManager
 
 
-class BandwidthAccountingComponent(Component):
+class BandwidthAccountingComponent(RestfulComponent):
     community: BandwidthAccountingCommunity
 
-    _rest_manager: RESTManager
     _ipv8: IPv8
 
     async def run(self):
@@ -28,9 +30,6 @@ class BandwidthAccountingComponent(Component):
         ipv8_component = await self.require_component(Ipv8Component)
         self._ipv8 = ipv8_component.ipv8
         peer = ipv8_component.peer
-
-        rest_component = await self.require_component(RESTComponent)
-        self._rest_manager = rest_component.rest_manager
 
         if config.general.testnet or config.bandwidth_accounting.testnet:
             bandwidth_cls = BandwidthAccountingTestnetCommunity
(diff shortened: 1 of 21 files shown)


kozlovsky:
retest this please

kozlovsky:
retest this please

src/tribler-core/tribler_core/components/restapi.py
```diff
@@ -1,9 +1,59 @@
...
+        super().__init__()
+        self.endpoint_attrs = set()
+
+    async def set_readable_status(self, readable_status):
+        rest_component = await self.get_component(RESTComponent)
+        if rest_component:
+            state_endpoint: StateEndpoint = rest_component.rest_manager.get_endpoint('state')
+            state_endpoint.readable_status = readable_status
+
+    async def init_endpoints(self, endpoints, values, ipv8=None, ipv8_endpoints=()):
```
ichorid, line 27: When reading through this code I thought at first that the method is modal: if `ipv8` arg is not None, act on IPv8 endpoints.

I'd rather split this method in two:
`init_endpoints` and `init_ipv8_endpoints`
Otherwise, the definition is misleading and unobvious

src/tribler-core/tribler_core/components/upgrade.py
```diff
@@ -22,8 +23,9 @@ async def run(self):
...
-        rest_component = await self.require_component(RESTComponent)
-        rest_component.rest_manager.get_endpoint('upgrader').upgrader = self.upgrader
-        rest_component.rest_manager.get_endpoint('state').readable_status = STATE_UPGRADING_READABLE
-
+        await self.init_endpoints(['upgrader'], [('upgrader', self.upgrader)])
+        await self.set_readable_status(STATE_UPGRADING_READABLE)
         await self.upgrader.run()
+
+    async def shutdown(self):
+        self.release_endpoints()
```
ichorid, line 31: I would expect `release_endpoints` to be called automatically on every call to `shutdown` of `RestfulComponent`. That's why we subclass stuff, right?

ichorid (changes requested):

src/tribler-core/tribler_core/components/restapi.py
drew2a: Agree 👍

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
```diff
@@ -14,10 +17,9 @@
...
 
 
-class BandwidthAccountingComponent(Component):
+class BandwidthAccountingComponent(RestfulComponent):
     community: BandwidthAccountingCommunity
 
-    _rest_manager: RESTManager
     _ipv8: IPv8
 
     async def run(self):
```
drew2a, line 25: If you add a class to the chain of inheritance, it is almost always better to call overridden methods of a base class, eg:

```python
class BandwidthAccountingComponent(RestfulComponent):
    community: BandwidthAccountingCommunity

    _ipv8: IPv8

    async def run(self):
        await super().run()
```

Overwise you change the behavior of base class kinda in an "illegal" way.

drew2a (changes requested):
I appreciate your care about developers in terms of "encapsulating logic to a base class" and simplify the code of components.

I am just afraid that it adds cognitive complexity to the components themselves.

Now operations with REST are not as obvious as they were before.

I would suggest just making a dependency of `RESTComponent` optional (with adding the corresponding code) without incapsulating it in a base class.

src/tribler-core/tribler_core/components/restapi.py
kozlovsky: Done

src/tribler-core/tribler_core/components/upgrade.py
kozlovsky: Done

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
kozlovsky: Agree, changed it to calling the base method

kozlovsky:
> I appreciate your care about developers in terms of "encapsulating logic to a base class" and simplifying the code of components.
>
> I am just afraid that it adds cognitive complexity to the components themselves.
>
> Now operations with REST are not as obvious as they were before.
>
> I would suggest just making a dependency of RESTComponent optional (with adding the corresponding code) without incapsulating it in a base class.

@drew2a in general I agree that inheritance adds some cognitive complexity, and try to avoid using it.

But in this specific case I think that having the logic of endpoint initialization repeated in many components is a bit cumbersome for the following reason:
1. It is possible, that RESTComponent is unavailable.
2. Even if RESTComponent if present, it is possible that some specific subset of rest endpoints which a component is trying to initialize is disabled, so it is necessary to check availability of some endpoint before initialization
3. During the shutdown we again need to skip cleanup of endpoints which were disabled.

For example, this fragment of TunnelComponent initialization code needs to check the endpoints availability:
```python
        rest_component = await self.get_component(RESTComponent)
        if rest_component:
            rest_component.rest_manager.get_endpoint('ipv8').endpoints['/tunnel'].initialize(self._ipv8)
            if download_component:
                rest_component.rest_manager.get_endpoint('downloads').tunnel_community = community

            debug_endpoint = rest_component.rest_manager.get_endpoint('debug')
            debug_endpoint.tunnel_community = community
```
Now it looks like this:
```python
    await self.init_endpoints(['downloads', 'debug'], [('tunnel_community', community)])
    await self.init_ipv8_endpoints(self._ipv8, ['tunnel'])
```

In my opinion, extracting endpoint initialization logic to a separate component makes the remaining code of components easier to understand.

kozlovsky:
retest this please

drew2a:
@kozlovsky you are right, agree.

By the way, to increase readability we may use named arguments, eg:

```python
    await self.init_endpoints(endpoints = ['downloads', 'debug'], values = [('tunnel_community', community)])
    await self.init_ipv8_endpoints(self._ipv8, endpoints = ['tunnel'])
```

or, even, use a dict instead a list of tules:

```python
    await self.init_endpoints(endpoints = ['downloads', 'debug'], values = {'tunnel_community': community})
    await self.init_ipv8_endpoints(self._ipv8, endpoints = ['tunnel'])
```

ichorid:
> @kozlovsky you are right, agree.
> 
> By the way, to increase readability we may use named arguments, eg:
> 
> ```python
>     await self.init_endpoints(endpoints = ['downloads', 'debug'], values = [('tunnel_community', community)])
>     await self.init_ipv8_endpoints(self._ipv8, endpoints = ['tunnel'])
> ```
> 
> or even use a dict instead a list of tules:
> 
> ```python
>     await self.init_endpoints(endpoints = ['downloads', 'debug'], values = {'tunnel_community': community})
>     await self.init_ipv8_endpoints(self._ipv8, endpoints = ['tunnel'])
> ```

@drew2a is right. Just use a single, dict argument.

kozlovsky:
@drew2a @ichorid I changed arguments according to your suggestion

ichorid (approved):

PR #6396, opened 2021-09-30
Add TagComponent

drew2a:
This PR adds TriblerComponent that includes:
* TagCommunity
* Tag entity in DB


src/tribler-core/tribler_core/components/tag/tag_component.py
```diff
@@ -0,0 +1,43 @@
...
+        community = TagCommunity(
+            ipv8_component.peer,
+            self._ipv8.endpoint,
+            self._ipv8.network,
+            metadata_store=metadata_store_component.mds
+        )
+        self.community = community
+
+        self._ipv8.add_strategy(community, RandomWalk(community), 30)
+        self._ipv8.add_strategy(community, RemovePeers(community), INFINITE)
```
qstokkink, line 34: **issue:** The target peers of the `RandomWalk` are equal to the default `Community` `max_peers`. Therefore, this code should also follow the default target peers of `20` instead of `30`.

**issue:** This code is removing peers from the common peer pool `self._ipv8.network`. There should really only ever be one strategy removing from the common peer pool (which is already handled by the `DiscoveryCommunity` churn strategy and modulated by the `GoldenRatio` strategy). This line should be removed.

**suggestion:**
```suggestion
        self._ipv8.add_strategy(community, RandomWalk(community), 20)
```

**thought:** I've seen these issues before in other PRs. Perhaps the default way to add new `Community` instances should be in some common base class or perhaps as a method in the IPv8 component?

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
```diff
@@ -0,0 +1,36 @@
...
+            creator_public_key=self.creator_public_key.encode("utf-8"),
+            creator_sign=self.creator_sign.encode("utf-8")
+        ))
+
+
+@vp_compile
+class TagMessage(VariablePayload):
+    msg_id = 1
+    format_list = ['varlenI']
+    names = ['json']
```
devos50, line 36: **Question:** Why serialize everything in a single JSON blob? For readability and consistency, Unless you have a good reason for this, I would suggest following the same procedure as the messages in other communities (e.g., [the popularity community](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/modules/popularity/payload.py#L21)) 👍 .

src/tribler-core/tribler_core/components/tag/community/tag_community.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TagCommunity(TriblerCommunity):
+    """
+    Community for disseminating tags across the network.
+    """
+
+    community_id = unhexlify('042020c5e5e2ee0727fe99d704b430698e308d98')
+
+    def __init__(self, *args, metadata_store: MetadataStore = None,
+                 gossip_interval_for_tags_in_sec=5, **kwargs):
```
devos50, line 24: **Suggestion:** Move this 5 seconds into a global constant (like `GOSSIP_RANDOM_PEERS_COUNT`).

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
```diff
@@ -0,0 +1,36 @@
...
+from pydantic import BaseModel
+
+from ipv8.messaging.lazy_payload import VariablePayload, vp_compile
+
+
+class TagModel(BaseModel):
+    infohash: str
+    tag: str
+    creator_public_key: str
+    creator_sign: str
```
devos50, line 12: **Issue:** There currently is no primitive to remove tags. Earlier, we had a brief discussion on whether to use different message types to do so, or to add an additional flag in each payload (e.g., `removed` that is set to `1` if the tag is removed). I would opt for adding this flag in a payload so that we have a single message type that is being gossiped around.

src/tribler-core/tribler_core/components/tag/tag_component.py
drew2a: Thank you for your explanation. The code has been fixed.

You are right it is not clear, what is the best way to create just a default `Community`

>should be in some common base class or perhaps as a method in the IPv8 component?

Something like this?

```python
class Ipv8Component(RestfulComponent):
    ipv8: IPv8
    peer: Peer
    dht_discovery_community: Optional[DHTDiscoveryCommunity] = None

    ...

    def add_default_community(self, community):
        self._ipv8.add_strategy(community, RandomWalk(community), 20)
        community.bootstrappers.append(self.make_bootstrapper())
```

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
drew2a: It is easier to operate with messages in JSON format and it increases redability as well.
I can (and will) add `pydantic` validators to it (eg: to validate string length limit). 
This format also will allow us to create custom behavior for processing different types of these messages in the future.

Like
```python
class TagModel(BaseModel):
    infohash: str
    tag: str
    creator_public_key: str
    creator_sign: str
    removed: Optional[bool] = False  # this field from the future
```

Unfortunately, ipv8 doesn't support any modern tools for serializing and validating messages.

src/tribler-core/tribler_core/components/tag/community/tag_community.py
drew2a: What is the reason for moving it to a global constant? 

If it passes to the constructor, I can customize class behavior in a natural way.

Example:

https://github.com/drew2a/tribler/blob/28c8c8c8965c4147a14814224f3cb411df3b2a13/src/tribler-core/tribler_core/components/tag/community/tests/test_tag_community.py#L33-L34

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
drew2a: Agree. Will add this (or something like this) in near future.

The PR is a draft. It is not ready yet.

I've published it only for one reason: to give a rough view of how it will look like.

src/tribler-core/tribler_core/components/tag/community/tag_community.py
```diff
@@ -0,0 +1,54 @@
...
+        dictionary = json.loads(payload.json.decode("utf-8"))
+        tag_model = TagModel.parse_obj(dictionary)
+        self.logger.debug(f'Write to db: {tag_model}')
+        try:
+            with db_session():
+                self.metadata_store.Tag(**tag_model.to_pony_dict())
+        except TransactionIntegrityError:  # ignore all duplicates
+            pass
+
+    def gossip_random_tags(self):
```
ichorid, line 45: **Suggestion**: consider using pull-based gossip to prevent the simplest forms of spam attacks.

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
devos50: @ichorid good point, thanks! I remember that I used JSON once for some experiments with the market community and I had issues with signature verification indeed (so I just ignored the signature validation outcome in the experiment setting).

src/tribler-core/tribler_core/components/metadata_store/db/orm_bindings/tag.py
```diff
@@ -0,0 +1,15 @@
...
+from pony.orm import composite_key
+
+
+def define_binding(db):
+    class Tag(db.Entity):
+        id = orm.PrimaryKey(int, auto=True)
+        infohash = orm.Required(bytes)
+        tag = orm.Required(str)
+        creator_public_key = orm.Required(bytes)
+        creator_sign = orm.Required(bytes)
```
ichorid, line 11: **Nitpick**: it is better to name these `public_key` and `signature` for two reasons:
 * `sign` is a verb while `signature` is a noun
 * it is better to stick to the same naming scheme Channels use, for consistency

ichorid (changes requested):

src/tribler-core/tribler_core/components/tag/community/tag_community.py
drew2a: I remember that sentence of yours.
However, after reasoning, I have come to the conclusion that in terms of "spam attacks"  it doesn't matter whether we use push or pull.

Both of these approaches put approximately the same load on the lower layer protocol (ipv8). We can filter spam messages only on the Community level.

Pull-based gossip will lead to the use of a combination of a dictionary (for storing outcoming requests) and a `NumberCache`.

Push-based gossip needs neither.

To add simple prevention of "spam attack" for "push-based" gossip we just may control the frequency of incoming messages and drop all messages that come frequently than it is accepted (or peers who have exceeded the limit).  

We also add a filter for the GUI interface "only show a tag if it has been suggested by two distinct identities".

From my POV it should be enough for the first version of Tag System.

That's why I choose push-based gossip. It gives roughly the same level of spam protection, but it is simplier.

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
devos50: > It is easier to operate with messages in JSON format and it increases readability as well.

IPv8 message definition also results in readable code, see for example the message definition in [`BandwidthTransactionPayload`](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/bandwidth_accounting/community/payload.py#L13). It makes it immediately clear which fields are in the payload, and in which order.

> This format also will allow us to create custom behavior for processing different types of these messages in the future.

I'm not sure how many different types of messages we will make in the future, but going for a full-flexible message format looks unnecessary to me at this point. We have (and had) many communities running with statically defined messages and the need to change the format of a particular message has not occurred often. And if a particular message should be changed, we either make a new message or can change the community identifier.

I see an advantage of using JSON in the payload though: it can sidestep an additional conversion between two classes. In the `BandwidthAccountingCommunity` for example, we have separate `BandwidthTransactionPayload` and `BandwidthTransaction` classes. It would be great if subclasses of `VariablePayload` could be used directly as pedantic `BaseModel`, but for now I do not see this as a must have.

Minor note: JSON serialization is slower than IPv8 serialization if not using an optimized JSON serialization library such as `orjson`.

src/tribler-core/tribler_core/components/tag/tag_component.py
qstokkink: Yes, that should work.

One nuance is that if---at any point in the future---someone makes any part of this code multithreaded, the bootstrapper should be added to the `community` before calling `add_strategy`. Switching the two lines in `add_default_community` would be "safer".

drew2a: OK, cool! I can apply it to all components.

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
qstokkink: I think all of the pros and cons of JSON have been mentioned here. I just wanted to quickly address the following:

> Unfortunately, ipv8 doesn't support any modern tools for serializing and validating messages.

IPv8 supports the use of pretty much any and every serialization you can think of; any overlay designer can use whatever custom serialization they like, as long as it converts to and from `bytes`. For example, if you would rather use a pydantic validator to guarantee that an infohash string is 20 bytes long instead of using the IPv8 `20s` format, you can do so. Official IPv8 policy: use whatever you like.

Here's a fully functional example that shows how to seamlessly hook in JSON serialization https://github.com/Tribler/py-ipv8/issues/969#issuecomment-797933526 Specifically, look at `MessageKeepDict` in the "Click me for full runnable example". In this example you can swap out `json.dumps()` and `json.loads()` if you want to use some other serialization method. If you don't care about message truncation, you can even save 4 more bytes by swapping out `varlenH` for `raw` in `MessageKeepDict`.

If you're reusing your custom serialization in multiple messages, you can also take custom serialization one step further and add a custom serialization entry to your `Community` instance's `Serializer` instance ([click me for example](https://github.com/Tribler/py-ipv8/blob/7a94e31c1036b7d13491dc63a8e8d55ed34f0c7d/ipv8/messaging/anonymization/community.py#L163)).

src/tribler-core/tribler_core/components/tag/community/tag_community.py
ichorid: > It gives roughly the same level of spam protection

No, it definitely does not give the same level of protection, unless you add lots of things like controlling the rate of incoming packets - which defies the whole point of simplicity. 

It will take one mischievous student to take down the whole network in flames, if you use **both** blind transitivity **and** push-based gossip, even if you add rate control.

Anyways, that's your choice. Let's just hope you're still with us when :hankey:  hits the fan :wink:

devos50: For comparison, we are using pull-based gossip in the `BandwidthAccountingCommunity`, see [here](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/bandwidth_accounting/community/bandwidth_accounting_community.py#L145). Unfortunately, I don't exactly remember the reasoning behind that design 😕 . We can discuss it further on Monday.

src/tribler-core/tribler_core/components/tag/community/tag_community.py
devos50: Yes, it's possible to customise this number, but extracting this number to a constant can improve readability and shows a developer that this is a default setting. It is also more in line with our current code practices, see for example [here](https://github.com/Tribler/py-ipv8/blob/master/ipv8/community.py#L32) and [here](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/modules/remote_query_community/tests/test_eva_protocol.py#L27). I dislike the declaration of constants without any context since we had too many of them in the past, which made it very hard to understand what's going on.

If we agree that extracting default constants to a global variable is unnecessary, we should at least try to be consistent 👍 .

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
drew2a: @qstokkink what I would really like to see in ipv8:

```python
@dataclass
class MyMessage:
    id: int = 1
    some_field: str = ''


class MyCommunity(Community):
        self.add_message_handler(MyMessage, self.on_message)

    @lazy_wrapper(TagMessage)
    def on_message(self, peer, payload):
        self.ez_send(peer, TagMessage(MyMessage(some_field='data')))
```

src/tribler-core/tribler_core/components/tag/db/tag_db.py
```diff
@@ -0,0 +1,164 @@
...
+
+from tribler_core.utilities.path_util import Path
+
+
+class Operation(IntEnum):
+    ADD = 1
+    REMOVE = 2
+
+
+class TagDatabase:
```
drew2a, line 16: @devos50 @kozlovsky 

The DB has been added. Feel free to review it.

If you see some misspellings or incorrectly worded sentences in the comments, please write about it.

src/tribler-core/tribler_core/components/tag/community/tag_community.py
drew2a: Fixed by extracting it to the const `DEFAULT_GOSSIP_INTERVAL_FOR_TAGS_IN_SEC`

src/tribler-core/tribler_core/components/tag/db/tag_db.py
drew2a: Tomorrow I'll add limits for the tags and tags operations. 
Still not sure how it should be implemented.

src/tribler-core/tribler_core/components/tag/tag_component.py
drew2a: Applied here: https://github.com/Tribler/tribler/pull/6402

All the changes have been applied to this branch as well.

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
drew2a: A primitive to remove tags has been added.

src/tribler-core/tribler_core/components/metadata_store/db/orm_bindings/tag.py
drew2a: Thank you.
Tables have been extracted from Channels.

src/tribler-core/tribler_core/components/tag/community/tag_payload.py
qstokkink: @drew2a thanks for the input. This feature has my support and is captured in this (open) issue (https://github.com/Tribler/py-ipv8/issues/1011). It even has [a prototype](https://github.com/Tribler/py-ipv8/issues/1011#issuecomment-872377297), if you're feeling adventurous. Hopefully, it becomes part of the main codebase soon.

drew2a: I've rewritten TagCommunity serialization by "old native ipv8 serialization".

src/tribler-core/tribler_core/components/tag/db/tag_db.py
kozlovsky: @drew2a to implement the constraint on the maximum number of tags operations, you can add an attribute `TorrentTag.operations_count = Required(int, default=0)`. You can modify it incrementally from the Python code when adding new operations. Or you can calculate operations count on the fly and check the value before adding a new operation to the database: `torrent_tag.operations.count()`. It is simpler but a bit less efficient.

The same is for the maximum number of tags for the torrent: you can add an attribute `Torrent.tags_count = Required(int, default=0)`, or just check `torrent.tags.count()` when necessary.

devos50:
> Please note: I will add more tests in the next PR. This PR should be merged to main to let @devos50 work on the GUI part.

Feel free to add these tests in this PR, I can just rebase my branch when you make changes (adding tests to this PR should not merge conflict with my work). 👍

drew2a:
retest this please

Issue #2894, opened 2017-04-04
transitive trust based on blockchain constructs


synctext:
```PhD Thesis direction```

By using the learn-by-doing methodology we will create a transitive trust solution.
Specifically we will design, implement, deploy, and incrementally improve transitive trust solution using our own append-only TrustChain solution.

Steps and dependancies:
- [x] True halves #2135 
- [x] live edges #2458 
- [ ] Gumby test coverage and deployment
- [ ] Remove Bloom filters and use Live Edge walking + record record + growing interval construct
![live_edge](https://cloud.githubusercontent.com/assets/325224/24652899/de5d25a0-1933-11e7-9bb3-f6f5640c4ee4.png)
- [x] Creat N-Hop-Sync across TrustChain to replace Full-Sync that fails to scale.
- [ ] Move TrustChain down from community level to replace discovery community. First, cut Dispersy into two : #2778
- [ ] Repair scalability of AllChannel with N-Hop-Sync

pimveldhuisen:
Some work on live trust edges based in dispersy MultiChainCommunity, can be found on this branch:
https://github.com/pimveldhuisen/tribler/tree/trust-edges
Corresponding with this gumby branch:
https://github.com/pimveldhuisen/gumby/tree/trust-edges
And this DAS experiment:
https://jenkins.tribler.org/job/Tribler/job/Experiments/job/multichain/job/trust-walker_das/

qstokkink:
Live edges implementation:
https://github.com/qstokkink/tribler/tree/live_edges
Live edges gumby experiment:
https://github.com/qstokkink/gumby/tree/live_edges

qstokkink:
test

qstokkink:
Isolated 1000 node DAS5 experiment is now up and running:

![trust_0hudoo2sq1dl564hzc41 7bdbsk](https://cloud.githubusercontent.com/assets/3630389/26276083/8e15b768-3d70-11e7-86fc-959eda4a717f.png)
![trust_0lex9dy13wfudbcf49y9f2hy42g](https://cloud.githubusercontent.com/assets/3630389/26276084/92b55120-3d70-11e7-8127-a1288f3b6fd7.png)
![trust_0lfvfjtsyhduckn7xft3v100g7a](https://cloud.githubusercontent.com/assets/3630389/26276085/98063fa4-3d70-11e7-9b5c-c93a53a73a7f.png)
[...]

PR #6402, opened 2021-10-01
Introduce "default init" and "unload" methods for Ipv8Component

drew2a:
Based on https://github.com/Tribler/tribler/pull/6396

```python

class Ipv8Component(RestfulComponent):
    ipv8: IPv8

    def initialise_community_by_default(self, community):
        community.bootstrappers.append(self.make_bootstrapper())

        # Value of `target_peers` must not be equal to the value of `max_peers` for this community.
        # This causes a deformed network topology and makes it harder for peers to connect to others.
        # More information: https://github.com/Tribler/py-ipv8/issues/979#issuecomment-896643760
        random_walk_max_peers = max(10, community.max_peers - 10)
        self.ipv8.add_strategy(community, RandomWalk(community), random_walk_max_peers)

    async def unload_community(self, community):
        await self.ipv8.unload_overlay(community)
```

### DIFF
--- a/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
+++ b/src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
@@ -1,7 +1,3 @@
-from ipv8.peerdiscovery.discovery import RandomWalk
-
-from ipv8_service import IPv8
-
 from tribler_common.simpledefs import STATEDIR_DB_DIR
 
 from tribler_core.components.bandwidth_accounting.community.bandwidth_accounting_community import (
@@ -17,17 +13,14 @@ from tribler_core.components.upgrade import UpgradeComponent
 class BandwidthAccountingComponent(RestfulComponent):
     community: BandwidthAccountingCommunity
 
-    _ipv8: IPv8
+    _ipv8_component: Ipv8Component
 
     async def run(self):
         await super().run()
         await self.get_component(UpgradeComponent)
+        self._ipv8_component = await self.require_component(Ipv8Component)
+
         config = self.session.config
-
-        ipv8_component = await self.require_component(Ipv8Component)
-        self._ipv8 = ipv8_component.ipv8
-        peer = ipv8_component.peer
-
         if config.general.testnet or config.bandwidth_accounting.testnet:
             bandwidth_cls = BandwidthAccountingTestnetCommunity
         else:
@@ -35,18 +28,19 @@ class BandwidthAccountingComponent(RestfulComponent):
 
         db_name = "bandwidth_gui_test.db" if config.gui_test_mode else f"{bandwidth_cls.DB_NAME}.db"
         database_path = config.state_dir / STATEDIR_DB_DIR / db_name
-        database = BandwidthDatabase(database_path, peer.public_key.key_to_bin())
-        community = bandwidth_cls(peer, self._ipv8.endpoint, self._ipv8.network,
-                                  settings=config.bandwidth_accounting,
-                                  database=database)
-        self._ipv8.add_strategy(community, RandomWalk(community), 20)
+        database = BandwidthDatabase(database_path, self._ipv8_component.peer.public_key.key_to_bin())
+        self.community = bandwidth_cls(self._ipv8_component.peer,
+                                       self._ipv8_component.ipv8.endpoint,
+                                       self._ipv8_component.ipv8.network,
+                                       settings=config.bandwidth_accounting,
+                                       database=database)
 
-        community.bootstrappers.append(ipv8_component.make_bootstrapper())
+        self._ipv8_component.initialise_community_by_default(self.community)
 
-        self.community = community
         await self.init_endpoints(endpoints=['trustview', 'bandwidth'],
-                                  values={'bandwidth_db': community.database, 'bandwidth_community': community})
+                                  values={'bandwidth_db': self.community.database,
+                                          'bandwidth_community': self.community})
 
     async def shutdown(self):
         await super().shutdown()
-        await self._ipv8.unload_overlay(self.community)
+        await self._ipv8_component.unload_community(self.community)
--- a/src/tribler-core/tribler_core/components/ipv8.py
+++ b/src/tribler-core/tribler_core/components/ipv8.py
@@ -89,13 +89,19 @@ class Ipv8Component(RestfulComponent):
             'asyncio', 'attestation', 'dht', 'identity', 'isolation', 'network', 'noblockdht', 'overlays'
         ])
 
-    def initialise_community_by_default(self, community):
+    def initialise_community_by_default(self, community, default_random_walk_max_peers=20):
         community.bootstrappers.append(self.make_bootstrapper())
 
-        # Value of `target_peers` must not be equal to the value of `max_peers` for this community.
+        # Value of `target_peers` must not be equal to the value of `max_peers` for the community.
         # This causes a deformed network topology and makes it harder for peers to connect to others.
         # More information: https://github.com/Tribler/py-ipv8/issues/979#issuecomment-896643760
-        random_walk_max_peers = max(10, community.max_peers - 10)
+        #
+        # Then:
+        # random_walk_max_peers should be less than community.max_peers:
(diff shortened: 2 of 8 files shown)


src/tribler-core/tribler_core/components/ipv8.py
```diff
@@ -89,28 +89,40 @@ async def run(self):
...
         if bootstrap_override := self.session.config.ipv8.bootstrap_override:
             address, port = bootstrap_override.split(':')
             args = {'ip_addresses': [(address, int(port))], 'dns_addresses': []}
         return DispersyBootstrapper(**args)
 
-    def init_peer_discovery_community(self):
+    def _init_peer_discovery_community(self):
         ipv8 = self.ipv8
         community = DiscoveryCommunity(self.peer, ipv8.endpoint, ipv8.network, max_peers=100)
+        community.bootstrappers.append(self.make_bootstrapper())
```
qstokkink, line 114: **nit:** seems like you can replace this with `self.initialise_community_by_default(community)` and also shed the `ipv8.add_strategy(community, RandomWalk(community), 20)`.
Same for `_init_dht_discovery_community`. Loses 2 LOC overall.

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
```diff
@@ -17,36 +13,34 @@
...
         database_path = config.state_dir / STATEDIR_DB_DIR / db_name
-        database = BandwidthDatabase(database_path, peer.public_key.key_to_bin())
-        community = bandwidth_cls(peer, self._ipv8.endpoint, self._ipv8.network,
-                                  settings=config.bandwidth_accounting,
-                                  database=database)
-        self._ipv8.add_strategy(community, RandomWalk(community), 20)
+        database = BandwidthDatabase(database_path, self._ipv8_component.peer.public_key.key_to_bin())
+        self.community = bandwidth_cls(self._ipv8_component.peer,
+                                       self._ipv8_component.ipv8.endpoint,
+                                       self._ipv8_component.ipv8.network,
```
qstokkink, line 34: **thought/nit:** We're injecting `self._ipv8_component.ipv8.network` as a dependency and then the component itself is overwriting its input with a `Network()` of its own creation. We should, at some point (probably not this PR) move the `Network()` overwrite to this file (same for `PopularityCommunity`).

src/tribler-core/tribler_core/components/tests/test_tribler_components.py
```diff
@@ -64,6 +63,25 @@ async def test_ipv8_component(tribler_config):
...
+
+    # test discovery_community enabled
+    tribler_config.gui_test_mode = False
+    tribler_config.discovery_community.enabled = True
+    session = Session(tribler_config, [MasterKeyComponent(), RESTComponent(), Ipv8Component()])
+    with session:
+        await session.start()
+
+        comp = Ipv8Component.instance()
+        assert comp._peer_discovery_community
```
qstokkink, line 83: **suggestion:** This test now breaks with the `Arrange/Act/Assert` pattern: what I see here is `Arrange/Assert/Arrange/Assert/Arrange/Assert`. By splitting this code at the comments (turning `# test dht disabled` into `async def test_ipv8_component_dht_disabled()` and `# test discovery_community enabled` into `async def test_ipv8_component_discovery_enabled()`) you can follow the AAA pattern with very little effort.

qstokkink (dismissed):
Nice 👍

In my opinion this is good to merge as-is. However, I would urge you to consider implementing my one suggestion already in this PR.

src/tribler-core/tribler_core/components/ipv8.py
drew2a: Agree. Is 20 target peers is necessary for the RandomWalk strategy for DiscoveryCommunity? 

Why I asking is because  if I replace it by `self.initialise_community_by_default(community)`, then `target_peers` become `max_peers-10 == 90`

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
drew2a: Yes, probably. For me purpose of overwriting a Network is unclear.
It is true for `PopularityCommunity` and `GigachannelCommunity` (as successors of `RemoteQueryCommunity`)
cc: @ichorid 

(I see it as out of the scope of this PR)

src/tribler-core/tribler_core/components/tests/test_tribler_components.py
drew2a: Agree! Will implement it soon.

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
ichorid: GigaChannelCommunity walks and deletes peers aggressively, so creates a separate Network instance to prevent interfering with other communities.

src/tribler-core/tribler_core/components/ipv8.py
qstokkink: Short answer: yes, 20 is a good idea.

Long answer: The DiscoveryCommunity is a meta-community that all peers join to quickly synchronize the other Communities they are a part of. This also forms a secondary bootstrap mechanism (which is why you generally don't want to deviate from the default parameterization of your Community and use a custom `Network()`). Actively walking to some peers in the DiscoveryCommunity  will allow you to get a little bit of diversity in your connections. However, walking to too many peers here will just give you a lot of peers you do not share Communities with. 20 seems to be a good number (also considering the network topologies of the other Communities this Community shares its peer pool with). You could even set this to 0. However, 90 may produce too many open connections if overlay identifiers start diverging in the network.

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
qstokkink: The reason for a "clean" `Network()` is the introduction of bias through custom parameterization of Communities. If you bias your peer discovery mechanism, you get a self-reinforcing effect that essentially forces some peers out of your network. It is not a problem if one peer deviates, but if you force a default on the entire network you will start seeing this. We've learned this the hard way https://github.com/Tribler/tribler/issues/2894#issuecomment-462808735 Therefore, for every Community that uses custom walks (including peer removal) or custom parameterization, we use a new `Network`.

src/tribler-core/tribler_core/components/ipv8.py
drew2a: Got it. What if we will rewrite `initialise_community_by_default` this way:

```python
    def initialise_community_by_default(self, community, default_random_walk_max_peers=20):
        community.bootstrappers.append(self.make_bootstrapper())

        # Value of `target_peers` must not be equal to the value of `max_peers` for the community.
        # This causes a deformed network topology and makes it harder for peers to connect to others.
        # More information: https://github.com/Tribler/py-ipv8/issues/979#issuecomment-896643760
        #
        # Then:
        # random_walk_max_peers should be less than community.max_peers:
        random_walk_max_peers = min(default_random_walk_max_peers, community.max_peers - 10)

        # random_walk_max_peers should be greater than 0
        random_walk_max_peers = max(0, random_walk_max_peers)
        self.ipv8.add_strategy(community, RandomWalk(community), random_walk_max_peers)
```

qstokkink: Seems good to me 👍

drew2a: Implemented

src/tribler-core/tribler_core/components/tests/test_tribler_components.py
drew2a: Implemented

qstokkink (approved):
Thanks for going out of your way to even implement the nit and optional suggestion. Looks good 👍

src/tribler-core/tribler_core/components/bandwidth_accounting/bandwidth_accounting_component.py
drew2a: Implemented here: https://github.com/Tribler/tribler/pull/6408

PR #6408, opened 2021-10-05
Move Network overwrite from community to component

drew2a:
@qstokkink 
> We're injecting self._ipv8_component.ipv8.network as a dependency and then the component itself is overwriting its input with a Network() of its own creation. We should, at some point (probably not this PR) move the Network() overwrite to this file (same for PopularityCommunity).

The `Network() overwrite` has been moved for:
* `PopularityCommunity`
* `GigachannelCommunity`

For `BandwidthAccountingCommunity` I didn't find any place with `Network()` overwriting.


ichorid (approved):

qstokkink (approved):
Nice, much cleaner 👍

Issue #6395, opened 2021-09-30
Content appearing and disappearing when refreshing a channel


devos50:
I noticed that when going to a channel in Tribler and refreshing it by clicking on the channel name in the breadcrumb navigator, sometimes content appears and disappears. While this is expected (since Tribler can discover new content), sometimes the amount of content is changing significantly. Note that it requires like 20+ or 30+ refreshes sometimes to trigger this behavior.

Below are two screenshots that were taken within the same channel. Refreshing it occasionally gives different content. Note that sometimes, item A in the first screenshot is not present in screenshot B, hinting that this might be a bug somewhere.

<img width="1074" alt="Schermafbeelding 2021-09-30 om 12 20 01" src="https://user-images.githubusercontent.com/1707075/135470145-09df6f91-4eb4-4213-b6d7-cb20e05e6e05.png">

<img width="1074" alt="Schermafbeelding 2021-09-30 om 12 20 06" src="https://user-images.githubusercontent.com/1707075/135470157-eb0459e8-35cb-4783-96df-714e8c295caa.png">

devos50:
I have the impression that #6415 even makes it easier to reproduce. Anyway, in like 20 refreshes of one of the most popular channels in Tribler, I get random content in like 5 refreshes.

ichorid:
I had to click it about 20-30 times to get to see the thing :shrug: . I guess it is highly dependent on network and PC performance.

PR #6415, opened 2021-10-05
Add anyio happy eyeballs gigachannel select

ichorid:


devos50:
TIL: [the happy eyeballs algorithm](https://datatracker.ietf.org/doc/html/rfc8305) 😁 

Out of curiosity, what's the exact problem this is addressing in Tribler? 🤔

drew2a:
@devos50 were you convinced in the course of the offline @ichorid's explanation that we have a problem that can be solved by the happy eyeballs algorithm?

ichorid:
> @ichorid Unfortunately, this isn't fixing #6395 for me. Random content is still appearing when refreshing the channel.

:facepalm:

ichorid:
> As #6395 is not fixed, I postpone the review procedure.

Like, indefinitely? :raised_eyebrow:

drew2a:
From my understanding this PR was a fix for #6395. If it doesn't fix the problem (as @devos50 mentioned), I as a reviewer should wait until PR be ready.

If it solves another issue, please, link it to the description.
As a reviewer I must understands the problem and your solution to approve or disapprove it.

PR #6419, opened 2021-10-06
Fixes #6358: correct component's shutdown

### LINKED ISSUE #6358: [Components] In some rare cases the method component.stop() can work incorrectly
During the refactoring (#6335), we found the following potential problem:

In some rare cases the method [component.stop()](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/base.py#L165) can work incorrectly.

The current implementation (simplified) is the following:

```python
class Session:
    async def start(self, failfast=True):
        coros = [comp.start() for comp in self.components.values()]
        await gather(*coros, return_exceptions=not failfast)

    async def shutdown(self):
        await gather(*[create_task(component.stop()) for component in self.components.values()])
```

That means whatever `comp.start()` was failed or succeeded, the corresponding `component.stop()` will be called during a shutdown.

That means `component.stop()` should distinguish initialized and uninitialized fields of the corresponding class (this is not currently being done).

kozlovsky:
This PR fixes #6358. Now component's shutdown code checks that component's attributes were actually initialized.

Also, shutdown notifications were removed from shutdown methods, as they do not display correct information: now component's shutdown process is asynchronous, and the last notification may not reflect the actual action performing.


src/tribler-core/tribler_core/components/restapi.py
```diff
@@ -109,12 +113,7 @@ def report_callback(text_long, sentry_event):
 
         CoreExceptionHandler.report_callback = report_callback
 
-        # We provide the REST API only after the essential endpoints (events, state and shutdown) and
-        # the exception handler were initialized
-        self.rest_manager = rest_manager
-
     async def shutdown(self):
```
drew2a, line 116: **nit**:  it doesn't relate to the PR, but it seems that `await super().shutdown()` is missed

src/tribler-core/tribler_core/components/version_check.py
```diff
@@ -17,5 +19,5 @@ async def run(self):
         self.version_check_manager.start()
 
     async def shutdown(self):
```
drew2a, line 21: **nit**: it doesn't relate to the PR, but it seems that `await super().shutdown()` is missed

src/tribler-core/tribler_core/components/version_check.py
```diff
@@ -8,6 +8,8 @@ class VersionCheckComponent(Component):
     version_check_manager: VersionCheckManager
 
     async def run(self):
```
drew2a, line 10: **nit**: it doesn't relate to the PR, but it seems that `await super().run()` is missed

src/tribler-core/tribler_core/components/ipv8/ipv8_component.py
```diff
@@ -30,6 +32,9 @@ class Ipv8Component(RestfulComponent):
     _peer_discovery_community: Optional[DiscoveryCommunity] = None
 
     async def run(self):
+        self.ipv8 = None
+        self.peer = None
+
         await super().run()
```
drew2a, line 38: **nit** from my POV it is safer (in a long term) to keep an inheritance chain as simple and as predictable as possible. Therefore I suggest always call the overridden `super()` method first, and any other operations second.
```suggestion
    async def run(self):
        await super().run()
        self.ipv8 = None
        self.peer = None


```

drew2a (changes requested):
👍 great work. Have some *nit* comments, but in general, PR is ready to merge.

src/tribler-core/tribler_core/components/version_check.py
kozlovsky: Fixed

src/tribler-core/tribler_core/components/restapi.py
kozlovsky: Fixed

drew2a (dismissed):

PR #6425, opened 2021-10-06
Fixed GUI dialog position on show

### LINKED ISSUE #6423: Request error dialog incorrectly rendered
When an error occurs during a HTTP request, Tribler can show a dialog with a traceback:

Schermafbeelding 2021-10-06 om 15 30 03

This dialog is not aligned properly and I cannot click the close button. When resizing the window, however, it jumps to the right location. We probably need a call to `dialog.resize()` immediately after showing the popup.

devos50:
Fixes #6423


devos50:
The pylint warning seems to be incorrect and can be ignored. The method is correctly overwritten 👍

PR #6432, opened 2021-10-07
Fixed status code check in REST API tests

### LINKED ISSUE #6431: Status code check in REST API tests is broken
It seems that after merging #6206, the status code check in the REST API tests is broken (see [here](https://github.com/tribler/tribler/commit/c8c307d5019e390276037c58b6dc6a2d10fb95ef#diff-95630493715b227737b018143e427c0cf40244934198f41a25b9a7a02ce0ec00R58) for the code change). The check is currently as follows:

```
assert expected_code, response == status
```

However, this is merely checking whether `expected_code` is true (i.e., non-zero), and if the assert fails, it will print `response == status` as the reason. We should fix this logic.

devos50:
Fixes #6431

### DIFF
--- a/src/tribler-core/tribler_core/restapi/base_api_test.py
+++ b/src/tribler-core/tribler_core/restapi/base_api_test.py
@@ -55,7 +55,7 @@ async def do_request(test_client, url, expected_code=200, expected_json=None,
                 print(response['error']['message'])
             else:
                 print(response['error'])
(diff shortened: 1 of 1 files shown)


ichorid:
In asserts, the left part is _the expected_ and the right one is _the actual_.

So, the correct way to fix it is:
`assert (expected_code, response) == status`

PR #6435, opened 2021-10-08
Removed popular torrents explanation box

### LINKED ISSUE #6389: Clean up 'popular torrents' page by moving the text to a toolbox
The 'popular torrents' page always shows a line of text explaining what the user sees on this page.

Schermafbeelding 2021-09-29 om 11 08 55

I believe this information can be removed from this page and instead, be shown behind an 'information' button. We do something similar here:

Schermafbeelding 2021-09-29 om 11 10 38

devos50:
This was harder than I thought. There were two complications when implementing this. First, I was unable to place the explanation box directly to the right of the text (as I did on the trust page). This is because the title widget is used for the channel navigation breadcrumb and modifying the width constraints breaks the breadcrumb computations. So I decided to put the widget to the right of the page.

Second, it seems that there is a bug that prevents customization of the QToolTip using CSS. The proper way to address this issue is by providing a custom mask but that's more complicated for now. The tooltip is not clipped correctly and there are some ugly borders visible. Fortunately, they blend with the background so they are hard to spot.

I believe the end result is acceptable and preferred over the old approach, but let me know if not (note that I cannot change the border radius/color due to the bug).

Schermafbeelding 2021-10-08 om 14 08 14

Fixes #6389

### DIFF
--- a/src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
+++ b/src/tribler-gui/tribler_gui/qt_resources/mainwindow.ui
@@ -102,7 +102,11 @@ color:#B5B5B5;
 }
 
 QCheckBox { color:#B5B5B5; }
-QCheckBox::indicator:unchecked {border: 1px solid #555;}</string>
+QCheckBox::indicator:unchecked {border: 1px solid #555;}
+
+QToolButton QToolTip {
+padding: 4px;
+}</string>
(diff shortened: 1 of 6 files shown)


devos50:
Ready for review. Hover events in the GUI are, unfortunately, very hard to test.

PR #6449, opened 2021-10-12
READY: Add tree-based view for torrent files in download dialog

### LINKED ISSUE #1396: [feature] Tree view to select files to download
Some torrents (including different kinds of anthologies) contain folder trees and sometimes only one folder may be of interest (is the download target). In this cases selecting files to download using the current "list view" can be devastating in terms of time. So what is needed in such cases is a tree view where each folder can be selected/unselected/selected partially.

(an example of such interface can be found in uTorrent; in fact, this issue is the reason why I have to use another client, too)

### LINKED ISSUE #6443: Trees for file selection in torrents
Every other GUI-based BitTorrent client out there uses a tree view for selecting/managing files in torrents. We're the only ones who use a plain list insteaf. Good luck selecting some files you need from a torrent with 40k entries (e.g. [some academic torrent](https://academictorrents.com/details/5a447ff50062194bd58dd11c0fedead59e6d873c)).

изображение

изображение

related to #4468 

originally reported in #1396

ichorid:
fixes #1396 
fixes #6443 

I had to try at least three different ways of adding progress bars to QTreeWidget. This one is the safest, it does not trigger the dreaded "QObject deleted" error. Drawing progress bars by using `setItemWidget` is still slow, so it is disabled for torrents with more than 1k entries.
The proper, fast and safe way to implement this kind of stuff is to move to use the Model-View architecture. However, that would prove too much work, at least for now.

Add download dialog (Note that sorting works, and that there is a counter for total selected files size and select/deselect all buttons are removed as unnecessary. "Download exists" label is removed too - the corresponding notification is now shown in the same "Loading metainfo" label)

изображение

изображение
изображение

UPDATE2:
I've tripped at least **three** different QT bugs while doing this :facepalm: The whole thing is built from workarounds. We really need to switch to Model-View eventually.


ichorid:
> Tree view when starting a download:
> 
>     * Can you make the ‘Included?’ Column auto-size to the required width or would this be too much work?

Yep, there is an option for that. I'll try it.

>     * There is no context regarding the numbers below the table (e.g., 3.3 MB / 889.0 MB). I understand that the first number is the total selected file size whereas the second number is the overall size of the download. This is not clear for users. We can add something like: **Selected:** 3.3 MB, **Total:** 889.0 MB.

Good point.

>     * Should we add ‘select all’ / ‘select none’ buttons?

`Select all` / `unselect all` buttons are unnecessary because clicking the top-level checkboxes magically :mage_man:  sets/unsets checkboxes for the whole subtree.

>     * I’m unable to select a bunch of files and exclude them for download. I used to be able to do so by right-clicking. The selection also seems to be a bit broken though (probably because you’re using the notorious `QTreeView`).

Well, you're not supposed to, really... Just check/uncheck the checkboxes - that is it. Selection is completely unnecessary in the case of "tree+checkboxes" solution: it requires the same amount of clicks to either click the checkboxes or first select them and then open the menu, etc.

ichorid:
retest this please

PR #6458, opened 2021-10-15
Implemented customized checkbox style across the GUI

### LINKED ISSUE #5624: Too low contrast on disabled checkboxes
7.5.2

Windows 10

1. Open a torrent file in Tribler with the open torrent file dialog

Should be able to tell disabled checked and disabled unchecked checkboxes apart. This is a bit difficult with the default grey on grey color scheme, as disabled state is usually represented by greying out controllers. But in Tribler, they're all grey. Not sure how to best fix it. Making the regular enabled checkbox check Tribler-orange and the disabled state black could help; but only when both are visible at the same time.

##### Actual behavior: #####
Tribler checkbox contrast

### LINKED ISSUE #6457: Settings page initializes with the wrong page
When starting Tribler and when going to the settings panel, it does not initialize on the 'general' tab:

Schermafbeelding 2021-10-15 om 17 32 17

This is because the `currentIndex` property of the `settings_stacked_widget` widget is not initialized correctly.

### LINKED ISSUE #6459: Customized checkbox not visible in Windows GUI tests
It seems that the customized checkbox introduced in #6449 is not visible in Windows.

**On Mac:**
screenshot_add_torrent_url_startdownload_dialog_files-2

**On Windows:**
screenshot_add_torrent_url_startdownload_dialog_files

(screenshots taken from the GUI tests)

I'm not sure about Linux since we are not running GUI tests on that platform.

devos50:
Inspired by #6449, I implemented customized checkbox styles across the GUI. The result of this change can be seen in the screenshot below:

Schermafbeelding 2021-10-15 om 18 30 49

There are styles for checked/unchecked checkboxes, as well for enabled/disabled states. Since in a frozen environment the image path is different, we are unable to embed the customization stylesheet in an `.ui` file. Instead, I extend the existing stylesheet when loading the `TriblerWindow` widget.

I also took the opportunity to fix the settings page that would open at a different index.

Fixes #5624 
Fixes #6457 
Fixes #6459

### DIFF
--- a/src/tribler-gui/tribler_gui/dialogs/startdownloaddialog.py
+++ b/src/tribler-gui/tribler_gui/dialogs/startdownloaddialog.py
@@ -79,7 +79,7 @@ class StartDownloadDialog(DialogContainer):
             image: url('%s');
         }
         """
-            % get_image_path('down_arrow_input.png')
+            % get_image_path('down_arrow_input.png', convert_slashes_to_forward=True)
         )
 
         if self.window().tribler_settings:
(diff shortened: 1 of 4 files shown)


devos50:
retest this please

Issue #6214, opened 2021-07-09
Tag based content discovery and crowdsourcing


ichorid:
After a thorough discussion, we came to the following architecture for the Tags system:
* every metadata entry should have one or several tags
* the tag assigning system maintains a local database of tags, based on frequencies
* whenever a user starts a download (from a channel or from an external world), if the download has less than 5 tags, Tribler _gently_ asks the user to add some tags (from suggested tags)
* if the user suggests a change to a torrent's tags, [a PR into the owner's channel](https://github.com/Tribler/tribler/issues/6208) is automatically created and propagated upstream
* the tag selection system should be bootstrapped from some collective knowledge pool, or from other users
* we should first create some wireframes to see how tags adding and search look like
* the main reason for tags system is **deduplication** and gentle nudging of users to organize in collectives

grimadas:
Comments regarding the tag-based organisation and why need this in the first place: 


## Why tag-centric organisation? 
 
- A more natural way to organize people not around channel but around real interest.  
- Most channels are GIGA, hard to moderate and even look at the huge amount of content. Too much dependency on the channel owner.  
- How many are using channel feature? Are they browsing even? 

## What is tag-centric even? 

Tag-centric organization is people,  content,  metadata. 
  
+ Content. A content that is associated with this tag, e.g. videos, audio, text data etc.  
+ Metadata.  Forums, chats and structure, labels etc.  
+ People. People are crucial part of this organisation. They might be consumers, contributors, servers or moderators. There are multiple roles.  


**The vision for Tribler: **
- Let's give more tools for people to organise communities based on tags. 
- Metrics: engagement, do users use features, the tag-communities are available etc.   
- We decrease time to create community. They will organise faster and with bigger scale.

synctext:
This issue re-does the work already conducted 14 years ago. But then it failed.
This is a big task, beyond a single developer. Read about this architecture [full article](https://pdfs.semanticscholar.org/999b/69584737d53f29e39743c724ae4dd52c0ccb.pdf):
![image](https://user-images.githubusercontent.com/325224/125473400-16ca1987-6a78-4180-8472-fc85d2913309.png)
The exact hard-core science details:
![image](https://user-images.githubusercontent.com/325224/125473754-e6017867-0aa9-4277-acb8-b821f257a25a.png)
[...]

PR #6462, opened 2021-10-18
Add Secondary Key to MasterKeyComponent

drew2a:
This PR adds a Secondary Key to the MasterKeyComponent.

In PR:
* Refactoring of MasterKeyComponent has been made
* The secondary key now has been using by TagCommunity
 
Related to #6214 and addresses the anonymity issue of associating content with Public Key and IP.


ichorid:
I suspect the whole PR could be replaced with a single line of:
```
 derived_key = crypto_kdf_derive_from_key(libnacl.crypto_box_SECRETKEYBYTES, 12345, "BLA", trustchain_keypair)
```

Hmm... :thinking:

drew2a:
> I suspect the whole PR could be replaced with a single line of...

It is not possible, because this PR contains not only new logic for a MasterKeyComponent (which can be implemented in one line), but also refactoring and changes to the TagCommunity.

PR #6504, opened 2021-10-27
Make requirements.txt a single source of trust for dependencies 

### LINKED ISSUE #6227: Make requirements.txt a single source of trust for dependencies
We have several places that describe dependencies (duplicates). 
In the case of adding a new dependency, it is easy to miss all the places necessary to change. 
Also, it is hard to maintain such a structure.

### Documentation
https://github.com/Tribler/tribler/blob/7954a2f42546ef9f2d70b40694fc7fa457368411/doc/development/development_on_linux.rst#L14-L16

https://github.com/Tribler/tribler/blob/7954a2f42546ef9f2d70b40694fc7fa457368411//doc/development/development_on_windows.rst#L62-L65

### Code

https://github.com/Tribler/tribler/blob/7954a2f42546ef9f2d70b40694fc7fa457368411/src/tribler-core/tribler_core/dependencies.py#L13-L37

### Requirements.txt

https://github.com/Tribler/tribler/blob/7954a2f42546ef9f2d70b40694fc7fa457368411/src/requirements.txt#L1-L24

Ref:
* https://github.com/Tribler/tribler/pull/6225
* https://github.com/Tribler/tribler/pull/6102
* https://github.com/Tribler/tribler/issues/6188

drew2a:
This PR fixes #6227 in three steps:

1. It fixes documentation by requesting a user to use `pip install -r requirements.txt` instead of installing a big list of dependencies, eg `pip3 install aiohttp aiohttp_apispec chardet configobj decorator libnacl matplotlib netifaces networkx pony psutil pyasn1 `.
 
image

2. It splits the main `requirements.txt` intro `requirements.txt` for GUI and `requirements.txt` for CORE:

image

image

3. It removes `dependency.py` logic from production, but as soon as the #6518 is solved it will no longer be an issue.
4. It adds "smart" mocking to the documentations builder.

### warning
This PR introduces the new dependency:

```pyhon
mock~=4.0.3
```


src/tribler-core/tribler_core/requirements.txt
```diff
@@ -0,0 +1,23 @@
...
+chardet
+libnacl
+pyasn1
+lz4
+pony
+service-identity
+decorator
+networkx
+cryptography
+cryptography
```
qstokkink, line 13:
```suggestion
cryptography
```

**suggestion/nit:** You could order all `requirements.txt` files alphabetically to make them easier to parse by humans.

src/tribler-common/tribler_common/dependencies.py
```diff
@@ -0,0 +1,107 @@
...
+from tribler_core.utilities.path_util import Path
+
+# fmt: off
+
+logger = logging.getLogger(__name__)
+
+Scope = Enum('Scope', 'core gui common')
+
+_frozen_env = is_frozen()
+_base_path = get_base_path().parent.parent  # this is the "/src" folder
```
xoriole, line 27: **issue:** `.parent.parent` affects the frozen path for requirements file.

drew2a: Thank you, fixed.

Also, I've added test for this case:

```python
async def test_dev_get_requirements_path():
    assert _get_requirements_path(scope=Scope.common, frozen=False) == Path(
        tribler_common.__file__).parent / 'requirements.txt'


@patch_dependencies_module(get_base_path, new=MagicMock(return_value=Path('')))
async def test_frozen_get_requirements_path():
    assert _get_requirements_path(scope=Scope.common, frozen=True) == Path('requirements-common.txt')
```

src/tribler-core/tribler_core/requirements.txt
drew2a: Thank you 👍  Implemented.

src/tribler-common/tribler_common/patch_import.py
```diff
@@ -0,0 +1,38 @@
...
+    :param modules: a list of prefixes of modules that should
+        be mocked, and an ImportError could not be raised for.
+    :param mock_kwargs: kwargs for MagicMock object.
+    :return: patch object
+    """
+
+    def try_import(module_name, *args, **kwargs):
+        try:
+            return _builtins_import(module_name, *args, **kwargs)
+        except:
```
kozlovsky, line 33: **issue**: The docstring says that the function disables the `ImportError` exception, while in fact, it catches all exceptions occurring during the module import.

**suggestion**: Can the function catch `ImportError` exception only? It may be less brittle, as it would not mask other exception types during the module import. Another option is changing the docstring and writing that the function will catch any exception during the import process.

drew2a: Yes, thank you, fixed.

kozlovsky (approved):

PR #6535, opened 2021-11-08
WIP: Simplify GUI-Core initial communication

### LINKED ISSUE #6526: GUI-Core startup communication is too complicated
Tribler's startup sequence is complex and fragile:
 * run the GUI, 
 * then GUI runs the Core,
 * then GUI tries to connect to the State endpoint (polling), 
 * then GUI tries to connect to the Events endpoint
 * then GUI waits for Upgrader events
 * then GUI waits for `TRIBLER_STARTED` event pushed over the events endpoint

This system is prone to race conditions. Also, the thing is further complicated by Core pushing error to the GUI through the Events enpoint.

The startup sequence can be refactored into a much simpler thing:
* run the GUI,
* then GUI **runs Upgrader as a separate app**
* then GUI runs the Core,
* then GUI connects to the Events endpoint

This change will require the following series of refactorings:
1. Move error reporting to the `tribler_common` module, stop passing errors to the GUI and instead raise them directly from the Core (we already do this to raise errors when the GUI is not available) (@drew2a , thanks for the suggestion!)
2. Move Upgrader into a separate application. Refactor it to exclude `asyncio` stuff and instead use signals and stdio to communicate with the GUI.
3. Remove unnecessary State endpoint and make Events endpoint always send a "Tribler version X" push message on connection.
4. _Bonus level_: Refactor initialization of REST endpoints to pure dependency injection - the REST component could be started last (or first, and then dynamically inject all the component-specific endpoints into it)

The results will be:
* removing all race conditions on startup
* more reliable bug reports
* greatly simplified Upgrader code
* release dependence of every component on Upgrader
* pave the way for a painless FastAPI migration

ichorid:
Fixes #6526

This PR consists of the following series of changes, all working towards the goal of making the process of starting Tribler GUI-to-Core communication as simple as possible:

1. ~Remove Dispersy-to-Pony upgrade routines (stats show there are no Dispersy-based Tribler instances left in the wild)~
2. Remove UpgradeComponent and add "upgrade" mode for `run_tribler.py` that runs upgrader as an independent console application. The upgrading application is a simple, synchronous application that writes its status to standard output and can be gracefully stopped by `SIGINT`.
3. Change `run_tribler.py` to use command-line arguments (`argparse`) instead of environment variables to select which mode to use: either one of full, core, or upgrade mode.
4. Remove `StateEndpoint` and all Tribler startup state-related code.
5. Refactor all the endpoints to use true dependency injection. No late assignment of attributes anymore!
5. Move the initialization of all the endpoints into the `RESTComponent`. The relationship between `RESTComponent` and other components are now reversed: `RESTComponent` now depends on every other component that has something to do with the endpoints.
6.  All the endpoints are now initialized in one go, in the `RESTComponent`
7. CoreManager in the GUI is now split into two parts: CoreManager and UpgradeManager. UpgraderManager runs `run_tribler.py` with `--upgrade` argument first, capturing and showing its standard output. When it finishes, CoreManager is run as usual.

Overall, these changes make the Tribler startup process very simple: 
1. start the GUI
2. GUI looks if there is running Core available: 
3. if the Core is already running, connect to its EventsEndpoint immediately, 
4. if there is no Core running, run it, then
5. run Upgrader, then
6. run Core, then 
7. connect to its EventsEndpoint immediately


ichorid:
retest this please

PR #6546, opened 2021-11-11
Make TEST_SENTRY_URL the single source of sentry test mode information

drew2a:
The intention of this PR is to simplify the Application Tester's error reporting by removing `requires_user_consent` field.

Initially, `requires_user_consent` has been introduced to prevent mis-switching to the test mode, which sends error messages automatically (without user participation) and it works in tandem with `TEST_SENTRY_URL` like:

```python
@staticmethod
    def can_send_automatically(error_reporting_requires_user_consent):
        test_sentry_url_is_defined = os.environ.get('TEST_SENTRY_URL', None) is not None
        return test_sentry_url_is_defined and not error_reporting_requires_user_consent
```

This PR renames `TEST_SENTRY_URL` to `TRIBLER_TEST_SENTRY_URL` which should be enough for preventing mis-switching to test mode without specifying `error_reporting_requires_user_consent`.

### DIFF
--- a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
+++ b/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import sys
 from contextlib import contextmanager
 from contextvars import ContextVar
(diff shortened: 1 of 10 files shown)


src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py
```diff
@@ -332,6 +333,14 @@ def get_actual_strategy():
         strategy = SentryReporter.thread_strategy.get()
         return strategy if strategy else SentryReporter.global_strategy
 
+    @staticmethod
+    def get_test_sentry_url():
+        return os.environ.get('TRIBLER_TEST_SENTRY_URL', None)
+
+    @staticmethod
+    def is_in_test_mode():
+        return SentryReporter.get_test_sentry_url() is not None
```
kozlovsky, line 342: **issue:** it may be convenient to treat empty value of an environment variable as a sign that it is not set

**suggestion**: we can treat the empty string the same way as a missing environment variable.

```suggestion
        return bool(SentryReporter.get_test_sentry_url())
```

kozlovsky (changes requested):

kozlovsky (approved):

xoriole (approved):

PR #6552, opened 2021-11-15
Add CoreRuntimeError to the error list that causes Tribler shutdown

### LINKED ISSUE #6521: [main] Tribler gets stuck on shutting down screen
How to reproduce:
1. Go to `notifier.py`
2. Add line `raise Exception('test')` to 
```python
        # We have to call the notifier callbacks through call_soon_threadsafe
        # because the notify method could have been called from a non-reactor thread
        raise Exception('test')
        self._loop.call_soon_threadsafe(self._notify, subject, *args)
```

Note: before the grand refactoring this trick did not lead to the stuck.

image

drew2a:
This PR fixes #6521 by introducing `CoreRuntimeError` and adding this error to the list that causes Tribler shutdown.

### DIFF
--- a/src/tribler-gui/tribler_gui/core_manager.py
+++ b/src/tribler-gui/tribler_gui/core_manager.py
@@ -20,6 +20,10 @@ START_FAKE_API = False
 SKIP_VERSION_CLEANUP = os.environ.get("SKIP_VERSION_CLEANUP", "FALSE").lower() == "true"
 
 
+class CoreRuntimeError(Exception):
+    """This error raises in case of tribler core finished with error"""
+
+
 class CoreManager(QObject):
     """
     The CoreManager is responsible for managing the Tribler core (starting/stopping). When we are running the GUI tests,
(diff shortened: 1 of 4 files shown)


src/tribler-gui/tribler_gui/core_manager.py
```diff
@@ -83,7 +87,7 @@ def on_core_finished(self, exit_code, exit_status):
                     self.core_traceback_timestamp,
                 )
 
-            raise RuntimeError(exception_msg)
+            raise CoreRuntimeError(exception_msg)
```
kozlovsky, line 90: Can you rename the error to CoreCrashedError to clarify that it does not inherit from RuntimeError?

drew2a:
This PR has been rebased to the release branch: https://github.com/Tribler/tribler/pull/6556

src/tribler-gui/tribler_gui/core_manager.py
kozlovsky: Or even better would be to have base exception class `CoreError` inherited from `RuntimeError` with two subclasses `CoreConnectTimeoutError` and `CoreCrashedError`

PR #6556, opened 2021-11-16
Add CoreRuntimeError to the error list that causes Tribler shutdown 

### LINKED ISSUE #6521: [main] Tribler gets stuck on shutting down screen

drew2a:
This PR fixes #6521 by introducing CoreRuntimeError and adding this error to the list that causes Tribler shutdown.

### DIFF
--- a/src/tribler-gui/tribler_gui/__init__.py
+++ b/src/tribler-gui/tribler_gui/__init__.py
@@ -9,6 +9,11 @@ from tribler_common.logger import setup_logging
 
 logger = logging.getLogger(__name__)
 
+
+class CoreError(Exception):
+    """This is the base class for exceptions that causes GUI shutdown"""
+
+
 def load_logger_config(log_dir):
     """
     Loads tribler-gui module logger configuration. Note that this function should be called explicitly to
--- a/src/tribler-gui/tribler_gui/__init__.py
+++ b/src/tribler-gui/tribler_gui/__init__.py
@@ -10,10 +10,6 @@ from tribler_common.logger import setup_logging
 logger = logging.getLogger(__name__)
 
 
-class CoreError(Exception):
-    """This is the base class for exceptions that causes GUI shutdown"""
-
-
 def load_logger_config(log_dir):
     """
     Loads tribler-gui module logger configuration. Note that this function should be called explicitly to
--- a/src/tribler-gui/tribler_gui/tests/test_core_manager.py
+++ b/src/tribler-gui/tribler_gui/tests/test_core_manager.py
@@ -1,0 +1,35 @@
+from unittest.mock import MagicMock, patch
+
+import pytest
+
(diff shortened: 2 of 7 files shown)


src/tribler-gui/tribler_gui/__init__.py
```diff
@@ -9,6 +9,11 @@
 
 logger = logging.getLogger(__name__)
 
+
+class CoreError(Exception):
+    """This is the base class for exceptions that causes GUI shutdown"""
```
drew2a, line 14: @kozlovsky I'm not sure that `__init__.py` is the best place for the `CoreError` class. If you have any suggestions feel free to give them to me.

kozlovsky: I think we should have `exceptions.py` file in `tribler_gui` similar to `exceptions.py` file in `tribler_core`

src/tribler-gui/tribler_gui/tests/test_core_manager.py
```diff
@@ -0,0 +1,35 @@
+from unittest.mock import MagicMock, patch
```
kozlovsky, line 1: **nitpick:** It seems the tests don't use magic methods of mock objects (such as `__iter__`, `__getitem__`, etc.), so the base `Mock` class should be enough.

kozlovsky, line 1: **praise** good test coverage, as always!

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/__init__.py
drew2a: exceptions.py has been added.

src/tribler-gui/tribler_gui/tests/test_core_manager.py
drew2a: unittest uses `MagicMock` as a default class for mock objects: https://docs.python.org/3/library/unittest.mock.html#the-mock-class

> The patch() decorators makes it easy to temporarily replace classes in a particular module with a Mock object. By default patch() will create a MagicMock for you. You can specify an alternative class of Mock using the new_callable argument to patch().

drew2a: 👍

drew2a:
retest this please

kozlovsky (approved):

PR #6558, opened 2021-11-16
Refactor/remove obsolete upgrades

ichorid:
This removes support for upgrading from Tribler versions prior to 7.5.


kozlovsky:
I checked the database, we don't have active users with Tribler version <7.5.2 after year 2020. So, dropping support for versions <7.5.2 should be safe.

kozlovsky (approved):

PR #6559, opened 2021-11-16
Run Upgrader in a GUI thread

ichorid:
Partially fixes #6525

This one moves some of the functions of Core Manager into the new Upgrade Manager class. The upgrader is now run in a GUI QThread for simplicity. Also, this changes `run_tribler.py` to use arguments instead of environment variables to indicate the running mode (Core vs GUI vs Core Test Mode). Arguments are safer and unambiguous.

The main point of this PR is to stop sending event notifications from the Core from upgrader and thus pave the road for simplification of Tribler startup.

Upgrade Manager is briefly run during GUI tests, quitting when not able to find an existing Tribler state dir. This is not a real test, but at least it covers all the relevant lines. The downside is that when running single tests from the GUI suite waiting for additional time could be a bit annoying.

Simplification of GUI tests uncovered some hidden race conditions, which required minor fixes.


ichorid:
retest this please

ichorid:
retest this please

src/tribler-core/run_tribler_upgrader.py
```diff
@@ -0,0 +1,53 @@
...
+
+    upgrader = TriblerUpgrader(state_dir, channels_dir, primary_key,
+                               update_status_callback=update_status_callback,
+                               interrupt_upgrade_event=interrupt_upgrade_event)
+    upgrader.run()
+
+
+if __name__ == "__main__":
+    def print_text(text):
+        print(text)
```
drew2a, line 41: **Question**: why are we using print here? Is it possible to use logger?

src/tribler-core/tribler_core/tests/test_start_core.py
```diff
@@ -1,29 +1,19 @@
...
-@patch('tribler_core.start_core.VersionHistory', new=MagicMock())
 @patch('tribler_core.start_core.set_process_priority', new=MagicMock())
 @patch('tribler_core.start_core.check_and_enable_code_tracing', new=MagicMock())
-@patch('tribler_core.start_core.core_session', new=MagicMock())
-@patch('tribler_core.start_core.ProcessChecker', new=MockedProcessChecker())
 @patch('asyncio.get_event_loop', new=MagicMock())
 @patch('tribler_core.start_core.TriblerConfig.load', new=MagicMock())
-@patch.object(MockedProcessChecker, 'remove_lock_file', create=True)
-async def test_start_tribler_core_no_exceptions(mocked_remove_lock_file):
+@patch.object(start_core, 'core_session')
```
drew2a, line 15: **Issue** it looks like an illegal way of using `patch.object`

You can use this instead:
```suggestion
@patch('tribler_core.start_core.core_session')
```

src/tribler-gui/tribler_gui/tests/test_gui.py
```diff
@@ -38,56 +39,36 @@ def api_port():
...
+        core_args=[str(RUN_TRIBLER_PY.absolute()), '--gui_test_mode'],
+    )  # pylint: disable=W0621
     app.set_activation_window(window)
     QTest.qWaitForWindowExposed(window)
 
     screenshot(window, name="tribler_loading")
-    wait_for_signal(window.core_manager.events_manager.tribler_started)
+    wait_for_signal(
+        window.core_manager.events_manager.tribler_started,
+        timeout=10,
```
drew2a, line 61: **Question** why did you use 10 as a timeout?

drew2a (changes requested):
This PR definitely looks like an improvement.

I've not reviewed the logic, because I don't understand it. I Hope @kozlovsky or @devos50 can do it properly.

I left some comments. Also, `pylint` should be satisfied.

src/run_tribler.py
```diff
@@ -46,23 +66,38 @@ def init_boot_logger():
...
     root_state_dir = get_root_state_directory()
     logger.info(f'Root state dir: {root_state_dir}')
 
+    api_port = os.environ.get('CORE_API_PORT')
+    api_key = os.environ.get('CORE_API_KEY')
+
     # Check whether we need to start the core or the user interface
-    if 'CORE_PROCESS' in os.environ:
-        logger.info('Running in "core" mode')
+    if parsed_args.mode in (RunMode.CORE_ONLY, RunMode.GUI_TEST_MODE):
```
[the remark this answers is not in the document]
ichorid: I would rather leave such changes for a separate PR...

src/run_tribler.py
```diff
@@ -77,8 +112,13 @@ def init_boot_logger():
...
 
-        from tribler_core.check_os import check_and_enable_code_tracing, check_environment, check_free_space, \
-            enable_fault_handler, error_and_exit, should_kill_other_tribler_instances
+        from tribler_core.check_os import (
+            check_and_enable_code_tracing,
+            check_environment,
+            check_free_space,
+            enable_fault_handler,
+            error_and_exit,
+        )
```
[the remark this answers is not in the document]
ichorid: This was for a long time on my list. But again, that would be a matter of a separate PR.

src/tribler-core/run_tribler_upgrader.py
ichorid: The new upgrade script is a convenience for testing and for people who run Tribler headless. The script must be run interactively, and the thing is going to print percentages to the console. Using the logger would be too much for that task (especially to log a long list of upgrade percentages).

src/tribler-gui/tribler_gui/tests/test_gui.py
ichorid: That's a residue of testing different timeout values. It is the same as the default. I will remove it.

ichorid:
@drew2a , the remaining lines will be very hard to cover. In fact, this PR _increases_ the coverage - the lines that are shown as non-covered were never covered before (they just count as new lines, because these were extracted from CoreManager).

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/tests/test_gui.py
```diff
@@ -38,56 +39,35 @@ def api_port():
...
+        core_args=[str(RUN_TRIBLER_PY.absolute()), '--gui_test_mode'],
+    )  # pylint: disable=W0621
     app.set_activation_window(window)
     QTest.qWaitForWindowExposed(window)
 
     screenshot(window, name="tribler_loading")
-    wait_for_signal(window.core_manager.events_manager.tribler_started)
+    wait_for_signal(
+        window.core_manager.events_manager.tribler_started,
+        flag=window.core_manager.events_manager.tribler_started,
```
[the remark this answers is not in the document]
ichorid: :facepalm:

src/tribler-gui/tribler_gui/upgrade_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        connect(self.stop_upgrade, self._stop_upgrade)
+
+    def _stop_upgrade(self):
+        self._upgrade_interrupted_event.interrupted = True
+
+    def _update_status_callback(self, text):
+        self.status_update.emit(text)
+
+    def run(self):
+        from run_tribler_upgrader import upgrade_state_dir
```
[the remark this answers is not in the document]
ichorid: Good point :+1:

src/tribler-core/tribler_core/start_core.py
```diff
@@ -180,7 +160,6 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, gui_test_mo
     if trace_logger:
         trace_logger.close()
 
-    process_checker.remove_lock_file()
     # Flush the logs to the file before exiting
     for handler in logging.getLogger().handlers:
         handler.flush()
```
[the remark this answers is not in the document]
ichorid: Again, I would rather do it in a separate PR.

kozlovsky (dismissed):

ichorid:
retest this please

src/tribler-gui/tribler_gui/event_request_manager.py
```diff
@@ -131,6 +129,7 @@ def on_finished(self):
         self.connect_timer.setSingleShot(True)
         self.connect_timer.timeout.connect(self.connect)
         self.connect_timer.start(500)
+        self.tribler_started_flag = False
```
kozlovsky, line 132: **issue:** It is a bit dangerous to unset the flag here because it can lead to a race condition. If `tribler_started_flag` is set to True and then back to False quickly, it might remain unnoticed, as if it was never set to True.

**suggestion:** Probably, it is not necessary to set the flag back to False. Or you can introduce a separate `tribler_stopped_flag` and set it to True here.

**thought:**  The way we process Qt signals may be unsafe in some places: if a signal is raised too quickly, a slot might be not attached to the signal at this moment yet. Ideally, we should have a corresponding asyncio Event for each one-time PyQt signal, and the corresponding slot should just set this event. Then the Tribler code can wait for this event, and it will work correctly without a race condition even if the signal was triggered before we start waiting for the event.

kozlovsky (changes requested):

src/tribler-gui/tribler_gui/event_request_manager.py
ichorid: Good point!
The problem is, we don't use `asyncio` in the GUI. The reactor is run by QT, and there is no such thing as a "single-shot event" in QT. So, we have to invent something of our own.
I'll try to make this a bit safer :+1:

ichorid: Also, the follow-up to this PR will remove the `TRIBLER_STARTED` and the event entirely :wink:

devos50 (commented):
Looks good - all my feedback has been implemented. I just have one suggestion left.

Before merging, please make sure the pylint errors are fixed 👍

src/run_tribler.py
```diff
@@ -1,14 +1,34 @@
...
+    GUI_TEST_MODE = auto()
+
+
+class RunTriblerArgsParser(argparse.ArgumentParser):
+    def __init__(self, *args, **kwargs):
+        kwargs['description'] = 'Run Tribler BitTorrent client'
+        super().__init__(*args, **kwargs)
+        mode = self.add_mutually_exclusive_group()
+        mode.add_argument('--core', action='store_const', dest='mode', const=RunMode.CORE_ONLY)
+        mode.add_argument('--gui_test_mode', action='store_const', dest='mode', const=RunMode.GUI_TEST_MODE)
```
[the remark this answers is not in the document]
ichorid: Good point...

devos50 (approved):

kozlovsky (approved):

PR #6560, opened 2021-11-17
Add exist_ok=True to mkdir

### LINKED ISSUE #6157: [7.9.0] FileExistsError: '.Tribler' (in run_tribler > check_os.enable_fault_handler)
[WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\<...>\\AppData\\Roaming\\.Tribler'

[Here](https://github.com/Tribler/tribler/blob/e52c5aca43dbc4df3854334a2f5e9350f88120bd/src/tribler-core/tribler_core/check_os.py#L213):

```python
def enable_fault_handler(log_dir):
    """
    Enables fault handler if the module is available.
    """
    logger.info(f'Enable fault handler: "{log_dir}"')
    try:
        import faulthandler

        if not log_dir.exists():
            path_util.makedirs(log_dir)
```

Happened two times with a month interval. Not very often, probably we can ignore it. Looks like some weird race condition?

drew2a:
This PR fixes #6157

It also extends `patch_import` function by adding two new arguments:
* **strict:** If `strict` is equal to True, then whenever importing module exists or not, it will be replaced
        by MagicMock. If `strict` is equal to False, then the real module will be return in case it is exist, and
        MagicMock in case the module doesn't exist.
* **always_raise_exception_on_import:** if set to True, then the import of the particular module always will raise
        the ImportError

These improvements to the `patch_import` are necessary to properly test `enable_fault_handler` function.

### DIFF
--- a/src/tribler-common/tribler_common/patch_import.py
+++ b/src/tribler-common/tribler_common/patch_import.py
@@ -11,28 +11,41 @@ from typing import List
 
 import mock
 
-
 __all__ = ['patch_import']
 
 _builtins_import = builtins.__import__
 
 
-def patch_import(modules=List[str], **mock_kwargs):
+def patch_import(modules=List[str], strict: bool = False, always_raise_exception_on_import=False, **mock_kwargs):
     """
     Mocks import statement, and disable ImportError if a module
     could not be imported.
     :param modules: a list of prefixes of modules that should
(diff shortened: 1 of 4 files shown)


drew2a:
retest this please

kozlovsky (approved):

PR #6563, opened 2021-11-18
Specify versions in `requirement.txt` files.

### LINKED ISSUE #6562: Specify versions in requirements.txt
As the next logical step after #6227 and #6518, we should specify versions for the libraries in all main `requirements.txt` in the project:
* https://github.com/Tribler/tribler/blob/main/src/requirements.txt
* https://github.com/Tribler/tribler/blob/main/src/requirements-test.txt
* https://github.com/Tribler/tribler/blob/main/doc/requirements.txt

Ref:
* https://pip.pypa.io/en/stable/cli/pip_install/#requirements-file-format
* https://www.python.org/dev/peps/pep-0440/#version-specifiers

drew2a:
This PR fixes #6562 by specifying versions in `requrement.txt` files.


drew2a:
retest this pleasee

devos50 (approved):
Looks good. My only concern is that we should regularly update these dependencies which adds one more item on the chores list (we tend to forget these kind of things).

We are using [Poetry](https://python-poetry.org) in the [BAMI](https://github.com/tribler/bami) project for dependency management. Poetry provides a bit more flexibility when specifying dependencies, and can also setup a virtual environment based on the specified dependencies. This can be helpful when we eventually are going to run tests/build Tribler in a virtual environment.

kozlovsky (changes requested):
When I fixed the 1-Click release, there were many unrelated problems when the build finished correctly, but the resulting binary refuses to work or crashes for different reasons.

Empirically, I managed to find such a combination of packages for which the build worked correctly:

```
PyQt5==5.15.4
PyQt5-Qt5==5.15.2
PyQt5-sip==12.8.0
pyqtgraph==0.11.0
PyQtWebEngine==5.15.2
```

As an example of the problem, there is a bug in the latest version of the PyQt5-sip library, resulting to the following error:

"RuntimeError: the sip module implements API v12.0 to v12.8 but the PyQt5.QtCore module requires API v12.9"
Similar to: https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1828923.html

I think we need to freeze the versions of these modules as well, even if we are not importing them directly in Tribler code.

drew2a:
@kozlovsky 5.15.4 QT doesn't work on my Mac.

```
PyQt5==5.15.4
```

```python
  File "/Users/<user>/Projects/github.com/Tribler/tribler/src/run_tribler.py", line 5, in <module>
    from PyQt5.QtCore import QSettings
ImportError: dlopen(/Users/<user>/Projects/github.com/Tribler/tribler/venv/lib/python3.9/site-packages/PyQt5/QtCore.abi3.so, 0x0002): Library not loaded: @rpath/QtCore.framework/Versions/5/QtCore
  Referenced from: /Users/<user>/Projects/github.com/Tribler/tribler/venv/lib/python3.9/site-packages/PyQt5/QtCore.abi3.so
  Reason: tried: '/Users/<user>/Projects/github.com/Tribler/tribler/venv/lib/python3.9/site-packages/PyQt5/Qt5/lib/QtCore.framework/Versions/5/QtCore' (no such file), '/Users/<user>/Projects/github.com/Tribler/tribler/venv/lib/python3.9/site-packages/PyQt5/Qt5/lib/QtCore.framework/Versions/5/QtCore' (no such file), '/usr/local/lib/QtCore.framework/Versions/5/QtCore' (no such file), '/usr/local/lib/QtCore.framework/Versions/5/QtCore' (no such file), '/Library/Frameworks/QtCore.framework/Versions/5/QtCore' (no such file), '/System/Library/Frameworks/QtCore.framework/Versions/5/QtCore' (no such file)
```

drew2a:
We can specify versions per platform eg:
```
SomeProject ==5.4 ; python_version < '3.8'
SomeProject; sys_platform == 'win32'
```

kozlovsky:
Well, you are right, this configuration is from our mac builder:

PyQt5==5.15.1
PyQt5-sip==12.8.1
pyqtgraph==0.12.3
PyQtWebEngine==5.15.2

It is actually a problem that our different builder machines currently has different set of libraries

Let's specify this set of libraries, and re-test them on different platforms in hope they will work on Windows and Linus as well

kozlovsky (approved):

PR #6534, opened 2021-11-08
Install pip dependencies during the build

xoriole:
Fixes https://github.com/Tribler/tribler/issues/6518

PR #6568, opened 2021-11-19
Fixes 1-Click release for 7.11

kozlovsky:
This PR together with #6545 contains fixes for various problems with Tribler builds.
With this fixes, 1-Click release Jenkins job is able to assemble executables and test them with application tester

### DIFF
--- a/build/win/makedist_win.bat
+++ b/build/win/makedist_win.bat
@@ -46,7 +46,7 @@ REM Arno: When adding files here, make sure tribler.nsi actually
 REM packs them in the installer .EXE
 
 python3 -m pip install --upgrade --no-deps --force-reinstall pydantic --no-binary pydantic
-python3 -m pip install --upgrade --no-deps --force-reinstall typing_extensions --no-binary typing_extensions
+python3 -m pip install --upgrade --no-deps --force-reinstall typing_extensions==3.10.0.2 --no-binary typing_extensions
 
 %PYTHONHOME%\Scripts\pyinstaller.exe tribler.spec
 
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -1,3 +1,8 @@
+# A fix for "LookupError: unknown encoding: idna" error.
+# Adding encodings.idna to hiddenimports is not enough.
+# https://github.com/pyinstaller/pyinstaller/issues/1113
+import encodings.idna  # pylint: disable=unused-import
+
 import logging.config
 import os
 import sys
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -2,7 +2,6 @@
 # Adding encodings.idna to hiddenimports is not enough.
 # https://github.com/pyinstaller/pyinstaller/issues/1113
 import encodings.idna  # pylint: disable=unused-import
-
 import logging.config
 import os
 import sys
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -140,6 +140,7 @@ if __name__ == "__main__":
             logger.info("Shutting down Tribler")
             if trace_logger:
                 trace_logger.close()
+
             # Flush all the logs to make sure it is written to file before it exits
             for handler in logging.getLogger().handlers:
                 handler.flush()
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -4,15 +4,15 @@ import sys
 
 from PyQt5.QtCore import QSettings
 
+from tribler_common.sentry_reporter.sentry_reporter import SentryReporter, SentryStrategy
+from tribler_common.sentry_reporter.sentry_scrubber import SentryScrubber
+
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


build/win/makedist_win.bat
```diff
@@ -46,7 +46,7 @@ REM Arno: When adding files here, make sure tribler.nsi actually
 REM packs them in the installer .EXE
 
 python3 -m pip install --upgrade --no-deps --force-reinstall pydantic --no-binary pydantic
-python3 -m pip install --upgrade --no-deps --force-reinstall typing_extensions --no-binary typing_extensions
+python3 -m pip install --upgrade --no-deps --force-reinstall typing_extensions==3.10.0.2 --no-binary typing_extensions
```
kozlovsky, line 49: Without --no-binary PyInstaller does not include this module into the Tribler exe file on Windows
The latest version of typing_extension (4.0.0) is currently not working as well.
After #6534 we can move it into requirements with --no-binary specification

src/tribler-core/tribler_core/check_os.py
```diff
@@ -89,8 +89,14 @@ def should_kill_other_tribler_instances(root_state_dir):
     if current_pid != old_pid and old_pid > 0:
         # If the old process is a zombie, simply kill it and restart Tribler
         old_process = psutil.Process(old_pid)
-        logger.info(f'Old process status: {old_process.status()}')
-        if old_process.status() == psutil.STATUS_ZOMBIE:
```
kozlovsky, line 93: Sometimes Tribler crashes on psutil.NoSuchProcess exception when reading `old_process.status()`

src/tribler-core/tribler_core/start_core.py
```diff
@@ -174,12 +174,15 @@ def start_tribler_core(base_path, api_port, api_key, root_state_dir, gui_test_mo
     loop = asyncio.get_event_loop()
     loop.set_exception_handler(CoreExceptionHandler.unhandled_error_observer)
 
-    loop.run_until_complete(core_session(config, components=list(components_gen(config))))
-
-    if trace_logger:
-        trace_logger.close()
-
-    process_checker.remove_lock_file()
```
kozlovsky, line 182: We must always remove the lock file on Tribler crash to re-run the application tester reliably, so try/finally block is necessary here.

src/tribler-gui/tribler_gui/core_manager.py
```diff
@@ -64,7 +64,12 @@ def on_core_read_ready(self):
         if b'Traceback' in raw_output:
             self.core_traceback = decoded_output
             self.core_traceback_timestamp = int(round(time.time() * 1000))
-        print(decoded_output.strip())  # noqa: T001
```
kozlovsky, line 67: Sometimes Tribler crashes when trying to write data to stdout during the shutdown

src/tribler-core/tribler_core/start_core.py
kozlovsky: Our lock file mechanics are broken, as it was built on the assumption that OS does not reuse PIDs before system restart. This is not true, Windows especially reuses process IDs quickly, and that leads to many non-obvious bugs. I will open a separate issue for that.

src/run_tribler.py
```diff
@@ -135,7 +139,10 @@ def init_boot_logger():
             logger.info("Shutting down Tribler")
             if trace_logger:
                 trace_logger.close()
+
             # Flush all the logs to make sure it is written to file before it exits
             for handler in logging.getLogger().handlers:
                 handler.flush()
+
+            SentryReporter.global_strategy = SentryStrategy.SEND_SUPPRESSED
```
kozlovsky, line 147: Without this Sentry suppression, it sometimes happens that an error takes place during the Tribler shutdown. Then Sentry tries to display the dialog window, but the Qt state is already partially destructed, which leads to mysterious crashes.

tribler.spec
```diff
@@ -93,18 +93,18 @@ hiddenimports = [
...
+    'pyaes',
     'pydantic',
-    'PyQt5.QtTest',
     'pyqtgraph',
     'pyqtgraph.graphicsItems.PlotItem.plotConfigTemplate_pyqt5',
     'pyqtgraph.graphicsItems.ViewBox.axisCtrlTemplate_pyqt5',
     'pyqtgraph.imageview.ImageViewTemplate_pyqt5',
+    'PyQt5.QtTest',
+    'requests',
+    'scrypt', '_scrypt',
```
devos50, line 106: I believe `pyaes`, `sqlalchemy` and `script` are not required anymore. These were dependencies for `bitcoinlib` (yes, we used to have a Bitcoin wallet in Tribler 😁 ) but we remove that library and we apparently forgot to remove them from `tribler.spec`.

devos50 (approved):
Great set of changes! 👍

tribler.spec
kozlovsky: Good to know! I'll remove them in a separate PR after additional testing

PR #6573, opened 2021-11-23
Simplify GUI-Core startup sequence

### LINKED ISSUE #6526: GUI-Core startup communication is too complicated

ichorid:
fixes #6526 

The third and the last part of #6535

4. Remove `StateEndpoint` and all Tribler startup state-related code.
5. Refactor all the endpoints to use true dependency injection. No late assignment of attributes anymore!
5. Move the initialization of all the endpoints into the `RESTComponent`. The relationship between `RESTComponent` and other components are now reversed: `RESTComponent` now depends on every other component that has something to do with the endpoints.
6.  All the endpoints are now initialized in one go, in the `RESTComponent`
7. CoreManager in the GUI is now split into two parts: CoreManager and UpgradeManager. UpgraderManager runs upgrader routines in a QThread. When it finishes, CoreManager is run as usual.


doc/extract_swagger.py
```diff
@@ -18,8 +18,8 @@ async def extract_swagger(destination_fn):
     config.api.http_enabled = False
     config.api.https_enabled = False
 
-    root_endpoint = RootEndpoint(config, middlewares=[ApiKeyMiddleware(config.api.key), error_middleware])
-    api_manager = RESTManager(config=config.api, root_endpoint=root_endpoint, state_dir=config.state_dir)
+    root_endpoint = RootEndpoint(middlewares=[ApiKeyMiddleware(config.api.key), error_middleware])
+    api_manager = RESTManager(config.api, root_endpoint, config.state_dir)
```
drew2a, line 22: **NIT** seems that this change is unnecessary.

src/tribler-common/tribler_common/patch_import.py
```diff
@@ -8,11 +8,11 @@
 """
 import builtins
 from typing import List
-
-import mock
+from unittest.mock import Mock, patch
```
drew2a, line 11: **Issue** : this is a part of the tribler_common package, so this can be used in other parts of code in production.

Therefore it seems to be wrong to change the mock library to `unittest.mock`

**Solution**: keep mock as a dependency as it was before.

src/tribler-common/tribler_common/patch_import.py
```diff
@@ -41,11 +41,11 @@ def try_import(module_name, *args, **kwargs):
             raise ImportError
 
         if strict:
-            return mock.MagicMock(**mock_kwargs)
+            return Mock(**mock_kwargs)
```
drew2a, line 44: **Question**: What is the reason to change `MagicMock` to `Mock`?

src/tribler-core/tribler_core/components/restapi/restapi_component.py
```diff
@@ -85,36 +65,63 @@ async def run(self):
...
+        key_component                  = await self.require_component(KeyComponent)
+        ipv8_component                 = await self.maybe_component(Ipv8Component)
+        libtorrent_component           = await self.maybe_component(LibtorrentComponent)
+        resource_monitor_component     = await self.maybe_component(ResourceMonitorComponent)
+        bandwidth_accounting_component = await self.maybe_component(BandwidthAccountingComponent)
+        gigachannel_component          = await self.maybe_component(GigaChannelComponent)
+        tag_component                  = await self.maybe_component(TagComponent)
+        tunnel_component               = await self.maybe_component(TunnelsComponent)
+        torrent_checker_component      = await self.maybe_component(TorrentCheckerComponent)
+        gigachannel_manager_component  = await self.maybe_component(GigachannelManagerComponent)
```
drew2a, line 82: **Note:** this formatting is nice, but it will be reformated with high probability by IDE on the next change.

src/tribler-core/tribler_core/components/metadata_store/restapi/metadata_endpoint_base.py
```diff
@@ -36,10 +37,10 @@
 
 
 class MetadataEndpointBase(RESTEndpoint):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, metadata_store: MetadataStore,  *args, tags_db: TagDatabase = None, **kwargs):
```
devos50, line 40: **Nit**: there are two spaces after `Metadatastore,`, you could remove one.

src/tribler-core/tribler_core/components/restapi/rest/debug_endpoint.py
```diff
@@ -41,12 +44,12 @@ class DebugEndpoint(RESTEndpoint):
     This endpoint is responsible for handing requests regarding debug information in Tribler.
     """
 
-    def __init__(self, *args, **kwargs):
-        RESTEndpoint.__init__(self, *args, **kwargs)
-        self.state_dir = None
-        self.log_dir = None
-        self.tunnel_community = None
-        self.resource_monitor = None
+    def __init__(self, state_dir: Path, log_dir: Path, tunnel_community:TunnelCommunity=None, resource_monitor:ResourceMonitor=None):
```
devos50: **Suggestion:** I'm not sure why pylint is not raising an issue here but we usually have a space after the `:` when using `typing`. For consistency, consider changing this for the last two attributes.

src/tribler-core/tribler_core/components/restapi/rest/events_endpoint.py
```diff
@@ -74,7 +74,9 @@ def __init__(self):
         self.undelivered_error: Optional[dict] = None
 
         # We need to know that Tribler completed its startup sequence
+        # FIXME: remove this
```
devos50: **Question:** I'm not sure sure what should be removed here and why.

src/tribler-core/tribler_core/components/tag/restapi/tags_endpoint.py
```diff
@@ -28,10 +28,10 @@ class TagsEndpoint(RESTEndpoint):
     Top-level endpoint for tags.
     """
 
-    def __init__(self, *args, **kwargs):
-        RESTEndpoint.__init__(self, *args, **kwargs)
-        self.db: Optional[TagDatabase] = None
-        self.community: Optional[TagCommunity] = None
+    def __init__(self, db: TagDatabase, community: TagCommunity):
+        super(TagsEndpoint, self).__init__()
```
devos50: **Suggestion:** You can just call `super()`.

src/tribler-core/tribler_core/components/restapi/restapi_component.py
```diff
@@ -1,82 +1,62 @@
...
 
     _events_endpoint: EventsEndpoint
-    _state_endpoint: StateEndpoint
     _core_exception_handler: CoreExceptionHandler = default_core_exception_handler
 
+    def maybe_add(self, path: str, endpoint_cls: Type[RESTEndpoint], *args, **kwargs):
+        """ Add the corresponding endpoint to the path in case there are no `NoneComponent`
+        in *args or **kwargs
+        """
+        if any(isinstance(arg, NoneComponent) for arg in list(args) + list(kwargs.values())):
```
drew2a, line 54: **Issue**: this expression is hard to read.
**Solution**:

```suggestion
        arguments_chain = chain(args, kwargs.values())
        if any(isinstance(arg, NoneComponent) for arg in arguments_chain):
```

drew2a (changes requested):
This is a nice improvement in general! I've left some comments.

devos50 (commented):
Good changes! I have a few minor comments here and there.

doc/extract_swagger.py
ichorid: (Removing `config` argument from `RootEndpoint` is necessary.)
The change from named to unnamed arguments is to conform to the way of indicating required vs optional arguments:
required args must be unnamed, while optional must be named.

src/tribler-common/tribler_common/patch_import.py
ichorid: From https://newbedev.com/python-mock-vs-unittest-mock-patch:

>The mock library has been integrated into the Python standard library from Python version 3.3 on as unittest.mock. They deliver the same functionality.

>Nowadays the (external) mock library is a backport of the version in the standard library. If you are using a recent version of Python and don't have any special version requirement, the version from the standard library should be preferred.

The change is due to `docs-build.sh` failing because of unresolved dependency to `mock`. I was facing the choice of:
* adding `mock` to `requirements.txt`
* switch to `unittest.mock`

I choose the latter option, because it removes the external dependency, which is a backport.

src/tribler-common/tribler_common/patch_import.py
ichorid: https://stackoverflow.com/questions/17181687/mock-vs-magicmock

In short, regular `Mock` does **not** provide so-called "magic methods". This has two advantages:
1. it is a bit lighter
2. if you don't _need_ to mock magic methods, you really **should not** mock them (e.g. you don't want your objects to behave like a container, don't you? :wink: )

src/tribler-core/tribler_core/components/restapi/restapi_component.py
ichorid: That's why I put `fmt: off` there.

src/tribler-core/tribler_core/components/metadata_store/restapi/metadata_endpoint_base.py
ichorid: Changed. For some reason, the linter did not catch that :thinking:

src/tribler-core/tribler_core/components/restapi/rest/debug_endpoint.py
ichorid: Again, the linter was silent... :sheep:

src/tribler-core/tribler_core/components/restapi/rest/events_endpoint.py
ichorid: outdated...

src/tribler-core/tribler_core/components/restapi/restapi_component.py
ichorid: FYI: https://stackoverflow.com/questions/3375307/how-to-disable-code-formatting-for-some-part-of-the-code-using-comments

doc/extract_swagger.py
drew2a: Is it a common practice? If yes, could you post a link to the resource that profs that?

devos50 (dismissed):

src/tribler-common/tribler_common/patch_import.py
ichorid: `build-docs.sh` tries uses  `doc/requirements.txt`. So, if you don't use the Jenkins job, but instead tries to build it locally, it fails with missing dependencies.

src/tribler-common/tribler_common/patch_import.py
drew2a: > it is a bit lighter

True

>if you don't need to mock magic methods, you really should not mock them (e.g. you don't want your objects to behave like a container, don't you? 😉 )

In the method above `MagicMock` is used as a mock class for the library patch. Therefore `MagicMock` will be returned as a library replacement.

In the case of the following library usage, `MagicMock` gives you a correct behavior:

```python
    import library
    print(len(library.some_set))
```

With your improvement, the example above will fail.

src/tribler-core/tribler_core/components/restapi/restapi_component.py
drew2a: > That's why I put fmt: off there.

It will not help with IDE formatting.

doc/extract_swagger.py
ichorid: изображение


https://www.python.org/dev/peps/pep-0570/#benefits-of-positional-only-parameters


An alternative opinion:
https://trstringer.com/python-named-arguments/

src/tribler-common/tribler_common/patch_import.py
drew2a: Now it is clear, thank you. Don't forget to remove actual `Mock` dependency from

https://github.com/ichorid/tribler/blob/b95ffaa55304c103858fe4dca9724c6433e303ab/src/tribler-common/tribler_common/requirements.txt#L1

drew2a (changes requested):
The requested changes were not addressed.

doc/extract_swagger.py
drew2a: The screenshot above (and link to the python.org) doesn't refer to 

>The change from named to unnamed arguments is to conform to the way of indicating required vs optional arguments: required args must be unnamed, while optional must be named.

ichorid: Let's put it a different way: in some cases (e.g. C-based functions), using named parameters can lead to incompatibility.
For instance, let's say one day we decide to replace some class definitions with C-based counterparts. Then, named arguments won't work anymore. CPython makes the false promise of allowing addressing positional arguments by name, as shown in the screenshot. Therefore, we must adhere to the safest option possible: address positional arguments by position, and addressing named arguments by name. Another argument goes for subclassing and avoiding Liskov violation:
изображение

Yet another argument is the one of _consistency_: this is the first time I see someone address positional arguments by name in our codebase. If we decide to allow addressing positional arguments by name, we must provide guidelines for that (e.g. "always address by name"). Otherwise, this is surprising and inconsistent.

src/tribler-common/tribler_common/patch_import.py
ichorid: > With your improvement, the example above will fail.

And that is exactly what we want. Tests must fail when something unexpected happens. When that happens, we replace the offending mock with `MagicMock`. But at least we do that for a reason.

src/tribler-core/tribler_core/components/restapi/restapi_component.py
ichorid: > It will not help with IDE formatting.

If you enable your IDE to do so (see the link I posted above), it will. Remind that we don't have an official guideline for the code to be friendly to a particular IDE-based formatting tool. Instead, we have the `black` pre-commit hook (which respects `fmt: off`).

doc/extract_swagger.py
drew2a: > Let's put it a different way: in some cases (e.g. C-based functions), using named parameters can lead to incompatibility.

Agree, but there are no C-based functions here. We should not treat all of our functions as C-basedso because of the potential presence of C-based functions somewhere.

> Yet another argument is the one of consistency: this is the first time I see someone address positional arguments by name in our codebase.

This change was introduced by @xoriole. I also use this approach during writing code for Tribler because it increases readability.

> Another argument goes for subclassing and avoiding Liskov violation:
I don't see any application of Liskov violation in our case. 
Even if you want to refactor some parts of Tribler by replacing them with C-based counterparts there are plenty of ways how to do it without Liskov violation. 
It is also an extremely unlikely scenario that we would change `RESTManager` to a C-based counterpart.

src/tribler-common/tribler_common/patch_import.py
drew2a: I'm not sure about your intention, but my intention was the following. I introduced this mechanism mostly for mocking libraries during swagger extraction (doc build).

You can not guarantee that libraries from our `requerements.txt` are not used in the way that I described above.

Also, it should never fail for the example above even for the tests because it is not unexpected behavior. It is expected behavior.

src/tribler-core/tribler_core/components/restapi/restapi_component.py
drew2a: Black pre-commit hook is not mandatory to use nor agreed to use by all the team.

I just noticed here, that a person who will change a line in this particular file will remove all formatting with a very high probability.

src/tribler-common/tribler_common/patch_import.py
ichorid: >Also, it should never fail for the example above even for the tests because it is not unexpected behavior. It is expected behavior.


Good point, I overlooked that.

src/tribler-core/tribler_core/components/restapi/restapi_component.py
ichorid: > will remove all formatting with a very high probability.

I'm fine with that. That's for that person and their PR reviewer to decide what to do with the code they change :wink:

doc/extract_swagger.py
ichorid: For the sake of keeping the team productive, and not discussing petty style issues, I'm reverting this change. In the future, I will not do this kind of change to prevent such useless disputes. :shrug:

src/tribler-common/tribler_common/patch_import.py
drew2a: I will add a test to document desirable behavior after your PR be merged.

src/tribler-common/tribler_common/tests/test_dependencies.py
```diff
@@ -37,7 +37,8 @@ async def test_get_dependencies():
     # assert that in each scope dependencies are exist
     assert list(get_dependencies(Scope.gui))
     assert list(get_dependencies(Scope.core))
-    assert list(get_dependencies(Scope.common))
+    # Disabled because the list of common dependencies is currently empty
+    #assert list(get_dependencies(Scope.common))
```
drew2a, line 41: **Nit** you can safely remove these lines.

src/tribler-common/tribler_common/tests/test_patch_import.py
```diff
@@ -26,13 +25,13 @@ async def test_mock_import_import_real_lib():
 @patch_import(['time'])
 async def test_mock_import_not_strict():
     import time
-    assert not isinstance(time, MagicMock)
+    assert not isinstance(time, Mock)
```
drew2a, line 28: **Issue**: after reverting back `patch_import` to the using of `MagicMock` class, this line also should be reverted.

src/tribler-common/tribler_common/tests/test_patch_import.py
```diff
@@ -26,13 +25,13 @@ async def test_mock_import_import_real_lib():
...
 
 
 @patch_import(['time'], strict=True)
 async def test_mock_import_strict():
     import time
-    assert isinstance(time, MagicMock)
+    assert isinstance(time, Mock)
```
drew2a, line 34: **Issue**: after reverting back `patch_import` to the using of `MagicMock` class, this line also should be reverted.

drew2a (changes requested):
Added some minor and easy-to-fix comments.

src/tribler-common/tribler_common/tests/test_patch_import.py
ichorid: Funny how this does not fail... :thinking:

src/tribler-common/tribler_common/tests/test_dependencies.py
ichorid: Done

drew2a (approved):
Thanks for addressing all the questions and for this improvement.

src/tribler-common/tribler_common/patch_import.py
drew2a: Here is the tests improvement: https://github.com/Tribler/tribler/pull/6588

Issue #6502, opened 2021-10-27
test_drop_silent_peer failing during tests on Mac


devos50:
See [here](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_mac_pytest/2292/testReport/junit/src.tribler-core.tribler_core.components.metadata_store.remote_query_community.tests.test_remote_query_community/TestRemoteQueryCommunity/test_drop_silent_peer/) for the detailled log.

```
Failed: Timeout >5.0s

self = <tribler_core.components.metadata_store.remote_query_community.tests.test_remote_query_community.TestRemoteQueryCommunity testMethod=test_drop_silent_peer>

    def setUp(self):
        random.seed(456)
        super().setUp()
        self.count = 0
        self.metadata_store_set = set()
>       self.initialize(BasicRemoteQueryCommunity, 2)

src/tribler-core/tribler_core/components/metadata_store/remote_query_community/tests/test_remote_query_community.py:63: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/pyipv8/ipv8/test/base.py:57: in initialize
    self.nodes = [self.create_node(*args, **kwargs) for _ in range(node_count)]
src/pyipv8/ipv8/test/base.py:57: in <listcomp>
    self.nodes = [self.create_node(*args, **kwargs) for _ in range(node_count)]
src/tribler-core/tribler_core/components/metadata_store/remote_query_community/tests/test_remote_query_community.py:72: in create_node
    metadata_store = MetadataStore(
src/tribler-core/tribler_core/components/metadata_store/db/store.py:220: in __init__
    self._db.execute(sql_create_fts_table)
<string>:2: in execute
    ???
/usr/local/lib/python3.8/site-packages/pony/utils/utils.py:61: in cut_traceback
    try: return func(*args, **kwargs)
/usr/local/lib/python3.8/site-packages/pony/orm/core.py:870: in execute
    return database._exec_raw_sql(sql, globals, locals, frame_depth=cut_traceback_depth+1, start_transaction=True)
/usr/local/lib/python3.8/site-packages/pony/orm/core.py:882: in _exec_raw_sql
    return database._exec_sql(adapted_sql, arguments, False, start_transaction)
/usr/local/lib/python3.8/site-packages/pony/orm/core.py:938: in _exec_sql
[...]

PR #6574, opened 2021-11-23
Use in-memory SQLite DB for Tags in GUI test mode

ichorid:
It is more efficient and convenient to use an in-memory SQLite database for testing.

### DIFF
--- a/src/tribler-core/tribler_core/components/tag/tag_component.py
+++ b/src/tribler-core/tribler_core/components/tag/tag_component.py
@@ -20,13 +20,9 @@ class TagComponent(RestfulComponent):
         self._ipv8_component = await self.require_component(Ipv8Component)
         key_component = await self.require_component(KeyComponent)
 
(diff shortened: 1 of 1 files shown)


ichorid:
retest this please

drew2a (changes requested):
This is a nice change, but tests should pass before the approval.

devos50:
Seems that it triggered #6502 twice.

devos50:
retest this please

devos50:
With this code, the `test_drop_silent_peer` test seems to pass on my (Mac) machine:

```
test_remote_query_community.py::TestRemoteQueryCommunity::test_drop_silent_peer 

======================== 1 passed, 17 warnings in 2.65s ========================
```

Note, however, that it takes 2.65 seconds to complete the test. The test might exceed the 5-second timeout if the machine is under heavy load. Since we are running tests in parallel, it might trigger the timeout.

ichorid:
@devos50 , indeed, the test takes too long. I'll handle it in a separate PR.

drew2a (approved):
👍

Issue #6581, opened 2021-11-24
Tribler logging systems need fixes/improvements.


kozlovsky:
Currently, the logging system has the following problems:
1. Incorrect configuration of stdout/stderr handlers: log messages with WARNING level are silenced and appear neither in stdin nor stdout (discovered by @ichorid).
2. Upgrader configuration is incorrect - the configuration is placed in `tribler_core/logger.yaml`, but the config file used with upgrader is `tribler_gui/logger.yaml`.
3. We have several logger config files with similar settings, and sometimes it is hard to understand which file should be modified.
4. The logger config is overcomplicated, and some settings are actually not used.
5. The code for logging setup is duplicated in GUI and Core with minimal changes.
6. GUI does not separate stdout/stderr output received from Core and redirects everything to stdout.

ichorid:
Note that indentation introduced in https://github.com/Tribler/tribler/pull/6567 is applied inconsistently for some reason.

```

	[PID:25100] 2021-11-25 10:40:59,170 - INFO - PopularityCommunity(13) - Init. Settings: enabled=True cache_dir='health_cache'.
	[PID:25100] 2021-11-25 10:40:59,245 - INFO - RESTManager(134) - Started HTTP REST API: http://localhost:52194
[PID:25060] 2021-11-25 10:40:59,260 - INFO - TriblerGUI(128) - Will connect to events endpoint
	[PID:25100] 2021-11-25 10:40:59,335 - INFO - Socks5Connection(99) - Client has sent METHOD REQUEST
[PID:25100] 2021-11-25 10:40:59,335 - INFO - Socks5Connection(181) - Accepting UDP ASSOCIATE request to 127.0.0.1:50376 (BIND addr 0.0.0.0:0)
[PID:25100] 2021-11-25 10:40:59,714 - INFO - TriblerTunnelCommunity(247) - Creating a new circuit of length 1 (type: DATA)

```

PR #6585, opened 2021-11-24
Fixes for Tribler logger

kozlovsky:
This PR fixes all issues with logging listed in #6581

Now:
1. GUI and Core load logger configuration from the same config file `tribler_common/logger.yaml`.
2. Logger file structure was simplified, unused code was removed.
3. Logger messages with the WARNING level are not dropping silently.
4. Core errors that GUI reads from Core stderr go to GUI stderr.


kozlovsky:
retest this please

ichorid:
retest this please

kozlovsky:
retest this please

src/tribler-common/tribler_common/logger.py
```diff
@@ -0,0 +1,67 @@
...
+
+import yaml
+
+LOG_CONFIG_FILENAME = 'logger.yaml'
+
+
+logger = logging.getLogger(__name__)
+
+
+# note: this class is used by src/tribler-common/tribler_common/logger/config.yaml
```
drew2a, line 14: 👍

src/tribler-common/tribler_common/logger.yaml
```diff
@@ -64,12 +61,11 @@ handlers:
...
-loggers:
-    TriblerGUI:
-        level: INFO
-        handlers: [console, error_console, info_memory_handler, error_memory_handler]
-        propagate: no
+# loggers:
+#    ExampleLoggerName:
+#        level: INFO
+#        handlers: [stdout_handler, stderr_handler, info_memory_handler, error_memory_handler]
+#        propagate: no
```
drew2a, line 71: **Nit:** Commented part can be safely removed.

src/tribler-common/tribler_common/logger.py
kozlovsky: Fixed the reference to point to the correct file :)

src/tribler-common/tribler_common/tests/test_logger.py
```diff
@@ -0,0 +1,85 @@
+import logging
+from unittest.mock import MagicMock, Mock, call, patch
+
+from tribler_common.logger import get_logger_config_path, setup_logging
+
+
+@patch('tribler_common.logger.__file__', '/a/b/c/logger.py')
```
drew2a, line 7: **Question:** Does this mean that `tribler_common.logger.__file__` will return `'/a/b/c/logger.py'?

src/tribler-gui/tribler_gui/core_manager.py
```diff
@@ -101,9 +112,10 @@ def start_tribler_core(self, core_args=None, core_env=None):
 
         self.core_process = QProcess()
         self.core_process.setProcessEnvironment(core_env)
-        self.core_process.setReadChannel(QProcess.StandardOutput)
-        self.core_process.setProcessChannelMode(QProcess.MergedChannels)
-        connect(self.core_process.readyRead, self.on_core_read_ready)
+        self.core_process.setReadChannel(QProcess.StandardOutput)  # not necessary?
```
drew2a, line 115: **Nit:**  this comment is not useful for future readers and may be deleted

drew2a (dismissed):

ichorid (approved):

src/tribler-common/tribler_common/logger.yaml
kozlovsky: It indeed can be removed, but I left it as an example of loggers section structure. It is primarily useful for suppressing some overly verbose logger. For example, a developer can set DEBUG verbosity in logs, but want to turn off `TunnelDispatcher` and `PyQt5.uic` loggers to reduce spam in console. Probably it is better for me to extend the comment and add detailed example for this use-case.

src/tribler-common/tribler_common/tests/test_logger.py
kozlovsky: It will return file path, which depends on machine's directory layout and operation system. In test I just use some abstract path to ensure that it will be handled correctly.

src/tribler-gui/tribler_gui/core_manager.py
kozlovsky: I suspect this line of code can be removed, but it requires further investigation.
It may be useful for readers to know of the code that the presence of this line does not have obvious reason, in case they want to refactor the code.
In principle, I can remove the line right now in hope that it will not break something non-obvious.

src/tribler-common/tribler_common/tests/test_logger.py
drew2a: 👍

drew2a (dismissed):

kozlovsky:
retest this please

ichorid (approved):

PR #6592, opened 2021-11-29
Replace random char generation by fake words

drew2a:
This PR affects only Tribler's work in GUI test mode.

This PR replaces random char generation by fake words:

image
image
image
image

### DIFF
--- a/src/tribler-core/tribler_core/components/metadata_store/utils.py
+++ b/src/tribler-core/tribler_core/components/metadata_store/utils.py
@@ -1,6 +1,7 @@
 import random
-import string
 import time
+
+from faker import Faker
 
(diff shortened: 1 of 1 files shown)


drew2a:
retest this please

devos50 (commented):
Even though it looks better, we have actually discovered bugs related to Unicode characters in the past. Not using non-ASCII characters in the GUI test mode might hide these bugs.

devos50 (approved):

drew2a:
> Even though it looks better, we have actually discovered bugs related to Unicode characters in the past. Not using non-ASCII characters in the GUI test mode might hide these bugs.

Agree but discovering such bugs is closer to the set of responsibilities of Application Tester rather then GUI test mode.

Issue #6582, opened 2021-11-24
[Tests] Error in Windows GUI tests "in connect connect(self.reply.readyRead, self.on_read_data)"


drew2a:
During the test of https://github.com/Tribler/tribler/pull/6580 the following error occurred:

### Stacktrace

```python
ERROR    ErrorHandler:error_handler.py:40 tribler_gui.utilities.CreationTraceback: 
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\tribler\AppData\Local\Programs\Python\Python38\Scripts\pytest.exe\__main__.py", line 7, in <module>
    sys.exit(console_main())
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\_pytest\config\__init__.py", line 180, in console_main
    code = main()
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\_pytest\config\__init__.py", line 157, in main
    ret = config.hook.pytest_cmdline_main(
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\pluggy\hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\pluggy\manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\pluggy\manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\pluggy\callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\_pytest\main.py", line 289, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\_pytest\main.py", line 240, in wrap_session
    session.exitstatus = doit(config, session) or 0
[...]

PR #5821, opened 2020-12-08
Improve GUI error handling

drew2a:
This PR improves GUI error handling by adding:
1. A user identification based on the user's public key
1. Breadcrumbs for QWidget and QDialog 

An example of an issue with these improvements: https://sentry.tribler.org/organizations/tribler/issues/265

PR #6596, opened 2021-11-30
Fixes for GUI startup sequence

ichorid:
This PR adds some fixes related to #6582 

Also, fixes error buffering broken after startup refactoring.

### DIFF
--- a/src/tribler-core/tribler_core/components/restapi/restapi_component.py
+++ b/src/tribler-core/tribler_core/components/restapi/restapi_component.py
@@ -126,6 +126,10 @@ class RESTComponent(Component):
             self._events_endpoint.on_tribler_exception(reported_error)
 
         self._core_exception_handler.report_callback = report_callback
+        # Reraise the unreported error, if there is one
+        if self._core_exception_handler.unreported_error:
+            report_callback(self._core_exception_handler.unreported_error)
+            self._core_exception_handler.unreported_error = None
 
     async def shutdown(self):
         await super().shutdown()
--- a/src/tribler-gui/tribler_gui/event_request_manager.py
+++ b/src/tribler-gui/tribler_gui/event_request_manager.py
@@ -45,6 +45,7 @@ class EventRequestManager(QNetworkAccessManager):
         self.shutting_down = False
         self.error_handler = error_handler
         self._logger = logging.getLogger(self.__class__.__name__)
+        # This flag is used to prevent race condition when starting GUI tests
         self.tribler_started_flag = False
         self.reactions_dict = {
             NTFY.CHANNEL_ENTITY_UPDATED.value: self.node_info_updated.emit,
@@ -61,6 +62,7 @@ class EventRequestManager(QNetworkAccessManager):
 
     def events_start_received(self, event_dict):
         if event_dict["version"]:
+            self.tribler_started_flag = True
             self.tribler_started.emit(event_dict["version"])
             # if public key format will be changed, don't forget to change it
             # at the core side as well
--- a/src/tribler-gui/tribler_gui/event_request_manager.py
+++ b/src/tribler-gui/tribler_gui/event_request_manager.py
@@ -3,18 +3,19 @@ import logging
 import time
 
 from PyQt5.QtCore import QTimer, QUrl, pyqtSignal
-from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkRequest
+from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkReply, QNetworkRequest
 
 from tribler_common.reported_error import ReportedError
 from tribler_common.sentry_reporter.sentry_reporter import SentryReporter
 from tribler_common.simpledefs import NTFY
 
-from tribler_gui.exceptions import CoreConnectTimeoutError
+from tribler_gui.exceptions import CoreConnectTimeoutError, CoreConnectionError
 from tribler_gui.utilities import connect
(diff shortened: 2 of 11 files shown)


ichorid:
retest this please

ichorid:
@drew2a , I removed Sentry log skipping, I guess it was too broad. Was it right?

drew2a:
@ichorid not sure. 

The line that you've removed solved the problem (https://github.com/Tribler/tribler/commit/69125a7ede3c1b4950400e295ada9f987581ac75) of error spam that occurs in the case of (I guess) lost connection with the core (ref: https://github.com/Tribler/tribler/pull/5821).

So from my point of view, this line should not be removed.

src/tribler-gui/tribler_gui/event_request_manager.py
```diff
@@ -71,7 +73,6 @@ def events_start_received(self, event_dict):
     def on_error(self, error, reschedule_on_err):
         self._logger.info(f"Got Tribler core error: {error}")
 
-        SentryReporter.ignore_logger(self._logger.name)
```
drew2a, line 74: **Issue:** without this line we will have unuseful garbage in the logs, see https://github.com/Tribler/tribler/pull/5821

drew2a (changes requested):

src/tribler-gui/tribler_gui/event_request_manager.py
ichorid: Addressed by logging only every 10th connection attempt.

ichorid:
retest this please

ichorid:
retest this please

src/tribler-core/tribler_core/components/restapi/restapi_component.py
```diff
@@ -126,6 +126,9 @@ def report_callback(reported_error: ReportedError):
             self._events_endpoint.on_tribler_exception(reported_error)
 
         self._core_exception_handler.report_callback = report_callback
+        # Reraise the unreported error, if there is one
+        if self._core_exception_handler.unreported_error:
+            report_callback(self._core_exception_handler.unreported_error)
```
drew2a, line 131: **Note** it is better to set `unreported_error` to None to indicate that the error has been reported.

```suggestion
            report_callback(self._core_exception_handler.unreported_error)
            self._core_exception_handler.unreported_error = None
```

src/tribler-gui/tribler_gui/event_request_manager.py
```diff
@@ -69,20 +75,23 @@ def events_start_received(self, event_dict):
                 SentryReporter.set_user(public_key.encode('utf-8'))
 
     def on_error(self, error, reschedule_on_err):
-        self._logger.info(f"Got Tribler core error: {error}")
+        if error == QNetworkReply.ConnectionRefusedError:
+            if self.remaining_connection_attempts % 10 == 0:
+                self._logger.info("Tribler Core refused connection, retrying...")
```
drew2a, line 80: **Issue** it is unclear, why should we log each 10th attempt?
From my POV  it is better either to remove logging (line 80) or to replace it by:

```suggestion
            self._logger.debug("Tribler Core refused connection, retrying...")
```

src/tribler-gui/tribler_gui/event_request_manager.py
```diff
@@ -118,14 +127,13 @@ def on_finished(self):
...
-        self.connect_timer = QTimer()
-        self.connect_timer.setSingleShot(True)
-        self.connect_timer.timeout.connect(self.connect)
-        self.connect_timer.start(500)
+        self.connect_timer.start(RECONNECT_INTERVAL_MS)
 
     def connect(self, reschedule_on_err=True):
-        self._logger.info("Will connect to events endpoint")
+        if self.remaining_connection_attempts % 10 == 0:
+            self._logger.info("Will connect to events endpoint")
```
drew2a, line 134: **Issue** same here. It is unclear why should we log each 10th attempt.

drew2a (changes requested):

src/tribler-gui/tribler_gui/event_request_manager.py
ichorid: If I understood you correctly before, your original concern was about the logger spamming the breadcrumbs list. Well, logging each 10th attempt solves that problem. Seeing the connection attempts at the "Info" level is really useful during development.

 I'm no expert with Sentry internals, but if you think changing the logging level to Debug mode will solve the breadcrumbs spam problem, so be it.

drew2a: Breadcrumbs are configurable. Now we use these settings:
https://github.com/Tribler/tribler/blob/151e7ab936187717786b05736cd0dcf4b455340a/src/tribler-common/tribler_common/sentry_reporter/sentry_reporter.py#L114-L117

ichorid: BTW, wouldn't enabling Sentry to capture Debug-level logs enhance our ability to debug problems? Or that will produce too much output?

ichorid:
retest this please

drew2a (approved):

src/tribler-gui/tribler_gui/event_request_manager.py
drew2a: I'm not sure, but I guess so :)

ichorid:
retest this please

PR #6600, opened 2021-12-01
Improved GUI dialog logic

### LINKED ISSUE #6597: Make sure GUI dialogs are managed by the `MainWindow` instance.
Our usage of dialogs in the GUI code is inconsistent. Specifically, we are creating dialog objects in different widgets classes, which can lead to race conditions and other hidden errors. We suspect that this might be a reason for the (recent) instability of the GUI tests.

I propose to manage all dialogs centrally, i.e., by a single manager. This manager keeps track of the open dialogs and allows us, amongst other features, to close all open dialogs when shutting down. We can also improve the GUI tests by implementing a `wait_for_dialog` method and remove a few hard-coded timeouts in the process.

devos50:
This PR improves how dialogs in the GUI are managed. Our logic to create, close and hide dialogs is quite inconsistent across the code base, leading to errors and hard-coded timeouts, e.g., in the GUI tests. Specifically, this PR makes the following changes:
- Introduces the static `DialogManager` class. When a dialog (a subclass of `DialogContainer`) is created, the dialog is automatically added to the `DialogManager`, and removed when the `close_dialog` method is called. This central approach removes the need to bind dialog objects to `QWidget` and allows us to close all open dialogs when closing Tribler (with the exception of the feedback dialog which is not a subclass of `DialogContainer`).
- Removed the binding of various widgets to `QWidget` classes.
- Improved the GUI tests by implementing a `wait_for_dialog` method that waits for the appearance/disappearance of a dialog with a specific class type. This removes the need for various `QWait` calls in the GUI tests and should make the GUI tests more robust.
- Extended the GUI tests with several tests that test the various dialogs we have in Tribler.

Feedback is welcome!

Fixes #6597

### DIFF
--- a/src/tribler-gui/tribler_gui/dialog_manager.py
+++ b/src/tribler-gui/tribler_gui/dialog_manager.py
@@ -6,14 +6,14 @@ class DialogManager:
     This class manages all dialogs that are created in the GUI.
     The only exception is the feedback dialog that is raised when an error occurs in the GUI.
     """
-    dialogs: List["DialogContainer"] = []  # Stack of open dialogs - recent dialogs are at the end of this list
+    dialogs: List["DialogContainer"] = []  # noqa: F821
 
     @staticmethod
-    def add_dialog(dialog: "DialogContainer") -> None:
+    def add_dialog(dialog: "DialogContainer") -> None:  # noqa: F821
         DialogManager.dialogs.append(dialog)
 
     @staticmethod
-    def remove_dialog(dialog: "DialogContainer") -> None:
+    def remove_dialog(dialog: "DialogContainer") -> None:  # noqa: F821
         if dialog in DialogManager.dialogs:
             DialogManager.dialogs.remove(dialog)
 
--- a/src/tribler-gui/tribler_gui/dialog_manager.py
+++ b/src/tribler-gui/tribler_gui/dialog_manager.py
@@ -1,0 +1,36 @@
+from typing import List, Optional, Type
+
+
+class DialogManager:
+    """
+    This class manages all dialogs that are created in the GUI.
+    The only exception is the feedback dialog that is raised when an error occurs in the GUI.
+    """
+    dialogs: List["DialogContainer"] = []  # Stack of open dialogs - recent dialogs are at the end of this list
+
+    @staticmethod
+    def add_dialog(dialog: "DialogContainer") -> None:
+        DialogManager.dialogs.append(dialog)
+
+    @staticmethod
+    def remove_dialog(dialog: "DialogContainer") -> None:
+        if dialog in DialogManager.dialogs:
+            DialogManager.dialogs.remove(dialog)
+
+    @staticmethod
+    def get_dialogs(diag_cls: Optional[Type]):
+        """
+        Return either all dialogs or dialogs with a specific type.
+        """
+        if not diag_cls:
+            return DialogManager.dialogs
+        return [dialog for dialog in DialogManager.dialogs if isinstance(dialog, diag_cls)]
+
+    @staticmethod
+    def close_all_dialogs(diag_cls: Optional[Type] = None) -> None:
+        """
+        Close all open dialogs or dialogs with a specific type.
+        """
+        to_close = DialogManager.get_dialogs(diag_cls) if diag_cls else DialogManager.dialogs
+        for dialog in to_close:
+            dialog.close_dialog()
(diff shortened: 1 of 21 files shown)


xoriole:
retest this please

xoriole:
retest this please

src/tribler-gui/tribler_gui/dialog_manager.py
```diff
@@ -0,0 +1,35 @@
+from typing import List, Optional, Type
+
+
+class DialogManager:
```
ichorid, line 4: I'd suggest doing it the way we did with TriblerRequestManager - create TriblerDialog class that will automatically register themselves with the singleton DialogManager.

src/tribler-gui/tribler_gui/dialogs/confirmationdialog.py
```diff
@@ -10,7 +10,7 @@
 
 
 class ConfirmationDialog(DialogContainer):
-    button_clicked = pyqtSignal(int)
+    button_clicked = pyqtSignal(DialogContainer, int)
```
ichorid, line 13: Do we really have to pass this `DialogContainer` thing everywhere?

src/tribler-gui/tribler_gui/dialog_manager.py
devos50: I've renamed `DialogContainer` to `TriblerDialog` (I'm not sure why it was named `DialogContainer` in the first place though). When creating a `TriblerDialog`, the constructor will automatically add this class to the `DialogManager`. When closing the dialog, it will unregister from the `DialogManager`. 👍

src/tribler-gui/tribler_gui/dialogs/confirmationdialog.py
devos50: Now that we removed the assignment of dialogs to object attributes, the callback sometimes needs the dialog object to complete its jobs. For example, a callback might require some of the written text in the dialog and check it before the dialog can be closed (this happens with the `AddTagsDialog`). Passing the dialog to the signal callback is an adequate approach to achieve this (but maybe I'm missing a much simpler alternative?).

ichorid: In general, functions/methods/procedures should avoid _extracting_ info from arguments. Otherwise, it is unclear what _exact_ part of the info the function expects from the container argument passed to it. If a child dialog class extends the DialogContainer class, it should overload the signal definition with exact used arguments. E.g.

```python
# Parent class
class DialogContainer:
button_clicked = pyqtSignal(int)

# Child class
InputDialogContainer(DialogContainer):
button_clicked = pyqtSignal(int, str)
```

devos50:
retest this please

devos50:
retest this please

devos50:
retest this please

src/tribler-gui/tribler_gui/dialogs/confirmationdialog.py
devos50: In this situation, the callbacks invoked when clicking one of the buttons might require arbitrary data from the dialog (especially when we create some custom dialogs with additional attributes). For example, a dialog button callback could require access to the text in the dialog input field. Now that we do not bind the dialog to a class attribute anymore, passing the dialog as part of the pyqtSignal invoked when clicking one of the buttons in the dialog is an alternative approach to make the callback aware of the appropriate dialog.

An alternative solution I was thinking of is binding the dialog to a lambda function when connecting the signal to the callback. So for example, instead of writing:

```
connect(dialog.button_clicked, self.on_start_download_action)
```

We would write something like:

```
connect(dialog.button_clicked, lambda action: self.on_start_download_action(dialog, action))
```

However, I believe this is not very readable and I prefer passing the dialog as part of the Qt signal emit. If you have a cleaner solution, please let me know and I can consider rewriting the callbacks.

devos50:
GUI tests still seem to be unstable and are failing for various reasons 😕

ichorid:
> GUI tests still seem to be unstable and are failing for various reasons, even after removing some of the `qWait` calls confused

Don't worry, we will nail it down eventually :+1:  We should just remove all possible causes of instability, one by one. Tedious, but worth it :fist_raised:

src/tribler-gui/tribler_gui/dialogs/confirmationdialog.py
ichorid: >  the callbacks invoked when clicking one of the buttons might require arbitrary data from the dialog 
> to make the callback aware of the appropriate dialog.

I'd say the solution should be to create a subclass of the Dialog class for each kind of dialog with callbacks being methods of the said subclass. This way we have a very clear understanding of what the callback does in the context of the dialog.

devos50:
This PR has become stale and outdated, so I'm closing it for now.

I will try to refactor/finish this PR when I have time 👍

PR #6602, opened 2021-12-01
Make --core and --gui-test-mode separate flags with all combinations possible

### LINKED ISSUE #6601: The `--gui-test-mode` option does not work as expected
When `run_tribler.py` is started with the `--gui-test-mode` option, the core process fails to run with the error:
```python
  File "/Users/martijndevos/Documents/tribler/src/run_tribler.py", line 102, in <module>
    start_core.start_tribler_core(api_port, api_key, state_dir,
  File "/Users/martijndevos/Documents/tribler/src/tribler-core/tribler_core/start_core.py", line 126, in start_tribler_core
    config.api.http_port = int(api_port)
TypeError: int() argument must be a string, a bytes-like object, or a number, not 'NoneType'
```

The reason for the error is the incorrect handling of the `--gui-test-mode` option. When Tribler is started with this option, it immediately starts in Core mode, while it should be started in GUI mode.

To fix the problem, `--core` and `--gui-test-mode` should be separate flags (not a mutually exclusive group), and `--gui-test-mode` flag should be processed by both GUI and Core processes.

kozlovsky:
Fixes #6601

Both GUI and Core processes should receive `--gui-test-mode` flag to handle it correctly

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -23,18 +23,12 @@ logger = logging.getLogger(__name__)
 # pylint: disable=import-outside-toplevel, ungrouped-imports
 
 
-class RunMode(Enum):
-    CORE_ONLY = auto()
-    GUI_TEST_MODE = auto()
-
-
 class RunTriblerArgsParser(argparse.ArgumentParser):
     def __init__(self, *args, **kwargs):
         kwargs['description'] = 'Run Tribler BitTorrent client'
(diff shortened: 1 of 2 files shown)


kozlovsky:
retest this please

kozlovsky:
retest this please

devos50 (approved):

PR #6608, opened 2021-12-02
Fix api_key encoding/decoding

kozlovsky:
This PR fixes #6607 by fixing encoding/decoding of `api_key`.

PR #6613, opened 2021-12-02
Fix api_key conversion in DebugWindow

kozlovsky:
Previously GUI process stores `api_key` value mostly as bytes (and sometimes as str, which caused problems), and now it after #6608 and #6611, it is always stored as str. When displaying `api_key` in DebugWindow, the conversion from bytes to str is not needed anymore, as `api_key` is already str.


kozlovsky:
retest this please

ichorid (approved):

PR #6620, opened 2021-12-03
Remove spam in pytest output

kozlovsky:
Currently, pytest output has many unimportant warnings from third-party libraries. We can safely ignore all of them. Displaying all these warnings makes understanding pytest results more complex than necessary. It is essential that if we get some new and important warning one day, we can easily miss it in this wall of existing warnings.

In this PR, I update the `pytest.ini` config file to suppress all unimportant warnings and also I suppress warnings in code when necessary.

Also, I switched the default logging level in tests from DEBUG to INFO. In most cases, DEBUG output is not helpful, as it mainly consists of spam generated by internal PyQt modules. So, most of the time, it makes test results harder to understand. If necessary, we can always switch it back to DEBUG level later when analyzing some specific test failure.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,6 +2,12 @@
 timeout = 5
 log_level = DEBUG
 log_cli_format = %(asctime)s %(levelname)s %(message)s
+markers =
+    guitest:Tests for GUI. Skipped by default, use --guitests option to enable them
+    tunneltest:Slow tests for tunnels. Skipped by default, use --tunneltests option to enable them
+    enable_https:Use HTTPS instead of HTTP in marked tests
+    api_key:Used by rest_manager fixture to inject api_key value
+
 filterwarnings =
     ignore:Passing field metadata as a keyword arg is deprecated:DeprecationWarning:marshmallow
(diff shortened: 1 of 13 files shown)


kozlovsky:
retest this please

src/tribler-core/tribler_core/components/restapi/restapi_component.py
```diff
@@ -41,6 +42,9 @@
 from tribler_core.utilities.unicode import hexlify
 
 
+warnings.filterwarnings("ignore", "Multiple schemas resolved to the name SpeedTestRequest", module=".*marshmallow.*")
```
devos50, line 45: Suggestion: I think we can fix this one by updating the IPv8 pointer (see [this](https://github.com/Tribler/py-ipv8/commit/56c4b2dc37f7498754a69f577325f01ab0ad9d7a) commit).

xoriole (approved):
Nice! 👍

kozlovsky:
retest this please

kozlovsky:
retest this please

PR #6629, opened 2021-12-07
Fix for #6603

### LINKED ISSUE #6603: Our Windows builder frequently leaves directories behind after the Tribler tests have finished
Our tests are writing files to the `/tmp` directory and are not properly removed, at least on our Windows 64-bit machine that run the tests. See the following screenshot:

Schermafbeelding 2021-12-01 om 12 34 25

There seem to be two kind of directories that are left behind:

- Directories named with ~10 random characters: they contain the following items:
Schermafbeelding 2021-12-01 om 13 06 18
This makes me suspect that these directories are left behind by a test that does something with downloading (e.g., our lib torrent or channel tests).
- Directories that contain libtorrent metadata when doing DHT lookups. The directory is created [here](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/libtorrent/download_manager/download_manager.py#L143) and should be cleaned-up [here](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/libtorrent/download_manager/download_manager.py#L190).

drew2a:
This PR fixes #6603.

`metadata_tmpdir` now passes through the constructor of `DownloadManager` for each test.

### DIFF
--- a/src/tribler-core/tribler_core/components/libtorrent/download_manager/download_manager.py
+++ b/src/tribler-core/tribler_core/components/libtorrent/download_manager/download_manager.py
@@ -140,7 +140,7 @@ class DownloadManager(TaskManager):
             self.dht_health_manager = DHTHealthManager(dht_health_session)
 
         # Make temporary directory for metadata collecting through DHT
-        self.metadata_tmpdir = Path.mkdtemp(suffix='tribler_metainfo_tmpdir')
+        self.metadata_tmpdir = self.metadata_tmpdir or Path.mkdtemp(suffix='tribler_metainfo_tmpdir')
 
         # Register tasks
         self.register_task("process_alerts", self._task_process_alerts, interval=1)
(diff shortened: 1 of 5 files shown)


drew2a:
retest this please

drew2a:
retest this please

src/tribler-core/tribler_core/components/libtorrent/download_manager/download_manager.py
```diff
@@ -140,7 +140,7 @@ def initialize(self):
             self.dht_health_manager = DHTHealthManager(dht_health_session)
 
         # Make temporary directory for metadata collecting through DHT
-        self.metadata_tmpdir = Path.mkdtemp(suffix='tribler_metainfo_tmpdir')
+        self.metadata_tmpdir = self.metadata_tmpdir or Path.mkdtemp(suffix='tribler_metainfo_tmpdir')
```
drew2a, line 143: This is the main change.

kozlovsky (approved):

PR #6647, opened 2021-12-13
GUI settings should store api_key as bytes for compatibility with previous versions of Tribler

kozlovsky:
In #6608 and #6611 `api_key` value type was changed from bytes to str. As it turns out, it can break previous releases, as previous releases expect that the value of `api_key` in GUI settings is stored as bytes. So, if a user installs the next release of Tribler (7.12+), or run Tribler from the `main` branch, and then attempts to run a previous version of Tribler (for example, release 7.11), the Tribler will crash on reading `api_key` value from settings.

The PR fixes this problem: `api_key` is now stored in GUI settings as bytes. New function `set_api_key` should correctly handle all possible scenarios when `api_key` was set in bytes or str, or not set at all.

### DIFF
--- a/src/tribler-gui/tribler_gui/tests/test_util.py
+++ b/src/tribler-gui/tribler_gui/tests/test_util.py
@@ -1,6 +1,10 @@
+from unittest.mock import MagicMock
 from urllib.parse import unquote_plus
 
-from tribler_gui.utilities import compose_magnetlink, dict_item_is_any_of, quote_plus_unicode, unicode_quoter
+import pytest
+
+from tribler_gui.utilities import compose_magnetlink, create_api_key, dict_item_is_any_of, format_api_key, \
+    quote_plus_unicode, set_api_key, unicode_quoter
 
 
 def test_quoter_char():
@@ -124,3 +128,29 @@ def test_is_dict_has():
     assert dict_item_is_any_of(d, 'k', ['v'])
     assert dict_item_is_any_of(d, 'k', ['v', 'a'])
     assert dict_item_is_any_of(d, 'k', ['a', 'v'])
+
+
+def test_create_api_key():
+    x = create_api_key()
+    assert len(x) == 32 and bytes.fromhex(x).hex() == x
+
+
+def test_format_api_key():
+    api_key = "abcdef"
+    x = format_api_key(api_key)
+    assert x == "abcdef"
+
+    api_key = b"abcdef"
+    x = format_api_key(api_key)
+    assert x == "abcdef"
+
+    api_key = 123
+    match_str = r"^Got unexpected value type of api_key from gui settings \(should be str or bytes\): int$"
+    with pytest.raises(ValueError, match=match_str):
+        format_api_key(api_key)
+
+
+def test_set_api_key():
+    gui_settings = MagicMock()
+    set_api_key(gui_settings, "abcdef")
+    gui_settings.setValue.assert_called_once_with("api_key", b"abcdef")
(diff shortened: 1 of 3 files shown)


src/tribler-gui/tribler_gui/utilities.py
```diff
@@ -222,6 +222,57 @@ def get_gui_setting(gui_settings, value, default, is_bool=False):
...
+            raise ValueError(f'Got unexpected value type of api_key from gui settings '
+                             f'(should be str or bytes): {type(api_key).__name__}')
+
+    if should_save:
+        api_key_bytes = api_key.encode('ascii')
+        gui_settings.setValue("api_key", api_key_bytes)
+
+    return api_key
+
+
```
drew2a, line 275: **Issue:** the inner logic of set_api_key is overcomplicated.
I propose to simplify set_api_key inner logic by applying the following approach:
```suggestion
def create_api_key() -> str:
    return os.urandom(16).hex()


def format_api_key(api_key) -> str:
    if isinstance(api_key, bytes):
        # In Tribler application, api_key is stored as a string. But in gui_settings api_key is stored as bytes,
        # for compatibility with previous versions of Tribler. Otherwise, a user can get an error if he rolls back
        # to a previous version of Tribler, that cannot read str values of api_key from GUI settings.
        return api_key.decode('ascii')

    if isinstance(api_key, str):
        # If we read api_keys from gui_settings as str, we need to save it as bytes, to restore
        # settings compatibility with previous versions of Tribler
        return api_key

    raise ValueError(f'Got unexpected value type of api_key from gui settings '
                     f'(should be str or bytes): {type(api_key).__name__}')


def set_api_key(gui_settings: QSettings, api_key: str):
    api_key_bytes = api_key.encode('ascii')
    gui_settings.setValue("api_key", api_key_bytes)
```

drew2a (changes requested):

src/tribler-gui/tribler_gui/tribler_window.py
```diff
@@ -147,11 +148,7 @@ def __init__(
         self.root_state_dir = Path(root_state_dir)
         self.gui_settings = settings
         api_port = api_port or int(get_gui_setting(self.gui_settings, "api_port", DEFAULT_API_PORT))
-        api_key = api_key or get_gui_setting(self.gui_settings, "api_key", hexlify(os.urandom(16)))
-        if isinstance(api_key, bytes):
-            # in QSettings, api_key can be stored as bytes, then we decode it to str
-            api_key = api_key.decode('ascii')
-        self.gui_settings.setValue("api_key", api_key)
+        api_key = set_api_key(self.gui_settings, api_key)
```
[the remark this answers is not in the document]
kozlovsky: I intended to not overwrite the `QSettings` api_key value with the same value, but if it is not necessary, then the code can be simplified indeed, thanks

src/tribler-gui/tribler_gui/utilities.py
kozlovsky: 👍

src/tribler-gui/tribler_gui/tests/test_util.py
```diff
@@ -124,3 +128,29 @@ def test_is_dict_has():
...
     assert dict_item_is_any_of(d, 'k', ['v', 'a'])
     assert dict_item_is_any_of(d, 'k', ['a', 'v'])
+
+
+def test_create_api_key():
+    x = create_api_key()
+    assert len(x) == 32 and bytes.fromhex(x).hex() == x
+
+
+def test_format_api_key():
```
drew2a, line 138: 👍

drew2a (approved):

PR #6654, opened 2021-12-14
[WIP] User identicons

### LINKED ISSUE #5154: Identicons for users and channels' authors
It will be much easier to identify channels and your own Tribler instance by using [identicons](https://drhus.github.io/awesome-identicons/).

If we opt to use identicons based on public key, which scheme should we choose? (see the link above for examples)
image

### LINKED ISSUE #6450: Hard to navigate through search results
image

The search results now are not very usable. 
1) They lack context. Where is the item from, descriptions, images. Anything that can help to distinguish. 
2) The items in the list are to close to each other. As a result for me this reads like one giant blob of text.

Suggestions:
- On click - expand the item, give context, then allow to download.  
- Consider different way to organise and present the data. For example, grid like structure. Items grouped by Category, Channels, or Tags.

ichorid:
fixes #5154 

изображение


devos50:
I had a closer look at this proposed feature and played a bit around in GUI test mode to get a better feeling for it.

First, I was wondering about the problem it would solve. Since #5154 has been linked, it would allow for quicker identification of the channel each torrent originates from.

I am mostly familiar with identifications from GitHub and they can help to quicker identify users. In the context of Tribler, however, I’m not sure if they serve their purpose, for the following reasons:
- When presented in a linear list, I feel overwhelmed by all these identicons and I mostly ignore them:

Schermafbeelding 2022-01-18 om 10 28 39

Then, when going to the list of torrents, I see all these identicons. I would think that each identicon would uniquely identify the torrent but that does not seem to be the case:

Schermafbeelding 2022-01-18 om 10 03 40

Multiple torrents are sharing the same identicon, which makes sense if you know that the identicon links to the channel, but that’s not clear for me at a first glance. There is a tooltip with the channel name, but for that to be effective, I think the identicon should also be clickable. Overall, I’m not convinced that they effectively solve the problem in their current form.
- We should be very careful with introducing new elements/columns in the tables that showcase content. Our tables already contain quite a few elements that might not be clear to the user. One can make the same argument for the recently added tags. However, that is a GUI element that has been heavily discussed and we tried to made them as non-intrusive as possible. Though I have to say that these identicons look ok in the ‘discovered channels’ page (they break the bland style), I think there are preferable alternatives.

Some alternatives I can think of:
- We used to have an additional part of the screen to show detailled torrent information (e.g., its trackers). This would have been the perfect place for that kind of information, but that part got removed (I don’t really remember why we did remove that though, there was probably a good reason for it).
- Maybe a cleaner solution is to show the channel name just under the title of a torrent, just like YouTube is doing. This text would then also be clickable and it would bring you to the channel the torrent is part of.

Note that my findings above are obviously subjective and other people might have a completely different experience when interacting with particular GUI elements. In general, I think it’s the best to not deviate too much from the user interfaces of established applications, and take inspiration from them. Are there any major applications that are embedding identicons in a scrollable list?

ichorid:
Identicons show channels authors. The purpose of this PR is to allow users to visually identify _trusted_ channel authors.

Regarding the source channel problem, there is another PR that solves that with [clickable tooltips](https://github.com/Tribler/tribler/pull/6713).

devos50:
We have discussed this functionality in the weekly developer meetings. While we appreciate the efforts and like identicons, the general consensus is that we do not have time to further pursue this approach at the moment. Instead, we would like to look into these kind of GUI improvements the moment when we are going to make changes to the way we organise and present content.

PR #6668, opened 2021-12-20
Set the correct range for Y-axis of the token balance graph

### LINKED ISSUE #6667: Incorrect token balance graph is displayed
Currently, my token balance graph looks the following way:

The actual balance is -3.9GB and changing, but on the graph, it is displayed as a line at -119MB. Manual scaling of the graph does not fix the problem.

This can be fixed by setting the proper range for the Y-axis of the graph. After the proper setting of the range, the same graph will look the following way:

Now you can see the correct labels on the Y-axis. The fix will be provided in a subsequent PR

kozlovsky:
This PR fixes #6667. Now `TrustPage.plot_absolute_values` dynamically sets the range for Y-axis depending on the received data.

After the fix, the graph displays data correctly, and Y-axis is adequately scaled.


kozlovsky:
retest this please

xoriole (approved):
Nice 👍

Issue #6686, opened 2021-12-25
Progress bars for file in the Downloads Details tabs freeze Tribler


ichorid:
A forum user [reported](https://forum.tribler.org/t/7-11-0-rc2-is-out/6946/4) that on slower machines drawing progress bars for a torrent of a dozen files makes Tribler unresponsive. That is confirmed on my developer machine with some testing. The problem is, we're using `QTreeWidget`, which becomes super-slow when replacing any column with a widget (such as a progress bar). To solve the problem, we must refactor the Details tab to use QT's Model-View-Delegate pattern.

drew2a:
I'm wondering whether #7705 fixed this issue, @qstokkink?

PR #6687, opened 2021-12-25
Disable progressbars for individual files

### LINKED ISSUE #6688: Tribler does not start - "TypeError: arguments did not match any overloaded call: QPoint(): too many arguments"
I am an Arch Linux user and did not use Tribler for some week and after that the app won't start with the following error in the log:
```
[PID:70515] 2021-12-26 17:45:42,157 - INFO - TriblerApplication(23) - Start Tribler application. Win id: "triblerapp". Sys argv: "['/opt/Tribler/run_tribler.py']"
[PID:70515] 2021-12-26 17:45:42,349 - INFO - TriblerApplication(51) - No running instances (socket error: 2)
[PID:70515] 2021-12-26 17:45:42,538 - INFO - tribler_gui(454) - Available Tribler translations {'Russian': 'ru_RU', 'Portuguese': 'pt_BR', 'Chinese': 'zh_CN', 'English': 'en_US'}
[PID:70515] 2021-12-26 17:45:42,539 - INFO - tribler_gui(455) - System language: ['hu-HU'], Tribler language: ['hu', 'hu-HU', 'hu-Latn-HU']
[PID:70515] 2021-12-26 17:45:42,539 - INFO - __main__(208) - Start Tribler Window
[PID:70515] 2021-12-26 17:45:42,572 - INFO - NetworkUtils(30) - Looking for first free port in range [52194..52294]
[PID:70515] 2021-12-26 17:45:42,572 - INFO - NetworkUtils(34) - 52194 is free
[PID:70515] 2021-12-26 17:45:43,984 - INFO - TriblerGUI(147) - Will connect to events endpoint
[PID:70515] 2021-12-26 17:45:43,989 - ERROR <error_handler:37> ErrorHandler.gui_error(): Traceback (most recent call last):
  File "/opt/Tribler/run_tribler.py", line 209, in <module>
    window = TriblerWindow(settings)
  File "/opt/Tribler/tribler-gui/tribler_gui/tribler_window.py", line 275, in __init__
    pos = self.gui_settings.value("pos", QPoint(center.x() - self.width() * 0.5, center.y() - self.height() * 0.5))
TypeError: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'

[PID:70515] 2021-12-26 17:45:43,991 - INFO - SentryReporter(286) - Event from exception: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'
[PID:70515] 2021-12-26 17:45:44,002 - INFO - SentryReporter(358) - Before send strategy: SentryStrategy.SEND_SUPPRESSED
Sentry is attempting to send 1 pending error messages
```
I am the Tribler package maintainer for the Arch Linux/AUR and at least one other user also reported the exact same issue with the same error message, so I thought this is an Arch-specific problem. I thought one of the Qt or Python update is the cause of this behavior, but I could not triangulate which one. I tried to downgrade several suspicious package which was upgraded in the meantime to pinpoint the issue, but no luck.
Beside version 7.10 I also tried 7.11-RC2, but it has the same issue.
I am pleased to provide more information If necessary to solve this problem.

ichorid:
related to #6686 

The only reasonable solution is to disable the progress bars completely

Also, fixes #6688

### DIFF
--- a/src/tribler-gui/tribler_gui/widgets/torrentfiletreewidget.py
+++ b/src/tribler-gui/tribler_gui/widgets/torrentfiletreewidget.py
@@ -107,12 +107,12 @@ class DownloadFileTreeWidgetItem(QTreeWidgetItem):
                 self.progress_bytes = self.progress_bytes - old_bytes + new_bytes
 
         if child_changed or force_update:
-            if self.file_size and self.progress_bytes:
+            if self.progress_bytes is not None and self.file_size:
                 self.file_progress = self.progress_bytes / self.file_size
                 self.setText(PROGRESS_COL, f"{self.file_progress:.1%}")
 
         # ACHTUNG! This can be _very_ slow for torrents with lots of files, hence disabled by default
-        # To draw progress bars with acceptable performance we'd have to implement use QT's MVC stuff
+        # To draw progress bars with acceptable performance we'd have to use QT's MVC stuff
         if draw_progress_bars:
             bar_container, progress_bar = create_progress_bar_widget()
             progress_bar.setValue(int(self.file_progress * 100))
--- a/src/tribler-gui/tribler_gui/dialogs/dialogcontainer.py
+++ b/src/tribler-gui/tribler_gui/dialogs/dialogcontainer.py
@@ -49,8 +49,8 @@ class DialogContainer(AddBreadcrumbOnShowMixin, QWidget):
             self.dialog_widget.setFixedWidth(self.width() - self.left_right_margin)
             self.dialog_widget.move(
                 QPoint(
-                    self.geometry().center().x() - self.dialog_widget.geometry().width() / 2,
-                    self.geometry().center().y() - self.dialog_widget.geometry().height() / 2,
+                    self.geometry().center().x() - self.dialog_widget.geometry().width() // 2,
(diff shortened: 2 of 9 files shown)


src/tribler-gui/tribler_gui/widgets/torrentfiletreewidget.py
```diff
@@ -107,12 +107,12 @@ def subtree_progress_update(self, updates, force_update=False, draw_progress_bar
                 self.progress_bytes = self.progress_bytes - old_bytes + new_bytes
 
         if child_changed or force_update:
-            if self.file_size and self.progress_bytes:
+            if self.file_size is not None and self.progress_bytes is not None:
```
kozlovsky, line 110: **question:** Is it safe regarding to possible `ZeroDivisionError`? Can `self.file_size` be zero here?

ichorid: Good point, it is safer to leave the check for non-zero value just in case :+1:

kozlovsky (approved):

PR #6702, opened 2022-01-03
Make notifier more generic

### LINKED ISSUE #5652: [dev] Make `Notifier` more general
Make `Notifier` more general

There are three points, that we can improve in the current Implementation of [Notifier](https://github.com/Tribler/tribler/blob/devel/src/tribler-core/tribler_core/notifier.py):

1. `Notifier` doesn't support `async/await`: 
```
```
But it seems wrong because we developing an Application that supports `async/await`.

2. `Notifier` prohibits us from introducing new topics without changing `NTFY`:
```
    def add_observer(self, subject, callback):
        assert isinstance(subject, NTFY)
```

What makes the `NTFY` kinda global event registry. 
In the future, this will lead to permanent merge conflicts, because every developer, that uses the `Notifier` should change `NTFY`.

It also makes it difficult to find the topics which belong to, say, a particular Community.

3. `Notifier.notify` method uses `*args` instead `**kwargs` which is more vulnerable to making `TypeError` just by misprint:
```
```

Example:
```
notifier = Notifier()
topic = ''

def _f(one, two, three):
    pass

notifier.add_observer('', _f)
notifier.notify(topic, 'one', 'two', 'three')  # ok
notifier.notify(topic, 'one', 'two', 'three')  # ok
notifier.notify(topic, 'one', 'two', '', 'three')  # TypeError: _f() takes 3 positional arguments but 4 were given

def _f1(one, two, three, four):
    pass

notifier.add_observer('', _f1)
notifier.notify(topic, 'one', 'two', 'three')  # TypeError: _f1() missing 1 required positional argument: 'four'
```

## Resuming

It would be nice if `Notifier` will be closer to the following:

```
class Notifier(object):
    def __init__(self):
        self.observers = defaultdict(lambda: [])

    def add_observer(self, subject, callback):
        self.observers[subject].append(callback)

    async def notify(self, subject, **kwargs):
        await asyncio.gather(*[callback(**kwargs) for callback in self.observers[subject]])
```

drew2a:
This PR closes #5652

### DIFF
--- a/src/tribler-core/tribler_core/notifier.py
+++ b/src/tribler-core/tribler_core/notifier.py
@@ -1,53 +1,44 @@
-"""
-Notifier.
-
-Author(s): Vadim Bulavintsev
-"""
 import logging
 from asyncio import get_event_loop
-
-from tribler_common.simpledefs import NTFY
+from collections import defaultdict
+from typing import Callable, Dict
 
 
 class Notifier:
-
     def __init__(self):
-        self._logger = logging.getLogger(self.__class__.__name__)
-        self.observers = {}
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.observers: Dict[str, set] = defaultdict(set)
+        # @ichorid:
         # We have to note the event loop reference, because when we call "notify" from an external thread,
         # we don't know anything about the existence of the event loop, and get_event_loop() can't find
         # the original event loop from an external thread.
-        self._loop = get_event_loop()
         # We remember the event loop from the thread that runs the Notifier
         # to be able to schedule notifications from external threads
+        self._loop = get_event_loop()
 
-    def add_observer(self, subject, callback):
-        assert isinstance(subject, NTFY)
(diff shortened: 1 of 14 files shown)


ichorid (dismissed):

src/tribler-core/tribler_core/notifier.py
```diff
@@ -1,53 +1,44 @@
...
+    def add_observer(self, topic: str, callback: Callable):
+        self.logger.debug(f"Add observer topic {topic}")
+        self.observers[topic].add(callback)
+
+    def remove_observer(self, topic: str, callback: Callable):
+        self.logger.debug(f"Remove observer topic {topic}")
+        self.observers[topic].discard(callback)
+
+    def notify(self, topic: str, *args, **kwargs):
+        def _notify(_topic, _kwargs, *_args):
```
drew2a, line 28: Remove args and kwargs

drew2a:
retest this please

kozlovsky (approved):

PR #6703, opened 2022-01-03
Fix missing XML declaration

overflw:
In `update_version_from_git.py` the current version is written into `metainfo.xml`. Unfortunately, the XML declaration is not generated after inserting the current version.


qstokkink:
ok to test

xoriole (approved):

devos50 (approved):

PR #6713, opened 2022-01-07
Add interactive tooltips

### LINKED ISSUE #6450: Hard to navigate through search results

ichorid:
Adds tooltips with clickable links to source channels to Popular and Search results pages.
partially fixes #6450 

tool

### DIFF
--- a/src/tribler-core/tribler_core/components/metadata_store/db/orm_bindings/channel_node.py
+++ b/src/tribler-core/tribler_core/components/metadata_store/db/orm_bindings/channel_node.py
@@ -275,6 +275,7 @@ def define_binding(db, logger=None, key=None):  # pylint: disable=R0915
 
             return self
 
+        @db_session
         def get_parent_nodes(self):
             full_path = {self: True}
             node = self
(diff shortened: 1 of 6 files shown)


devos50:
We have discussed this functionality in the weekly developer meetings. Similar to #6654, the consensus is that we do not have time to further pursue this approach at the moment. Instead, we would like to look into these kind of GUI improvements the moment when we are going to make changes to the way we organise and present content.

ichorid:
I've spent **a week** solving this. This PR is completely working, with no crashes. The only thing left to do is add GUI test. There is literally nothing to "pursue", just add a test. It does not even change the GUI appearance, so there is nothing to discuss in that regard either.

It makes me very sad to see how the team discards an important UX improvement that is basically ready and only requires a little bit of developer time to finish.

PR #6716, opened 2022-01-10
link detection fix

Dmole:
link detection fix

### DIFF
--- a/src/tribler.sh
+++ b/src/tribler.sh
@@ -2,8 +2,15 @@
 # Run Tribler from source tree
 
 script_path() {
-  DIR="${1%/*}"
-  (cd "$DIR" && echo "$(pwd -P)")
+  if readlink --help 2>/dev/null | grep -q canonicalize ; then
+    # Linux can support following soft links
+    FILE="$(readlink -f "$1")"
+    echo "${FILE%/*}"
+  else
+    # MacOS
+    DIR="${1%/*}"
+    (cd "$DIR" && pwd -P)
+  fi
 }
 
 UNAME="$(uname -s)"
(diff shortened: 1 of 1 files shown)


xoriole:
ok to test

src/tribler.sh
```diff
@@ -2,8 +2,8 @@
 # Run Tribler from source tree
 
 script_path() {
-  DIR="${1%/*}"
-  (cd "$DIR" && echo "$(pwd -P)")
+  FILE="$(readlink -f "$1")"
```
xoriole, line 5: While using `readlink -f` makes it possible to run tribler.sh from any path, the command is not available on macOS which is why we avoided using it earlier.

xoriole (changes requested):
Thank you for the PR. It works nicely on Linux but not on macOS so can not accept as it is, therefore requesting changes

src/tribler.sh
Dmole: readlink must have been added for me by brew, I'll make it optional in the PR.

drew2a:
I've converted this PR to "draft" as it is still in WIP state.

Dmole:
> I've converted this PR to "draft" as it is still in WIP state.

Let me know what needs to change to get it merged.

drew2a:
> Let me know what needs to change to get it merged.

As far as I know -- it is necessary to address all @xoriole requests and re-request his review 👍
image

drew2a:
@Dmole please, update your branch to the latest main. The current test failures are caused by changes from https://github.com/Tribler/tribler/pull/6799.

drew2a:
Give me 1 hour and I'll fix the test run (no action needed from your side).

drew2a:
Ok, I've fixed test runs. Now tests from the latest main should work fine.

xoriole:
@Dmole I have tested the PR on both Linux and macOS, and it works. Please rebase the PR to get it merged.

xoriole (approved):

PR #6718, opened 2022-01-12
Tags auto generation

drew2a:
This PR is a part of #6214 and it introduces automatic tags extraction from torrent titles.

The following rules have been added:
1. Tags in square brackets: `title [tag1, tag2, tag3]`
2. Tags in parentheses: `title (tag1, tag2, tag3)`
3. A tag in an extension: `title.tag`

The tags could be extracted for all new torrents and they could be extracted for existing torrents in the background. 

The extraction procedure for all new tags is pretty straightforward, they are just processed at the moment when they adding to MDS.

The background extraction procedure is a bit more complicated. This extraction makes by processing a batch of items. Every 10 seconds a batch of 1000 items has been processed. In the case tag rules processor rich the upper bound of DB, it starts from the beginning, but with increased interval (20 seconds) and increased batch size (2000 items).

Why 1000 items and 10 seconds? It is not too heavy for CPU and with these values, 360k items will be processed within the hour.

To distinguish processed items, the `tag_version` column has been added to `TorrentMetadata`:

```python
    class TorrentMetadata(db.MetadataNode):
        ...
        tag_version = orm.Required(int, default=0)
```

Therefore it could be used in the future when we will process multiple incoming rules (we should just increase this value by 1 for each new rule set).


src/tribler-core/tribler_core/components/tag/rules/tag_rules.py
```diff
@@ -0,0 +1,42 @@
+import re
+from typing import AnyStr, Iterable, Optional, Pattern, Sequence
+
+from tribler_core.components.tag.community.tag_validator import is_valid_tag
+
+square_brackets = re.compile(r'\[([^\[\]]+)]')
+brackets = re.compile(r'\(([^()]+)\)')
```
kozlovsky, line 7: **nitpick:** should be called "parentheses" instead. Or better, something like "parentheses_re".

kozlovsky (changes requested):

src/tribler-core/tribler_core/components/tag/rules/tag_rules.py
drew2a: Fixed.

src/tribler-core/tribler_core/components/tag/db/tag_db.py
```diff
@@ -7,18 +7,25 @@
 from pony.utils import between
 
 from tribler_core.components.tag.community.tag_payload import TagOperation, TagOperationEnum
+from tribler_core.utilities.pony_utils import get_or_create
 from tribler_core.utilities.unicode import hexlify
 
+CLOCK_START_VALUE = 0
+
+# we picked `-1` as a value because it is allows manually created tags get a higher priority
+CLOCK_FOR_AUTOGENERATED_TAGS = CLOCK_START_VALUE - 1
```
kozlovsky, line 16: Probably this logic is unnecessary and can be removed, as we also have `is_auto_generated` field

kozlovsky (changes requested):

src/tribler-core/tribler_core/components/tag/db/tag_db.py
drew2a: Fixed.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+            operation = TagOperation(
+                infohash=infohash,
+                operation=TagOperationEnum.ADD,
+                clock=CLOCK_START_VALUE,
+                creator_public_key=PUBLIC_KEY_FOR_AUTO_GENERATED_TAGS,
+                tag=tag
+            )
+            # we want auto generated operation to act like a normal operation
+            # therefore we use 2 as a `counter_increment` to immediately pass
+            # SHOW_THRESHOLD
```
kozlovsky, line 96: It is better to move these lines inside the `add_auto_generated_tag_operation` method and pass `infohash` and `tag` as arguments. This way, the method will encapsulate the details of `TagOperation`.

kozlovsky (changes requested):

src/tribler-core/tribler_core/components/metadata_store/db/store.py
```diff
@@ -136,14 +136,15 @@
...
     def __init__(
-        self,
-        db_filename: Union[Path, type(MEMORY_DB)],
-        channels_dir,
-        my_key,
-        disable_sync=False,
-        notifier=None,
-        check_tables=True,
-        db_version: int = CURRENT_DB_VERSION,
+            self,
```
devos50, line 139: **Question:** My impression is that these changed lines should be indented with four spaces. Why are we adding an additional four spaces per line here?

src/tribler-core/tribler_core/components/metadata_store/db/store.py
```diff
@@ -190,7 +191,11 @@ def sqlite_disable_sync(_, connection):
 
         self.MetadataNode = metadata_node.define_binding(self._db)
         self.CollectionNode = collection_node.define_binding(self._db)
-        self.TorrentMetadata = torrent_metadata.define_binding(self._db)
+        self.TorrentMetadata = torrent_metadata.define_binding(
+            self._db,
+            notifier=notifier,
+            tag_version=tag_version
```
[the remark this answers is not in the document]
devos50: Same remark here 👍 . Maybe even drop the word `version` since, strictly speaking, it is indeed not a version.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+    CLOCK_START_VALUE, PUBLIC_KEY_FOR_AUTO_GENERATED_TAGS,
+    TagDatabase,
+)
+from tribler_core.components.tag.rules.tag_rules import extract_only_valid_tags
+from tribler_core.notifier import Notifier
+
+LAST_PROCESSED_TORRENT_ID = 'last_processed_torrent_id'
+
+
+class TagRulesProcessor(TaskManager):
```
devos50, line 22: **Suggestion:** It seems that we now apply rules when 1) tags are coming in from the network, and 2) passively in the background. We might consider also extracting tags for unprocessed torrents when the user requests torrents with the REST API (if it's not adding too much latency). This would ensure that there are tags visible for the content that users see. The next time the user requests the same torrents, the rules are already applied to the requested torrents.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+        for tag in tags:
+            operation = TagOperation(
+                infohash=infohash,
+                operation=TagOperationEnum.ADD,
+                clock=CLOCK_START_VALUE,
+                creator_public_key=PUBLIC_KEY_FOR_AUTO_GENERATED_TAGS,
+                tag=tag
+            )
+            # we want auto generated operation to act like a normal operation
+            # therefore we use 2 as a `counter_increment` to immediately pass
```
devos50, line 95: **Question:** Does this still apply after the merge of https://github.com/Tribler/tribler/pull/6742?

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+            # SHOW_THRESHOLD
+            self.db.add_auto_generated_tag_operation(operation=operation)
+
+    def _schedule_new_process_batch_round(self):
```
devos50, line 99: **Question/issue:** It seems that the tag processor is a background task that is continuously running and periodically checks if there are new torrents available in the database to process. I think, however, that we can consider the tag processor as a one-shot task for now, mostly since the set of rules is fixed, and because new incoming torrents will be processed immediately. When we add additional rules through a new release, we can increase the version number of the processor and run the processor again on the database.

When we are also going to include rules created by users themselves, then we indeed need a continuous process. But since the rules are static at the moment, I would opt for running the processor once and when it has processed all torrents in the database, its task is done.

If we want to keep it continuously running, then I agree with @kozlovsky 's comment regarding the increase of the interval/batch size.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+
+LAST_PROCESSED_TORRENT_ID = 'last_processed_torrent_id'
+
+
+class TagRulesProcessor(TaskManager):
+    # this value must be incremented in the case of new rules set has been applied
+    version: int = 1
+
+    def __init__(self, notifier: Notifier, db: TagDatabase, mds: MDS.MetadataStore,
+                 batch_size: int = 1000, interval: float = 10):
```
devos50, line 27: **Suggestion:** Extract these values to a global variable.

src/tribler-core/tribler_core/conftest.py
```diff
@@ -186,7 +186,7 @@ def metadata_store(tmp_path):
 
 @pytest.fixture
 def tags_db():
-    db = TagDatabase()
+    db = TagDatabase(create_tables=True)
```
devos50, line 189: **Question:** Do we really have to pass this keyword argument every time we create the database? Could we instead make a `create_tables` kwarg with default value `True` in the constructor of `TagDatabase`?

devos50 (changes requested):
Overall, great work 👍 . I left a few minor comments and some concerns regarding the tag processor.

devos50:
Could [this error](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_win64_pytest/3046/testReport/junit/src.tribler-core.tribler_core.components.key.tests/test_key_component/test_get_private_key_filename/) be related to this PR (Windows tests)?

```
src.tribler-core.tribler_core.components.key.tests.test_key_component.test_get_private_key_filename (from pytest)

pyfuncitem = <Function test_get_private_key_filename>

    def pytest_pyfunc_call(pyfuncitem):  # type: ignore[no-untyped-def]
        """Run coroutines in an event loop instead of a normal function call."""
        fast = pyfuncitem.config.getoption("--aiohttp-fast")
        if asyncio.iscoroutinefunction(pyfuncitem.function):
            existing_loop = pyfuncitem.funcargs.get(
                "proactor_loop"
            ) or pyfuncitem.funcargs.get("loop", None)
            with _runtime_warning_context():
                with _passthrough_loop_context(existing_loop, fast=fast) as _loop:
                    testargs = {
                        arg: pyfuncitem.funcargs[arg]
                        for arg in pyfuncitem._fixtureinfo.argnames
                    }
>                   _loop.run_until_complete(pyfuncitem.obj(**testargs))

c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\aiohttp\pytest_plugin.py:186: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
c:\users\tribler\appdata\local\programs\python\python38\lib\contextlib.py:120: in __exit__
    next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    @contextlib.contextmanager
    def _runtime_warning_context():  # type: ignore[no-untyped-def]
        """Context manager which checks for RuntimeWarnings.
    
        This exists specifically to
        avoid "coroutine 'X' was never awaited" warnings being missed.
    
        If RuntimeWarnings occur in the context a RuntimeError is raised.
        """
        with warnings.catch_warnings(record=True) as _warnings:
            yield
            rw = [
                "{w.filename}:{w.lineno}:{w.message}".format(w=w)
                for w in _warnings
                if w.category == RuntimeWarning
            ]
            if rw:
>               raise RuntimeError(
                    "{} Runtime Warning{},\n{}".format(
                        len(rw), "" if len(rw) == 1 else "s", "\n".join(rw)
                    )
                )
E               RuntimeError: 1 Runtime Warning,
E               c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\aiohttp\test_utils.py:560:coroutine 'interval_runner' was never awaited

c:\users\tribler\appdata\local\programs\python\python38\lib\site-packages\aiohttp\pytest_plugin.py:143: RuntimeError
```

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+            # we want auto generated operation to act like a normal operation
+            # therefore we use 2 as a `counter_increment` to immediately pass
+            # SHOW_THRESHOLD
+            self.db.add_auto_generated_tag_operation(operation=operation)
+
+    def _schedule_new_process_batch_round(self):
+        self.logger.info('All items in TorrentMetadata have been processed.')
+        self.mds.set_value(LAST_PROCESSED_TORRENT_ID, '0')
+        self.logger.info('Set last_processed_torrent_id to 0')
+        self.interval *= 2
```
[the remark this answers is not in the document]
drew2a: My intention was to decrease computations during the time. 
But you are right, we can remove this logic.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
```diff
@@ -0,0 +1,109 @@
...
+        added = 0
+        for torrent in batch:
+            added += self.process_torrent_title(torrent.infohash, torrent.title)
+            torrent.tag_version = self.version
+            processed += 1
+
+        self.logger.info(f'Processed: {processed} titles. Added {added} tags.')
+        max_row_id = self.mds.get_max_rowid()
+
+        is_beyond_the_boundary = end > max_row_id
```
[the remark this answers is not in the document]
drew2a: Your suggestion is pretty valid.

> On the other side, scanning of already processed torrents does not do much harm; it should be a pretty light query. So probably, it is not worth complicating the logic.

Agree. For now, it seems that keeping the current logic is better than writing a new, a bit more "optimized" one.

src/tribler-core/tribler_core/components/metadata_store/db/store.py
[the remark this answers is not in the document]
drew2a: Ok, gentlemens!

src/tribler-core/tribler_core/components/metadata_store/db/store.py
drew2a: It is [black formatting](https://black.readthedocs.io/en/stable/) 🤷

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: My previous version of this PR had this logic.

I abandoned this logic because assumed that (1) and (2) are good enough as (1) works for all new torrents and (2) processes the whole DB within the first day after the update.

So I decided not to add a third point of roughly the same logic that will be useful only within the first hours after the update.

But if @devos50 and @kozlovsky vote to return the logic back, I'm ok with it.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Yes, agree, this was a feature engeneering from my side. I'll remove this logic :)

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: No, I should change these comments. Thanks.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Implemented.

src/tribler-core/tribler_core/components/metadata_store/db/store.py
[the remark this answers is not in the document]
drew2a: Fixed.

src/tribler-core/tribler_core/components/tag/db/tag_db.py
```diff
@@ -7,18 +7,24 @@
 from pony.utils import between
 
 from tribler_core.components.tag.community.tag_payload import TagOperation, TagOperationEnum
+from tribler_core.utilities.pony_utils import get_or_create
 from tribler_core.utilities.unicode import hexlify
 
+CLOCK_START_VALUE = 0
+
+# we picked `-1` as a value because it is allows manually created tags get a higher priority
```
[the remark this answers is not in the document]
drew2a: Deleted.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Comments have been removed.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Done

src/tribler-core/tribler_core/conftest.py
drew2a: Yes, we could :) I'll implement it.

drew2a: Done.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Done.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Done.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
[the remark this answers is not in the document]
drew2a: Fixed.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
drew2a: Removed

drew2a:
@devos50 

> Could this error be related to this PR (Windows tests)?

I guess it could, but I'm not sure. 
Anyway, I've rewritten the execution of the TagProcessor's tests (the possible cause of the error), so will see.

src/tribler-core/tribler_core/components/tag/rules/tag_rules_processor.py
kozlovsky: I think it might be good to add tags when retrieving torrents via REST API, as it will demonstrate tags to the user immediately after he starts using the new version of Tribler.

devos50: As an alternative, we might show the user that tags are being processed in the background so the user is aware that something is happening, and more tags will be visible after the processor is finished (we did the same when updating from legacy channels). Still, I prefer for tags to work without waiting for the completion of the processor and it would improve the user experience, even if it is logic that is only useful a few hours after an upgrade.

drew2a: Done

src/tribler-core/tribler_core/components/metadata_store/restapi/metadata_endpoint_base.py
```diff
@@ -68,6 +71,20 @@ def sanitize_parameters(cls, parameters):
             sanitized['metadata_type'] = frozenset(mtypes)
         return sanitized
 
+    def process_regular_torrent(self, entry):
```
drew2a, line 74: @kozlovsky: rename to `extract_tags`

drew2a: done

kozlovsky (approved):

devos50 (approved):

devos50:
🤔 maybe wait with merging until the PR pipeline can complete?

kozlovsky:
retest this please

kozlovsky:
retest this please

drew2a:
retest this please

drew2a:
retest this please

PR #6724, opened 2022-01-14
Add session context manager

drew2a:
During the implementation of #6718 I came up with the idea of simplification component's tests.

Instead of writing tests like 

```python
@pytest.mark.asyncio
async def test_bandwidth_accounting_component(tribler_config):
    components = [KeyComponent(), Ipv8Component(), BandwidthAccountingComponent()]
    session = Session(tribler_config, components)
    with session:
        await session.start()

        comp = BandwidthAccountingComponent.instance()
        assert comp.started_event.is_set() and not comp.failed
        assert comp.community
        assert comp._ipv8_component

        await session.shutdown()
```

We could write:

```python
@pytest.mark.asyncio
async def test_bandwidth_accounting_component(tribler_config):
    components = [KeyComponent(), Ipv8Component(), BandwidthAccountingComponent()]
    with Session(tribler_config, components) as session:
        async with in_session(session):
            comp = BandwidthAccountingComponent.instance()
            assert comp.started_event.is_set() and not comp.failed
            assert comp.community
            assert comp._ipv8_component
```

it becomes possible by introducing `in_session` context manager:
```python
@asynccontextmanager
async def in_session(session):
    try:
        yield await session.start()
    finally:
        await session.shutdown()
```

### DIFF
--- a/src/tribler-core/run_bandwidth_crawler.py
+++ b/src/tribler-core/run_bandwidth_crawler.py
@@ -9,10 +9,11 @@ from pathlib import Path
 from ipv8.loader import IPv8CommunityLoader
 
 from tribler_common.simpledefs import STATEDIR_DB_DIR
-from tribler_core.config.tribler_config import TriblerConfig
+
 from tribler_core.components.bandwidth_accounting.db.database import BandwidthDatabase
-from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
 from tribler_core.components.bandwidth_accounting.settings import BandwidthAccountingSettings
+from tribler_core.config.tribler_config import TriblerConfig
+from tribler_core.modules.bandwidth_accounting.launcher import BandwidthCommunityLauncher
 from tribler_core.start_core import Session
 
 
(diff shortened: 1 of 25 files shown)


kozlovsky (changes requested):
I like the idea, and it nicely simplifies the code of tests. But the suggested API looks non-intuitive to me:

```python
with Session(tribler_config, components) as session:
    async with in_session(session):
        ...
```

It may be hard to guess what `async with in_session(session)` exactly means.

What about extending `Session.start` method to return a context manager? Then you can write the code in the following style:

```python
await with Session(tribler_config, components).start():
    ...
```

or, in seldom cases when we want to access a component instance before the session start:

```python
session = Session(tribler_config, components)
with session:
        a = ComponentA.instance()
        ...
        with session.start():
            ...
```

With this API, it should be intuitive that exiting from the context manager should close the started session in some way, similar to closing files after exiting from `with open(...)` context manager.

drew2a:
At the end of the collaboration with @kozlovsky the following changes have been implemented:

Session context manager becomes:
```python
@asynccontextmanager
async def session_manager(session: Session):
     with session:  # set the current session as a default session
        try:
            await session.start_components()  # on enter
            yield session
        finally:
            await session.shutdown()  # on leave
```

Tests become:

```python
    async with Session(tribler_config, components).start():
        comp = Component.instance()
        assert comp.started_event.is_set() and not comp.failed
        assert comp.version_check_manager```

drew2a:
retest this please

kozlovsky (approved):

PR #6730, opened 2022-01-18
Fix run_bandwidth_crawler script

kozlovsky:
In this PR I fix the script to run bandwidth crawler. It initializes BandwidthDatabase and BandwidthAccountingCommunity in a bit different way comparing to Tribler itself, so I added `crawler_mode` flag to `BandwidthAccountingComponent`. Another options was to add an option to BandwidthAccounting section of TriblerConfig, but the option does not look important enough to complicate the general config.


kozlovsky:
retest this please

drew2a (dismissed):

PR #6731, opened 2022-01-18
Fix unstable test_channels_peers_mapping_drop_excess_peers

kozlovsky:
On fast machine, the test `[test_channels_peers_mapping_drop_excess_peers](https://github.com/Tribler/tribler/blob/0ee609e142956b7d0c47054e1d9259450a1a20cc/src/tribler-core/tribler_core/components/gigachannel/community/tests/test_gigachannel_community.py#L231)` sometimes sets `peer.last_response` time to the same value for all peers, and fails after that. In this fix, I explicitly set a different time for each peer to avoid this type of failure.


devos50 (approved):
👍

PR #6736, opened 2022-01-20
Remove allow_exit_node_checkbox

drew2a:
Based on yesterday's discussion, this PR removes allow_exit_node_checkbox from UI as potentially dangerous for a user.

image


devos50:
If we are to remove this option from the GUI, should we provide some documentation on how to enable the exit node functionality if a user really wants this?

xoriole (approved):
I think we can remove this

drew2a:
@devos50 everything should be documented :)

PR #6738, opened 2022-01-20
Fix/6700

### LINKED ISSUE #6700: Error when adding a download that contains URL-escaped characters in the filename
When importing a download with escaped chapters in the filename, I see the following error:

Schermafbeelding 2022-01-02 om 21 47 11

I'm also unable to start the download of this torrent (it will raise another error that the file cannot be found).

**To Reproduce**
1. Create a torrent in the download directory named `bbb%20%21abc.torrent`.
2. Double-click the torrent to add it to Tribler.

drew2a:
This PR fixes #6700.

The problem was the interaction between GUI and Core by using REST. In particular, the problem was in the manual URI serializing by using `quote_plus_unicode` and manual URI deserializing by using `uri_to_path`.
I removed both serializing and deserializing procedures by replacing them with the native `aiohttp` mechanism (by using `params`). The current `quote/unquote` processing is performed by https://github.com/aio-libs/yarl/.

The funny fact is that they had a bug for `yarl` below `v1.6.3` and our manual `quote/unquote` mechanism (now removed) had kinda the same bug.

During investigation procedures, two functions have been extracted from `run_tribler.py`:
1. `run_core`
2. `run_gui`

During the PR the following bugs were fixed:
1. `import encodings.idna` removes by IDE as unused
2. `SentryStrategy.SEND_SUPPRESSED` has been set for the core from the GUI application

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -25,6 +25,7 @@ class RunTriblerArgsParser(argparse.ArgumentParser):
     def __init__(self, *args, **kwargs):
         kwargs['description'] = 'Run Tribler BitTorrent client'
         super().__init__(*args, **kwargs)
+        self.add_argument('torrent', help='torrent file to download', default='', nargs='?')
         self.add_argument('--core', action="store_true")
         self.add_argument('--gui-test-mode', action="store_true")
 
@@ -65,6 +66,7 @@ if __name__ == "__main__":
     init_sentry_reporter()
 
     parsed_args = RunTriblerArgsParser().parse_args()
+    logger.info(f'Run Tribler: {parsed_args}')
 
     # Get root state directory (e.g. from environment variable or from system default)
     from tribler_common.osutils import get_root_state_directory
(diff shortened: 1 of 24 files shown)


kozlovsky (changes requested):
Great finding, but it seems that the fix is incomplete. When I launch `run_tribler.py`  with a torrent path as an argument and the torrent name contains URI-encoded characters, Tribler still cannot download the torrent. For example, if I try to download the file `ubuntu%2021.10-desktop-amd64.iso.torrent` (note a `%20` sequence which represents a space character), I see the following error after Tribler is started:

image

As the bug is related to URI-escaped characters, I think it is better to add a test with a torrent file name that contains such characters.

drew2a:
retest this please

drew2a:
retest this please

drew2a:
retest this please

drew2a:
Our tests use `yarl 1.4.2` But we need at least `yarn 1.6.3` because of https://github.com/aio-libs/yarl/issues/517

image

drew2a:
retest this please

kozlovsky (approved):

kozlovsky (approved):

PR #6757, opened 2022-02-04
Use ipv8 as an external dependency

drew2a:
This PR is a part of #6505.

It changes the way the `ipv8` module is used in Tribler from "sub-module" to "external dependency".

I've requested a review from the involved participants. In case, the participants agree with the change, I will continue with the implementation.

The future steps will be:
1. Merging this branch into the main
2. Changing corresponding Jenkins jobs

**The question**: I'm not sure about the ordering of the future steps. Should I merge the branch first, and change the corresponding Jenkins jobs second, or the opposite?

### DIFF
--- a/doc/build-docs.sh
+++ b/doc/build-docs.sh
@@ -9,10 +9,6 @@ then
   echo "Please run this script from project root as:\n./doc/build-docs.sh"
 fi
 
-# Update git modules if necessary
-git submodule sync
-git submodule update --init --force --recursive src/pyipv8
-
 # all commands are executed from the doc directory
 cd doc
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "py-ipv8"]
-	path = src/pyipv8
-	url = https://github.com/Tribler/py-ipv8.git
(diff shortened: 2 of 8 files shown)


xoriole (dismissed):
Since IPv8 is quite stable, I'm in favor of using IPv8 as an external dependency

qstokkink (changes requested):
The `.gitmodules` file should also be removed, this will likely make the tests run.

That aside, **praise:** after 9 long years, finally no more submodules 🙏

qstokkink (dismissed):

doc/build-docs.sh
```diff
@@ -11,7 +11,6 @@ fi
 
 # Update git modules if necessary
 git submodule sync
```
kozlovsky, line 13: I think these lines can be removed as well?

kozlovsky (changes requested):

doc/build-docs.sh
devos50: Then we also don't need the `--recursive` flag anymore when cloning Tribler for experiments (although having it doesn't hurt) 👍

devos50 (approved):

doc/build-docs.sh
drew2a: Yes, thank you. Removed.

kozlovsky (approved):

PR #6773, opened 2022-02-16
Docker support

xoriole:
# Docker support

- Run docker as non-root user
- Map local `.Tribler` to docker container so that downloads runs inside the container.
- Add a Github action to build and publish docker container to docker registry whenever the `main` branch is updated. The built image is published to 
https://hub.docker.com/r/triblercore/triblercore

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,19 +6,23 @@ FROM triblercore/libtorrent:1.2.10-x
 RUN apt update && apt upgrade -y
 RUN apt install -y libsodium23 python3-pip git
 
+RUN useradd -ms /bin/bash user
+USER user
+WORKDIR /home/user
+
 # Then, install pip dependencies so that it can be cached and does not
 # need to be built every time the source code changes.
 # This reduces the docker build time.
-RUN mkdir /requirements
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -6,19 +6,23 @@ FROM triblercore/libtorrent:1.2.10-x
 RUN apt update && apt upgrade -y
 RUN apt install -y libsodium23 python3-pip git
 
+RUN useradd -ms /bin/bash user
```
kozlovsky, line 9: **question:** username "user" looks too general to me, can we use something more specific, like "dockeruser" or "docker"?

kozlovsky (changes requested):

Dockerfile
xoriole: It was intentional to make the user general and non-root as the user does not require any special rights

kozlovsky (approved):

drew2a:
This PR could lead to this issue: https://stackoverflow.com/questions/64059610/how-to-resolve-refusing-to-allow-an-oauth-app-to-create-or-update-workflow-on

PR #6809, opened 2022-03-17
Fix a crash in the Events tab of the Debug pane

### LINKED ISSUE #6808: [7.11] Crash in the Events tab of the Debug pane
**Describe the bug**
Currently the Events tab of the Debug window is not working

**To Reproduce**
1. Run Tribler
2. Click on "Debug" button
3. Open the "Events" tab
4. Tribler crashes with the exception:

```python
  File "C:\dev\tribler\src\tribler\gui\utilities.py", line 403, in trackback_wrapper
    raise exc from CreationTraceback(traceback_str)
  File "C:\dev\tribler\src\tribler\gui\utilities.py", line 400, in trackback_wrapper
    callback(*args, **kwargs)
  File "C:\dev\tribler\src\tribler\gui\debug_window.py", line 199, in tab_changed
    self.run_with_timer(self.load_events_tab)
  File "C:\dev\tribler\src\tribler\gui\debug_window.py", line 155, in run_with_timer
    call_fn()
  File "C:\dev\tribler\src\tribler\gui\debug_window.py", line 644, in load_events_tab
    item.setText(0, f"{event_dict['type']}")
KeyError: 'type'
```

The exception is triggered in GUI test `test_debug_pane`, but the test considered as passed.

The reason for the error is that the corresponding dict element is currently named "topic" and not "type".

kozlovsky:
This PR fixes #6808

### DIFF
--- a/src/tribler/gui/debug_window.py
+++ b/src/tribler/gui/debug_window.py
@@ -18,7 +18,6 @@ import libtorrent
 import psutil
 
 from tribler.core.utilities.utilities import has_bep33_support
-
 from tribler.gui.defs import DEBUG_PANE_REFRESH_TIMEOUT, GB, MB
 from tribler.gui.dialogs.confirmationdialog import ConfirmationDialog
 from tribler.gui.event_request_manager import received_events as tribler_received_events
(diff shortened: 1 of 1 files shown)


devos50 (approved):

src/tribler/gui/debug_window.py
```diff
@@ -641,7 +640,7 @@ def load_events_tab(self):
         for event_dict, timestamp in tribler_received_events:
             item = QTreeWidgetItem(self.window().events_tree_widget)
             item.setData(0, Qt.UserRole, event_dict)
-            item.setText(0, f"{event_dict['type']}")
+            item.setText(0, f"{event_dict['topic']}")
```
drew2a, line 643: **Nitpick, noncritical:** probably it is safer to use `get` here...

```suggestion
            item.setText(0, f"{event_dict.get('topic')}")
```

drew2a (changes requested):

src/tribler/gui/debug_window.py
kozlovsky: I agree

drew2a (approved):
👍

PR #6820, opened 2022-03-21
Handle possible duplicates in random tags generated by tag_torrent

### LINKED ISSUE #6819: Unstability in test_get_my_channel_tags
Sometimes [`test_get_my_channel_tags`](https://github.com/Tribler/tribler/blob/50c9571947ba951f7cfe8a3b77cd1e72379532a0/src/tribler/core/components/metadata_store/restapi/tests/test_channels_endpoint.py#L699) fail with the [following error](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_linux_pytest/3239/console):

```python
    async def test_get_my_channel_tags(metadata_store, mock_dlmgr_get_download, my_channel, rest_api):  # pylint: disable=redefined-outer-name
        """
        Test whether tags are correctly returned over the REST API
        """
        with db_session:
            json_dict = await do_request(
                rest_api,
                'channels/%s/%d?metadata_type=%d'
                % (hexlify(my_channel.public_key), my_channel.id_, REGULAR_TORRENT),
                expected_code=200,
            )
    
        assert len(json_dict['results']) == 9
        for item in json_dict['results']:
>           assert len(item["tags"]) >= 2
E           AssertionError: assert 1 >= 2
E            +  where 1 = len(['effect'])
```

The reason for the error is the following: [`my_channel`](https://github.com/Tribler/tribler/blob/dd8a622f8b8f8be18e0bad24c430b16326e1d0f1/src/tribler/core/components/metadata_store/restapi/tests/conftest.py#L45) fixture uses the [`tag_torrent`](https://github.com/Tribler/tribler/blob/dd8a622f8b8f8be18e0bad24c430b16326e1d0f1/src/tribler/core/components/metadata_store/utils.py#L43) helper function. That function constructs a list of random tags and associates them with a specified torrent's infohash. It expects the number of torrents to be at least 2. But sometimes random tag names generated by faker can repeat, and the actual number of tags will be just one.

To fix the problem, the `tag_torrent` function should handle possible duplicates in a list of random tags.

kozlovsky:
Fixes #6819

### DIFF
--- a/src/tribler/core/components/metadata_store/utils.py
+++ b/src/tribler/core/components/metadata_store/utils.py
@@ -41,10 +41,21 @@ def get_random_word(min_length=0):
 
 
 def tag_torrent(infohash, tags_db, tags=None, suggested_tags=None):
-    tags = tags or [get_random_word(min_length=MIN_TAG_LENGTH)
-                    for _ in range(random.randint(2, 6))]
-    suggested_tags = suggested_tags or [get_random_word(min_length=MIN_TAG_LENGTH)
-                                        for _ in range(random.randint(1, 3))]
+    if tags is None:
+        tags_count = random.randint(2, 6)
+        tags = []
+        while len(tags) < tags_count:
+            tag = get_random_word(min_length=MIN_TAG_LENGTH)
+            if tag not in tags:
+                tags.append(tag)
+
+    if suggested_tags is None:
+        suggested_tags_count = random.randint(1, 3)
+        suggested_tags = []
+        while len(suggested_tags) < suggested_tags_count:
+            tag = get_random_word(min_length=MIN_TAG_LENGTH)
+            if tag not in suggested_tags:
+                suggested_tags.append(tag)
 
     def _add_operation(_tag, _op, _key):
         operation = TagOperation(infohash=infohash, tag=_tag, operation=_op, clock=0,


kozlovsky:
retest this please

src/tribler/core/components/metadata_store/utils.py
```diff
@@ -41,10 +41,21 @@ def get_random_word(min_length=0):
...
+            if tag not in tags:
+                tags.append(tag)
+
+    if suggested_tags is None:
+        suggested_tags_count = random.randint(1, 3)
+        suggested_tags = []
+        while len(suggested_tags) < suggested_tags_count:
+            tag = get_random_word(min_length=MIN_TAG_LENGTH)
+            if tag not in suggested_tags:
+                suggested_tags.append(tag)
```
drew2a, line 58: **NIT, noncritical:** you use the same pattern twice. It is possible to extract this pattern:

```python
def generate_sentence_with_fixed_word_length(words_count: int, min_word_length) -> List[str]:
    sentence = []
    while len(sentence) < words_count:
        word = get_random_word(min_length=min_word_length)
        if word not in sentence:
            sentence.append(word)

    return sentence
```

and refactor the code block to:

```suggestion
    if tags is None:
        tags_count = random.randint(2, 6)
        tags = generate_sentence_with_fixed_word_length(tags_count, MIN_TAG_LENGTH)

    if suggested_tags is None:
        suggested_tags_count = random.randint(1, 3)
        suggested_tags = generate_sentence_with_fixed_word_length(suggested_tags_count, MIN_TAG_LENGTH)
```

drew2a (changes requested):

src/tribler/core/components/metadata_store/utils.py
kozlovsky: Thank you for the suggestion, but I'm not persuaded we should do it, for the following reasons:

1) A usual "rule of thumb" is ["if the same code is repeated three or more times, consider to extract it into a function"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)). Here the same pattern is repeated only twice.

2) The version with function does not make the code shorter.

3) It is not obvious that the function will be reusable in other places. It may be totally ok to have repeated words in a sentence in other situations.

4) Right now all repeated patterns are inside the same function, extracting the code to a second function makes the code "less local" and so a bit more complex to analyze (as the reader cannot assume that the new function is not used in other places).

drew2a: Agree, but the rationale behind referenced "rule of thumb" is the following:

> Attempting premature refactoring risks selecting a wrong abstraction, which can result in worse code as new requirements emerge and will eventually need to be refactored again.

In the code above suggested abstraction is clear and correct. It extracts exactly the same logic to the function with a very clear intention.

> The version with function does not make the code shorter.

Agree, but shorter doesn't mean better. 
The suggested refactoring code is easier to read because it doesn't require from reader to analyze the code blocs and to determine the pattern.

> It is not obvious that the function will be reusable in other places. In other situations it may be totally ok to have repeated words in a sentence.

Agree, you can make it local.

> Right now all repeated patterns are inside the same function, extracting the code to a second function makes the code "less local" and so a bit more complex to analyze (as the reader cannot assume that the new function is not used from other places).

Agree, you can make it local.

drew2a (approved):

PR #6821, opened 2022-03-21
Replace `get` by `select.limit` in BandwidthDatabase.get_latest_transaction

### LINKED ISSUE #6802: [7.11] MultipleObjectsFoundError: Multiple objects were found. Use BandwidthTransaction.select(...) to retrieve them
Sentry Issue: [TRIBLER-DK](https://sentry.tribler.org/organizations/tribler/issues/1148/?referrer=github_integration)

```
MultipleObjectsFoundError: Multiple objects were found. Use BandwidthTransaction.select(...) to retrieve them
(12 additional frame(s) were not displayed)
...
  File "pony\orm\core.py", line 4014, in get
  File "pony\orm\core.py", line 4121, in _find_one_
  File "pony\orm\core.py", line 4181, in _find_in_db_
  File "pony\orm\core.py", line 4301, in _fetch_objects
  File "pony\utils\utils.py", line 106, in throw
```

drew2a:
This PR fixes #6802.

It seems possible to create a multiple transactions for the given pair or `public_key_a` and `public_key_b` by setting `bandwidth_database.store_all_transactions` to `True`:

https://github.com/Tribler/tribler/blob/ec35106cc618ef60f73a1925a78812cc1549f515/src/tribler/core/components/bandwidth_accounting/db/transaction.py#L151-L161

That's why `BandwidthDatabase.get_latest_transaction` should be ready for this situation.

### DIFF
--- a/src/tribler/core/components/bandwidth_accounting/db/database.py
+++ b/src/tribler/core/components/bandwidth_accounting/db/database.py
@@ -1,7 +1,8 @@
 from pathlib import Path
 from typing import List, Optional, Union
 
-from pony.orm import Database, count, db_session, desc, select, sum
+from pony import orm
+from pony.orm import Database, count, db_session, desc, select
 
 from tribler.core.components.bandwidth_accounting.db import history, misc, transaction as db_transaction
 from tribler.core.components.bandwidth_accounting.db.transaction import BandwidthTransactionData
@@ -119,7 +120,7 @@ class BandwidthDatabase:
         :param public_key: The public key of the peer of which we want to determine the total taken.
         :return The total amount of bandwidth taken by the specified peer, in bytes.
         """
-        return sum(transaction.amount for transaction in self.BandwidthTransaction
+        return orm.sum(transaction.amount for transaction in self.BandwidthTransaction
                    if transaction.public_key_a == public_key)
 
     @db_session
@@ -129,7 +130,7 @@ class BandwidthDatabase:
         :param public_key: The public key of the peer of which we want to determine the total given.
         :return The total amount of bandwidth given by the specified peer, in bytes.
         """
-        return sum(transaction.amount for transaction in self.BandwidthTransaction
+        return orm.sum(transaction.amount for transaction in self.BandwidthTransaction
                    if transaction.public_key_b == public_key)
 
     @db_session
--- a/src/tribler/core/components/bandwidth_accounting/db/database.py
+++ b/src/tribler/core/components/bandwidth_accounting/db/database.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 from typing import List, Optional, Union
 
-from pony.orm import Database, count, db_session, select, sum
+from pony.orm import Database, count, db_session, desc, select, sum
 
 from tribler.core.components.bandwidth_accounting.db import history, misc, transaction as db_transaction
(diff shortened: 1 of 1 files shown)


drew2a:
retest this please

drew2a:
retest this please

drew2a:
retest this please

src/tribler/core/components/bandwidth_accounting/db/database.py
```diff
@@ -91,8 +93,12 @@ def get_latest_transaction(self, public_key_a: bytes, public_key_b: bytes) -> Ba
         :param public_key_b: The public key of the party receiving the bandwidth.
         :return The latest transaction between the two specified parties, or None if no such transaction exists.
         """
-        db_obj = self.BandwidthTransaction.get(public_key_a=public_key_a, public_key_b=public_key_b)
-        return BandwidthTransactionData.from_db(db_obj) if db_obj else None
+        db_txs = select(tx for tx in self.BandwidthTransaction
+                        if public_key_a == tx.public_key_a and public_key_b == tx.public_key_b) \
+            .order_by(desc(self.BandwidthTransaction.timestamp)) \
```
kozlovsky, line 98: **nitpick:** I suspect it is safer to order transactions by a sequence number:
```suggestion
            .order_by(lambda tx: desc(tx.sequence_number)) \
```

src/tribler/core/components/bandwidth_accounting/db/database.py
```diff
@@ -91,8 +93,12 @@ def get_latest_transaction(self, public_key_a: bytes, public_key_b: bytes) -> Ba
...
+            .limit(1)
+
+        return BandwidthTransactionData.from_db(db_txs[0]) if len(db_txs) == 1 else None
```
kozlovsky, line 101: **nitpick:** It is a bit more convenient to use `.first()` instead of `.limit(1)`, this way the next line will be a bit simpler:

```suggestion
            .first()

        return BandwidthTransactionData.from_db(db_tx) if db_tx else None
```

`.first` returns the first object (if exists) or None

src/tribler/core/components/bandwidth_accounting/db/database.py
```diff
@@ -114,7 +120,7 @@ def get_total_taken(self, public_key: bytes) -> int:
         :param public_key: The public key of the peer of which we want to determine the total taken.
         :return The total amount of bandwidth taken by the specified peer, in bytes.
         """
-        return sum(transaction.amount for transaction in self.BandwidthTransaction
+        return orm.sum(transaction.amount for transaction in self.BandwidthTransaction
```
kozlovsky, line 123: **nitpick:** It is also possible to use `sum` method of a query instead of the `sum` function to avoid `orm` import:
`return select(transaction.amount for transaction in ...).sum()`

kozlovsky (approved):

kozlovsky (approved):

devos50 (approved):
Thanks for fixing this! We indeed missed this case 👍 .

By default, `store_all_transactions` should be set to `False` on the end-user side so I don't think this would affect end users under normal circumstances. We actually added the `store_all_transactions` option to perform network analytics and to get insights in the dynamics of bandwidth transfers (since we would like to analyse as much transactions as possible). However, a user is able enable this flag by enabling the crawler mode of the bandwidth component. Maybe this is what happened and triggered the bug.

PR #6824, opened 2022-03-21
For test run with '--no-parallel' argument, skip all "normal" tests

drew2a:
Tribler's tests have been executed in Jenkins by the following commands:

```shell
pytest src/tribler/core -v --tx '12*popen//execmodel=eventlet' --cov=tribler --cov-report=xml:../output/coverage.xml --junitxml=test_report.xml
pytest src/tribler/core -v --no_parallel --cov=tribler --cov-report=xml:../output/coverage.xml --cov-append
```

1. During the first run, all tests that have the `@no_parallel` mark should be excluded from the execution.
2. During the second run, all tests that don't have the `@no_parallel` mark should be excluded from the execution.

This PR makes the second statement correct.


kozlovsky:
retest this please

PR #6827, opened 2022-03-22
Refactor add_fake_torrents_channels fixture

### LINKED ISSUE #6616: Assert in `test_get_popular_torrents` (during the tests on Mac)
When testing #6600, I noticed the following stack trace, also see [here](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_mac_pytest/2722/testReport/junit/src.tribler-core.tribler_core.components.metadata_store.restapi.tests/test_channels_endpoint/test_get_popular_torrents/):

```
assert 29 == 30   +29   -30

add_fake_torrents_channels = None, mock_dlmgr_get_download = None
mock_dlmgr = <Mock id='4611317520'>
rest_api = <aiohttp.test_utils.TestClient object at 0x112dcebe0>

    async def test_get_popular_torrents(add_fake_torrents_channels, mock_dlmgr_get_download, mock_dlmgr, rest_api):
        """
        Test getting the list of popular torrents. The list is served as contents of a pseudo-channel
        """
        json_dict = await do_request(rest_api, 'channels/popular_torrents', expected_code=200)
    
        def fields(d, *args):
            return {key: d[key] for key in args}
    
        seeders_orig_order = [fields(d, 'type', 'num_seeders', 'num_leechers') for d in json_dict['results']]
    
        def sort_key(d):
            a = 1 if d["type"] == CHANNEL_TORRENT else 2 if d["type"] == COLLECTION_NODE else 3
            b = -d["num_seeders"]
            c = -d["num_leechers"]
            return (a, b, c)
    
        assert seeders_orig_order == sorted(seeders_orig_order, key=sort_key)
>       assert len(json_dict['results']) == 30  # torrents 1, 3, 5 in each of 10 channels
E       assert 29 == 30
E         +29
E         -30

src/tribler-core/tribler_core/components/metadata_store/restapi/tests/test_channels_endpoint.py:289: AssertionError
```

drew2a:
This PR fixes #6616 by removing random (which could lead to some rarely occurred bugs) from `add_fake_torrents_channels` fixture.

### DIFF
--- a/src/tribler/core/components/metadata_store/restapi/tests/conftest.py
+++ b/src/tribler/core/components/metadata_store/restapi/tests/conftest.py
@@ -1,3 +1,4 @@
+import hashlib
 from time import time
 
 from ipv8.keyvault.crypto import default_eccrypto
@@ -11,34 +12,44 @@ from tribler.core.components.metadata_store.utils import tag_torrent
 from tribler.core.utilities.utilities import random_infohash
 
 
+@db_session
+def create_channel(metadata_store, title, torrents_per_channel=5, local_version=0, subscribed=True):
+    def sha1_hash(value: str) -> bytes:
+        return hashlib.sha1(value.encode('utf-8')).digest()
+
+    def add_torrent_to_channel(torrent_title, seeders, self_checked):
+        t = metadata_store.TorrentMetadata(origin_id=channel.id_, title=torrent_title,
+                                           infohash=sha1_hash(torrent_title), sign_with=key)
+        t.health.seeders = seeders
+        t.health.self_checked = self_checked
+
+        one_week = 60 * 60 * 24 * 7
+        now = int(time())
+
+        t.health.last_check = now - one_week if self_checked else now
+
+    key = default_eccrypto.generate_key('curve25519')
+
+    channel = metadata_store.ChannelMetadata(title=title, subscribed=subscribed,
+                                             num_entries=torrents_per_channel,
+                                             infohash=sha1_hash(title), id_=123,
+                                             sign_with=key, version=10,
+                                             local_version=local_version)
+
+    for torrent_i in range(torrents_per_channel):
+        add_torrent_to_channel(f'torrent{channel.title}{torrent_i}', seeders=torrent_i,
+                               self_checked=bool(torrent_i % 2))
+
+
+@pytest.fixture
+def add_subscribed_and_not_downloaded_channel(metadata_store):
+    create_channel(metadata_store, 'Subscribed and not downloaded', subscribed=True, local_version=0)
+
+
 @pytest.fixture
 def add_fake_torrents_channels(metadata_store):
-    infohashes = []
-
-    torrents_per_channel = 5
-    # Add a few channels
-    with db_session:
-        for ind in range(10):
-            ext_key = default_eccrypto.generate_key('curve25519')
-            channel = metadata_store.ChannelMetadata(
-                title='channel%d' % ind,
-                subscribed=(ind % 2 == 0),
-                num_entries=torrents_per_channel,
-                infohash=random_infohash(),
-                id_=123,
-                sign_with=ext_key,
-                version=10,
-                local_version=(ind % 11),
-            )
-            for torrent_ind in range(torrents_per_channel):
-                rand_infohash = random_infohash()
-                infohashes.append(rand_infohash)
-                t = metadata_store.TorrentMetadata(
-                    origin_id=channel.id_, title='torrent%d' % torrent_ind, infohash=rand_infohash, sign_with=ext_key
-                )
-                t.health.seeders = int.from_bytes(t.infohash[:2], byteorder="big")
-                t.health.self_checked = bool(torrent_ind % 2 == 1)
-                t.health.last_check = int(time()) - (60 * 60 * 24 * 7 if torrent_ind % 2 else 0)
+    for i in range(10):
+        create_channel(metadata_store, f'channel{i}', subscribed=bool(i % 2), local_version=i)
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/metadata_store/restapi/tests/test_channels_endpoint.py
```diff
@@ -65,8 +65,7 @@ async def test_get_channels(rest_api, add_fake_torrents_channels, mock_dlmgr, me
     mock_dlmgr.download_exists = lambda *args: None
     json_dict = await do_request(rest_api, 'channels')
     assert len(json_dict['results']) == 10
-    # Default channel state should be METAINFO_LOOKUP
-    assert json_dict['results'][-1]['state'] == CHANNEL_STATE.METAINFO_LOOKUP.value
+    assert json_dict['results'][0]['state'] == CHANNEL_STATE.UPDATING.value
```
kozlovsky, line 68: It seems that our tests sometimes check multiple things at the same time. This test not only checks that it is possible to retrieve channels from the corresponding endpoint but also checks that a channel with local_version equal to zero has ` CHANNEL_STATE.METAINFO_LOOKUP` state by default.

**issue:** After the change, the fixture does not return any channel with the `METAINFO_LOOKUP` state anymore. Probably it still should.

**suggestion:** To fix it, the `subscribed` attribute formula should be `subscribed=bool(channel_i % 2 == 0)`, so `channel0` that have `local_version` zero is not subscribed. After that, the state of the last channel will be `CHANNEL_STATE.METAINFO_LOOKUP` again.

src/tribler/core/components/metadata_store/restapi/tests/test_channels_endpoint.py
```diff
@@ -272,6 +271,7 @@ async def test_get_popular_torrents(add_fake_torrents_channels, mock_dlmgr_get_d
     Test getting the list of popular torrents. The list is served as contents of a pseudo-channel
     """
     json_dict = await do_request(rest_api, 'channels/popular_torrents', expected_code=200)
+    assert len(json_dict['results']) == 20
```
kozlovsky, line 274: **nitpick:** Maybe add a comment why the result's length is 20? Like:
```suggestion
    # torrents2 & torrent4 in each of 10 channels (but not torrent0, as it has 0 seeders)
    assert len(json_dict['results']) == 20
```

src/tribler/core/components/metadata_store/restapi/tests/conftest.py
```diff
@@ -13,32 +14,45 @@
...
+        for channel_i in range(10):
+            channel_title = f'channel{channel_i}'
+            channel_hash = sha1_hash(channel_title)
+            key = default_eccrypto.generate_key('curve25519')
+
             channel = metadata_store.ChannelMetadata(
-                title='channel%d' % ind,
-                subscribed=(ind % 2 == 0),
+                title=channel_title,
+                subscribed=bool(channel_i % 2),
```
kozlovsky, line 31: **issue:** changing the formula for subscribed affects the resulted torrents in unexpected way, I'd revert it to:
```suggestion
                subscribed=bool(channel_i % 2 == 0),
```

src/tribler/core/components/metadata_store/restapi/tests/conftest.py
```diff
@@ -13,32 +14,45 @@
 
 @pytest.fixture
 def add_fake_torrents_channels(metadata_store):
-    infohashes = []
+    def sha1_hash(value: str) -> bytes:
+        byte_array = bytearray(value, encoding='utf-8')
+        return hashlib.sha1(byte_array).digest()
```
kozlovsky, line 19: **nitpick:** can be written a bit more concise:
```suggestion
        return hashlib.sha1(value.encode('utf8')).digest()
```

kozlovsky (changes requested):

src/tribler/core/components/metadata_store/restapi/tests/test_channels_endpoint.py
drew2a: Well, I don't change the internal logic of channels. 
This change only shows that with new (simplified) test data the state of the channel is `UPDATING` and `PREVIEW` instead of `METAINFO_LOOKUP` which shows that `METAINFO_LOOKUP` is not a default state.

The states of all channels in this test are following:

```python
00 = {str} 'Updating'
01 = {str} 'Updating'
02 = {str} 'Updating'
03 = {str} 'Updating'
04 = {str} 'Updating'
05 = {str} 'Updating'
06 = {str} 'Updating'
07 = {str} 'Updating'
08 = {str} 'Updating'
09 = {str} 'Preview'
```

src/tribler/core/components/metadata_store/restapi/tests/conftest.py
drew2a: Could you give an example to prove the claim "changing the formula affects the resulted torrents in an unexpected way"?
Tests are passing.

src/tribler/core/components/metadata_store/restapi/tests/conftest.py
drew2a: Two-liner is easier to read than one-liner :)

drew2a:
retest this please

src/tribler/core/components/metadata_store/restapi/tests/conftest.py
kozlovsky: Sometimes two lines are indeed much clearer than one line. But I'm not sure this is that case. `sha1` accepts bytes, and the most straightforward way to convert string to bytes is to call the string's `encode` method. The difference between `bytearray` and `bytes` is that `bytearray` is mutable. So when someone creates `bytearray` instead of bytes, it can be expected that he wants to somehow change the value of the bytearray. Otherwise, it just looks strange.

drew2a: Ok, agree, thanks for the explanation.

kozlovsky (approved):

Issue #6706, opened 2022-01-04
Improvement Request: Align flake8 import order with Pycharm rules


drew2a:
During each PR we test new code for compliance with the `flake8` rules (see: https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_flake8)

The majority of Tribler's developers use PyCharm as the main IDE. PyCharm uses a different set of rules for import ordering than flake8 which often led to conflicts like [this](https://jenkins-ci.tribler.org/job/GH_Tribler_PR_Tests/job/PR_flake8/3795/artifact/output/index.html):

![image](https://user-images.githubusercontent.com/13798583/148057435-afb1e723-f775-4a60-a29d-8f098fdf7cfe.png)

![image](https://user-images.githubusercontent.com/13798583/148057468-d49c0296-baa0-4d25-912b-2ba3aca1c0ee.png)

It seems safe and convenient to align flake8 import order rules with Pycharm rules.

### Additional information

Flake8 rules are described here: 
https://github.com/Tribler/tribler/blob/1ff369756367ad49348ebab37a1ded95d1279244/.flake8#L1-L5

Flake8 is executed by running the following command:

```python
python3 /usr/local/bin/flake8 --format=html --htmldir=../output $FILES
```

drew2a:
https://pypi.org/project/flake8-import-order/

drew2a:
Implemented

PR #6837, opened 2022-03-28
Remove pre-commit

drew2a:
This PR implements changes that were requested in #6706 and aligns checks with the default `pycharm` scheme.

Also, it removes `pre-commit` and `black` as unused.


.flake8
```diff
@@ -1,5 +1,4 @@
 [flake8]
 max-line-length = 120
-select=E901,E999,F821,F822,F823,I100,I101,I201,I202,T001,T002,T003,T004
-ignore=E203,W503,C0330
-application-import-names=tribler,run_tribler
+ignore=E127, E128, E131, E731, E722, E501, F401, F841,  F722 , I202
```
kozlovsky, line 3: It seems it is not correct to specify both `select` and ìgnore`. In stat case, the `ignore` section just got ignored.
Also, the previous `select` list of enabled warnings did not include some useful warnings, for example:

```
./src\tribler\core\components\libtorrent\download_manager\download.py:714:5:
F811 redefinition of unused 'get_upload_mode' from line 705

./src\tribler\core\components\libtorrent\download_manager\download.py:718:5:
F811 redefinition of unused 'set_upload_mode' from line 709

./src\tribler\gui\tests\test_error_handler.py:68:1:
F811 redefinition of unused 'test_gui_core_connect_timeout_error' from line 59
```
I think the best way is to define extend-ignore list with errors we want to ignore:
```suggestion
extend-ignore =
  I202, # Additional newline in a group of imports
  E127, # continuation line over-indented for visual indent
  E128, # continuation line under-indented for visual indent
  E131, # continuation line unaligned for hanging indent
  F722, # syntax error in forward annotation '20s'
  E731, # do not assign a lambda expression, use a def
  F401, # Pylint-checked, module imported but unused
  F841, # Pylint-checked, local variable is assigned to but never used
  E722, # Pylint-checked, do not use bare 'except'
  E501 # Pylint-checked, line too long
```

I also added comments to error codes

kozlovsky (changes requested):

kozlovsky (approved):

Issue #6845, opened 2022-04-04
Improve the EVA protocol: kozlovsky edition


kozlovsky:
@drew2a I reviewed PR #6831 and did not find problems in this PR, but I think I found the other significant issue in EVA protocol unrelated to the PR changes.

The problem is in the `EVAProtocol.on_data` method:

```python
async def on_data(self, peer, payload):
    ...
    can_be_handled = transfer.block_number == payload.block_number - 1
    if not can_be_handled or transfer.nonce != payload.nonce:
        return
    ...
```

As I understand it, this method expects to receive the incoming messages in strictly sequential order, and a new message should have a block number that is one more than the previous one. At the same time, the `on_acknowledgement` method sends outgoing messages in batches of 16, where 16 is `window_size`:

```python
async def on_acknowledgement(self, peer: Peer, payload: Acknowledgement):
    ...
    for block_number in range(transfer.block_number, transfer.block_number + transfer.window_size):
        ...
        self.community.eva_send_message(peer, Data(block_number, transfer.nonce, data))
```

As messages are UDP packets, the network can deliver the messages of the same batch in any order; it is incorrect to expect they will be received in the same sequence. The current implementation of the `on_data` method will reject many messages as "out-of-order", and it can significantly slow down the data transfer. We can add a diagnostic counter of how many messages are received out-of-order, I suspect the number may be significant if there is some traffic in the network.

The `on_data` method should accept any message whose number lies inside the current window. The method can put incoming messages in an array of `window_size`, using the block number to calculate the message's offset in the array. Initially, it can be initialized as an array of `None`s, and after all messages of the current window are received, they can be joined together and appended to the result binary string. This way, EVAProtocol performance can potentially be much better under a heavy load.

---

Also, a couple of minor notices:

First:

```python
transfer.window_size = max(self.MIN_WINDOWS_SIZE, min(payload.window_size, self.binary_size_limit))
```
It does not look correct to calculate `min` between `window_size` and `binary_size_limit` because `window_size` is a dimensionless quantity (the number of messages), while binary_size_limit` is the number of bytes. Apples and oranges.

Second:

```python
transfer.data_binary += payload.data_binary
```
This is potentially slow because it copies the entire previous data each time a new piece is appended. The correct way is to accumulate binary chunks in a list:
```python
transfer.data_list.append(payload.data_binary)
```
And join all chunks to a single binary string at the end in a single step:
```python
result = ""b.join(transfer.data_list)
````

---
Additional from @devos50 

[...]

PR #6878, opened 2022-04-20
EVA: add incoming array

drew2a:
This PR addresses #6845 

1. Adds an array for incoming messages
2. Corrects calculating of `window_size` limits
3. Improves the final binary array calculation

### DIFF
--- a/src/tribler/core/components/ipv8/eva_protocol.py
+++ b/src/tribler/core/components/ipv8/eva_protocol.py
@@ -31,7 +31,7 @@ An example of use:
 >>>     async def on_error(self, peer, exception):
 ...         self.logger.error(f'Error has been occurred: {exception}')
 """
-__version__ = '2.0.0'
+from __future__ import annotations
 
 import asyncio
 import logging
@@ -42,11 +42,13 @@ from collections import defaultdict, deque
 from dataclasses import dataclass
 from enum import Enum, auto
 from random import SystemRandom
-from typing import Awaitable, Callable, Dict, Optional, Type
+from typing import Awaitable, Callable, Dict, List, Optional, Type
 
 from ipv8.lazy_community import lazy_wrapper
 from ipv8.messaging.lazy_payload import VariablePayload, vp_compile
 from ipv8.types import Peer
+
+__version__ = '2.1.0'
 
 logger = logging.getLogger('EVA')
 
@@ -68,7 +70,7 @@ class Acknowledgement(VariablePayload):
 @vp_compile
 class Data(VariablePayload):
     format_list = ['I', 'I', 'raw']
-    names = ['block_number', 'nonce', 'data']
+    names = ['number', 'nonce', 'data']
 
 
 @vp_compile
@@ -144,6 +146,36 @@ class TransferResult:
     data: bytes
     nonce: int
 
+    def __str__(self):
+        return f'TransferResult(peer={self.peer}, info: {self.info}, data hash: {hash(self.data)}, nonce={self.nonce})'
+
+
+class TransferWindow:
+    def __init__(self, start: int, size: int):
+        self.blocks: List[Optional[bytes]] = [None] * size
+
+        self.start = start
+        self.processed: int = 0
+        self.last_window: bool = False
+
+    def add(self, index: int, block: bytes):
+        if self.blocks[index] is not None:
+            return
+        self.blocks[index] = block
+        self.processed += 1
+
+    def is_finished(self) -> bool:
+        return self.processed == len(self.blocks)
+
+    def consecutive_blocks(self):
+        for block in self.blocks:
+            if block is None:
+                break
+            yield block
+
+    def __str__(self):
+        return f'{{start: {self.start}, processed: {self.processed}, size: {len(self.blocks)}}}'
+
 
 class Transfer:  # pylint: disable=too-many-instance-attributes
     """The class describes an incoming or an outgoing transfer"""
@@ -151,25 +183,25 @@ class Transfer:  # pylint: disable=too-many-instance-attributes
     NONE = -1
 
     def __init__(self, transfer_type: TransferType, info: bytes, data: bytes, data_size: int, block_count: int,
-                 nonce: int, peer: Peer, protocol, future: Optional[Future] = None, window_size: int = 0,
-                 updated: float = 0):
+                 nonce: int, peer: Peer, protocol: EVAProtocol, future: Optional[Future] = None, updated: float = 0):
         """ This class has been used internally by the EVA protocol"""
         self.type = transfer_type
         self.info = info
         self.data = data
+        self.data_list: List[bytes] = []  # for incoming transfers only
         self.data_size = data_size
         self.block_count = block_count
-        self.block_number = Transfer.NONE
         self.future = future
         self.peer = peer
         self.nonce = nonce
-        self.window_size = window_size
         self.updated = updated
         self.protocol = protocol
 
+        self.window: Optional[TransferWindow] = None
+
         self.attempt = 0
-        self.acknowledgement_number = 0
         self.terminated = False
+        self.acknowledgement_received = False
 
     def finish(self):
         result = TransferResult(peer=self.peer, info=self.info, data=self.data, nonce=self.nonce)
@@ -179,7 +211,7 @@ class Transfer:  # pylint: disable=too-many-instance-attributes
         if self.terminated:
             return
 
-        logger.debug(f'Terminate. Peer: {self.peer}. Transfer: {self}')
+        logger.debug(f'Terminate. Result: {result}.')
 
         container = self.protocol.incoming if self.type == TransferType.INCOMING else self.protocol.outgoing
         container.pop(self.peer, None)
@@ -194,8 +226,28 @@ class Transfer:  # pylint: disable=too-many-instance-attributes
         self.peer = None
         self.protocol = None
         self.future = None
+        self.data_list = None
+        self.window = None
 
         self.terminated = True
+
+    def create_window(self):
+        if not self.window:
+            self.window = TransferWindow(start=0, size=self.protocol.window_size)
+            logger.debug(f'New window: {self.window}')
+            return
+
+        self.data_list.extend(self.window.consecutive_blocks())
+        self.window = TransferWindow(
+            start=len(self.data_list),
+            size=self.protocol.window_size
+        )
+        logger.debug(f'Next window: {self.window}')
+
+    def get_block(self, number: int) -> bytes:
+        start_position = number * self.protocol.block_size
+        stop_position = start_position + self.protocol.block_size
+        return self.data[start_position:stop_position]
 
     def _terminate_with_result(self, result: TransferResult):
         if self.future:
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
+            return
+        self.data[index] = block
+        self.processed += 1
+
+    def finished(self) -> bool:
+        return self.processed == self.size
+
+    def flush(self, data_list: List[bytes]) -> int:
+        added = 0
+        for number in range(self.size):
```
kozlovsky, line 170: Can be written as `for block in self.data:`

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
+    def add(self, index: int, block: bytes):
+        if self.data[index] is not None:
+            return
+        self.data[index] = block
+        self.processed += 1
+
+    def finished(self) -> bool:
+        return self.processed == self.size
+
+    def flush(self, data_list: List[bytes]) -> int:
```
kozlovsky, line 168: **thought:** It may be a tiny bit more flexible to have a method that returns a list of consecutive blocks instead of modifying the `data_list` as a side effect. In that case, inside the method, you can write something like:

```
def consecutive_blocks(self):
    for block in self.data:
        if block is None:
            break
        yield block
```

And outside of the method, you can do `transfer.data_list.extend(window.consecutive_blocks())`.

In that case, when creating a `TransferWindow`, instead of calculating `start=self.window.start + added` we can do `start=len(transfer.data_list)`, so calculating `added` is unnecessary.

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
+        return f'TransferResult(peer={self.peer}, info: {self.info}, data hash: {hash(self.data)}, nonce={self.nonce})'
+
+class TransferWindow:
+    def __init__(self, start: int, size: int):
+        self.data: List[Optional[bytes]] = [None for _ in range(size)]
+
+        self.size = size
+        self.start = start
+        self.processed: int = 0
+        self.last_window: bool = False
```
kozlovsky, line 157: **thought:** The `TransferWindow` itself does not use the `last_window` flag, so in my opinion, it is better to move it to `Transfer` because it is possible that in further simplifications, it may be removed entirely and replaced with something like `if len(transfer.data_list) == number_of_packets_to_receive: ...`

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
+
+    def add(self, index: int, block: bytes):
+        if self.data[index] is not None:
+            return
+        self.data[index] = block
+        self.processed += 1
+
+    def finished(self) -> bool:
```
kozlovsky, line 165: It is better to rename it to `is_finished`

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
         self.block_count = block_count
-        self.block_number = Transfer.NONE
         self.future = future
         self.peer = peer
         self.nonce = nonce
-        self.window_size = window_size
         self.updated = updated
         self.protocol = protocol
 
+        self.window: Optional[TransferWindow] = None
```
kozlovsky, line 202: For incoming and outgoing transfers, the lists of attributes are pretty different, so it looks like they should be presented as two separate subclasses, `IncomingTransfer` and `OutgoingTransfer`.

kozlovsky (changes requested):

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: Yes, this is on my TODO list. But I'm not sure that it should be done in this PR.

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
     data: bytes
     nonce: int
 
+    def __str__(self):
+        return f'TransferResult(peer={self.peer}, info: {self.info}, data hash: {hash(self.data)}, nonce={self.nonce})'
+
+class TransferWindow:
+    def __init__(self, start: int, size: int):
+        self.data: List[Optional[bytes]] = [None for _ in range(size)]
```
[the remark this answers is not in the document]
drew2a: Changed.

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
+
+        self.size = size
```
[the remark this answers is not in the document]
drew2a: Well... A window by definition is not about bytes but about packets. So in this case, `size` looks like the correct name. 

I've replaced the property `self.size` with `len(self.blocks)` as you suggested but left `size` in the constructor.

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
-        self.window_size = window_size
         self.updated = updated
         self.protocol = protocol
 
+        self.window: Optional[TransferWindow] = None
+
         self.attempt = 0
-        self.acknowledgement_number = 0
         self.terminated = False
+        self.acknowledgment_received = False
```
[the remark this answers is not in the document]
drew2a: fixed

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -492,77 +542,70 @@ async def on_acknowledgement(self, peer: Peer, payload: Acknowledgement):
...
-            logger.warning("Cannot handle incoming acknowledgement from peer %s - nonce mismatch", peer)
+            logger.warning(f'Cannot handle incoming acknowledgement from peer {peer} - nonce mismatch.')
             return
 
-        transfer.block_number = payload.number
-        if transfer.block_number > transfer.block_count:
+        transfer.acknowledgment_received = True
+        transfer.updated = time.time()
+
+        is_final_acknowledgment = payload.number > transfer.block_count
```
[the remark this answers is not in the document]
drew2a: fixed

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -144,32 +144,66 @@ class TransferResult:
...
 
 class Transfer:  # pylint: disable=too-many-instance-attributes
     """The class describes an incoming or an outgoing transfer"""
 
     NONE = -1
 
     def __init__(self, transfer_type: TransferType, info: bytes, data: bytes, data_size: int, block_count: int,
-                 nonce: int, peer: Peer, protocol, future: Optional[Future] = None, window_size: int = 0,
-                 updated: float = 0):
+                 nonce: int, peer: Peer, protocol, future: Optional[Future] = None, updated: float = 0):
```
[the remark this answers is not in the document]
drew2a: Magic :)

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: 👍

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: Agree, your design is better. 
I partially dislike my implementation of `flush` because it modified `data_list` and makes it mutable (which we should avoid).

drew2a: I've implemented your suggestion.

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: renamed

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -194,9 +228,29 @@ def terminate(self, result: Optional[TransferResult] = None, exception: Optional
...
+            return
+
+        added = self.window.flush(self.data_list)
+        self.window = TransferWindow(
+            start=self.window.start + added,
+            size=self.protocol.window_size
+        )
+        logger.debug(f'Next window: {self.window}')
+
+    def block(self, number: int) -> bytes:
```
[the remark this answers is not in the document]
drew2a: renamed

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -194,9 +228,29 @@ def terminate(self, result: Optional[TransferResult] = None, exception: Optional
         self.peer = None
         self.protocol = None
         self.future = None
+        self.data_list = None
+        self.window = None
 
         self.terminated = True
 
+    def create_window(self):
```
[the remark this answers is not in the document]
drew2a: This is on my TODO list.

I can do it in this PR (for me it is ok). It just increases the difficulty of review.

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: I know. I placed it into the `Window` class just because of the feeling that this is safer rather than placing it into the `Transfer` class.
My explanation for this feeling is the following: A `Window` creates often. A `Transfer` creates only once. In case of any mistakes in the implementation of EVA, a more frequently created object (window) could level up some of the issues.

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -492,77 +542,70 @@ async def on_acknowledgement(self, peer: Peer, payload: Acknowledgement):
...
+        if transfer.window.finished():
+            logger.debug(f'Window finished: {transfer.window}')
+            if not transfer.window.last_window:
+                self.send_acknowledgement(transfer)
+                return
+
+            logger.debug('Finish transfer')
             self.send_acknowledgement(transfer)
+            transfer.data = b''.join(transfer.data_list)
+            transfer.finish()
```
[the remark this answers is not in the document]
drew2a: Yeap. Maybe you are right and we should encapsulate some logic from `Protocol` to `Transfer`. 

I'll try to do so and let you know is it worth it or not.

src/tribler/core/components/ipv8/eva_protocol.py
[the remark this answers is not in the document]
kozlovsky: I'm Ok to have it in this PR

kozlovsky (approved):

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: Will be addressed here: https://github.com/Tribler/tribler/pull/6882

src/tribler/core/components/ipv8/eva_protocol.py
drew2a: Will be addressed here: https://github.com/Tribler/tribler/pull/6882

src/tribler/core/components/ipv8/eva_protocol.py
```diff
@@ -194,9 +228,29 @@ def terminate(self, result: Optional[TransferResult] = None, exception: Optional
...
 
         self.terminated = True
 
+    def create_window(self):
+        if not self.window:
+            self.window = TransferWindow(start=0, size=self.protocol.window_size)
+            logger.debug(f'New window: {self.window}')
+            return
+
+        added = self.window.flush(self.data_list)
```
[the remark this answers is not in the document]
drew2a: Will be addressed here: https://github.com/Tribler/tribler/pull/6883

src/tribler/core/components/ipv8/eva_protocol.py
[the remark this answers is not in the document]
drew2a: Will be addressed here: https://github.com/Tribler/tribler/pull/6883

PR #6887, opened 2022-04-26
Split the EVA protocol file

drew2a:
This PR addresses https://github.com/Tribler/tribler/issues/6845

Only splitting. No new code has been added.


src/tribler/core/components/ipv8/eva/exceptions.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+from typing import Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from tribler.core.components.ipv8.eva.transfer import Transfer
```
kozlovsky, line 6: Cool, I didn't know that `TYPE_CHECKING` allows to solve cyclic imports when types are needed for type checks only

kozlovsky (approved):

src/tribler/core/components/ipv8/eva/exceptions.py
drew2a: 🚀

PR #6913, opened 2022-05-25
Refactoring of workflows

drew2a:
This PR adds two new workflows:

* `!PR.yml` for the event `on: [pull_request]`
* `!main.yml` for the event `on: [push: [branches: [main]]]`

In these workflows, there are lists of child jobs that must be called for these events. It must simplify the understanding of which workflows runs on which event.

Eg:

```yml
name: Main

on:
  push:
    branches:
      - main

jobs:
  codacy:
    uses: ./.github/workflows/codacy.yml

  coverage:
    uses: ./.github/workflows/coverage.yml

  docker:
    uses: ./.github/workflows/docker-image.yml

  pytest:
    uses: ./.github/workflows/pytest.yml
```

All jobs besides these two are now triggered by the `workflow_call` event: https://docs.github.com/en/actions/using-workflows/reusing-workflows


xoriole (approved):

kozlovsky (approved):

drew2a:
@xoriole @kozlovsky I've added ubuntu build to the `!PR.yml`

PR #6923, opened 2022-06-02
Add pip cache for GitHub Actions

drew2a:
This PR adds cache for pip python dependencies. It should decrease the time of job runs.

(this improvement saves ~1m for each run)

Ref:
* https://github.com/actions/setup-python/issues/330
* https://docs.github.com/en/actions/creating-actions/about-custom-actions#composite-actions
* https://github.com/syphar/restore-virtualenv

### DIFF
--- a/.github/actions/pyenv/action.yml
+++ b/.github/actions/pyenv/action.yml
@@ -1,0 +1,37 @@
+name: pyenv
+description: Setup python and install required dependencies to venv with cache
+inputs:
+  python-version:
+    default: '3.8'
+    description: 'Python version'
+    required: false
+
+  requirements:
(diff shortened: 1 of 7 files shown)


xoriole (commented):
Nice! How can the cache be cleared if we need to?

drew2a:
@xoriole in the current implementation it is not possible.

But for cache renewal like this:

```shell
python -m pip install --upgrade pip
pip install -r requirements.txt
```

You can just comment this line:
https://github.com/drew2a/tribler/blob/a48031e768de099ec2ad955051bd3adc859e36b1/.github/actions/pyenv/action.yml#L33

xoriole (approved):

PR #6928, opened 2022-06-08
Build Tribler binaries for macOS

drew2a:
This PR adds the `build_mac.yml` workflow that can be run manually.

Also, this PR performs the last step of `update_version_from_git.py` refactoring.

This change is backward-compatible with Jenkins one-click-release, but we must add the following lines to Jenkins before the call `makedist_mac.sh`:

```
          git describe | python -c "import sys; print(next(sys.stdin).lstrip('v'))" > .TriblerVersion
          git rev-parse HEAD > .TriblerCommit

          python ./build/update_version.py -r .
```
Ref:
* https://stackoverflow.com/questions/8970981/attach-license-file-to-dmg-under-osx
* https://apple.stackexchange.com/questions/387000/filename-is-damaged-and-cant-be-opened-you-should-move-it-to-the-bin-when-o


drew2a:
It works, but produced applications can't be run because of "Tribler.app is damaged and can’t be opened. you should move it to the bin" error.

It can be solved by running `xattr -cr ./Tribler.app`

xoriole (approved):

PR #6936, opened 2022-06-14
Fix  "error: unrecognized arguments"

### LINKED ISSUE #6929: [Ubuntu, 7.12-RC1] Can't open torrent file from browser
Ubuntu 22.04
Tribler 7.12-RC1

When trying to open a torrent file directly from Firefox's "Downloads" list, a window with the following text appears:
`{'exc_info': (<class 'tribler.gui.exceptions.CoreCrashedError'>, CoreCrashedError('The Tribler core has unexpectedly finished with exit code 2 and status: 0!\nLast core output: \n usage: tribler [-h] [--core] [--gui-test-mode] [--allow-code-injection]\n               [--trace-exception] [--trace-debug] [--testnet]\n               [--chant-testnet] [--trustchain-testnet] [--tunnel-testnet]\n               [torrent]\ntribler: error: unrecognized arguments: /home/bob/Downloads/cool_torrentfile.torrent'), <traceback object at 0x7f68db489580>), 'attachments': []}.`

Note the crash happens only when Tribler is closed (i.e. when it should be started automatically when clicking on a torrent file). When Tribler is already running, the torrent is opened correctly.

drew2a:
This PR fixes #6929

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -73,8 +73,8 @@ def init_boot_logger():
 if __name__ == "__main__":
     init_boot_logger()
     init_sentry_reporter()
-
-    parsed_args = RunTriblerArgsParser().parse_args()
+    
+    parsed_args, _ = RunTriblerArgsParser().parse_known_args()
     logger.info(f'Run Tribler: {parsed_args}')
 
     # Get root state directory (e.g. from environment variable or from system default)


drew2a:
@kozlovsky could you please confirm that it works on built windows?

kozlovsky (changes requested):
The fix is indeed working; I can confirm it on Windows. But I don't think this is the proper way to handle the issue. The proposed fix just ignores arguments that ArgsParser does not understand, and it may be error-prone. Like, if later GUI starts specifying new options when running Core, any typo in an argument will remain unnoticed, as Core will silently ignore unknown arguments.

I debugged the problem and found the reason for the error. It is in [this line](https://github.com/Tribler/tribler/blob/410fb735d3758ccf8ac4745e0f2cd3c89a31f2e9/src/tribler/gui/core_manager.py#L98) of CoreManager:

```python
self.core_process.start(sys.executable, core_args)
```

There is a difference in arguments between a python script and a compiled executable. For script, arguments can look like
```python
["myscript.py", "arg1", "arg2"]
```
and `sys.executable` is something like `C:/Python39/python.exe`.

For the same code in the compiled executable, arguments will be
```python
["myscript.exe", "arg1", "arg2"]
```
and `sys.executable` is `myscript.exe`.

That means that n Tribler executing `self.core_process.start(sys.executable, core_args)` from the compiled binary runs `tribler.exe` with arguments `["tribler.exe", "some.torrent", "--core"]`. So the proper fix, in my opinion, is to remove duplicate `"tribler.exe"` from the argument list when running a code from compiled binary.

drew2a:
The proper fix is here: https://github.com/Tribler/tribler/pull/6937

Issue #6938, opened 2022-06-16
GitHub Action for testing latest IPv8 changes


qstokkink:
**Is your feature request related to a problem? Please describe.**
There is no easy way to check if Tribler works with the latest IPv8 version on GitHub. The implication is that I have "no idea" if Tribler crashes and burns when I release a new IPv8 version (of course I manually test changes, but this only goes so far).

**Describe the solution you'd like**
I'd like to have a GitHub Action that lets me easily run Tribler's test suite with the latest version of IPv8. This lets me make changes to IPv8 for Tribler interoperability before the production branch crashes and burns due to IPv8.

**Describe alternatives you've considered**
I'm testing locally, but this is not quite as easy as launching a GitHub Action and this approach only ensures that changes work on my machine.

**Additional context**
N/A

PR #6943, opened 2022-06-22
Add `test_latest_ipv8.yml`

drew2a:
This PR solves #6938


qstokkink (approved):
**praise:** Perfect in every way; very much approved.

PR #6955, opened 2022-06-28
Simplify Session class

### LINKED ISSUE #6356: [Components] Session implementation have to be simplified
During the refactoring (#6335), we found the following potential problem:

Class [session](https://github.com/Tribler/tribler/blob/main/src/tribler-core/tribler_core/components/base.py#L35) operates with multiple stacks to make it possible to perform async test calls:

```python
class Session:
    _next_session_id = count(1)
    _default: Optional[Session] = None
    _stack: List[Session] = []

    @staticmethod
    def _get_default_session() -> Session:
        if Session._default is None:
            raise SessionError("Default session was not set")
        return Session._default

    def set_as_default(self):
        Session._default = self

    @staticmethod
    def unset_default_session():
        Session._default = None

    @staticmethod
    def current() -> Session:
        if Session._stack:
            return Session._stack[-1]
        return Session._get_default_session()
```
The potential problem here is that this complexity can lead to non-obvious bugs that are difficult to reproduce and debug. 

For production use this logic is unnecessary. 
We talk to @kozlovsky about the Session offline, and he has a solution for that.

drew2a:
This PR closes #6356 by simplifying the `Session` class. The ability to use multiple sessions has been removed.

### DIFF
--- a/src/tribler/core/components/session.py
+++ b/src/tribler/core/components/session.py
@@ -4,11 +4,10 @@ import logging
 import os
 import sys
 from asyncio import Event, create_task, gather, get_event_loop
-from contextlib import asynccontextmanager
 from pathlib import Path
 from typing import Dict, List, Optional, Type, TypeVar
 
-from tribler.core.components.base import Component, ComponentError
+from tribler.core.components.component import Component, ComponentError
 from tribler.core.config.tribler_config import TriblerConfig
 from tribler.core.utilities.crypto_patcher import patch_crypto_be_discovery
 from tribler.core.utilities.install_dir import get_lib_path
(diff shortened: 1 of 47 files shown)


src/tribler/core/components/session.py
```diff
@@ -0,0 +1,129 @@
...
+from tribler.core.utilities.notifier import Notifier
+from tribler.core.utilities.simpledefs import STATEDIR_CHANNELS_DIR, STATEDIR_DB_DIR
+
+
+class SessionError(Exception):
+    pass
+
+
+@asynccontextmanager
+async def session_manager(session: Session):
```
kozlovsky, line 25: I think session_manager can be removed as well

kozlovsky (changes requested):
Great changes!

src/tribler/core/components/session.py
drew2a: Thank you, removed.

kozlovsky (approved):

PR #6960, opened 2022-07-04
Fix Tribler shutdown after closing of FeedbackDialog

kozlovsky:
Currently, in case of exception, if SentryReporter is in test mode, after reporting an exception it calls [`AppManager.quit_application()`](https://github.com/Tribler/tribler/blob/da025dc5b5e651064bd4e54cfe2a2d29f367bf20/src/tribler/gui/dialogs/feedbackdialog.py#L199), which in turn calls `QApplication.quit()`.

This logic is incorrect, as the Tribler core process, managed by a `QProcess` instance, terminated immediately without any possibility to do cleanup work, like removing the `triblerd.lock` file or saving the libtorrent state.

The correct behavior is to stop the Core process the usual way and then shut down GUI. This PR fixes the shutdown logic, allowing the Core to shut down properly.

### DIFF
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -143,10 +143,15 @@ class CoreManager(QObject):
 
         self.shutting_down = True
         self._logger.info("Stopping Core manager")
-        if self.core_process or self.core_connected:
+
+        need_to_shutdown_core = (self.core_process or self.core_connected) and not self.core_finished
+        if need_to_shutdown_core:
             self._logger.info("Sending shutdown request to Tribler Core")
             self.events_manager.shutting_down = True
             TriblerNetworkRequest("shutdown", lambda _: None, method="PUT", priority=QNetworkRequest.HighPriority)
+        elif self.should_quit_app_on_core_finished:
+            self._logger.info('Core finished, quitting GUI application')
+            self.app_manager.quit_application()
 
     def on_core_finished(self, exit_code, exit_status):
         self.core_running = False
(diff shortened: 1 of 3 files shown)


src/tribler/gui/core_manager.py
```diff
@@ -143,10 +143,12 @@ def stop(self, quit_app_on_core_finished=True):
 
         self.shutting_down = True
         self._logger.info("Stopping Core manager")
-        if self.core_process or self.core_connected:
+        if (self.core_process or self.core_connected) and not self.core_finished:
             self._logger.info("Sending shutdown request to Tribler Core")
             self.events_manager.shutting_down = True
             TriblerNetworkRequest("shutdown", lambda _: None, method="PUT", priority=QNetworkRequest.HighPriority)
+        elif self.should_quit_app_on_core_finished:
+            self.app_manager.quit_application()
```
drew2a, line 151: I'm not sure, but Tribler should crash on this line because you have removed the line `self.app_manager: AppManager = parent.app_manager` from FeedbackDialog class.

drew2a (changes requested):

src/tribler/gui/core_manager.py
```diff
@@ -143,10 +143,12 @@ def stop(self, quit_app_on_core_finished=True):
 
         self.shutting_down = True
         self._logger.info("Stopping Core manager")
-        if self.core_process or self.core_connected:
+        if (self.core_process or self.core_connected) and not self.core_finished:
```
drew2a, line 146: NIT: for me it is more clear
```suggestion
        need_to_shutdown_core = (self.core_process or self.core_connected) and not self.core_finished
        if need_to_shutdown_core:
```

drew2a (approved):

xoriole (approved):

src/tribler/gui/core_manager.py
kozlovsky: Thanks, changed the condition this way

PR #6965, opened 2022-07-05
Fix `extract_swagger.py`

### LINKED ISSUE #6674: Endpoints documentation pages empty
The read the docs pages for all the endpoints are empty. For eg.

https://tribler.readthedocs.io/en/latest/restapi/createtorrent.html
https://tribler.readthedocs.io/en/latest/restapi/debug.html

drew2a:
This PR fixes #6674

Two main changes:
1. `extract_swagger.py` now runs full Tribler session instead of just `RestManager`. In this session only two components are presented: `[ReporterComponent, RESTComponent]`.
2. `RESTComponent` now can be run in "swagger_mode" which means all endpoints should be initialized but without real components.

image


src/tribler/core/components/restapi/restapi_component.py
```diff
@@ -115,10 +120,11 @@ async def run(self):
                        metadata_store_component.mds)
         self.maybe_add('/tags', TagsEndpoint, db=tag_component.tags_db, community=tag_component.community)
 
-        ipv8_root_endpoint = IPV8RootEndpoint()
-        for _, endpoint in ipv8_root_endpoint.endpoints.items():
-            endpoint.initialize(ipv8_component.ipv8)
-        self.root_endpoint.add_endpoint('/ipv8', ipv8_root_endpoint)
+        if not isinstance(ipv8_component, NoneComponent):
```
kozlovsky, line 123: If I understand correctly, with this logic `/ipv8` endpoint with its children will not be included in the generated API doc. How about including `IPV8RootEndpoint` unconditionally and then skipping `endpoint.initialize(...)` calls to all children endpoints in swagger mode?

src/tribler/core/components/restapi/restapi_component.py
```diff
@@ -51,12 +52,15 @@ def maybe_add(self, path: str, endpoint_cls: Type[RESTEndpoint], *args, **kwargs
...
         arguments_chain = chain(args, kwargs.values())
-        if any(isinstance(arg, NoneComponent) for arg in arguments_chain):
-            self.logger.info("Endpoint %s can't load component, skipping init", path)
+        need_to_skip = any(isinstance(arg, NoneComponent) for arg in arguments_chain)
+        if need_to_skip and not self.swagger_mode:
+            self.logger.warning("Skipped")
             return
 
         self.root_endpoint.add_endpoint(path, endpoint_cls(*args, **kwargs))
+        self.logger.info("OK")
```
kozlovsky, line 63: **nitpick:** A single `"OK"` log line may be uninformative. How about changing it to something like "Endpoint XXX added" or removing it entirely?

doc/extract_swagger.py
```diff
@@ -1,34 +1,42 @@
...
+        proto.data_received(b'GET /docs/swagger.json HTTP/1.1\r\n'
+                            b'Connection: close\r\n\r\n')
+        await proto.start()
+        logging.info('Proto has been started.')
 
-    root_endpoint = RootEndpoint(middlewares=[ApiKeyMiddleware(config.api.key), error_middleware])
-    api_manager = RESTManager(config=config.api, root_endpoint=root_endpoint, state_dir=config.state_dir)
-    await api_manager.start()
+        response = fp.getvalue()
+        logging.info(f'Response: {response}')
```
kozlovsky, line 36: **nitpick:** Big response output can spam logs. I'd use `debug` log level instead of `info`.

kozlovsky (changes requested):
Looks good! I'd add the `/ipv8` endpoint to the Swagger documentation as well (if I understand correctly, it is currently skipped)

src/tribler/core/components/restapi/restapi_component.py
```diff
@@ -43,6 +43,7 @@
 class RESTComponent(Component):
     rest_manager: RESTManager = None
     root_endpoint: RootEndpoint = None
+    swagger_mode: bool = False
```
xoriole, line 46: `swagger_mode` seems unclear to me. By swagger, the following UI comes to mind where users can test out REST APIs. By having this variable as default false, it could be assumed that the UI might not be available as well, which is not the case. I'd suggest renaming it something in line with extraction related.
swagger-ui-tribler

drew2a: @xoriole what name would you prefer?

src/tribler/core/components/restapi/restapi_component.py
drew2a: From my POV lines like "Endpoint XXX added" just after "Adding XXX..." lead only to the pollution of the logs.

doc/extract_swagger.py
drew2a: Done

src/tribler/core/components/restapi/restapi_component.py
kozlovsky: Maybe `swagger_doc_extraction_mode`?

kozlovsky (approved):

src/tribler/core/components/restapi/restapi_component.py
xoriole: Alexander's suggestion sounds good

drew2a: Done

PR #6694, opened 2021-12-29
Fix 6653

drew2a:
This PR fixes #6653 by switching `SentryReporter` to a normal class instead of a singleton.

Tests of `SentryReporter` have been rewritten.

As a part of `SentryReporter`'s tests refactoring, the ability to pass a module-as-a-string has been added to `patch_import`.

PR #6975, opened 2022-07-20
Fixes Tribler does not send crash reports to Sentry

### LINKED ISSUE #6974: [pre-7.12] Tribler crashes are not reported to Sentry anymore
During the release preparation, I discovered that Tribler 7.12 does not send crash reports to Sentry anymore:
* Instead of displaying the FeedbackDialog that allows editing crash reports before sending, Tribler shows a simplified message box with minimal error details.
Screenshot 2022-07-20 at 22 39 58

* In the next simplified dialog box, Tribler suggests reporting the error, but even if the user press "Yes", the error is not reported
Screenshot 2022-07-20 at 22 40 22

To reproduce the problem, you can press the `Ctrl+O` shortcut that triggers a not-yet-fixed bug in GUI, or you can press the`Ctrl+Alt+Shift+G` shortcut that triggers a test GUI error that is handled just like a typical GUI error.

* In case of a Core-related exception, the normal FeedbackDialog shows correctly, but the error still does not get reported to Sentry after the user presses the "Send" button.

The issue should be fixed before the release, or we do not get reports about Tribler crashes.

kozlovsky:
This PR fixes #6974

As it turns out, in #6694, SentryReporter stops being a singleton. Several instances of it are created in different places in the code, one in [`tribler_gui.__init__`](https://github.com/Tribler/tribler/blob/0901e0e2bd24d41e173d9ce95333971985ea60bb/src/tribler-gui/tribler_gui/__init__.py#L6), another in [`CoreExceptionHandler`](https://github.com/Tribler/tribler/blob/0901e0e2bd24d41e173d9ce95333971985ea60bb/src/tribler-core/tribler_core/components/reporter/exception_handler.py#L43) that is imported by [`run_tribler`](https://github.com/Tribler/tribler/blob/0901e0e2bd24d41e173d9ce95333971985ea60bb/src/run_tribler.py#L17) and so presents in GUI process as well. The reporter that's `init` method is called in [`init_sentry_reporter`](https://github.com/Tribler/tribler/blob/0901e0e2bd24d41e173d9ce95333971985ea60bb/src/run_tribler.py#L33) is not the same as [`gui_sentry_reporter`](https://github.com/Tribler/tribler/blob/0901e0e2bd24d41e173d9ce95333971985ea60bb/src/tribler-gui/tribler_gui/__init__.py#L6).

This PR switches to use the single instance of SentryReporter, `default_sentry_reporter`. It fixes the problem, and crash reports can be delivered to Sentry again.

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -8,8 +8,7 @@ import logging.config
 import os
 import sys
 
-from tribler.core.components.reporter.exception_handler import default_core_exception_handler
-from tribler.core.sentry_reporter.sentry_reporter import SentryStrategy
+from tribler.core.sentry_reporter.sentry_reporter import default_sentry_reporter, SentryStrategy
 from tribler.core.sentry_reporter.sentry_scrubber import SentryScrubber
 
 logger = logging.getLogger(__name__)
@@ -46,7 +45,7 @@ def init_sentry_reporter():
     as a URL for sending sentry's reports while a Tribler client running in
     test mode
     """
-    sentry_reporter = default_core_exception_handler.sentry_reporter
+    sentry_reporter = default_sentry_reporter
     from tribler.core.version import sentry_url, version_id
     test_sentry_url = sentry_reporter.get_test_sentry_url()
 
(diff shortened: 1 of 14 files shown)


kozlovsky:
Close in favor of #6976

PR #7008, opened 2022-08-17
Rewrite window geometry restoring

### LINKED ISSUE #6973: Tribler interface seems wonky
**Describe the bug**
I use both Linux and Windows. But, the Windows version seems wonky, and I cannot find the setting button. 

**To Reproduce**
1. Go to Start menu
2. Search on Tribler
3. double click on Tribler icon
4. I cannot find the setting

**Expected behavior**
The UI should should similar with the Linux version

**Screenshots**
image

 - OS: Windows 10
 - Tribler's version 7.11

**Additional context**

drew2a:
This PR fixes #6973 by rewriting the logic of restoring Tribler window position and size.
Now if any part of the Tribler window is outside of screen boundaries, the window moves to the position (0, 0)

### DIFF
--- a/src/tribler/gui/tribler_window.py
+++ b/src/tribler/gui/tribler_window.py
@@ -12,9 +12,8 @@ from PyQt5.QtCore import (
     QCoreApplication,
     QDir,
     QObject,
-    QPoint,
     QRect,
-    QStringListModel,
+    QSize, QStringListModel,
     QTimer,
     QUrl,
     Qt,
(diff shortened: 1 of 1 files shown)


src/tribler/gui/tribler_window.py
```diff
@@ -395,6 +383,39 @@ def sigint_handler(*_):
...
+            pos = self.gui_settings.value("pos", self.pos())
+            self._logger.info(f'Restored window position: {pos}')
+
+            window_geometry = QRect(pos, size)
+            union: QRect = screen_geometry | window_geometry
+            delta_x = union.width() - screen_geometry.width()
+            delta_y = union.height() - screen_geometry.height()
+            actual_position = QPoint(
+                pos.x() + delta_x,
+                pos.y() + delta_y
```
kozlovsky, line 411: This logic is incorrect. `delta_x` and `delta_y` are always non-negative, as the union is always bigger than `screen_geometry`.
When `delta_x` and `delta_y` are added to `pos`, `pos` moves right and down and never left and up. If the right and bottom part of the Tribler window is placed out of the screen, each time the Tribler is restarted, the Tribler windows will move further and further right and down until it is entirely out of sight.

kozlovsky (changes requested):

src/tribler/gui/tribler_window.py
drew2a: Thank you, you are right. 

I've removed the last commit.

kozlovsky (approved):

PR #7021, opened 2022-08-29
Fix/stop core on components startup exception

### LINKED ISSUE #7013: [7.12.1-RC1] CoreConnectTimeoutError due to a crash in REST API Component
If a component's `start` method throws an exception, it should lead to a Tribler Core shutdown for most components. However, with the current logic, the Core shutdown does not happen immediately; it is postponed until Tribler GUI can connect to Tribler Core and retrieve information about the Core exception to show the user a feedback dialog.

It works well for most components, but the REST API component is a special case. If the REST API component crashes with an exception, Tribler GUI will not be able to communicate with Tribler Core and retrieve the exception. Because of this, we can see the following behavior in case of exception in the REST API component:

1. Tribler GUI launches Tribler Core;
2. Tribler Core session starts components;
3. REST API component crashes with an exception;
4. `CoreExceptionHandler` remembers `unreported_error` to report it to Tribler GUI later when it can connect. The error is suppressed, and the Core continues to work.
5. EventRequestManager cannot connect to Tribler Core, as REST API is not working, and eventually raises `CoreConnectTimeoutError`.
6. The resulted error has no information about the actual reason for the error, as GUI was not able to connect to the Core

Examples:
https://sentry.tribler.org/organizations/tribler/issues/1352/events/daaa2a25ff0d46fd843a60bcbe83e4bf/
https://sentry.tribler.org/organizations/tribler/issues/1352/events/ec7fb3e184d3491ebd5d0c30f0476de1/

To reproduce the problem, you can switch to a release branch and raise an exception from the `start` method of `RESTComponent`.

To fix the issue, Core should shut down in case of an exception in `RESTComponent`. Then GUI will see that the Core is finished and can take the correct error traceback from the Core stderr. The simplest way to do this is to not add special handling to `RESTComponent` and to immediately shut down the Core on any exception in any component with the `tribler_should_stop_on_component_error` flag equal to True (currently, this flag leads to Core shutdown, but only after GUI successfully connected to Core).

The last Core stderr output does not necessarily contain the correct Core error traceback, as Core can write additional messages to stderr during the shutdown. To reliable retrieve the error, it is necessary to concatenate at least several latest stderr writes.

kozlovsky:
Fixes #7013. Stops Tribler Core on Session component's startup exception. Returns non-zero exit code from the Core process.

### DIFF
--- a/src/tribler/core/sentry_reporter/sentry_reporter.py
+++ b/src/tribler/core/sentry_reporter/sentry_reporter.py
@@ -280,7 +280,7 @@ class SentryReporter:
         Returns:
             the event that has been saved in `_before_send` method
         """
-        self._logger.info(f"Event from exception: {exception}")
+        self._logger.debug(f"Event from exception: {exception}")
 
         if not exception:
             return {}
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -211,11 +211,14 @@ class CoreManager(QObject):
     @staticmethod
     def format_error_message(exit_code: int, exit_status: int) -> str:
         message = f"The Tribler core has unexpectedly finished with exit code {exit_code} and status: {exit_status}."
-        try:
-            string_error = os.strerror(exit_code)
-        except ValueError:
-            # On platforms where strerror() returns NULL when given an unknown error number, ValueError is raised.
-            string_error = 'unknown error number'
+        if exit_code == 1:
+            string_error = "Application error"
+        else:
+            try:
+                string_error = os.strerror(exit_code)
+            except ValueError:
+                # On platforms where strerror() returns NULL when given an unknown error number, ValueError is raised.
+                string_error = 'unknown error number'
         message += f'\n\nError message: {string_error}'
         return message
 
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -2,6 +2,7 @@ import logging
 import os
 import re
 import sys
+from collections import deque
 from pathlib import Path
 from typing import Optional
 
@@ -14,6 +15,9 @@ from tribler.gui.event_request_manager import EventRequestManager
 from tribler.gui.exceptions import CoreCrashedError
 from tribler.gui.tribler_request_manager import TriblerNetworkRequest
 from tribler.gui.utilities import connect
+
+
+CORE_OUTPUT_DEQUE_LENGTH = 10
 
 
 class CoreManager(QObject):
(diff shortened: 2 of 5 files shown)


src/tribler/core/sentry_reporter/sentry_reporter.py
```diff
@@ -280,7 +280,7 @@ def event_from_exception(self, exception) -> Dict:
         Returns:
             the event that has been saved in `_before_send` method
         """
-        self._logger.info(f"Event from exception: {exception}")
+        self._logger.debug(f"Event from exception: {exception}")
```
kozlovsky, line 283: To avoid spamming logs with multiple copies of the same error traceback

src/tribler/gui/core_manager.py
```diff
@@ -47,8 +51,8 @@ def __init__(self, root_state_dir: Path, api_port: int, api_key: str, app_manage
         self.should_quit_app_on_core_finished = False
 
         self.use_existing_core = True
-        self.last_core_stdout_output: str = ''
-        self.last_core_stderr_output: str = ''
+        self.last_core_stdout_output: deque = deque(maxlen=CORE_OUTPUT_DEQUE_LENGTH)
+        self.last_core_stderr_output: deque = deque(maxlen=CORE_OUTPUT_DEQUE_LENGTH)
```
kozlovsky, line 55: The last write to stderr does not necessarily contain the actual the actual exception traceback, so we store several writes.

src/tribler/gui/core_manager.py
```diff
@@ -196,18 +202,24 @@ def kill_core_process_and_remove_the_lock_file(self):
...
+        else:
+            try:
+                string_error = os.strerror(exit_code)
+            except ValueError:
+                # On platforms where strerror() returns NULL when given an unknown error number, ValueError is raised.
+                string_error = 'unknown error number'
         message += f'\n\nError message: {string_error}'
-
-        quoted_output = re.sub(r'^', '> ', last_core_output, flags=re.MULTILINE)
-        message += f"\n\nLast core output:\n{quoted_output}"
```
kozlovsky, line 210: This PR moves the rendering of the last core output to `gui_error`, as we need it not only for `CoreCrashedError` but for other subclasses of CoreError as well, for example, for `CoreConnectTimeoutError`.

src/tribler/core/sentry_reporter/sentry_reporter.py
drew2a: NIT. As an alternative option, you can use `SentryReporter.ignore_logger()`

drew2a (approved):

xoriole (approved):

src/tribler/core/sentry_reporter/sentry_reporter.py
kozlovsky: As I understand it, `SentryReporter.ignore_logger()` affects what we send to Sentry, but does not affect our log files.

PR #7022, opened 2022-08-29
Add core stdout breadcrumb

drew2a:
This PR changes the process of collecting CORE core breadcrumbs.

Instead of collecting them on the core side, they are added to the GUI Sentry Reported as they appear.

"Normal" error: https://sentry.tribler.org/organizations/tribler/issues/1599
CoreCrashError: https://sentry.tribler.org/organizations/tribler/issues/1546

The purpose of this improvement is to investigate `CoreConnectionError` issues like https://github.com/Tribler/tribler/issues/6989

### DIFF
--- a/src/tribler/core/sentry_reporter/sentry_scrubber.py
+++ b/src/tribler/core/sentry_reporter/sentry_scrubber.py
@@ -47,6 +47,10 @@ class SentryScrubber:
 
         self._compile_re()
 
+    @staticmethod
+    def remove_breadcrumbs(event: Dict) -> Dict:
+        return delete_item(event, BREADCRUMBS)
+
     def _compile_re(self):
         """Compile all regular expressions."""
         for folder in self.home_folders:
--- a/src/tribler/core/sentry_reporter/tests/test_sentry_scrubber.py
+++ b/src/tribler/core/sentry_reporter/tests/test_sentry_scrubber.py
@@ -318,3 +318,19 @@ def test_scrub_list(scrubber):
 
     assert scrubber.scrub_entity_recursively(['/home/username/some/']) == ['/home/<highlight>/some/']
     assert 'username' in scrubber.sensitive_occurrences
+
+
+def test_remove_breadcrumbs():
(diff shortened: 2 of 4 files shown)


kozlovsky (changes requested):
Thanks for the PR! In my opinion, two things need to be fixed:
1. I think we should not add Core stdout output to breadcrumbs, as it contains INFO-level logs that are very spammy and will flood the sentry report with unimportant info. As a result, useful GUI-related breadcrumbs will be cut off, as we limit the number of breadcrumbs added to the report. Instead, we should add `last_core_stderr_output`, which is handled in the `on_core_stderr_read_ready` function below. If Core generates an exception, it will be in stderr.
2. The level for some log messages should be changed from `ERROR` to probably `WARNING` for two reasons:
    1. They do not contain a "real" error relevant to the crash we want to report via Sentry. Usually, we ignore them, disregarding the current `ERROR` level of messages.
    2. They potentially contain sensitive information.
    Specifically, I'm talking about the `on_tracker_error_alert` handler that sends an ERROR-level message to logs. It shows torrent names in breadcrumbs, which is not good for privacy. I think we should change the level of this log message to `WARNING`.

drew2a:
> I think we should not add Core stdout output to breadcrumbs, as it contains INFO-level logs that are very spammy and will flood the sentry report with unimportant info. As a result, useful GUI-related breadcrumbs will be cut off, as we limit the number of breadcrumbs added to the report. Instead, we should add last_core_stderr_output, which is handled in the on_core_stderr_read_ready function below. If Core generates an exception, it will be in stderr.

Well, we already put into breadcrumbs all kinds of spammy messages (info, warning, error) both from the Core and from the GUI:

image

drew2a:
> The level for some log messages should be changed from ERROR to probably WARNING for two reasons:

Maybe, but it is not a part of the current PR. 
This PR is dedicated to adding breadcrumbs from the Core stdout.

kozlovsky (approved):

PR #7049, opened 2022-09-12
Hide token balance value

### LINKED ISSUE #7048: Hide token balance statistics until the balance algorithm is fixed
The current formula for the token balance calculation leads to surprising results, where most Tribler users see a negative token balance even if they are "good citizens".

Only exit nodes can have a positive balance with the current Tribler implementation.

The balance the user sees should motivate him to some behavior that is desirable to a Tribler community. The current balance algorithm basically encourages a user to be an exit node. That contradicts the latest changes in Tribler when an option to become an exit node was removed to provide better safety for a user who does not understand the consequences of being an exit node.

For balance to be useful, it should use a different algorithm that encourages users to a helpful behavior. Probably it should be adding information about new popular torrents.

We should hide the token balance value until the balance algorithm is fixed to avoid user confusion.

kozlovsky:
This PR fixes #7048

### DIFF
--- a/src/tribler/gui/tribler_window.py
+++ b/src/tribler/gui/tribler_window.py
@@ -536,7 +536,7 @@ class TriblerWindow(QMainWindow):
 
         self.top_menu_button.setHidden(False)
         self.left_menu.setHidden(False)
-        self.token_balance_widget.setHidden(False)
+        # self.token_balance_widget.setHidden(False)  # restore it after the token balance calculation is fixed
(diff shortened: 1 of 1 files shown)


drew2a (approved):

devos50 (approved):

ichorid:
I've already lost count of how many times we turned the thing on and off :facepalm: 
I bet it will be turned on in 7.13 and turned off again in 7.14 :rofl:

Issue #7002, opened 2022-08-17
"DatabaseError: database disk image is malformed" after Tribler was terminated during the upgrade process


kozlovsky:
When a new version of Tribler copies data files from a folder for a previous version (like, "7.11" -> "7.12"), the process of copying is not atomic. If Tribler is terminated during that process, the database file can be in a half-copied state.

If Tribler is launched after it was terminated during the file copying, the Tribler sees that the new state folder is already present and tries to use it. That results in the following error:

```python
ErrorHandler.gui_error(): Traceback (most recent call last):
  File "C:\dev\tribler\src\tribler\gui\upgrade_manager.py", line 34, in run
    self._upgrade_state_dir(
  File "C:\dev\tribler\src\tribler\run_tribler_upgrader.py", line 40, in upgrade_state_dir
    upgrader.run()
  File "C:\dev\tribler\src\tribler\core\upgrade\upgrade.py", line 95, in run
    self.upgrade_pony_db_8to10()
  File "C:\dev\tribler\src\tribler\core\upgrade\upgrade.py", line 307, in upgrade_pony_db_8to10
    if not database_path.exists() or get_db_version(database_path) >= 10:
  File "C:\dev\tribler\src\tribler\core\upgrade\db8_to_db10.py", line 247, in get_db_version
    cursor.execute('SELECT value FROM MiscData WHERE name == "db_version"')
sqlite3.DatabaseError: database disk image is malformed
```

To prevent this type of error, the process of copying files to a new state folder should be atomic - files should be copied to a folder with a temporal name, and then the folder should be renamed to the proper name after all files are copied successfully.

PR #7055, opened 2022-09-20
Provide atomicity of state dir copying

kozlovsky:
This PR fixes the source of #7002 by providing atomicity for state directory copying. To achieve that, it first copies the state into a temporary dir and then renames it to a correct name after the copying process is finished.

Also, it adds logging to the version manager, making it possible to see upgrade-related events in error reports.

### DIFF
--- a/src/tribler/core/upgrade/version_manager.py
+++ b/src/tribler/core/upgrade/version_manager.py
@@ -94,6 +94,7 @@ class TriblerVersion:
         self.last_launched_at = last_launched_at
         self.root_state_dir = root_state_dir
         self.directory = self.get_directory()
+        self.tmp_copy_directory = self.get_tmp_copy_directory()
         self.prev_version_by_time = None
         self.prev_version_by_number = None
         self.can_be_copied_from = None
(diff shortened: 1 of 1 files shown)


src/tribler/core/upgrade/version_manager.py
```diff
@@ -103,6 +108,9 @@ def __repr__(self):
     def get_directory(self):
         return self.root_state_dir / ('%d.%d' % self.major_minor)
 
+    def get_tmp_copy_directory(self):
+        return self.root_state_dir / ('%d.%d_tmp_copy' % self.major_minor)
```
drew2a, line 112: NIT: Maybe it is a bit safer to generate some random postfix for this temporary directory
```suggestion
        major = self.major_minor[0]
        minor = self.major_minor[1]
        temporary_dir_name = f'{major}.{minor}_tmp_copy_{random.randrange(100)}'
        return self.root_state_dir / temporary_dir_name
```

drew2a (approved):

xoriole (approved):

PR #7060, opened 2022-09-20
Add "subclass match" for session.get_instance()

drew2a:
This PR changes the method `Session.get_instance()` in such a way that it returns not only instances with direct class match, but instances with subclass match.

This will allow crawlers to make changes to components and communities without making an actual change to the Tribler.

Example:

```python
import asyncio
from pathlib import Path

from tribler.core.components.ipv8.ipv8_component import Ipv8Component
from tribler.core.components.key.key_component import KeyComponent
from tribler.core.components.metadata_store.metadata_store_component import MetadataStoreComponent
from tribler.core.components.tag.tag_component import TagComponent
from tribler.core.config.tribler_config import TriblerConfig
from tribler.core.utilities.tiny_tribler_service import TinyTriblerService


class CustomTagComponent(TagComponent):
    # Apply all necessary changes to the Tag Component here:
    ...


class Service(TinyTriblerService):
    def __init__(self, working_dir):
        super().__init__(TriblerConfig(state_dir=working_dir),
                         working_dir=working_dir,
                         components=[KeyComponent(), Ipv8Component(), MetadataStoreComponent(), CustomTagComponent()])


def run_tribler():
    service = Service(
        working_dir=Path(__file__).parent / 'tribler',
    )

    loop.create_task(service.start_tribler())
    try:
        loop.run_forever()
    finally:
        loop.run_until_complete(loop.shutdown_asyncgens())
        loop.close()


if __name__ == "__main__":
    run_tribler()

```


src/tribler/core/components/tests/test_base_component.py
```diff
@@ -141,3 +122,36 @@ class ComponentB(Component):
         assert isinstance(component_b, NoneComponent)
         assert isinstance(component_b.any_attribute, NoneComponent)
         assert isinstance(component_b.any_attribute.any_nested_attribute, NoneComponent)
+
+
+def test_get_instance_direct_match(tribler_config: TriblerConfig):
+    class AncestorOfB(ComponentB):
```
xoriole, line 128: - Since this is a subclass, It would better to rename it to something like `DerivedB` or `ChildrenOfB` or `SubClassOfB`. I consider ancestor is not be a right word.
- Same applies for other examples in this test file.

xoriole (changes requested):
Nice PR!
Love it 👍

src/tribler/core/components/tests/test_base_component.py
drew2a: Thank you for the suggestion. Applied.

xoriole (approved):

src/tribler/core/components/session.py
```diff
@@ -51,7 +51,20 @@ async def __aexit__(self, *_):
...
+        if direct_match := self.components.get(comp_cls):
+            return direct_match
+
+        # try to find a subclass match
+        candidates = {c for c in self.components if issubclass(c, comp_cls)}
+
+        if not candidates:
+            return None
+        if len(candidates) >= 2:
+            raise KeyError(f'Found more that two subclasses for the class {comp_cls}')
```
kozlovsky, line 64: **nitpick:** It is better to use a new exception class (something like `MultipleComponentsFound`) inherited from `ComponentError`.

With the KeyError exception, it sometimes becomes possible that the method is expected to raise KeyError in some situations (like, multiple components found), but also it raises the same KeyError exception for a totally unrelated reason (incorrect access to some list or dict), and it goes unnoticed.

Also, the standard [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError) exception has the following description: "Raised when a mapping (dictionary) key is not found in the set of existing keys." There we have a different situation when multiple objects are found, so using KeyError looks incorrect.

kozlovsky (approved):

src/tribler/core/components/session.py
drew2a: Thank you for this suggestion. It has been implemented.

PR #7062, opened 2022-09-23
Fix/7053

### LINKED ISSUE #7053: Tribler becomes unresponsive with a large `tags.db`
**Describe the bug**
When using Tribler with a large `tags.db` (mine is 4.58 GB), Tribler becomes unresponsive fast, and everything blocks when doing searches/browsing through channels.

**To Reproduce**
1. Search for some content, like 'ubuntu' and scroll down. I noticed that the first request to `/search` completes pretty fast, but scrolling down in the list of search results and loading the subsequent batches is _very_ slow and blocks the Tribler core entirely.

**Expected behavior**
Tribler should remain responsive with a large `tags.db`.

 - OS: macOS Monterey 12.3
 - Tribler's version [e.g. 7.10.0]: latest commit on `main` branch.

drew2a:
This PR fixes #7053 by changing the approach of picking up random tags for the gossip.

The approach was suggested by @kozlovsky and can be described as:

```python
    def _get_random_tag_operation(self) -> Optional[Entity]:
        """ Get random tag operation from the DB.
        """
        max_id = get_max(self.instance.TorrentTagOp, column_name='id')
        if not max_id:
            return None

        random_id = random.randint(1, max_id)
        return self.instance.TorrentTagOp.get(id=random_id)
```

On a big dataset (~5Gb) this change gives x6000-x7000 increase of speed for the method `get_tags_operations_for_gossip`.


kozlovsky (approved):
Looks good; I like it!

Some minor notes:

1. Instead of `_get_random_tag_operation` it is also possible to use `TorrentTagOp.select_random(1)` , but the difference is insignificant.

2. It may be good to add a test for a case when the database is not empty but no operations satisfy the condition. The result should be empty.

PR #7094, opened 2022-10-20
Fix pyenv GitHub Action

### LINKED ISSUE #7091: pyenv GitHub Action seems to be broken
[pyenv GitHub Action ](https://github.com/Tribler/tribler/blob/main/.github/actions/pyenv/action.yml)seems to be broken.
See: https://github.com/Tribler/tribler/actions/runs/3288237714/jobs/5418361820

image

The workaround is to add `invalidate-cache: true` https://github.com/Tribler/tribler/pull/7090/commits/200c562242e43c1d849a298fbcbba086a86c0148 for corresponding yml files

drew2a:
This PR fixes #7091 by adding `ref_name` of a GitHub event as a `custom_cache_key_element` for the `restore-virtualenv` action.

This should help to isolate caches between branches.

### DIFF
--- a/.github/actions/pyenv/action.yml
+++ b/.github/actions/pyenv/action.yml
@@ -17,7 +17,7 @@ inputs:
     required: false
 
   invalidate-cache:
-    default: 'true' # temporary set to true. see: https://github.com/Tribler/tribler/issues/7091
+    default: 'false'
     description: 'Force create a virtualenv'
     required: false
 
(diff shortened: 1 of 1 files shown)


drew2a:
I have not received a response from  @xoriole which is why I requested reviews from @devos50 and @kozlovsky 🤝

devos50 (approved):

xoriole (approved):

PR #7101, opened 2022-10-21
Refactoring for the New Version Dialog

### LINKED ISSUE #7077: Tribler is asking to upgrade version too frequently.
> New version available
> Version 7.12.1 of Tribler is available.Do you want to visit the website to download the newest version?
> [Ignore] [Later] [OK]

I press "Later", expecting to see the notice again in a day or a week. But it seems to postpone it only by an hour or so.

drew2a:
This PR adds a refactoring for #7077

image

The actions for the buttons are the following:
1. Ignore: ignore this version at all. During the next Tribler run, all notifications about this version will be silently ignored.
2. Later: ignore the version check for the current Tribler run.
3. Ok: go to the Tribler website. The next version check will be performed 6 hours later.

### DIFF
--- a/src/tribler/gui/i18n/ru_RU.ts
+++ b/src/tribler/gui/i18n/ru_RU.ts
@@ -632,7 +632,7 @@ If unsure, press &apos;No&apos;. You will be able to remove those directories fr
     </message>
     <message>
         <location filename="../tribler_window.py" line="649"/>
-        <source>Version %s of Tribler is available.Do you want to visit the website to download the newest version?</source>
+        <source>Version %s of Tribler is available. Do you want to visit the website to download the newest version?</source>
         <translation>Доступна новая версия Tribler: %s. Перейти к веб-сайту разработчиков чтобы загрузить её?</translation>
     </message>
     <message>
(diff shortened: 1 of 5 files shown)


vi:
Note: also hacked up a simpler pull request about it: #7102

drew2a:
@vi thank you for contributing! 
I've implemented an almost identical solution with a small refactoring.

:)

drew2a:
I've accepted and merged https://github.com/Tribler/tribler/pull/7102
In this PR I left only refactoring.

src/tribler/gui/i18n/ru_RU.ts
```diff
@@ -632,7 +632,7 @@ If unsure, press &apos;No&apos;. You will be able to remove those directories fr
     </message>
     <message>
         <location filename="../tribler_window.py" line="649"/>
-        <source>Version %s of Tribler is available.Do you want to visit the website to download the newest version?</source>
+        <source>Version %s of Tribler is available. Do you want to visit the website to download the newest version?</source>
```
xoriole, line 635: Maybe you can add the correct translation for this message as well here.

xoriole (approved):

src/tribler/gui/i18n/ru_RU.ts
drew2a: It is already here. I can confirm that it is correct.

Issue #7132, opened 2022-10-31
[Tests] Timeout in Windows GUI tests


drew2a:
https://github.com/Tribler/tribler/actions/runs/3363436019/jobs/5576574735 in https://github.com/Tribler/tribler/pull/7129

```python
============================= test session starts =============================
platform win32 -- Python 3.8.10, pytest-7.1.2, pluggy-1.0.0 -- c:\users\runneradmin\.virtualenvs\.venv\scripts\python.exe
cachedir: .pytest_cache
Using --randomly-seed=1
rootdir: D:\a\tribler\tribler, configfile: pytest.ini
plugins: anyio-3.3.4, Faker-14.1.0, looptime-0.2, aiohttp-1.0.4, asyncio-0.18.3, freezegun-0.4.2, mock-3.7.0, profiling-1.7.0, randomly-3.11.0, timeout-2.1.0
asyncio: mode=auto
timeout: 300.0s
timeout method: thread
timeout func_only: False
collecting ... collected 47 items

src/tribler/gui/tests/test_core_manager.py::test_on_core_finished_raises_error PASSED [  2%]
src/tribler/gui/tests/test_core_manager.py::test_decode_raw_core_output PASSED [  4%]
src/tribler/gui/tests/test_core_manager.py::test_on_core_read_ready_os_error_suppressed PASSED [  6%]
src/tribler/gui/tests/test_core_manager.py::test_format_error_message PASSED [  8%]
src/tribler/gui/tests/test_core_manager.py::test_on_core_finished_calls_quit_application PASSED [ 10%]
src/tribler/gui/tests/test_core_manager.py::test_on_core_stdout_read_ready PASSED [ 12%]
src/tribler/gui/tests/test_core_manager.py::test_on_core_stderr_read_ready PASSED [ 14%]
src/tribler/gui/tests/test_error_handler.py::test_gui_info_type_in_handled_exceptions PASSED [ 17%]
src/tribler/gui/tests/test_error_handler.py::test_core_should_stop PASSED [ 19%]
src/tribler/gui/tests/test_error_handler.py::test_core_error PASSED      [ 21%]
src/tribler/gui/tests/test_error_handler.py::test_gui_core_connect_timeout_error PASSED [ 23%]
src/tribler/gui/tests/test_error_handler.py::test_gui_error_suppressed PASSED [ 25%]
src/tribler/gui/tests/test_error_handler.py::test_gui_error_tribler_stopped PASSED [ 27%]
[...]

PR #7141, opened 2022-11-02
Fix Windows GUI tests

drew2a:
This PR fixes https://github.com/Tribler/tribler/issues/7132 

Refs:
* https://github.com/pyvista/setup-headless-display-action
* https://github.com/jurplel/install-qt-action

### DIFF
--- a/.github/workflows/guitest.yml
+++ b/.github/workflows/guitest.yml
@@ -14,7 +14,7 @@ on:
         required: false
 
       enable-profiling:
-        default: true
+        default: false
         type: boolean
         required: false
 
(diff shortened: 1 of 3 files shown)


kozlovsky (approved):

.github/workflows/guitest.yml
```diff
@@ -14,7 +14,7 @@ on:
         required: false
 
       enable-profiling:
-        default: true
+        default: false
```
drew2a, line 17: I suppose that the problem is solved, and that is why profiling for GUI tests has been disabled.

.github/workflows/guitest.yml
```diff
@@ -25,10 +25,14 @@ jobs:
       fail-fast: false
       matrix: ${{fromJson(inputs.matrix)}}
 
+    defaults:
+      run:
+        shell: bash
```
drew2a, line 30: This is the fix for Windows GUI tests. 
Without it, the wrong env had been used and this led to the execution of the whole test suite.

src/tribler/gui/tests/conftest.py
```diff
@@ -8,7 +8,7 @@ def pytest_addoption(parser):
 
 def pytest_collection_modifyitems(config, items):
     for item in items:
-        item.add_marker(pytest.mark.timeout(30))
+        item.add_marker(pytest.mark.timeout(60))
```
drew2a, line 11: This is the fix that should solve the timeout problem.

kozlovsky (approved):

Issue #7146, opened 2022-11-04
[Build]  macOS-10.15 environment is deprecated


drew2a:
GitHub deprecated `macOS-10.15`:

```
[PR: .github#L1](https://github.com/Tribler/tribler/pull/7144/files#annotation_5744268134)
The macOS-10.15 environment is deprecated, consider switching to macos-11(macos-latest), macos-12 instead. For more details see https://github.com/actions/virtual-environments/issues/5583
```

Therefore we can not use our macOS builder with GitHub Actions.
https://github.com/Tribler/tribler/blob/main/build/mac/makedist_macos.sh

Even though we can still build binaries on Jenkins, we must sooner or later update our macOS build scripts to be able to build Tribler on modern macOS.

As far as I remember, problems with moving to a newer macOS version lie in these lines

https://github.com/Tribler/tribler/blob/fcea30e6d75818f0a38c68e4eb3cb64f97c1c831/build/mac/makedist_macos.sh#L106-L107

<img width="757" alt="image" src="https://user-images.githubusercontent.com/13798583/200021537-4e4e8a70-770b-4f2b-ba4b-2f21769da4aa.png">

drew2a:
I can rewrite the current macOS build script by script that uses [InstallBuilder](https://installbuilder.com/) (I already did this at my last job).

We can get a  "special" license:

https://installbuilder.com/purchase.html
> In addition to our standard commercial licenses, we offer special licenses for academic purposes and open-source projects. We also offer small business discounts to companies with less than 5 employees. For more information, contact us at [sales@installbuilder.com](mailto:sales@installbuilder.com).

drew2a:
Other options: 
* https://stackoverflow.com/questions/11487596/making-macos-installer-packages-which-are-developer-id-ready
* http://chromasoft.blogspot.com/2010/02/building-dmg-installer-for-mac-simple.html

PR #7147, opened 2022-11-04
Disable macOS build

drew2a:
This PR related to #7146


xoriole (commented):
Instead of commenting out the Github action job, the following options would be preferable:
1) Updating the if-statement that evaluates to false to the job, or
1) Removing it altogether since we have in git

drew2a:
The macOS section has been removed.

xoriole (approved):

PR #7156, opened 2022-11-09
Remove mocks of stdout and stderr

### LINKED ISSUE #7140: [Tests] AssertionError: Expected 'print' to be called once. Called 7551 times.
https://github.com/Tribler/tribler/actions/runs/3379038407/jobs/5610070695#step:4:256 in https://github.com/Tribler/tribler/pull/7139

```python
src/tribler/core/logger/tests/test_logger.py F

=================================== FAILURES ===================================
_________________________ test_setup_logging_exception _________________________

__wrapped_mock_method__ = <function NonCallableMock.assert_called_once_with at 0x7f2518bbb9d0>
args = (<MagicMock name='print' id='139796884481792'>, 'Error in loading logger config. Using default configs. ', 'ZeroDivisionError: ')
kwargs = {'file': <MagicMock name='stderr' id='139796840503808'>}
__tracebackhide__ = True
msg = "Expected 'print' to be called once. Called 7551 times.\nCalls: [call('Traceback (most recent call last):\\n', file=<M...logger config. Using default configs. ', 'ZeroDivisionError: ', file=<MagicMock name='stderr' id='139796840503808'>)]."
__mock_self = <MagicMock name='print' id='139796884481792'>
actual_args = ('Error in loading logger config. Using default configs. ', 'ZeroDivisionError: ')
actual_kwargs = {'file': <MagicMock name='stderr' id='139796840503808'>}
introspection = '', @py_assert2 = None, @py_assert1 = None

    def assert_wrapper(
        __wrapped_mock_method__: Callable[..., Any], *args: Any, **kwargs: Any
    ) -> None:
        __tracebackhide__ = True
        try:
>           __wrapped_mock_method__(*args, **kwargs)


../../../.virtualenvs/.venv/lib/python3.8/site-packages/pytest_mock/plugin.py:414: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <MagicMock name='print' id='139796884481792'>
args = ('Error in loading logger config. Using default configs. ', 'ZeroDivisionError: ')
kwargs = {'file': <MagicMock name='stderr' id='139796840503808'>}
msg = "Expected 'print' to be called once. Called 7551 times.\nCalls: [call('Traceback (most recent call last):\\n', file=<M...logger config. Using default configs. ', 'ZeroDivisionError: ', file=<MagicMock name='stderr' id='139796840503808'>)]."

    def assert_called_once_with(self, /, *args, **kwargs):
        """assert that the mock was called exactly once and that that call was
        with the specified arguments."""
        if not self.call_count == 1:
            msg = ("Expected '%s' to be called once. Called %s times.%s"
                   % (self._mock_name or 'mock',
                      self.call_count,
                      self._calls_repr()))
>           raise AssertionError(msg)
E           AssertionError: Expected 'print' to be called once. Called 7551 times.
```

drew2a:
This PR fixes #7140


src/tribler/core/logger/tests/test_logger.py
```diff
@@ -18,23 +17,13 @@ def test_get_logger_config_path():
...
+    config_path = tmp_path / 'non_existent.conf'
+    assert not config_path.exists()
 
-    logger.info.assert_called_once_with(
-        "Load logger config: app_mode=<app-mode>, " "config_path=<config-path>, dir=<log-dir>"
-    )
-    print_.assert_called_once_with("Logger config not found in <config-path>. Using default configs.", file=stderr)
-    basic_config.assert_called_once_with(level=logging.INFO, stream=stdout)
+    setup_logging('', Path(''), config_path)
+    assert not mocked_safe_load.called
```
kozlovsky, line 26: The new version of the test does not look clear to me. It is not entirely obvious why `yaml.safe_load` is tested, and is it enough to test just `yaml.safe_load` (and will continue to be enough in the future)

I think it is better to test wether `logging.basicConfig(...)` and/or `ogging.config.dictConfig(...)` were called, something like this:

```python
@patch('tribler.core.logger.logger.logger')
@patch('builtins.print')
@patch('logging.basicConfig')
@patch('logging.config.dictConfig')
def test_setup_logging_no_config(dict_config: Mock, basic_config: Mock, print_: Mock, logger: Mock):
    config_path = MagicMock()
    config_path.exists.return_value = False
    config_path.__str__.return_value = '<config-path>'

    setup_logging('<app-mode>', '<log-dir>', config_path)

    logger.info.assert_called_once_with(
        "Load logger config: app_mode=<app-mode>, " "config_path=<config-path>, dir=<log-dir>"
    )
    print_.assert_called_once_with("Logger config not found in <config-path>. Using default configs.", file=sys.stderr)
    basic_config.assert_called_once_with(level=logging.INFO, stream=sys.stdout)
    dict_config.assert_not_called()
```

src/tribler/core/logger/tests/test_logger.py
```diff
@@ -67,30 +56,14 @@ def test_setup_logging(logger: Mock, dict_config: Mock, yaml_safe_load: Mock):
...
-    config_path.exists.return_value = True
-
-    setup_logging('<app-mode>', log_dir, config_path)
-
-    logger.info.assert_called_once_with(
-        "Load logger config: app_mode=<app-mode>, " "config_path=<config-path>, dir=<log-dir>"
-    )
-    print_.assert_called_once_with('Error in loading logger config. Using default configs. ', 'ZeroDivisionError: ', file=stderr)
-    basic_config.assert_called_once_with(level=logging.INFO, stream=stdout)
+    assert mocked_format_error_description.called
```
kozlovsky, line 66: Same here, it is not obvious why the test for `format_error_description` call is necessary and enough (now and in the future). It looks more safe to test wether `basicConfig(...)/dictConfig(...)` were called, something like this:

```python
@patch('tribler.core.logger.logger.logger')
@patch('builtins.print')
@patch('logging.basicConfig')
@patch('logging.config.dictConfig')
def test_setup_logging_exception(dict_config: Mock, basic_config: Mock, print_: Mock, logger: Mock):
    error = ZeroDivisionError()

    log_dir = MagicMock()
    log_dir.__str__.return_value = '<log-dir>'
    log_dir.exists.return_value = True
    log_dir.joinpath.side_effect = error

    config_path = MagicMock()
    config_path.__str__.return_value = '<config-path>'
    config_path.exists.return_value = True

    setup_logging('<app-mode>', log_dir, config_path)

    logger.info.assert_called_once_with(
        "Load logger config: app_mode=<app-mode>, " "config_path=<config-path>, dir=<log-dir>"
    )
    print_.assert_called_once_with('Error in loading logger config. Using default configs. ', 'ZeroDivisionError: ',
                                   file=sys.stderr)
    basic_config.assert_called_once_with(level=logging.INFO, stream=sys.stdout)
    dict_config.assert_not_called()
```

kozlovsky (changes requested):

src/tribler/core/logger/tests/test_logger.py
drew2a: Both versions of tests (the current and the proposed) do the same. They check what branch of execution is executed in this scenario.

For checking this I use a mock of `yaml.safe_load` function and you offered to check loggers. 

Both versions are unclear, but mine is shorter. 

I could add comments with an explanation of the test.

drew2a: Please take a look at this most recent example of fragility tests based on logging mocks: 
https://github.com/Tribler/tribler/actions/runs/3437151109/jobs/5731538805#step:8:64

kozlovsky: > Please take a look at this most recent example of fragility tests based on logging mocks

I agree that sometimes logging-related tests may be brittle. Often it means that the test should be rewritten in a more robust way. For example, in the test you are referring to, the correct assertion can be written as:

```python
    for rt in record_tuples:
        assert rt in caplog.record_tuples
```

IMO, the fact that sometimes incorrectly written logging-related checks are brittle does not mean that testing any logging-related functionality should be excluded, especially if the test itself directly relates to logging. `caplog` is an official fixture provided by `pytest` and not a "hack".

src/tribler/core/logger/tests/test_logger.py
kozlovsky: Ideally, the test should test two different things:
* that the code reaches the correct final state or performs the expected actions;
* that all code paths are covered.

In the case of  `test_setup_logging_exception`, checking just the call of `format_error_description` is intended to test that the code visited a specific code path. But the test does not check the desired result, namely that  `logging.basicConfig(...)` was called as a result of the action. It is possible that later, after some new refactoring, `tribler.core.logger.logger.format_error_description` will still be called, but `logging.basicConfig(...)` will not be called anymore. The test that only intends to achieve coverage does not check this. It makes the codebase more brittle, as it can stop performing the expected actions while still passing tests.

src/tribler/core/logger/tests/test_logger.py
kozlovsky: > Both versions are unclear, but mine is shorter.

The PR version that checks `yaml.safe_load` is intended to check that the code follows the specific code path, but it does not check the final desired state, namely that the `logger.basicConfig(...)` is called.

Tests that check code paths and not the desired state make the codebase more brittle, as it is possible that after some new refactoring, the code path is still executed (to the checked point), but the desired state ceases to be reached.

So, while it is possible to have coverage-only-oriented tests, it is not usually a good idea, in my opinion.

src/tribler/core/logger/tests/test_logger.py
drew2a: I've added comments and checks on the `logging.basicConfig(...)` call.

src/tribler/core/logger/tests/test_logger.py
drew2a: I've added comments and checks on the `logging.basicConfig(...)` call.

kozlovsky (approved):

PR #7160, opened 2022-11-11
Limit the log size

### LINKED ISSUE #7158: /tmp .log file filling up disk
Tribler 7.11.0

I am getting multi-gigabyte log files in the /tmp directory. Eventually they fill up the disk and crash my system (Ubuntu 18.04)

Is there any way to turn off these log files?
Limit their size?

drew2a:
This PR fixes #7158

1. It limits log files to 100KB each with a backup count equal to `1`. That means the overall log size will be 800KB.
$Core+GUI =(2 \cdot Error+2 \cdot Info) + (2 \cdot Error+2 \cdot Info) = 2 \cdot (2 \cdot 100KB+2 \cdot 100KB) = 800 KB  $

2. It completely removes the `resources.log` logic as it is no longer used.
3. It adds a migration for removing all old logs.

Refs:
* https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -14,7 +14,7 @@ on:
         required: false
 
       enable_profiling:
-        default: true
+        default: false
         type: boolean
         required: false
 
--- a/src/tribler/core/components/resource_monitor/implementation/core.py
+++ b/src/tribler/core/components/resource_monitor/implementation/core.py
@@ -1,10 +1,8 @@
-import os
 import time
 from collections import deque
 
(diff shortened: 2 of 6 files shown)


.github/workflows/pytest.yml
```diff
@@ -14,7 +14,7 @@ on:
         required: false
 
       enable_profiling:
-        default: true
+        default: false
```
drew2a, line 17: It seems that profiling adds a bit of instability to the test suite. Disable it by default.

xoriole (approved):

PR #7162, opened 2022-11-14
Find the running core port and use it in GUI

xoriole:
This PR adds a PortChecker class that finds the closest port opened by a process identified by given pid and the base port. 
A callback can also be set which is triggered when the port is detected.
```python3
    port_checker = PortChecker(pid, base_port, callback)
    port_checker.start_checking()

    # The detected port can be retrieved as:
    port_checker.detected_port
```
The port checker is used by the CoreManager to find the core API port. 

When the port is detected, request manager and event manager are updated so they can operate with the correct port. 

Further, it also adds two environment variables `TRIBLER_API_PORT_INITIAL` and `TRIBLER_API_PORT_DETECTED` with values initial and detected api port values. This is so that these variables will be made available in sentry report if there is a crash.

Fixes https://github.com/Tribler/tribler/issues/7137

### DIFF
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -14,7 +14,8 @@ from tribler.gui import gui_sentry_reporter
 from tribler.gui.app_manager import AppManager
 from tribler.gui.event_request_manager import EventRequestManager
 from tribler.gui.exceptions import CoreCrashedError
-from tribler.gui.tribler_request_manager import ShutdownRequest
+from tribler.gui.port_checker import PortChecker
+from tribler.gui.tribler_request_manager import ShutdownRequest, request_manager
 from tribler.gui.utilities import connect
 
 
@@ -56,6 +57,8 @@ class CoreManager(QObject):
         self.last_core_stderr_output: deque = deque(maxlen=CORE_OUTPUT_DEQUE_LENGTH)
 
         connect(self.events_manager.core_connected, self.on_core_connected)
+
+        self.port_checker = None
 
     def on_core_connected(self, _):
         if self.core_finished:
@@ -122,7 +125,17 @@ class CoreManager(QObject):
     def on_core_started(self):
         self.core_started = True
         self.core_running = True
+        self.start_port_checker()
         self.events_manager.connect(reschedule_on_err=True)  # retry until REST API is ready
+
+    def start_port_checker(self):
+        core_process_id = self.core_process.processId()
+        self.port_checker = PortChecker(core_process_id, self.api_port, self.port_checker_callback)
+        self.port_checker.start_checking()
+
+    def port_checker_callback(self, detected_port):
+        request_manager.port = detected_port
+        self.events_manager.update_port(detected_port)
 
     def on_core_stdout_read_ready(self):
         if self.app_manager.quitting_app:
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -136,6 +136,8 @@ class CoreManager(QObject):
     def port_checker_callback(self, detected_port):
         request_manager.port = detected_port
         self.events_manager.update_port(detected_port)
+        os.environ['TRIBLER_API_PORT_INITIAL'] = str(self.api_port)
+        os.environ['TRIBLER_API_PORT_DETECTED'] = str(detected_port)
 
     def on_core_stdout_read_ready(self):
         if self.app_manager.quitting_app:
--- a/src/tribler/gui/event_request_manager.py
+++ b/src/tribler/gui/event_request_manager.py
@@ -37,9 +37,10 @@ class EventRequestManager(QNetworkAccessManager):
 
     def __init__(self, api_port, api_key, error_handler):
         QNetworkAccessManager.__init__(self)
-        url = QUrl("http://localhost:%d/events" % api_port)
-        self.request = QNetworkRequest(url)
-        self.request.setRawHeader(b'X-Api-Key', api_key.encode('ascii'))
+        self.api_port = api_port
+        self.api_key = api_key
+        self.request = None
+
         self.start_time = time.time()
         self.connect_timer = QTimer()
         self.current_event_string = ""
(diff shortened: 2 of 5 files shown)


src/tribler/gui/port_checker.py
```diff
@@ -0,0 +1,56 @@
...
+        self.checker_timer.setSingleShot(True)
+        connect(self.checker_timer.timeout, self.on_timeout)
+        self.checker_timer.start(CHECK_INTERVAL_MS)
+
+    def on_timeout(self):
+        detected_port = self.detected_port()
+        if detected_port:
+            self.checker_timer.stop()
+            self.callback(detected_port)
+        elif self.checker_timer:
```
kozlovsky, line 34: There were some print statements in the previous version of `PortChecker`. With them, it was possible to see that `PortChecker` never stops on Windows. First, it detects an incorrect `processId` (as I described above), and then it continues to check open ports again and again even after `Tribler` has already started and GUI has connected to Core.

So, I have two suggestions here:
1. It is useful to have some logging output from `PortChecker` so we can detect that it runs longer than expected in case the process ID was detected incorrectly.
2. After the GUI process is able to connect to Core successfully, it should explicitly stop `PortChecker`.

src/tribler/gui/port_checker.py
```diff
@@ -0,0 +1,56 @@
...
+            self.checker_timer.start(CHECK_INTERVAL_MS)
+
+    def detected_port(self):
```
kozlovsky, line 37: So, as I described above, to detect the PID value, we can do here:
```python
            process_checker = ProcessChecker(self.root_state_dir)
            pid = process_checker._get_pid_from_lock()
```
(for this, we need to pass `root_state_dir` as a constructor argument for `PortChecker`)

src/tribler/gui/event_request_manager.py
```diff
@@ -37,9 +37,10 @@ class EventRequestManager(QNetworkAccessManager):
 
     def __init__(self, api_port, api_key, error_handler):
         QNetworkAccessManager.__init__(self)
-        url = QUrl("http://localhost:%d/events" % api_port)
-        self.request = QNetworkRequest(url)
-        self.request.setRawHeader(b'X-Api-Key', api_key.encode('ascii'))
+        self.api_port = api_port
+        self.api_key = api_key
+        self.request = None
```
kozlovsky, line 42: Nitpick: it is possible to create a request right now and skip the `if self.request is None` check later:
```python
        self.request = self.create_request()
```

src/tribler/gui/port_checker.py
```diff
@@ -0,0 +1,56 @@
...
+        self.process_id = pid
+        self.checker_timer = QTimer()
+        self.checker_timer.setSingleShot(True)
+        connect(self.checker_timer.timeout, self.on_timeout)
+        self.checker_timer.start(CHECK_INTERVAL_MS)
+
+    def on_timeout(self):
+        detected_port = self.detected_port()
+        if detected_port:
+            self.checker_timer.stop()
```
kozlovsky, line 32: Is it necessary to call `checker_timer.stop()` here if the timer is single-shot?

src/tribler/gui/port_checker.py
```diff
@@ -0,0 +1,56 @@
...
+    def setup_with_pid(self, pid):
+        self.process_id = pid
+        self.checker_timer = QTimer()
+        self.checker_timer.setSingleShot(True)
+        connect(self.checker_timer.timeout, self.on_timeout)
+        self.checker_timer.start(CHECK_INTERVAL_MS)
+
+    def on_timeout(self):
+        detected_port = self.detected_port()
+        if detected_port:
```
kozlovsky, line 31: It is not necessary right now, but I'd like to store the detected port as a `PortChecker` attribute. It may be useful for debugging purposes, and also, it looks a bit safer if we add a check: if the port was already detected, do not call the callback again if, for some reason, the method was executed again.

kozlovsky (changes requested):

xoriole:
retest this please

xoriole:
retest this please

src/tribler/gui/core_manager.py
```diff
@@ -122,8 +125,20 @@ def start_tribler_core(self):
     def on_core_started(self):
         self.core_started = True
         self.core_running = True
+        self.start_port_checker()
         self.events_manager.connect(reschedule_on_err=True)  # retry until REST API is ready
 
+    def start_port_checker(self):
+        core_process_id = self.core_process.processId()
```
kozlovsky, line 132: Can you add a comment for future developers that when using `venv` on Windows, the PID value returned from the `QProcess.processId` is incorrect, as it is the PID of the `venv` wrapper process and not the PID of the Core process itself? Something like:

```suggestion
        # When using venv on Windows, the Core process PID value returned by QProcess is incorrect,
        # as it is the PID of the venv wrapper process and not the PID of the actual Core process.
        # In that case, PortChecker finds nothing, and the default port is used.
        core_process_id = self.core_process.processId()
```

src/tribler/gui/core_manager.py
```diff
@@ -118,6 +125,7 @@ def start_tribler_core(self):
         connect(self.core_process.finished, self.on_core_finished)
         self._logger.info(f'Start Tribler core process {sys.executable} with arguments: {core_args}')
         self.core_process.start(sys.executable, core_args)
+        self.port_checker.setup_with_pid(self.core_process.processId())
```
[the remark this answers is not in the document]
kozlovsky: As it turns out, the reason it does not work correctly on my PC is venv. Venv creates a separate `python.exe` wrapper that calls the actual Python inside. So, the PID of the Core process that `QProcess` returns is the PID of the venv wrapper and not of the actual Core process. As most Tribler users use the compiled version of Tribler, they should not experience this problem. Nevertheless, we should be aware of this venv behavior.

On my machine, the PR is working, but not in the way it may be expected. As the actual PID of the Core process is different from what `QProcess` returns, the PortChecker cannot find the correct port, so it never calls the `EventRequestManager.update_port` method. Instead, the timeout of `EventRequestManager.connect_timer` forces the request with the default port value, which turns out to be the correct one.

kozlovsky (changes requested):

kozlovsky:
Based on the ideas and code from this PR, PR #7251 was implemented with a different way of passing information about the opened port

Issue #7164, opened 2022-11-14
[Tests] test_close_process_errors. ProcessLookupError: [Errno 3] assume no such process 


drew2a:
Error occurred here: https://github.com/Tribler/tribler/actions/runs/3463390209/jobs/5783556952#step:7:1451 in https://github.com/Tribler/tribler/pull/7163 
```python
src\tribler\core\utilities\tests\test_process_checker.py .......F

================================== FAILURES ===================================
__________________________ test_close_process_errors __________________________

self = <psutil._pswindows.Process object at 0x0000017ACE94C3C0>
args = ('inet',), kwargs = {}

    @functools.wraps(fun)
    def wrapper(self, *args, **kwargs):
        try:
>           return fun(self, *args, **kwargs)

c:\users\runneradmin\.virtualenvs\.venv\lib\site-packages\psutil\_pswindows.py:679: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <psutil._pswindows.Process object at 0x0000017ACE94C3C0>, kind = 'inet'

    @wrap_exceptions
    def connections(self, kind='inet'):
>       return net_connections(kind, _pid=self.pid)

c:\users\runneradmin\.virtualenvs\.venv\lib\site-packages\psutil\_pswindows.py:1005: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

kind = 'inet', _pid = 6096

    def net_connections(kind, _pid=-1):
        """Return socket connections.  If pid == -1 return system-wide
        connections (as opposed to connections opened by one process only).
        """
        if kind not in conn_tmap:
            raise ValueError("invalid %r kind argument; choose between %s"
                             % (kind, ', '.join([repr(x) for x in conn_tmap])))
        families, types = conn_tmap[kind]
>       rawlist = cext.net_connections(_pid, families, types)
E       ProcessLookupError: [Errno 3] assume no such process (originated from psutil_pid_is_running)

c:\users\runneradmin\.virtualenvs\.venv\lib\site-packages\psutil\_pswindows.py:359: ProcessLookupError

During handling of the above exception, another exception occurred:

checker = <tribler.core.utilities.process_checker.ProcessChecker object at 0x0000017ACECD8E50>
[...]

Issue #7134, opened 2022-11-01
[Tests] Timeout in `pytest` tests


drew2a:
It is the most common bug now.

Ubuntu.
https://github.com/Tribler/tribler/actions/runs/3367858797/jobs/5585934301 in https://github.com/Tribler/tribler/pull/7112

```python
============================= test session starts ==============================
platform linux -- Python 3.8.14, pytest-7.1.2, pluggy-1.0.0
Using --randomly-seed=3036501063
rootdir: /home/runner/work/tribler/tribler, configfile: pytest.ini
plugins: freezegun-0.4.2, aiohttp-1.0.4, timeout-2.1.0, anyio-3.3.4, mock-3.7.0, randomly-3.11.0, asyncio-0.18.3, Faker-[14](https://github.com/Tribler/tribler/actions/runs/3367858797/jobs/5585934301#step:5:15).1.0, looptime-0.2
timeout: 300.0s
timeout method: signal
timeout func_only: False
asyncio: mode=auto
collected 1147 items

src/tribler/core/components/metadata_store/restapi/tests/test_channels_endpoint.py . [  0%]
........................................                                 [  3%]
src/tribler/core/components/bandwidth_accounting/tests/test_bandwidth_endpoint.py . [  3%]
.                                                                        [  3%]
src/tribler/core/components/libtorrent/tests/test_download_config.py ... [  4%]
......                                                                   [  4%]
src/tribler/core/components/socks_servers/socks5/tests/test_server.py .. [  4%]
.....                                                                    [  5%]
src/tribler/core/components/restapi/rest/tests/test_debug_endpoint.py .. [  5%]
.........s.                                                              [  6%]
src/tribler/core/components/bandwidth_accounting/tests/test_bandwidth_accounting_component.py . [  6%]
                                                                         [  6%]
src/tribler/core/sentry_reporter/tests/test_sentry_tools.py ............ [  7%]
..............                                                           [  8%]
src/tribler/core/components/knowledge/community/tests/test_operations_requests.py . [  8%]
....                                                                     [  9%]
src/tribler/core/components/metadata_store/tests/test_channel_metadata.py . [  9%]
...............................                                          [ 11%]
src/tribler/core/components/socks_servers/socks5/tests/test_udp_connection.py . [ 11%]
.                                                                        [ 12%]
src/tribler/core/components/socks_servers/tests/test_socks_servers_component.py . [ 12%]
                                                                         [ 12%]
[...]

Issue #7111, opened 2022-10-27
[Tests] TestRemoteQueryCommunity.test_process_rpc_query_match_many OSError: [Errno 9] Bad file descriptor


drew2a:
Error occurred here https://github.com/Tribler/tribler/pull/7110 in https://github.com/Tribler/tribler/actions/runs/3337026709/jobs/5522868389
```python
=================================== FAILURES ===================================
__________ TestRemoteQueryCommunity.test_process_rpc_query_match_many __________
../../../.virtualenvs/.venv/lib/python3.8/site-packages/asynctest/case.py:303: in run
    self._unset_loop()
../../../.virtualenvs/.venv/lib/python3.8/site-packages/asynctest/case.py:207: in _unset_loop
    self.loop.close()
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/asyncio/unix_events.py:58: in close
    super().close()
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/asyncio/selector_events.py:92: in close
    self._close_self_pipe()
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/asyncio/selector_events.py:100: in _close_self_pipe
    self._ssock.close()
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/socket.py:500: in close
    self._real_close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <socket.socket [closed] fd=-1, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
_ss = <class '_socket.socket'>

    def _real_close(self, _ss=_socket.socket):
        # This function should not reference any globals. See issue #808164.
>       _ss.close(self)
E       OSError: [Errno 9] Bad file descriptor

/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/socket.py:494: OSError
----------------------------- Captured stdout call -----------------------------
[PID:1691] 2022-10-27 11:37:43,476 - INFO - BasicRemoteQueryCommunity(13) - Init. Settings: None.
[PID:1691] 2022-10-27 11:37:43,500 - INFO - BasicRemoteQueryCommunity(13) - Init. Settings: None.
[...]

Issue #7063, opened 2022-09-23
[Tests] sqlite3.OperationalError: disk I/O error


drew2a:
os: ubuntu-latest
https://github.com/Tribler/tribler/actions/runs/3111710669/jobs/5044306829#step:4:245

```python
__________________ ERROR at setup of test_get_channels_peers ___________________

func = <function SQLiteProvider.set_transaction_mode at 0x7ff83fea0dc0>
provider = <pony.orm.dbproviders.sqlite.SQLiteProvider object at 0x7ff83dbe0520>
args = (<sqlite3.Connection object at 0x7ff83f832300>, <pony.orm.core.SessionCache object at 0x7ff83dbe0ee0>)
kwargs = {}
dbapi_module = <module 'sqlite3' from '/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/sqlite3/__init__.py'>
should_retry = False

    @decorator
    def wrap_dbapi_exceptions(func, provider, *args, **kwargs):
        dbapi_module = provider.dbapi_module
        should_retry = False
        try:
            try:
                if provider.dialect != 'SQLite':
                    return func(provider, *args, **kwargs)
                else:
                    provider.local_exceptions.keep_traceback = True
>                   try: return func(provider, *args, **kwargs)

../../../.virtualenvs/.venv/lib/python3.8/site-packages/pony/orm/dbapiprovider.py:55: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

provider = <pony.orm.dbproviders.sqlite.SQLiteProvider object at 0x7ff83dbe0520>
connection = <sqlite3.Connection object at 0x7ff83f832300>
cache = <pony.orm.core.SessionCache object at 0x7ff83dbe0ee0>

    @wrap_dbapi_exceptions
    def set_transaction_mode(provider, connection, cache):
        assert not cache.in_transaction
        if cache.immediate:
            provider.acquire_lock()
        try:
            cursor = connection.cursor()
    
            db_session = cache.db_session
            if db_session is not None and db_session.ddl:
                cursor.execute('PRAGMA foreign_keys')
                fk = cursor.fetchone()
                if fk is not None: fk = fk[0]
                if fk:
                    sql = 'PRAGMA foreign_keys = false'
                    if core.local.debug: log_orm(sql)
                    cursor.execute(sql)
[...]

PR #7186, opened 2022-11-17
Add `pytest-sentry` for pytest runs

drew2a:
This PR helps us to collect statistics about flaky tests.

> Let's say you have a testsuite with some flaky tests that randomly break your CI build due to network issues, race conditions or other stuff that you don't want to fix immediately. The known workaround is to retry those tests automatically, for example using [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures).
One concern against plugins like this is that they just hide the bugs in your testsuite or even other code. After all your CI build is green and your code probably works most of the time.
pytest-sentry tries to make that choice a bit easier by tracking flaky test failures in a place separate from your build status. Sentry is already a good choice for keeping tabs on all kinds of errors, important or not, in production, so let's try to use it in testsuites too.

For the `pytest-sentry` it is necessary to specify `PYTEST_SENTRY_DSN`. 
There are several options for implementing it in our GitHub Actions. We discussed all of them within the `dev` team and came to a conclusion to store the `PYTEST_SENTRY_DSN` as a repository secret. 
For the secret to be accessible from `!PR.yml` it is necessary to change the environment in which this workflow runs from unprivileged to privileged (from `pull_request` to `pull_request_target`). To protect access from the forks PR to the main secrets, the `gatekeeper` job was introduced. 
It is only possible to pass the gatekeeper for users with the "write" (or higher) permission.

In the case, the user doesn't have write access, the gatekeeper will not allow the user to perform some of the tests:

image

Related:
* https://github.com/Tribler/tribler/issues/7164
* https://github.com/Tribler/tribler/issues/7134
* https://github.com/Tribler/tribler/issues/7111
* https://github.com/Tribler/tribler/issues/7063

Refs: 
* https://github.com/getsentry/pytest-sentry
* https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests
* https://blog.gitguardian.com/github-actions-security-cheat-sheet/
* https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
* https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
* https://github.com/lannonbr/repo-permission-check-action
* https://docs.github.com/en/actions/using-workflows/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow

### DIFF
--- a/.github/workflows/!PR.yml
+++ b/.github/workflows/!PR.yml
@@ -1,7 +1,7 @@
 name: PR
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - synchronize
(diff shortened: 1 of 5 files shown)


xoriole (approved):
:+1:

Issue #7177, opened 2022-11-16
Tribler does not close if left idle for a few hours


xoriole:
If Tribler is left idle for a few hours on the home screen, it does not shut down properly when trying to close it.

**The request from GUI to Core to shutdown is canceled, causing Tribler to be stuck at closing.**

```
[PID:605050] 2022-11-16 14:34:37,316 - INFO - CoreManager(187) - Stopping Core manager
[PID:605050] 2022-11-16 14:34:37,317 - INFO - CoreManager(203) - Sending shutdown request to Tribler Core
[PID:605050] 2022-11-16 14:34:37,320 - WARNING - root(252) - Request from GUI to Core was canceled: http://localhost:52195/shutdown
[PID:605050] 2022-11-16 14:34:37,321 - WARNING - CoreManager(205) - Re-sending shutdown request to Tribler Core
```

OS: Ubuntu 22.04
Branch: main

The full logs are attached here: 
[2022-11-16-tribler.logs.txt](https://github.com/Tribler/tribler/files/10022526/2022-11-16-tribler.logs.txt)

PR #7202, opened 2022-11-23
Skip shutdown request clearing on Tribler shutdown

xoriole:
Tribler can get stuck on shutdown if the request to shutdown is canceled already at the GUI level. Issue and logs are present in https://github.com/Tribler/tribler/issues/7177

This PR excludes the shutdown request from being cleared while shutting down making sure that the shutdown request is sent to the Core. 

Fixes https://github.com/Tribler/tribler/issues/7177

### DIFF
--- a/src/tribler/gui/tribler_request_manager.py
+++ b/src/tribler/gui/tribler_request_manager.py
@@ -86,7 +86,7 @@ class TriblerRequestManager(QNetworkAccessManager):
 
     def clear(self, skip_shutdown_request=True):
         for req in list(self.requests_in_flight.values()):
-            if skip_shutdown_request and req.is_shutdown_request():
+            if skip_shutdown_request and isinstance(req, ShutdownRequest):
                 continue
             req.cancel_request()
 
@@ -190,11 +190,6 @@ class TriblerNetworkRequest(QObject):
         # Pass the newly created object to the manager singleton, so the object can be dispatched immediately
         request_manager.add_request(self)
 
-    def is_shutdown_request(self):
-        return "shutdown" in self.url \
-               and self.method == "PUT" \
-               and self.priority == QNetworkRequest.HighPriority
-
     def on_finished(self, request):
         if self.reply is None:
             # Fix for rare race condition where reply is overwritten by e.g. previous call to destruct
@@ -273,3 +268,8 @@ class TriblerFileDownloadRequest(TriblerNetworkRequest):
             priority=QNetworkRequest.LowPriority,
             decode_json_response=False,
         )
+
+
+class ShutdownRequest(TriblerNetworkRequest):
+    def __init__(self, *args, **kwargs):
+        super().__init__("shutdown", *args, **kwargs, method="PUT", priority=QNetworkRequest.HighPriority)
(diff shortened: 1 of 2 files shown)


src/tribler/gui/tribler_request_manager.py
```diff
@@ -188,6 +190,11 @@ def __init__(
         # Pass the newly created object to the manager singleton, so the object can be dispatched immediately
         request_manager.add_request(self)
 
+    def is_shutdown_request(self):
+        return "shutdown" in self.url \
+               and self.method == "PUT" \
+               and self.priority == QNetworkRequest.HighPriority
```
drew2a, line 196: NIT: Is it necessary to check a method and a priority to detect the shutdown request? I'm just wondering if the priority changes, then this request still is the shutdown request, right?

What if we derive the class `ShutdownRequest` from `TriblerNetworkRequest` like:
```python
class ShutdownRequest(TriblerNetworkRequest):
        super().__init__("shutdown", *args, **kwargs, method="PUT", priority=QNetworkRequest.HighPriority)
```

And use it in the cycle above:

```python
    def clear(self, skip_shutdown_request=True):
        for req in list(self.requests_in_flight.values()):
            if skip_shutdown_request and isinstance(req, ShutdownRequest):
                continue
            req.cancel_request()
```

In this case, the check is much more correct and future-proof.
The downside of this approach is that you should change all `shutdown` usage from

```python

                TriblerNetworkRequest("shutdown", shutdown_request_processed, method="PUT",
                                      priority=QNetworkRequest.HighPriority, on_cancel=send_shutdown_request)
```

to 

```python
                ShutdownRequest(shutdown_request_processed, on_cancel=send_shutdown_request)```

Comment (xoriole) [reply src/tribler/gui/tribler_request_manager.py:196]: It is a good idea, will make the changes.

**Review by drew2a — APPROVED** [review]

**Review by kozlovsky — APPROVED** [review]
First, I like the changes in this PR. Instead of re-sending a canceled shutdown request, now the code does not cancel the request, and it looks much more logical than before.

With this, I do not fully understand the origin of the problem in issue #7177. The log that I see looks ok to me, and the code behaves as expected:

```python
[PID:605050] 2022-11-16 14:34:37,316 - INFO - CoreManager(187) - Stopping Core manager
[PID:605050] 2022-11-16 14:34:37,317 - INFO - CoreManager(203) - Sending shutdown request to Tribler Core
[PID:605050] 2022-11-16 14:34:37,320 - WARNING - root(252) - Request from GUI to Core was canceled: http://localhost:52195/shutdown
[PID:605050] 2022-11-16 14:34:37,321 - WARNING - CoreManager(205) - Re-sending shutdown request to Tribler Core
```

The shutdown request was canceled, and then it was re-sent again. It looks contrived but should work, so the log lines look as expected.

But are these the last lines of the log file? Then it is strange. If the shutdown request was not re-sent properly, I expect additional lines in the log, at least from the working Core.

And, if Core ignored the re-sent shutdown request, then the problem may be on the Core side.

So, if we have a problem with the shutdown, I'm not sure these changes should fix it, but the code definitely looks better than before.

xoriole:
As discussed offline, this PR might not fix the issue but still improves the code so I'll merge it.

Issue #7187, opened 2022-11-18
A download in GUI fluctuates between "0%" and "Seeding".


vi:
Sometimes a download in Triber 7.12.0 or 7.12.1 is shows alternately as if it were not really started ("Downloading", "0%") and the actual progress ("Downloading" with actual percentage or "Seeding" if it is finished). Pathenthesised number of seeds and peers also fluctuates between different nonzero numbers for broken and correct sets of download status data.

Both upload and download byte counters are small, but nonzero in the broken stats set. Both counters are also slowly rise over time, as if it is very slowly downloading the file second time. 

Individual counter for a file may stay "0%" even though whole download is being shown "100%".

Clicking around tabs showing download details also resulted in this:

> An error occurred during the request:
> index out of range
> [Close]

(`ERROR <tribler_request_manager:225> root.on_finished(): REST API error: {'error': 'index out of range'}`)

drew2a:
Could be related: https://github.com/Tribler/tribler/issues/5930

drew2a:
Could be a duplicate of 
- #5617

qstokkink:
This should be resolved in the new Tribler 8.0 web-based front end.

PR #7205, opened 2022-11-29
Refactoring of the `TriblerNetworkRequest`

drew2a:
Related to https://github.com/Tribler/tribler/issues/7187

The main changes in this PR are the following:
1. It fixes small logic mistakes in methods `TriblerNetworkRequest.on_finish()` and `TriblerNetworkRequest.cancel()`
2. It removes automatic request addition in the constructor of a `TriblerNetworkRequest`. This change opens an opportunity to write correct tests for `TriblerNetworkRequest`.

Minor changes:
1. Renamed and replaced `TriblerNetworkRequest` and `TriblerRequestManager`.
2. Extended logging.

The PR looks huge but it is mostly because the way how a `TriblerNetworkRequest` should be added to  `TriblerRequestManager` has been changed.

From
```python
Request()
```

To
```python
request = Request()
manager.add(request)
```


ichorid:
For the love of Guido, why not just `manager.add(Request())`?! :exploding_head:

drew2a:
@ichorid 
> For the love of Guido, why not just manager.add(Request())?! 🤯

There are always several ways of writing code.

I used this example above because it better describes the actual change.

This is
```python
        request = Request(
            endpoint="ipv8/asyncio/drift",
            on_finish=self.on_ipv8_health_enabled,
            data={
                "enable": True
            },
            method=Request.PUT
        )

        request_manager.add(request)
```

better than:


```python
        request_manager.add(
            Request(
                endpoint="ipv8/asyncio/drift",
                on_finish=self.on_ipv8_health_enabled,
                data={
                    "enable": True
                },
                method=Request.PUT
            )
        )
```

ichorid:
Even in cases like this?
```python
        request = Request("statistics/tribler", self.on_tribler_statistics)
        request_manager.add(request)
```
vs
```python
        request_manager.add(Request("statistics/tribler", self.on_tribler_statistics))
```

To me, the second one [reads almost like natural text](https://en.wikipedia.org/wiki/Literate_programming#Order_of_human_logic,_not_that_of_the_compiler).

drew2a:
> Even in cases like this?

Yes.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        if message := error.get('message'):
+            return message
+
+        return json.dumps(d)
+
+    def clear(self, skip_shutdown_request=True):
+        for req in list(self.active_requests.values()):
+            if skip_shutdown_request and isinstance(req, ShutdownRequest):
+                continue
+            req.cancel_request()
```
kozlovsky, line 119: Forgot to rename: it should be `req.cancel()` now

src/tribler/gui/widgets/downloadspage.py
```diff
@@ -459,12 +494,16 @@ def on_export_download(self, checked):
...
+
         selected_item = self.selected_items[:1]
         if action == 0 and selected_item:
             filename = self.dialog.dialog_widget.dialog_input.text()
-            TriblerFileDownloadRequest(
+            FileDownloadRequest(
                 f"downloads/{selected_item[0].download_info['infohash']}/torrent",
-                lambda data: self.on_export_download_request_done(filename, data),
+                on_finish,
             )
```
kozlovsky, line 508: Forgot to call `request_manager.add(request)`. This is an example that a new API can be a bit error-prone, as the call of `request_manager.add` can be easily forgotten.

For that reason, I suggest adding lightweight methods to `RequestManager`: `get`, `post`, `put`, etc., that create a corresponding request object and immediately add it to the request manager. With these methods, it would be impossible to forget to call `request_manager.add`. As a side benefit, it makes the codebase more concise again.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+    pyqt signal is fired when the response data is ready.
+    """
+
+    window = None
+
+    def __init__(self, limit: int = 50, timeout_interval: int = 15):
+        QNetworkAccessManager.__init__(self)
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+        self.active_requests = {}
```
kozlovsky, line 35: `active_requests` is a dict in which a key and a value are always the same object, a `Request` instance. That means it actually should be `set` instead of `dict`.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        self.performed_requests: Dict[Request, int] = LimitedOrderedDict(limit=200)
```
kozlovsky, line 36: I pretty much like the implementation of `LimitedOrderedDict`; it looks pretty nice!

With this, the idea of keeping status codes as dictionary values looks a bit strange. It is much easier to understand the code that works with status codes if `status_code` is kept as a `request` attribute.

If `request.status_code` is converted to an attribute, then using `deque` again becomes the most straightforward way to store the latest requests.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+    def add(self, request: Request):
+        if len(self.active_requests) > self.limit:
+            self._drop_timed_out_requests()
+
+        self.active_requests[request] = request
+        self.performed_requests[request] = 0
+        request.manager = self
+        request.url = self._get_base_url() + request.endpoint
+        request.url += f'?{self._urlencode(request.url_params)}' if request.url_params else ''
+        self.logger.info(f'Request: {request}')
```
kozlovsky, line 54: It looks a bit nicer if the manager is assigned by calling a new `request.set_manager(...)` method. In this case, it becomes possible to move the URL-handling logic to the Request class. It looks clearer if the `Request` class itself handles its `url` attribute.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        return data
+
+    @staticmethod
+    def _urlencode_single(key, value):
+        utf8_key = quote_plus(str(key).encode('utf-8'))
+        # Convert bool values to ints
+        if isinstance(value, bool):
+            value = int(value)
+        utf8_value = quote_plus(str(value).encode('utf-8'))
+        return f"{utf8_key}={utf8_value}"
```
kozlovsky, line 146: Instead of these methods, it is possible to use the standard `urlencode` function with the same result. While doing this, it is necessary to ensure that `True`/`False` values are encoded as `1`/`0`, but it is easy to do.

src/tribler/gui/network/request/request.py
```diff
@@ -0,0 +1,126 @@
...
+        self.priority = priority
+        self.method = method
+        self.capture_errors = capture_errors
+        self.raw_response = raw_response
+        self.data: Optional[bytes] = data
+        self.raw_data = data
+        if isinstance(self.data, Dict):
+            self.data = json.dumps(data)
+        if isinstance(self.data, str):
+            self.data = self.data.encode('utf8')
```
kozlovsky, line 59: Here, `raw_data` is the original data provided with `Request`, and `data` is an encoded version of it.

Usually, the meaning of `data` and `raw_data` is the other way around. For example, this is what ChatGPT thinks about it:

> If a Request object has attributes `data` and `raw_data`, what can be the value of these attributes?

> The data attribute of a Request object is typically used to store parsed and processed data that has been extracted from the request, such as form data or JSON data. The raw_data attribute, on the other hand, would likely store the original, raw data that was sent in the request, such as the bytes of a file or a string of XML or JSON. The exact values that these attributes can hold would depend on the implementation of the Request object.

Therefore, I suggest swapping these attributes to avoid confusion for future programmers.

src/tribler/gui/network/request/request.py
```diff
@@ -0,0 +1,126 @@
...
+
+        connect(self.on_finished_signal, on_finish)
+        connect(self.on_cancel_signal, on_cancel)
+
+        self.reply: Optional[QNetworkReply] = None  # to hold the associated QNetworkReply object
+        self.manager: Optional[RequestManager] = None
+        self.url: str = ''
+
+        # Pass the newly created object to the manager singleton, so the object can be dispatched immediately
```
kozlovsky, line 68: This comment is obsolete and should be removed

src/tribler/gui/network/request/request.py
```diff
@@ -0,0 +1,126 @@
...
+    PATCH = 'PATCH'
+    DELETE = 'DELETE'
+
+    # This signal is called if we receive some real reply from the request
+    # and if the user defined a callback to call on the received data.
+    # We implement the callback as a signal call and not as a direct callback
+    # because we want the request object be deleted independent of what happens
+    # during the callback call.
+    on_finished_signal = pyqtSignal(object)
+    on_cancel_signal = pyqtSignal()
```
kozlovsky, line 30: `on_cancel_signal` is obsolete and can be removed without any harm. It was used originally to re-send the shutdown request in case it was canceled. But the current logic avoids canceling the shutdown request, so it is unnecessary to re-send it now.

src/tribler/gui/network/request_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        request.reply = self.sendCustomRequest(qt_request, request.method.encode("utf8"), buf)
+        buf.setParent(request.reply)
+
+        connect(request.reply.finished, request._on_finished)  # pylint: disable=protected-access
+
+    def remove(self, request: Request):
+        self.active_requests.pop(request, None)
+
+        if request.reply:
+            request.reply.deleteLater()
```
kozlovsky, line 78: It looks strange that `RequestManager` handles the `reply` attribute of the `Request` object here. It looks better if the `Request` object handles its attribute itself. Therefore, I suggest to move these logic to `Request._delete`

src/tribler/gui/widgets/channelcontentswidget.py
```diff
@@ -133,14 +131,19 @@ def on_channel_committed(self, response):
             self.update_labels()
 
     def commit_channels(self, checked=False):  # pylint: disable=W0613
-        TriblerNetworkRequest("channels/mychannel/0/commit", self.on_channel_committed, method='POST')
+        request = Request(
+            endpoint="channels/mychannel/0/commit",
+            on_finish=self.on_channel_committed,
+            method=Request.POST
+        )
+        request_manager.add(request)
```
kozlovsky, line 139: The explicit creation of `Request` object looks a bit verbose, and (as I showed for `FileDownloadRequest`) it can be error-prone. Adding simple helper methods to `RequestManager` can make the code more concise and less error-prone:

```python
request_manager.post("channels/mychannel/0/commit", on_finish=self.on_channel_committed)
```

src/tribler/gui/network/request/file_download_request.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+from PyQt5.QtNetwork import QNetworkRequest
+
+from tribler.gui.network.request.request import Request
+
+
+class FileDownloadRequest(Request):
```
kozlovsky, line 8: I think `FileDownloadRequest` is unnecessary, as it is possible to do the same with a base `Request` class

src/tribler/gui/widgets/downloadspage.py
kozlovsky, line 508: It is not necessary to use a separate `FileDownloadRequest` class as the same can be expressed with the base Request object:
```python
request_manager.get(f"downloads/{selected_item[0].download_info['infohash']}/torrent",
                    on_finish, priority=QNetworkRequest.LowPriority, raw_response=True)

```

src/tribler/gui/network/request/shutdown_request.py
```diff
@@ -0,0 +1,8 @@
+from PyQt5.QtNetwork import QNetworkRequest
+
+from tribler.gui.network.request.request import Request
+
+
+class ShutdownRequest(Request):
```
kozlovsky, line 6: Instead of having a separate `ShutdownRequest` class, I suggest adding a `cancellable` attribute to the base `Request` class. This makes code more generic, and allows to make `tribler.gui.network.request` a module instead of a package

kozlovsky (changes requested):

kozlovsky (approved):

PR #7208, opened 2022-11-30
Add outputs for the gatekeeper job

drew2a:
Related to https://github.com/Tribler/tribler/pull/6716


drew2a:
Unfortunately, it is not possible to implement it this way.
https://github.com/orgs/community/discussions/26636

PR #7211, opened 2022-12-01
Add check on empty folder

### LINKED ISSUE #7041: Request Error while pressing the "download" button
**Describe the bug**
I see `Request Error` while pressing the "download" button (no path specified)

**To Reproduce**
1. Open the application
2. Search for "ubuntu"
3. Press the `download` button
4. Do not enter a path for the downloading
5. Press the download button

**Expected behavior**
I expected a message box that tells me "Please specify the path to the downloading" or the `download` button should be disabled until a user not specifies the path.

**Screenshots**

image

image

drew2a:
This PR fixes #7041

### DIFF
--- a/src/tribler/gui/dialogs/startdownloaddialog.py
+++ b/src/tribler/gui/dialogs/startdownloaddialog.py
@@ -275,6 +275,12 @@ class StartDownloadDialog(DialogContainer):
             )
         else:
             download_dir = self.dialog_widget.destination_input.currentText()
+            self.logger.info(f'Download folder: {download_dir}')
+            if not download_dir:
+                text = tr("Please specify the path to the download directory")
+                title = tr("The path is not specified")
+                ConfirmationDialog.show_message(self.dialog_widget, title, text, "OK")
+                return 
             is_writable, error = is_dir_writable(download_dir)
             if not is_writable:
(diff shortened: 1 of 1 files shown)


src/tribler/gui/dialogs/startdownloaddialog.py
```diff
@@ -275,6 +275,12 @@ def on_download_clicked(self, checked):
             )
         else:
             download_dir = self.dialog_widget.destination_input.currentText()
+            self.logger.info(f'Download folder: {download_dir}')
+            if not download_dir:
+                text = tr("Please specify the path to the download directory")
+                title = tr("The download directory is empty")
```
kozlovsky, line 281: The `text` is good, but the `title` is a bit confusing. It can be read as "the download directory exists but contain no files", while the intended meaning is "the path to the download directory is not specified". I think it is better to reword it.

kozlovsky (changes requested):

src/tribler/gui/dialogs/startdownloaddialog.py
drew2a: Changed to "The path is not specified"

kozlovsky (approved):

Issue #7065, opened 2022-09-26
[7.12.1] Exception in RESTComponent.start(): OSError:[Errno 10048] error while attempting to bind on address ('127.0.0.1', 52194)


drew2a:
```python
tribler.gui.utilities.CreationTraceback: , 
  File "run_tribler.py", line 100, in <module>, 
  File "tribler\gui\start_gui.py", line 77, in run_gui, 
  File "tribler\gui\upgrade_manager.py", line 137, in on_worker_finished, 
  File "tribler\gui\utilities.py", line 411, in trackback_wrapper, 
  File "tribler\gui\core_manager.py", line 117, in start_tribler_core, 
The above exception was the direct cause of the following exception:, 
Traceback (most recent call last):, 
  File "tribler\gui\utilities.py", line 414, in trackback_wrapper, 
  File "tribler\gui\utilities.py", line 411, in trackback_wrapper, 
  File "tribler\gui\core_manager.py", line 241, in on_core_finished, 
tribler.gui.exceptions.CoreCrashedError: The Tribler core has unexpectedly finished with exit code 1 and status: 0., 
Error message: Application error, 
Last Core output:, 
> apispec\ext\marshmallow\common.py:139: UserWarning: Multiple schemas resolved to the name SpeedTestRequest. The name has been modified. Either manually add each of the schemas with a different name or provide a custom schema_name_resolver., 
> , 
> Exception in RESTComponent.start(): OSError:[Errno 10048] error while attempting to bind on address ('127.0.0.1', 52194): only one usage of each socket address (protocol/network address/port) is normally permitted, 
> [PID:4548] 2022-09-26 09:32:48,723 - ERROR <base:233> RESTComponent.start(): Exception in RESTComponent.start(): OSError:[Errno 10048] error while attempting to bind on address ('127.0.0.1', 52194): only one usage of each socket address (protocol/network address/port) is normally permitted, 
> Traceback (most recent call last):, 
>   File "tribler\core\components\base.py", line 224, in start, 
>   File "tribler\core\components\restapi\restapi_component.py", line 128, in run, 
>   File "tribler\core\components\restapi\rest\rest_manager.py", line 129, in start, 
>   File "aiohttp\web_runner.py", line 121, in start, 
>   File "asyncio\base_events.py", line 1463, in create_server, 
[...]

PR #7212, opened 2022-12-02
[WIP] Add process_locker

kozlovsky:
Tribler has mechanisms that should ensure that only a single instance of a GUI/Core process is running:
* For GUI process [`QtSingleApplication`](https://github.com/Tribler/tribler/blob/58c6b9015e3226d86fc6148548f4fcb799b9f258/src/tribler/gui/single_application.py#L15) is responsible for that
* For the Core process, the uniqueness is controlled by ['ProcessChecker'](https://github.com/Tribler/tribler/blob/3428cd4b77c9c2e8c4e1eee8e3e5b886440d37ab/src/tribler/core/utilities/process_checker.py#L29) with the help of a `single_tribler_instance` context manager.

Unfortunately, both of these mechanisms are vulnerable to race conditions, so it is possible to run two Tribler instances simultaneously, which leads to bizarre bugs like [this](https://sentry.tribler.org/organizations/tribler/issues/1444/?project=7). The error itself does not provide any clue that the actual reason for it is the simultaneous start of two Tribler instances. I was able to reproduce it, but with a totally different error. That means that probably many other bizarre errors we sometimes observe have the same origin.

Because of this, we need to implement a proper way to ensure the uniqueness of Tribler processes, and this PR does this (and provides some other nice features that I'll describe a bit later)


drew2a:
Could be related: https://github.com/Tribler/tribler/issues/7065 https://github.com/Tribler/tribler/issues/7069

src/run_tribler.py
```diff
@@ -83,6 +83,8 @@ def init_boot_logger():
     logger.info(f'Root state dir: {root_state_dir}')
 
     api_port = os.environ.get('CORE_API_PORT')
+    api_port = int(api_port) if api_port else None
```
kozlovsky, line 86: The previous code version did not handle the situation when the `CORE_API_PORT` value is not set for the Core process.

src/tribler/core/components/reporter/exception_handler.py
```diff
@@ -111,6 +112,7 @@ def unhandled_error_observer(self, _, context):
                 should_stop=should_stop
             )
             self.logger.error(f"Unhandled exception occurred! {reported_error}\n{reported_error.long_text}")
+            process_locker.set_error(exc=exception)
```
kozlovsky, line 115: When an exception is caught, it is added to the description of the current process

src/tribler/core/exceptions.py
```diff
@@ -8,9 +8,6 @@
 class TriblerException(Exception):
     """Super class for all Tribler-specific Exceptions the Tribler Core throws."""
 
-    def __str__(self):
-        return str(self.__class__) + ': ' + Exception.__str__(self)
```
kozlovsky, line 12: There is no point in handling the conversion to `str` for some exceptions differently from all other exceptions

src/tribler/core/sentry_reporter/sentry_reporter.py
```diff
@@ -219,6 +220,10 @@ def send_event(self, event: Dict = None, post_data: Dict = None, sys_info: Dict
         reporter['events'] = extract_dict(sys_info, r'^(event|request)')
         reporter[SYSINFO] = {key: sys_info[key] for key in sys_info if key not in reporter['events']}
 
+        process_locker = get_global_process_locker()
+        if process_locker:
+            reporter['last_processes'] = [p.describe() for p in process_locker.get_last_processes()]
```
kozlovsky, line 225: An example of the section content in the error report:
```
[
GuiProcess(pid=3480, version='7.12.1-GIT', started='2022-12-21 13:34:06', api_port=52194, duration='0:00:33', exit_code=0), 
CoreProcess(pid=23584, gui_pid=3480, version='7.12.1-GIT', started='2022-12-21 13:34:13', api_port=52194, duration='0:00:25', exit_code=0), 
GuiProcess(canceled, pid=8116, version='7.12.1-GIT', started='2022-12-21 13:34:28', duration='0:00:00', exit_code=1, error='Tribler GUI application is already running'), 
GuiProcess(active, pid=15460, version='7.12.1-GIT', started='2022-12-21 13:36:38', api_port=52194, error='Core TriblerCoreTestException: Tribler Core Test Exception'), 
CoreProcess(pid=8616, gui_pid=15460, version='7.12.1-GIT', started='2022-12-21 13:36:45', api_port=52194, duration='0:00:17', exit_code=1, error='TriblerCoreTestException: Tribler Core Test Exception')
]
```

src/tribler/core/start_core.py
```diff
@@ -177,10 +179,25 @@ def run_core(api_port, api_key, root_state_dir, parsed_args):
     logger.info('Running Core' + ' in gui_test_mode' if parsed_args.gui_test_mode else '')
     load_logger_config('tribler-core', root_state_dir)
 
+    gui_pid = GuiProcessWatcher.get_gui_pid()
```
kozlovsky, line 182: GUI process passes its PID to the core process by setting the `TRIBLER_GUI_PID` environment variable that the core process reads. Here I re-use this logic to determine the PID of the GUI process; the determined GUI PID is added to debug information about the Core process.

src/tribler/core/utilities/process_locker.py
```diff
@@ -0,0 +1,362 @@
...
+                           exit_code=exit_code, error_msg=error_msg, error_info=from_json(error_info),
+                           other_params=from_json(other_params))
+
+    def to_dict(self) -> dict:
+        d = dict(rowid=self.rowid, pid=self.pid, kind=self.kind.value, active=self.active, canceled=self.canceled,
+                 app_version=self.app_version, started_at=self.started_at, creator_pid=self.creator_pid,
+                 api_port=self.api_port, finished_at=self.finished_at, exit_code=self.exit_code,
+                 error_msg=self.error_msg, error_info=self.error_info, shutdown_request_pid=self.shutdown_request_pid,
+                 shutdown_request_at=self.shutdown_requested_at, other_params=self.other_params)
+        return {key: value for key, value in d.items() if value is not None}
```
drew2a, line 111: NIT: maybe it is easier to derive the class from the `dataclass` to omit this conversion?

src/tribler/core/start_core.py
```diff
@@ -177,10 +179,25 @@ def run_core(api_port, api_key, root_state_dir, parsed_args):
...
+        process_locker.sys_exit(1, msg)
+
+    if api_port is None:
+        msg = 'api_port is not specified for a core process'
+        logger.error(msg)
+        process_locker.sys_exit(1, msg)
+
+    process_locker.set_api_port(api_port)
+
     with single_tribler_instance(root_state_dir):
```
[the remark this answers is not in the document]
drew2a: > Keeping the previous logic is still necessary for some time to ensure compatibility with previous Tribler versions. Later it should become possible to combine & simplify the logic

Let's discuss it when you have time

src/tribler/gui/start_gui.py
```diff
@@ -67,22 +72,25 @@ def run_gui(api_port, api_key, root_state_dir, parsed_args):
...
         logger.exception(ie)
-        error_and_exit("Import Error", f"Import error: {ie}")
+        show_system_popup("Import Error", f"Import error: {ie}")
+        process_locker.sys_exit(1, exc=ie)
 
     except TriblerException as te:
         logger.exception(te)
-        error_and_exit("Tribler Exception", f"{te}")
+        show_system_popup("Tribler Exception", f"{te.__class__.__name__}: {te}")
+        process_locker.sys_exit(1, exc=te)
```
drew2a, line 93: This block could be replaced by

```python
    except (ImportError, TriblerException) as e:
        logger.exception(e)
        exception_name = e.__class__.__name__
        show_system_popup(exception_name, f"{exception_name}: {e}")
        process_locker.sys_exit(1, exc=e)
```

But is still unclear why `ImportError` and `TriblerException` are special :)

src/tribler/gui/start_gui.py
```diff
@@ -41,14 +42,18 @@ def run_gui(api_port, api_key, root_state_dir, parsed_args):
...
+    check_environment()
+    check_free_space()
+
+    process_locker = ProcessLocker(root_state_dir, ProcessKind.GUI)
+    another_process_is_active = not process_locker.current_process.active
+    set_global_process_locker(process_locker)   # to be able to add information about exception to the process info
+    try:
         app_name = os.environ.get('TRIBLER_APP_NAME', 'triblerapp')
-        app = TriblerApplication(app_name, sys.argv)
+        app = TriblerApplication(app_name, sys.argv, another_process_is_active)
```
drew2a, line 56: I guess I understand your intention to pass the argument `another_process_is_active` to `TriblerApplication` but maybe it is not worth it in the end.

It is used only as:

```python
  ...
  if app.is_running():
            # if an application is already running, then send the command line
            # argument to it and close the current instance
            ...
```

That is why I would suggest removing the passing `another_process_is_active` logic completely and using `not process_locker.current_process.active` instead:

```python
  ...
  if another_process_is_active
            # if an application is already running, then send the command line
            # argument to it and close the current instance
            ...
```

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+            logger.warning(f'Retrying after the error: {e.__class__.__name__}: {e}')
+            return func(*args, **kwargs)
+    new_func: func
+    return new_func
+
+
+class ProcessManager:
+    def __init__(self, root_dir: Path, process_kind: ProcessKind, creator_pid: Optional[int] = None, **other_params):
+        self.root_dir = root_dir
+        self.filename = self._get_file_name(root_dir)
```
drew2a, line 83: NIT:

It is not obvious that the filename is a DB file path.
Suggestion: rename to the `db_filepath`

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+        self.current_process = TriblerProcess.current_process(process_kind, creator_pid, **other_params)
+        self.primary_process = self.atomic_get_primary_process(process_kind, self.current_process)
+
+    @classmethod
+    def _get_file_name(cls, root_dir: Path) -> Path:  # The method is added for easier testing
```
drew2a, line 88: NIT:

This method could be replaced by the following construction (that should be same convenient for testing):

```python
class ProcessManager:
    def __init__(self, root_dir: Path, process_kind: ProcessKind, creator_pid: Optional[int] = None,
                 db_file_name: str = DB_FILENAME, **other_params):
        self.root_dir = root_dir
        self.filename = root_dir / DB_FILENAME
        self.current_process = TriblerProcess.current_process(process_kind, creator_pid, **other_params)
        self.primary_process = self.atomic_get_primary_process(process_kind, self.current_process)
```

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def connect(self) -> sqlite3.Connection:
+        connection = sqlite3.connect(str(self.filename))
+        try:
+            connection.execute('BEGIN EXCLUSIVE TRANSACTION')
+            connection.execute(CREATE_SQL)
+            return connection
+        except:  # noqa: E722
+            connection.close()
+            raise
```
drew2a, line 99: It seems that the function `connect` uses only by the function `transaction`.

That is why it could be renamed to `_connect` and the except block could be removed safely.

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+                connection.close()  # pragma: no cover
+            if isinstance(e, sqlite3.DatabaseError):
+                self.filename.unlink(missing_ok=True)
+            raise
+
+    @with_retry
+    def atomic_get_primary_process(self, kind: ProcessKind,
+                                  current_process: Optional[TriblerProcess] = None) -> Optional[TriblerProcess]:
+        primary_process = None
+        with self.transaction() as connection:  # pylint: disable=not-context-manager  # false Pylint alarm
```
drew2a, line 122: NIT Why did you not use `transaction` as a variable name?
```suggestion
        with self.transaction() as transaction:  # pylint: disable=not-context-manager  # false Pylint alarm
```

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+            if current_process is not None:
+                if primary_process is None:
+                    current_process.primary = 1
+                    primary_process = current_process
+                else:
+                    current_process.primary = 0
+                    current_process.canceled = 1
+                current_process.save(connection)
+
+            return primary_process
```
drew2a, line 144: This function contains quite complex logic. As an attempt to simplify it (without changing):

```suggestion
        def _load_primary_process() -> Optional[TriblerProcess]:
            cursor = connection.execute("""
                SELECT * FROM processes WHERE kind = ? and "primary" = 1 ORDER BY rowid DESC LIMIT 1
            """, [kind.value])
            row = cursor.fetchone()
            if row is None:
                return None

            process = TriblerProcess.from_row(row)
            if process.is_running():
                return process

            process.primary = 0
            process.save(connection)
            return None

        with self.transaction() as connection:  # pylint: disable=not-context-manager  # false Pylint alarm
            primary_process = _load_primary_process()

            if current_process is None:
                return primary_process

            if primary_process is None:
                current_process.primary = 1
                current_process.save(connection)
                return current_process

            current_process.primary = 0
            current_process.canceled = 1
            current_process.save(connection)
            return primary_process
```

drew2a (changes requested):

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+                else:
+                    previous_primary_process.primary = 0
+                    previous_primary_process.save(connection)
+
+            if current_process is not None:
+                if primary_process is None:
+                    current_process.primary = 1
+                    primary_process = current_process
+                else:
+                    current_process.primary = 0
```
drew2a, line 140: I'm not sure that this line is correct.

src/tribler/core/start_core.py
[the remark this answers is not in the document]
kozlovsky: I removed the previous ProcessChecker logic

src/tribler/core/sentry_reporter/sentry_reporter.py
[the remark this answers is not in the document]
kozlovsky: Done

src/tribler/core/start_core.py
```diff
@@ -177,10 +179,25 @@ def run_core(api_port, api_key, root_state_dir, parsed_args):
...
     load_logger_config('tribler-core', root_state_dir)
 
+    gui_pid = GuiProcessWatcher.get_gui_pid()
+    process_locker = ProcessLocker(root_state_dir, ProcessKind.Core, gui_pid)
+    set_global_process_locker(process_locker)
+
+    if not process_locker.current_process.active:
+        msg = f'Another Core process with PID {process_locker.active_process.pid} is already running'
+        logger.warning(msg)
+        process_locker.sys_exit(1, msg)
```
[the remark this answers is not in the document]
kozlovsky: I renamed active to primary, not it should be clearer. A developer should use the new `process_manager.sys_exit()` in all cases after the process_manager is instantiated and the current process' information is stored in the database. If the standard `sys.exit` is used instead, nothing critical happens, but the process's final state will not be reflected in the database. It should not break anything, but the future error report will contain less helpful information.

src/tribler/core/utilities/process_locker.py
```diff
@@ -0,0 +1,362 @@
...
+from pathlib import Path
+from threading import Lock
+from typing import ContextManager, List, Optional
+
+import psutil
+from decorator import contextmanager
+
+from tribler.core.version import version_id
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
kozlovsky: Any logger instance is a singleton and a kind of a global variable. It does not matter if the logger is initialized inside the class; it is still the "global variable".

I put the logger inside the class now, but I still keep it in the module-level variable for it, as I need it inside the module-level `set_error` function.

src/tribler/core/utilities/process_locker.py
```diff
@@ -0,0 +1,362 @@
...
+def to_json(value) -> Optional[str]:
+    if value is None:
+        return None
+    return json.dumps(value)
+
+
+def from_json(value) -> Optional[dict]:
+    if value is None:
+        return None
+    return json.loads(value)
```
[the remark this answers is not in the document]
kozlovsky: I completely removed these functions, together with the fields they were used for

src/tribler/core/utilities/process_locker.py
kozlovsky: I considered this option, but I think it is more like a "normal" class now and contains enough logic to not be derived from the `dataclass`.

src/tribler/gui/start_gui.py
kozlovsky: Done

src/tribler/core/utilities/tribler_process.py
```diff
@@ -0,0 +1,183 @@
...
+                  self.shutdown_request_pid, self.shutdown_requested_at, self.finished_at,
+                  self.exit_code, self.error_msg, self._to_json(self.error_info),
+                  self._to_json(self.other_params), self.rowid, prev_version, self.pid, self.kind.value,
+                  self.app_version, self.started_at])
+            if cursor.rowcount == 0:
+                logger.error(f'Row {self.rowid} with row version {prev_version} was not found')
+
+    def _before_insert_check(self):
+        if self.row_version:
+            logger.error(f"The `row_version` value for a new process row should not be set. Got: {self.row_version}")
```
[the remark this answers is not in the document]
kozlovsky: I removed the method

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+
+from decorator import contextmanager
+
+from tribler.core.utilities.tribler_process import ProcessKind, TriblerProcess
+
+logger = logging.getLogger(__name__)
+
+DB_FILENAME = 'processes.sqlite'
+
+CREATE_SQL = """
```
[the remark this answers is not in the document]
kozlovsky: I extracted long SQL commands to a separate submodule `sql_scripts`

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def set_error(error_msg: Optional[str] = None, error_info: Optional[dict] = None,
+              exc: Optional[Exception] = None, replace: bool = False):
+    process_manager = get_global_process_manager()
+    if process_manager is None:
+        logger.warning('Cannot set error for process locker: no process locker global instance is set')
+    else:
+        process_manager.current_process.set_error(error_msg, error_info, exc, replace)
+        process_manager.save(process_manager.current_process)
```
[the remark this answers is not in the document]
kozlovsky: Done

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def with_retry(func):
```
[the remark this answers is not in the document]
kozlovsky: Done

src/tribler/core/utilities/tribler_process_manager.py
```diff
@@ -0,0 +1,175 @@
...
+            return func(*args, **kwargs)
+        except sqlite3.Error as e:
+            logger.warning(f'Retrying after the error: {e.__class__.__name__}: {e}')
+            return func(*args, **kwargs)
+    new_func: func
+    return new_func
+
+
+class ProcessManager:
+    def __init__(self, root_dir: Path, process_kind: ProcessKind, creator_pid: Optional[int] = None, **other_params):
```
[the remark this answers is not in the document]
kozlovsky: I removed it together with other unused attributes

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: Done

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: I removed the method

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: I refactored `_connect` and `transaction` methods to a simpler `connect` method

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: The new context manager is called `connect` instead of `transaction`, so it should be more intuitive now

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: Thanks for the suggestion, I refactored it this way

src/tribler/core/utilities/tribler_process_manager.py
kozlovsky: It was for "additional safety", but you are right that it is not necessary, and I removed it

src/tribler/core/components/session.py
```diff
@@ -27,7 +27,7 @@ class Session:
     def __init__(self, config: TriblerConfig = None, components: List[Component] = (),
                  shutdown_event: Event = None, notifier: Notifier = None, failfast: bool = True):
         # deepcode ignore unguarded~next~call: not necessary to catch StopIteration on infinite iterator
-        self.exit_code = 0
+        self.exit_code = None
```
drew2a, line 30: It seems that it is possible to return `None` instead of `0` here: https://github.com/kozlovsky/tribler/blob/process_locker/src/tribler/core/start_core.py#L120

src/tribler/core/utilities/process_manager/manager.py
```diff
@@ -0,0 +1,157 @@
...
+    with _lock:
+        return global_process_manager
+
+
+def set_error(error: Union[str | Exception], replace: bool = False):
+    process_manager = get_global_process_manager()
+    if process_manager:
+        process_manager.current_process.set_error(error, replace)
+    else:
+        logger.warning('Cannot set error for process locker: no process locker global instance is set')
```
drew2a, line 42: NIT: could be replaced:

```suggestion
    if process_manager := get_global_process_manager():
        process_manager.current_process.set_error(error, replace)
    else:
        logger.warning('Cannot set error for process locker: no process locker global instance is set')
```

Or, even, by:

```python
def set_error(error: Union[str | Exception], replace: bool = False):
    if process_manager := get_global_process_manager():
        process_manager.current_process.set_error(error, replace)
        return

    logger.warning('Cannot set error for process locker: no process locker global instance is set')
```

src/tribler/core/utilities/process_manager/manager.py
```diff
@@ -0,0 +1,157 @@
...
+        self.connection: Optional[sqlite3.Connection] = None
+        self.current_process = TriblerProcess.current_process(self, process_kind, creator_pid)
+        self.primary_process = self.atomic_get_primary_process(self.current_process)
+
+    @property
+    def logger(self) -> logging.Logger:
+        """
+        Used by the `with_retry` decorator
+        """
+        return logger
```
drew2a, line 59: It seems more straightforward, to use a logger, initialized as a class field instead of creating a property just to pass a logger, created on a module level.

src/tribler/core/components/session.py
kozlovsky: I'm not sure I understand. In PR it already returns `None` and not `0`. Or you mean to replace `return session.exit_code` with `return session.exit_code or None`?

src/tribler/core/utilities/process_manager/manager.py
kozlovsky: I changed it to a field

src/tribler/gui/start_gui.py
kozlovsky: I changed the logic to make it more straightforward:
```python
current_process_is_primary = process_manager.current_process.become_primary()
...
app = TriblerApplication(app_name, sys.argv, start_local_server=current_process_is_primary)
...
if not current_process_is_primary and app.connected_to_previous_instance:
    logger.info('GUI Application is already running. Passing a torrent file path to it.')
    ...
```

src/tribler/core/components/session.py
kozlovsky: To me it looks clearer is the `exit_code` is initially set to None, with the meaning `the exit code is not set`, comparing to converting `0` to `None` at a latter stage.

src/tribler/core/utilities/process_manager/manager.py
kozlovsky: Done

src/tribler/core/utilities/process_manager/manager.py
kozlovsky: Agreed, I changed logger property to a simple field

drew2a (changes requested):
I left some minor comments that are not related to the core PR logic.

kozlovsky:
I think I addressed @drew2a comments

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+@patch.object(logger, 'warning')
+def test_global_process_manager(warning: Mock, process_manager: ProcessManager):
+    assert get_global_process_manager() is None
+
+    set_global_process_manager(process_manager)
+    assert get_global_process_manager() is process_manager
+
+    set_global_process_manager(None)
+    assert get_global_process_manager() is None
```
drew2a, line 94: This test uses a singleton for storing the global process manager. This could lead to a flaky behavior for the test suite as we had before.

The test as well not tests the main functionality of setter (the locking).

That's why it is better to just remove it.

drew2a (changes requested):

src/tribler/gui/start_gui.py
```diff
@@ -38,8 +38,12 @@ def run_gui(api_port, api_key, root_state_dir, parsed_args):
         logger.info('Enabling a workaround for Ubuntu 21.04+ wayland environment')
         os.environ["GDK_BACKEND"] = "x11"
 
-    # Set up logging
-    load_logger_config('tribler-gui', root_state_dir)
```
kozlovsky, line 40: As it turns out, the previous code contains a bug. Tribler Core/GUI processes should not load logger config before determining whether the current process is the primary or not. Otherwise, primary and secondary processes start writing to log files, and the logging module does not support writing to the same log file from multiple processes. It leads to random errors during the log rotation, like, using an incorrect file descriptor.

In this PR, I fix the problem by first checking whether the current process is primary and then initializing file-based logging for a primary process only.

src/tribler/core/utilities/tests/test_utilities.py
```diff
@@ -286,14 +286,18 @@ def test_add_url_param_clean():
...
 
-def test_load_logger(tmpdir):
+@patch('logging.config.dictConfig')
+def test_load_logger(dict_config: Mock, tmpdir):
     """
     Test loading the Tribler logger configuration.
     """
-    logger_count = len(logging.root.manager.loggerDict)
     load_logger_config('test', tmpdir)
-    assert len(logging.root.manager.loggerDict) >= logger_count
```
kozlovsky, line 295: As it turns out, the test that checks the loading of the logger configuration was incorrect for two reasons:
1. It modifies the global configuration of the logger handlers, which can affect the subsequent tests (the logger configuration does not completely reset after each test)
2. The `assert` check does not check anything relevant. The `logger_count` value is the same before and after the test. What changed during the logger configuration is handlers and not loggers, so we should check handlers and not the loggers count.

The new version of the test mocks the `dictConfig` function, so the logger configuration does not change, and the state of the subsequent tests is not affected.

src/tribler/core/utilities/process_manager/tests/test_manager.py
kozlovsky: I'm ok with removing the test, but I don't see the reason for it, so I suggest discussing it one more time. In my opinion, it is better to have this test for better coverage of the code that handles the global singleton. Without the coverage, it may be possible that the logic will be broken in some later refactoring.

The test does not contain any complex logic and returns the global singleton to a previous state right after the test. 

> The test as well not tests the main functionality of setter (the locking).

I don't think the lock is the main functionality of `get_global_process_manager`/`set_global_process_manager`. There are two types of locks - long and short. If the code contains a non-trivial code inside a long lock, it may spend too much time inside it, and another thread can freeze for too long to acquire it. An even worse scenario is a deadlock when two threads acquire two different locks in an improper order.

Neither of these potential problems is possible in the current test. The logic inside the lock is very simple and contains just a single Python statement:
```python
    with _lock:
        global_process_manager = process_manager
```
and
```
    with _lock:
        return global_process_manager
```
It is an example of a short lock that only covers a simple instruction to protect access to a global variable to make read/write operation atomic. It is just not possible to stay inside the lock for too long. There is no non-trivial function call inside the lock that can delay the exit from the lock. So, in this case, the lock is not a "main" functionality of the function; it is a trivial and very short lock for protecting the atomic access to a global variable. The lock functionality is already tested in Python's test suite, and we can quickly check the code inside the lock to see that it does not have any potential to create a problem.

kozlovsky: Update: I added a try/finally block inside the test to ensure it has no chance to affect the following tests in case of an exception. In any case, the test suite will fail in case of exception, so there is no chance that after this test, we can have a mysterious problem in a subsequent test, but this test passes without an error. So, even in the (highly improbable) scenario that this test will fail and affect the global state, it will also be marked as failed and should explain the actual reason for any other error.

src/tribler/gui/start_gui.py
```diff
@@ -25,7 +27,7 @@
 
 
 def run_gui(api_port, api_key, root_state_dir, parsed_args):
-    logger.info('Running GUI' + ' in gui_test_mode' if parsed_args.gui_test_mode else '')
+    logger.info(f"Running GUI in {'gui_test_mode' if parsed_args.gui_test_mode else 'normal mode'}")
```
kozlovsky, line 30: Previously, because of the prcedence rules, an empty string was logged in normal mode. Basically:
```python
('Running GUI' + ' in gui_test_mode') if parsed_args.gui_test_mode else ('')
```

src/tribler/core/utilities/process_manager/tests/test_manager.py
drew2a: Why don't you use simple singleton creation for the `ProcessManager` as we have done for the `ExceptionHandler`?

https://github.com/Tribler/tribler/blob/main/src/tribler/core/components/reporter/exception_handler.py#L131 

In this case, it is not necessary to test the setter and the getter.

drew2a: > I don't think the lock is the main functionality of get_global_process_manager/set_global_process_manager.

Why then do use locks? What corner case do you want to cover?

drew2a: >  I need to create TriblerProcess before it, and I prefer to create the TriblerProcess instance in tribler.gui.start_gui/tribler.core.start_core instead.

It is not necessary. You can create a global singleton before and set the current process to it after.

kozlovsky: > Why then do use locks? What corner case do you want to cover?

It is necessary to have a lock when it is potentially possible to work with the same global variable from two different threads

> It is not necessary. You can create a global singleton before and set the current process to it after.

The the API will be more complicated, as it becomes necessary to make the current_process optional and check it in all places

Ok, I do not agree with this, but I removed the test to avoid long disputes

kozlovsky:
I addressed the changes

drew2a (approved):
Great job. It can fix around 8 current issues :)

Issue #7282, opened 2023-01-20
Migration to python 3.12


drew2a:
Python 3.11 offers some performance improvements that are good to have: https://docs.python.org/3/whatsnew/3.11.html#faster-runtime.

Python 3.11 also is a must-have for all Apple Silicone owners if they want to build Tribler easily: https://github.com/Tribler/tribler/issues/7252 https://github.com/Tribler/tribler/pull/7281

Currently, the following libraries from Tribler's dependencies are not ready for this migration:
- [x] ipv8: https://github.com/Tribler/py-ipv8/issues/1113
- [x] ponyORM: https://github.com/ponyorm/pony/issues/668
- [x] libtorrent: https://github.com/arvidn/libtorrent/discussions/7244 https://github.com/arvidn/libtorrent/issues/7536

ichorid:
The real problem is Libtorrent: building it with Python 3.11 requires Boost 1.81, which was published only recently. However, the PIP version of Libtorrent is pretty old and does not support 3.11.

Related discussions:
[Instructions for building Libtorrent with Python 3.11 bindings](https://github.com/arvidn/libtorrent/discussions/7187)

["update pypi libtorrent"](https://github.com/arvidn/libtorrent/discussions/7244)

["Who is responsible for which Python PIP package?"](https://github.com/arvidn/libtorrent/discussions/7178)

drew2a:
@ichorid thank you for the information! I've updated the issue's description.

Solomon1732:
I'll add that Python 3.10 added match statements: [What's new in python 3.10](https://docs.python.org/3/whatsnew/3.10.html), [specification](https://peps.python.org/pep-0634/), [motivations and rationale](https://peps.python.org/pep-0635/), [tutorial](https://peps.python.org/pep-0636/). It can negate the need for many `if-elif` chains when writing code.

Solomon1732:
There looks to be a difference between [this](https://github.com/arvidn/libtorrent/blob/2b6360320b1b1bf3a28b72c9866071b2b155ac21/pyproject.toml#L5) somewhat older libtorrent version and [this](https://github.com/arvidn/libtorrent/blob/RC_2_0/pyproject.toml) 2.0 branch. Namely, python 3.11 isn't skipped anymore. That said, I don't know if said branch is a stable release or a release candidate. Its release page isn't clear about it:
https://github.com/arvidn/libtorrent/releases

Solomon1732:
[...]

PR #7285, opened 2023-01-24
Update `ipv8` to the version `2.10`

drew2a:
This PR updates the ipv8 version to the `2.10`

Related to #7282


qstokkink:
https://github.com/Tribler/tribler/blob/ad7b6345439e0d59fc804117d291ae966ac6059b/src/tribler/core/components/tunnel/community/tunnel_community.py#L251

Should be changed to:

```python
  elif self.request_cache.has("retry", payload.circuit_id): 
```

xoriole (approved):

PR #7286, opened 2023-01-25
`Torrent Checker` refactoring

### LINKED ISSUE #6131: False-negative checks on a rare content
How to reproduce:
1. Open Tribler
1. Check the content of `fma_medium_april` channel (the content will be mostly "red")
1. Try to download the red content
1. The content will be downloaded successfully

image
image

I see the described behavior on my test channel, but at the same time, I see this behavior on the content in other channels.

### LINKED ISSUE #7287: [main] Recheck torrent health is broken
Latest main.

**Describe the bug**
GUI logic for rechecking torrent health is broken

**To Reproduce**
1. Search for "ubuntu"
2. Right-click on a torrent
3. Click on "Recheck health"
4. Nothing happens

**Expected behavior**
Torrent's health status should change.

**Screenshots**
image

image

drew2a:
This PR related to #6131, fixes #7287

Changes:
* logging has been extended
* fixed types inconsistency
* fixed `run_tribler_headless.py` (it was run without components)
* fixed `check_random_tracker` (the result of checks were never saved in DB)
* `app_mode` added to logging
* fixed `FakeDHTSession` logic inconsistency
* fixed  recheck torrent health logic (@kozlovsky)
* a tooltip was added to show the freshness of the seeders/leechers info (@kozlovsky)

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
@@ -5,7 +5,7 @@ from typing import Awaitable
 from ipv8.taskmanager import TaskManager
 
 from tribler.core.components.libtorrent.utils.libtorrent_helper import libtorrent as lt
-from tribler.core.components.torrent_checker.torrent_checker.torrentchecker_session import InfohashHealth
+from tribler.core.components.torrent_checker.torrent_checker.dataclasses import InfohashHealth
 from tribler.core.utilities.unicode import hexlify
 
 
--- a/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
@@ -1,4 +1,5 @@
 import math
+import time
 from asyncio import Future
 from typing import Awaitable
 
@@ -64,7 +65,8 @@ class DHTHealthManager(TaskManager):
         seeders = DHTHealthManager.get_size_from_bloomfilter(bf_seeders)
         peers = DHTHealthManager.get_size_from_bloomfilter(bf_peers)
         if not self.lookup_futures[infohash].done():
-            self.lookup_futures[infohash].set_result(InfohashHealth(infohash=infohash, seeders=seeders, leechers=peers))
+            health = InfohashHealth(infohash, last_check=int(time.time()), seeders=seeders, leechers=peers)
+            self.lookup_futures[infohash].set_result(health)
 
         self.lookup_futures.pop(infohash, None)
 
--- a/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
@@ -6,7 +6,7 @@ from typing import Awaitable
 from ipv8.taskmanager import TaskManager
 
 from tribler.core.components.libtorrent.utils.libtorrent_helper import libtorrent as lt
-from tribler.core.components.torrent_checker.torrent_checker.dataclasses import InfohashHealth
+from tribler.core.components.torrent_checker.torrent_checker.dataclasses import HealthInfo
 from tribler.core.utilities.unicode import hexlify
 
 
@@ -27,7 +27,7 @@ class DHTHealthManager(TaskManager):
         self.outstanding = {}  # Map from transaction_id to infohash
         self.lt_session = lt_session
 
-    def get_health(self, infohash, timeout=15) -> Awaitable[InfohashHealth]:
+    def get_health(self, infohash, timeout=15) -> Awaitable[HealthInfo]:
         """
         Lookup the health of a given infohash.
         :param infohash: The 20-byte infohash to lookup.
@@ -65,7 +65,7 @@ class DHTHealthManager(TaskManager):
         seeders = DHTHealthManager.get_size_from_bloomfilter(bf_seeders)
         peers = DHTHealthManager.get_size_from_bloomfilter(bf_peers)
         if not self.lookup_futures[infohash].done():
-            health = InfohashHealth(infohash, last_check=int(time.time()), seeders=seeders, leechers=peers)
+            health = HealthInfo(infohash, last_check=int(time.time()), seeders=seeders, leechers=peers)
             self.lookup_futures[infohash].set_result(health)
 
         self.lookup_futures.pop(infohash, None)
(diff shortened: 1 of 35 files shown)


hbiyik:
Thanks for this, as someone who has spent some time in this part of code, i had seen lots of issues there. Not cosmetic-wise but flow-wise as well. Sorry can't provide an actual feedback it ihas been a long while, but i appreciate the bravery. I hope it can motive :)

drew2a:
@hbiyik thank you for your support! It's quite motivating!

src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
```diff
@@ -413,72 +423,59 @@ async def scrape(self):
...
+    async def connect_to_tracker(self) -> TrackerResponse:
+        tasks = [self.download_manager.dht_health_manager.get_health(infohash, timeout=self.timeout) for infohash in
+                 self.infohash_list]
+        return TrackerResponse(url=DHT, responses=await gather(*tasks))
+
+
+TrackerSessionType = TypeVar('TrackerSessionType', bound=TrackerSession)
+
+
+def create_tracker_session(tracker_url, timeout, proxy, socket_manager) -> TrackerSessionType:
```
kozlovsky, line 470: I think we should not use TypeVar here. TypeVar is useful for specifying a relationship between different arguments. Or, for example, for specifying that the result type should be the same as the type of some argument.

In this function, the result type is not dependent on a type of any argument, so it is possible to directly say that the result type is `TrackerSession`

src/tribler/core/components/libtorrent/download_manager/dht_health_manager.py
```diff
@@ -18,18 +19,17 @@ def __init__(self, lt_session):
...
-        self.bf_peers = {}          # Map from infohash to (final) peers bloomfilter
-        self.outstanding = {}       # Map from transaction_id to infohash
+        self.lookup_futures = {}  # Map from binary infohash to future
+        self.bf_seeders = {}  # Map from infohash to (final) seeders bloomfilter
+        self.bf_peers = {}  # Map from infohash to (final) peers bloomfilter
+        self.outstanding = {}  # Map from transaction_id to infohash
         self.lt_session = lt_session
 
-    def get_health(self, infohash, timeout=15):
+    def get_health(self, infohash, timeout=15) -> InfohashHealth:
```
kozlovsky, line 28: It looks like the result type should be `Awaitable[InfohashHealth]`

src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
```diff
@@ -32,18 +38,22 @@
...
-    Creates a tracker session with the given tracker URL.
-    :param tracker_url: The given tracker URL.
-    :param timeout: The timeout for the session.
-    :return: The tracker session.
-    """
-    tracker_type, tracker_address, announce_page = parse_tracker_url(tracker_url)
+@dataclass
+class TrackerResponse:
+    url: str
+    responses: List[InfohashHealth]
```
kozlovsky, line 44: It looks pretty confusing when a`response` object with the type of `TrackerResponse` has a `responses` attribute. Even more confusing to see a `responses` list of `TrackerResponse` objects in which each object has the `responses` attribute. It may be a bit hard to understand what `responses` means in a specific place.

How about renaming this attribute to something like `torrent_health_list`?

src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
drew2a: Fixed

src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
drew2a: Renamed

src/tribler/core/components/torrent_checker/torrent_checker/torrent_checker.py
```diff
@@ -114,67 +120,69 @@ async def check_random_tracker(self):
...
+            exception_str = str(e).replace('\n]', ']')
+            self._logger.warning(f"Got session error for the tracker: {session.tracker_url}\n{exception_str}")
             self.tracker_manager.update_tracker_info(session.tracker_url, False)
-            e.tracker_url = session.tracker_url
             raise e
+        finally:
+            await self.clean_session(session)
 
     @property
     def torrents_checked(self):
```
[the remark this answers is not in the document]
kozlovsky: The `load_torrents_checked_from_db` method loads all "recently checked" torrents, that is, torrents checked later than four hours ago from the previous Tribler run. As it turns out, due to some amazing combination of bugs that I'll describe in a separate issue, the database contains hundreds of "local" torrents with a "self_checked" flag set to `True`, a huge number of seeders and incorrect "future" last_check time. These torrents are also selected from the DB because "future" time is "later than four hours ago", so they are considered recently checked torrents. As a result, immediately after the Tribler launch PopularityCommunity starts spreading invalid health records to other peers.

To mitigate this, `load_torrents_checked_from_db` should not load invalid health objects with future dates from the DB, so the `last_check` value should be not only "more than four hours ago" but also "not in the future". And separately, invalid health records should be cleaned in the database.

src/tribler/core/components/torrent_checker/torrent_checker/torrent_checker.py
```diff
@@ -281,159 +290,156 @@ def get_valid_trackers_of_torrent(self, torrent_id):
...
-    def _update_torrent_result(self, response):
-        infohash = response['infohash']
-        seeders = response['seeders']
-        leechers = response['leechers']
-        last_check = response['last_check']
-
-        self._logger.debug("Update result %s/%s for %s", seeders, leechers, hexlify(infohash))
-
+    def update_torrent_health(self, health: InfohashHealth):
+        self._logger.info(f'Update torrent health: {health}')
```
kozlovsky, line 434: This method should not update health if the information is incorrect, for example (and most importantly) when the "last_check" time is future. Because of time drift, it is possible to have unsynchronized time between machines, so we should accept records that are, for example, just a few seconds in the future, but the tolerable difference should not be too large.

kozlovsky: Sometimes the torrent was checked locally just recently, but other peers have incorrect health records with big numbers of seeders for this torrent and continue spreading these records. When such health records arrive on our machine, we should ignore them to avoid overriding the correct health. So, even if incoming health information looks correct, the local node should ignore it if recently it already checks this torrent and has fresh health info about it.

kozlovsky (changes requested):

kozlovsky (approved):

PR #7300, opened 2023-02-21
Fix `asyncio.create_task()` calls

### LINKED ISSUE #7299: Potential `asyncio.create_task` issue
The Tribler codebase could be affected by this bug:

> If you have ever used asyncio.create_task you may have created a bug for yourself that is challenging (read almost impossible) to reproduce. If it occurs, your code will likely fail in unpredictable ways.

> The root cause of this [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug) is that if you don't hold a reference to the task object returned by create_task then the task may disappear without warning when Python runs garbage collection. In other words, the code in your task will stop running with no obvious indication why.

https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/

drew2a:
This PR fixes #7299

* `shutdown` method has been added to the `RestEndpoint` base class
* `List[Task]` has been added as a result value of the `AsyncGroup.add()` call
* Calls of `asyncio.create_task` have been fixed

### DIFF
--- a/src/tribler/core/components/restapi/rest/events_endpoint.py
+++ b/src/tribler/core/components/restapi/rest/events_endpoint.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import time
 from asyncio import CancelledError
@@ -8,7 +9,6 @@ from aiohttp import web
 from aiohttp_apispec import docs
 from ipv8.REST.schema import schema
 from ipv8.messaging.anonymization.tunnel import Circuit
-from ipv8.taskmanager import TaskManager, task
 from marshmallow.fields import Dict, String
 
 from tribler.core import notifications
@@ -39,7 +39,7 @@ topics_to_send_to_gui = [
 
 
 @froze_it
-class EventsEndpoint(RESTEndpoint, TaskManager):
+class EventsEndpoint(RESTEndpoint):
     """
     Important events in Tribler are returned over the events endpoint. This connection is held open. Each event is
     pushed over this endpoint in the form of a JSON dictionary. Each JSON dictionary contains a type field that
@@ -47,8 +47,7 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
     """
 
     def __init__(self, notifier: Notifier, public_key: str = None):
-        RESTEndpoint.__init__(self)
-        TaskManager.__init__(self)
+        super().__init__()
         self.events_responses: List[RESTStreamResponse] = []
         self.app.on_shutdown.append(self.on_shutdown)
         self.undelivered_error: Optional[dict] = None
@@ -59,7 +58,8 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
 
     def on_notification(self, topic, *args, **kwargs):
         if topic in topics_to_send_to_gui:
-            self.write_data({"topic": topic.__name__, "args": args, "kwargs": kwargs})
+            data = {"topic": topic.__name__, "args": args, "kwargs": kwargs}
+            self.async_group.add(self.write_data(data))
 
     def on_circuit_removed(self, circuit: Circuit, additional_info: str):
         # The original notification contains non-JSON-serializable argument, so we send another one to GUI
@@ -69,10 +69,7 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
                                                     additional_info=additional_info)
 
     async def on_shutdown(self, _):
-        await self.shutdown_task_manager()
-
-    async def shutdown(self):
-        await self.shutdown_task_manager()
+        await self.shutdown()
 
     def setup_routes(self):
         self.app.add_routes([web.get('', self.get_events)])
@@ -101,7 +98,6 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
     def has_connection_to_gui(self):
         return bool(self.events_responses)
 
-    @task
     async def write_data(self, message):
         """
         Write data over the event socket if it's open.
@@ -124,7 +120,7 @@ class EventsEndpoint(RESTEndpoint, TaskManager):
     def on_tribler_exception(self, reported_error: ReportedError):
         message = self.error_message(reported_error)
         if self.has_connection_to_gui():
-            self.write_data(message)
+            self.async_group.add(self.write_data(message))
         elif not self.undelivered_error:
             # If there are several undelivered errors, we store the first error as more important and skip other
             self.undelivered_error = message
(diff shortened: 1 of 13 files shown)


src/tribler/core/utilities/tiny_tribler_service.py
```diff
@@ -51,7 +53,7 @@ async def start_tribler():
         if fragile:
             make_async_loop_fragile(loop)
 
-        loop.create_task(start_tribler())
+        _ = loop.create_task(start_tribler())
```
kozlovsky, line 56: The code looks correct, but for future developers, the importance of this "underscore" variable might not be clear, and at some moment, it may be removed as part of some refactoring.

For this reason, it may be better to have a named variable. As IDE and linters can complain about unused variable, it is possible to use a new attribute instead:
```python
self.main_task = loop.create_task(start_tribler())
```
(I think it is also possible to use `self.async_group.add` here instead of `loop.create_task`)

src/tribler/core/utilities/tiny_tribler_service.py
```diff
@@ -97,8 +99,9 @@ async def _terminate_by_timeout(self):
 
     def _graceful_shutdown(self):
         self.logger.info("Shutdown gracefully")
-        task = asyncio.create_task(self.session.shutdown())
-        task.add_done_callback(lambda result: self._stop_event_loop())
+        tasks = self.async_group.add(self.session.shutdown())
```
kozlovsky, line 102: A side question unrelated to this PR: do we need to pass multiple coroutines to the `async_group.add` method? It may be easier to handle the result if we only pass a single coroutine on each call of `async_group.add`:
```python
shutdown_task = self.async_group.add(self.session.shutdown())
```
I did a quick search, and it seems we never pass multiple coros to `async_group.add`

src/tribler/core/components/restapi/rest/rest_endpoint.py
```diff
@@ -14,16 +23,24 @@ class RESTEndpoint:
...
 
-    def add_endpoint(self, prefix, endpoint):
+    def add_endpoint(self, prefix: str, endpoint: RESTEndpoint | EventsEndpoint | IPV8RootEndpoint):
         self.endpoints[prefix] = endpoint
         self.app.add_subapp(prefix, endpoint.app)
 
+    async def shutdown(self):
+        for endpoint in self.endpoints.values():
+            if isinstance(endpoint, RESTEndpoint):  # IPV8RootEndpoint doesn't have a shutdown method
+                await endpoint.shutdown()
```
kozlovsky, line 40: Here we shut down endpoints one by one. I think shutting down all endpoints in parallel should be more efficient. Something like:
```python
shutdown_group = AsyncGroup()
for endpoint in self.endpoint.values():
    if isinstance(endpoint, RESTEndpoint):
        shutdown_group.add(endpoint.shutdown())
await shutdown_group.wait()
await self.async_group.cancel()
```

src/tribler/core/utilities/tiny_tribler_service.py
kozlovsky: nitpick: Another side note, unrelated to the current PR: it may be good to rename the `AsyncGroup.add` method to `AsyncGroup.add_task`. This way, we can find all places where it is called. Right now, it is hard to search for all places where `AsyncGroup.add` is called because the `add` name is too generic.

kozlovsky (changes requested):
Overall looks good. I think I found some other places where `asyncio.create_task` is used dangerously:

* In `Service._graceful_shutdown`, file `hidden_peer_discovery.py`:
```python
    def _graceful_shutdown(self):
        task = asyncio.create_task(self.on_tribler_shutdown())
        task.add_done_callback(lambda result: TinyTriblerService._graceful_shutdown(self))
```
the variable is created, but its lifetime is shorter than the lifetime of the task.

* In `Service._graceful_shutdown`, file `speed_test_exit.py` - the same.

drew2a:
@kozlovsky thank you for your review. Two calls of `Service._graceful_shutdown` were fixed.

src/tribler/core/components/restapi/rest/events_endpoint.py
```diff
@@ -69,10 +69,7 @@ def on_circuit_removed(self, circuit: Circuit, additional_info: str):
                                                     additional_info=additional_info)
 
     async def on_shutdown(self, _):
-        await self.shutdown_task_manager()
-
-    async def shutdown(self):
-        await self.shutdown_task_manager()
+        await self.shutdown()
```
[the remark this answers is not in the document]
drew2a: Thank you! Fixed by removing `TaskManager` from EventsEndpoint's base classes.

src/tribler/core/components/restapi/restapi_component.py
```diff
@@ -153,6 +153,9 @@ async def shutdown(self):
         if self._events_endpoint:
             await self._events_endpoint.shutdown()
```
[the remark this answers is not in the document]
drew2a: Thank you! I've removed the explicit call for `self._events_endpoint.shutdown()`.

I'm not sure that we should explicitly check whether shutdown is called or not. The reason for that is the fact that the current implementation works well with repeated calls.

src/tribler/core/components/restapi/rest/rest_endpoint.py
```diff
@@ -14,16 +23,24 @@ class RESTEndpoint:
...
 
     def setup_routes(self):
         pass
 
-    def add_endpoint(self, prefix, endpoint):
+    def add_endpoint(self, prefix: str, endpoint: RESTEndpoint | EventsEndpoint | IPV8RootEndpoint):
         self.endpoints[prefix] = endpoint
         self.app.add_subapp(prefix, endpoint.app)
 
+    async def shutdown(self):
```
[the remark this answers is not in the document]
drew2a: The inheritance from TaskManager has been removed 👍.
About the flag... I'm not sure, let's discuss it.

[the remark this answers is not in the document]
drew2a: Fixed

src/tribler/core/utilities/tiny_tribler_service.py
drew2a: Fixed

src/tribler/core/utilities/tiny_tribler_service.py
drew2a: Will fix it in the next PR.

src/tribler/core/components/restapi/rest/rest_endpoint.py
drew2a: Implemented

kozlovsky (approved):

PR #7306, opened 2023-02-27
Improve the AsyncGroup

drew2a:
This PR adds changes to AsyncGroup, requested in https://github.com/Tribler/tribler/pull/7300


src/tribler/core/utilities/async_group/async_group.py
```diff
@@ -3,38 +3,38 @@
...
-    ... ])
+    >>> group.add_task(void())
+    >>> group.add_task(void())
+    >>> group.add_task(void())
     >>> await group.cancel()
     """
 
     def __init__(self):
         self._futures: Set[Future] = set()
+        self._cancelled = False
```
kozlovsky, line 26: **nitpick**: While both "canceled" and "cancelled" are acceptable for the past tense of cancel, the version with one L is more common in American English, while the version with two L's is more common in British English.

In Tribler code (as in programming in general), American English is used more often than British English. For example, we use "color" and not "colour", "center" and not "centre". (On the other side, we are currently using both "behavior" and "behaviour" with the same frequency).

For that reason, it may be better to use "canceled" instead of "cancelled", and "CanceledException" instead of "CancelledException".

src/tribler/core/utilities/async_group/async_group.py
```diff
@@ -3,38 +3,38 @@
...
 
-    def add(self, *coroutines) -> List[Task]:
+    def add_task(self, coroutine) -> Task:
```
kozlovsky, line 28: It is possible to add an argument type here
```suggestion
    def add_task(self, coroutine: Coroutine) -> Task:
```

src/tribler/core/utilities/async_group/async_group.py
```diff
@@ -56,6 +61,10 @@ async def cancel(self) -> List[Future]:
 
         return active
 
+    @property
+    def cancelled(self):
+        return self._cancelled
+
     def _done_callback(self, future: Future):
         self._futures.remove(future)
```
kozlovsky, line 69: It should be safe to use `remove` here, but it is probably a bit safer to use `discard` instead, as an official example for `asyncio.create_task` [suggests doing in done callback](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

src/tribler/core/utilities/async_group/async_group.py
```diff
@@ -3,38 +3,38 @@
 from contextlib import suppress
 from typing import Iterable, List, Set
 
+from tribler.core.utilities.async_group.exceptions import CancelledException
+
 
 class AsyncGroup:
```
kozlovsky, line 9: In the previous PR, `asyncio.create_task(...)` was replaced with `async_group.add_task(...)` whenever possible to avoid a situation when a task is garbage-collected and disappears before its execution ends.

But now, another case is possible when an async group itself is garbage-collected before all its tasks are finished. In that case, tasks managed by this async group can be garbage collected before they are finished.

Usually, it should not be a problem, as an async group is typically created as a field of another object with a long life span. So, someone holds a reference to AsyncGroup long enough to prevent its early garbage collection. But theoretically, some async groups can be garbage collected too early.

I'd like to see (maybe in some future PR) a `__del__` method of `AsyncGroup` that checks if this group manages some active futures. In that case, it should log an error message (it can't throw an exception, as [exceptions are suppressed inside `__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__))

kozlovsky (changes requested):

src/tribler/core/utilities/async_group/async_group.py
drew2a: Implemented

src/tribler/core/utilities/async_group/async_group.py
drew2a: Implemented

src/tribler/core/utilities/async_group/async_group.py
drew2a: Done

kozlovsky (approved):

drew2a:
@kozlovsky thank you for your help! I really appreciate the effort you have contributed to this PR.

PR #7313, opened 2023-03-08
Fix and speed up health checks

### LINKED ISSUE #7287: [main] Recheck torrent health is broken

kozlovsky:
This PR fixes #7287. There were several reasons for the bug:
1. When no trackers for a specific infohash were able to provide health info, the resulting health record with zero seeders and zero leechers had an incorrect `last_check` timestamp value of 0.
2. The checks in `HealthInfo.should_update()` were also incorrect, and the logic sometimes decided to keep the previous health info and not replace it with the new health info.
3. In such situations, due to the third bug, the notification was not sent to the GUI; as a result, the GUI status for the torrent was stuck in the "Checking..." state indefinitely.

Also, this PR fixes the long checking time by updating the torrent health immediately. Previously, TorrentChecker aggregated all results from all possible trackers before providing the new health info. Some trackers return result almost immediately, while other wait for up to 20 seconds. Because of this, all health checks previously took around 20 seconds, and now in many cases, they are updated almost immediately and then corrected with results from additional trackers if necessary.

### DIFF
--- a/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
+++ b/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import Tuple
+
 from pony import orm
 from tribler.core.components.torrent_checker.torrent_checker.dataclasses import HealthInfo
 
@@ -28,5 +30,9 @@ def define_binding(db):
         def to_health(self) -> HealthInfo:
             return HealthInfo(infohash=self.infohash, last_check=self.last_check,
                               seeders=self.seeders, leechers=self.leechers)
+
+        @property
+        def seeders_leechers_last_check(self) -> Tuple[int, int, int]:
+            return self.seeders, self.leechers, self.last_check
 
     return TorrentState
--- a/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
+++ b/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
@@ -23,10 +23,9 @@ def define_binding(db):
         @classmethod
         def from_health(cls, health: HealthInfo):
             return cls(infohash=health.infohash, seeders=health.seeders, leechers=health.leechers,
-                       last_check=health.last_check)
+                       last_check=health.last_check, self_checked=health.self_checked)
 
         def to_health(self) -> HealthInfo:
-            return HealthInfo(infohash=self.infohash, last_check=self.last_check,
-                              seeders=self.seeders, leechers=self.leechers)
+            return HealthInfo(self.infohash, self.seeders, self.leechers, self.last_check, self.self_checked)
 
     return TorrentState
--- a/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
+++ b/src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Tuple
-
 from pony import orm
 from tribler.core.components.torrent_checker.torrent_checker.dataclasses import HealthInfo
 
@@ -30,9 +28,5 @@ def define_binding(db):
         def to_health(self) -> HealthInfo:
             return HealthInfo(infohash=self.infohash, last_check=self.last_check,
                               seeders=self.seeders, leechers=self.leechers)
-
-        @property
-        def seeders_leechers_last_check(self) -> Tuple[int, int, int]:
-            return self.seeders, self.leechers, self.last_check
 
     return TorrentState
--- a/src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
+++ b/src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
@@ -1,6 +1,6 @@
 import time
 from dataclasses import dataclass, field
-from typing import List, Tuple
+from typing import List
 
 import human_readable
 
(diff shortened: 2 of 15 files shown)


src/tribler/core/components/metadata_store/db/orm_bindings/torrent_state.py
```diff
@@ -29,4 +31,8 @@ def to_health(self) -> HealthInfo:
             return HealthInfo(infohash=self.infohash, last_check=self.last_check,
                               seeders=self.seeders, leechers=self.leechers)
 
+        @property
+        def seeders_leechers_last_check(self) -> Tuple[int, int, int]:
```
drew2a, line 35: NIT: maybe it is better to replace Tuple with Dataclass? It would be more obvious in terms of guessing what is [int, int, int].

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -1,14 +1,17 @@
...
 from tribler.core.utilities.unicode import hexlify
 
 
-TOLERABLE_TIME_DRIFT = 60  # one minute
-HOUR = 60 * 60
+MINUTE = 60
+HOUR = MINUTE * 60
+TOLERABLE_TIME_DRIFT = MINUTE  # When receiving health from another peer, how far the timestamp can be in the future?
+TORRENT_CHECK_WINDOW = MINUTE  # When asking multiple trackers in parallel, we ignore this time difference in responses
+HEALTH_FRESHNESS_SECONDS = 4 * HOUR  # Number of seconds before a torrent health is considered stale. Default: 4 hours
```
drew2a, line 14: This is great that you have added this comments 👍

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -46,19 +49,57 @@ def infohash_hex(self):
     def is_valid(self) -> bool:
         return self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
 
-    def should_update(self, torrent_state, self_checked=False):
-        if self.last_check <= torrent_state.last_check:
-            # The torrent state in the DB is already fresher than this health
-            return False
+    @property
+    def seeders_leechers_last_check(self) -> Tuple[int, int, int]:
```
drew2a, line 53: The same note here, about Dataclass.

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -46,19 +49,57 @@ def infohash_hex(self):
...
+        return self.seeders, self.leechers, self.last_check
+
+    def should_update(self, torrent_state, self_checked=False) -> bool:
```
drew2a, line 56: Overall it is a good fix, but let's try to make this function simpler :)

drew2a (changes requested):
Overall, this PR offers a good fix for the health check, but let's try to make the `should_update` function simpler :)

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
kozlovsky: I added a few methods to HealthInfo class that should make the code easier to read: `health.just_checked()`, `health.old()`, `health.older_than(other)`, `health.much_older_than(other)`

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -46,19 +56,45 @@ def infohash_hex(self):
...
-        if self.last_check <= torrent_state.last_check:
-            # The torrent state in the DB is already fresher than this health
+    def old(self) -> bool:
+        now = int(time.time())
+        return self.last_check < now - HEALTH_FRESHNESS_SECONDS
+
+    def older_than(self, other) -> bool:
+        return self.last_check < other.last_check - TORRENT_CHECK_WINDOW
+
+    def much_older_than(self, other) -> bool:
```
drew2a, line 66:
```suggestion
    def much_older_than(self, other: HealthInfo) -> bool:
```

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -46,19 +56,45 @@ def infohash_hex(self):
...
         return self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
 
-    def should_update(self, torrent_state, self_checked=False):
-        if self.last_check <= torrent_state.last_check:
-            # The torrent state in the DB is already fresher than this health
+    def old(self) -> bool:
+        now = int(time.time())
+        return self.last_check < now - HEALTH_FRESHNESS_SECONDS
+
+    def older_than(self, other) -> bool:
```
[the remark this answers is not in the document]
kozlovsky: Fixed

src/tribler/core/components/torrent_checker/torrent_checker/torrent_checker.py
```diff
@@ -155,15 +155,15 @@ async def check_random_tracker(self):
             response = await self.get_tracker_response(session)
         except Exception as e:  # pylint: disable=broad-except
             self._logger.warning(e)
+            # TODO: should we update the number of seeders to zero here?
```
drew2a, line 158: I'm not sure. An exception here shows that the tracker is unreachable. Which says nothing about the torrent health itself.

drew2a (changes requested):
The current version of PR looks good to me.

`codecov/patch` issues should be addressed (maybe not all of them, but as much as possible).

src/tribler/core/components/torrent_checker/torrent_checker/torrent_checker.py
kozlovsky: I agree, I removed the comment

drew2a (approved):
👍

PR #7315, opened 2023-03-10
Fix long shutdown

### LINKED ISSUE #7314: [main] Long Tribler's shutdown
**Describe the bug**
During the Tribler's shutdown, it freezes on this line:
https://github.com/Tribler/tribler/blob/4c7c3aaed69cda21355401a1fef415ceed6b20ca/src/tribler/core/components/restapi/rest/rest_manager.py#L169

**To Reproduce**
1. Open Tribler 
3. Close Tribler

**Screenshots**
image

**Additional context**
The bug doesn't reproduce while running Tribler headless.

drew2a:
This PR fixes #7314

### DIFF
--- a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
@@ -4,12 +4,9 @@ from contextlib import suppress
 from pathlib import PurePosixPath
 
 from aiohttp import web
-
 from aiohttp_apispec import docs, json_schema
-
 from ipv8.REST.schema import schema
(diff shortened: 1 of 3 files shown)


kozlovsky (changes requested):
The fixes look good but are a bit incomplete. Right now, it forcefully closes the opened connections with a reduced timeout. I think [the proper closing of the EventsEndpoint streamed response](https://github.com/Tribler/tribler/issues/7314#issuecomment-1467528740) is a better way to handle this. Then the short timeout is not as necessary, as all connections are able to close properly without hitting the timeout.

kozlovsky (approved):

PR #7319, opened 2023-03-14
Fix GUI downloads status update

### LINKED ISSUE #7318: Download start: statuses switching in unexpected order
**Describe the bug**
When a user clicks on the start button, the download status switches from "stopped"->"downloading"->"stoped"->"downloading"

**To Reproduce**
1. Add a download
2. Click on "stop"
2. Click on "start"

**Expected behavior**
The status should switch only once from "stopped" to "downloading".

**Screenshots**
image
start_download

 - OS: macOS
 - Tribler's version [main]

**Additional information**
You can see almost the same behavior when pressing the "stop" button as well.

drew2a:
This PR fixes #7318

The biggest part of this PR is a refactoring of `DLSTATUS`.  This change introduces `Status(Enum)` which makes it possible to write stricter code and reduce duplication of the status definitions.

Also, a new field was added to json response of  `async def get_downloads(self, request)` to reduce unnecessary conversions from a status code to a status string and back. In simple words this change makes it possible to remove the following function:
```python
    def get_raw_download_status(self):
        return dlstatus_strings.index(self.download_info["status"])
```

The bugfix itself is the following change (simplified):

from:

```python
    def on_download_stopped(self, json_result):
        if json_result and "modified" in json_result:
            for selected_item in self.selected_items:
                if selected_item.download_info["infohash"] == json_result["infohash"]:
                    selected_item.download_info['status'] = "DLSTATUS_STOPPED"
                    selected_item.update_item()
                    self.update_downloads()
```

to:

```python
    def on_download_stopped(self, json_result):
        if json_result and "modified" in json_result:
            for selected_item in self.selected_items:
                if selected_item.download_info["infohash"] == json_result["infohash"]:
                    selected_item.update_item()
                    self.update_downloads()
```

The root cause of the bug is the resynchronization of a download state between the GUI and the core.


src/tribler/gui/widgets/downloadspage.py
```diff
@@ -239,13 +233,12 @@ def update_download_visibility(self):
             filter_match = self.window().downloads_filter_input.text().lower() in item.download_info["name"].lower()
             is_channel = item.download_info["channel_download"]
             if self.filter == DOWNLOADS_FILTER_CHANNELS:
-                item.setHidden(not is_channel or not filter_match)
+                ahide = not (is_channel and filter_match)
```
kozlovsky, line 236: **nitpick:** Is "ahide" correct English word?

src/tribler/core/utilities/simpledefs.py
```diff
@@ -6,29 +6,19 @@
...
+    DLSTATUS_ALLOCATING_DISKSPACE = 0
+    DLSTATUS_WAITING4HASHCHECK = 1
+    DLSTATUS_HASHCHECKING = 2
+    DLSTATUS_DOWNLOADING = 3
+    DLSTATUS_SEEDING = 4
+    DLSTATUS_STOPPED = 5
+    DLSTATUS_STOPPED_ON_ERROR = 6
+    DLSTATUS_METADATA = 7
+    DLSTATUS_CIRCUITS = 8
+    DLSTATUS_EXIT_NODES = 9
```
kozlovsky, line 20: The `Status` name looks too generic and unclear; I think it is better to use `DownloadStatus` instead. Also, the `DLSTATUS_` prefix before each name looks verbose; I think it is possible to remove it. As a result, instead of, for example, `Status.DLSTATUS_SEEDING`, we'll have `DownloadStatus.SEEDING`, which has the same length and is more descriptive.

```suggestion
class DownloadStatus(Enum):
    ALLOCATING_DISKSPACE = 0
    WAITING4HASHCHECK = 1
    HASHCHECKING = 2
    DOWNLOADING = 3
    SEEDING = 4
    STOPPED = 5
    STOPPED_ON_ERROR = 6
    METADATA = 7
    CIRCUITS = 8
    EXIT_NODES = 9
```

kozlovsky (changes requested):
Looks good; I only have minor comments about naming

src/tribler/gui/widgets/downloadspage.py
drew2a: I've taken it from the definition:
```python
    def setHidden(self, ahide: bool) -> None: ...
```

kozlovsky: I suspect in the original C++ definition it was a prefix for "argument" to avoid name conflicts inside a function. Probably in Tribler code it is better to name the variable just "hide"

drew2a: Renamed

kozlovsky (changes requested):
The DLSTATUS_XXX statuses are also described in the [doc/restapi/introduction.rst](https://github.com/Tribler/tribler/blob/4f9cfa1851ca66ec330ce8eec571a401235e0a5e/doc/restapi/introduction.rst#L52) file, probably should be renamed here as well. Also, I see they are mentioned in comments in [`DownloadState.get_progress & get_status`](https://github.com/Tribler/tribler/blob/4a4b019a59b66e763acaf55f315509c8ad65299c/src/tribler/core/components/libtorrent/download_manager/download_state.py#L59)

kozlovsky (approved):

PR #7320, opened 2023-03-14
Apply autoformat + optimise imports

drew2a:
This PR is a final part of https://github.com/Tribler/tribler/issues/6706

Now we don't have any open PR/release or branches. It is a good time to format the codebase.


kozlovsky (changes requested):
Looks nice! I think at least some of Codacy Static Code Analysis warnings should be fixed manually

Issue #7295, opened 2023-02-20
Torrent's health zombie records with invalid last check time from the future infect Tribler nodes, spread from machine to machine, and can't be fixed by local health checks


kozlovsky:
While reviewing #7286, I discovered a very interesting bug when incorrect health information with a future date and sometimes a huge number of seeders spreads from peer to peer infecting new peers and spreading even further. This becomes possible through a combination of several bugs in `TorrentChecker`, `MetadataStore`, and `PopularityCommunity`.

As I currently understand it, the bug works the following way:

1. Some Tribler clients' machines have totally incorrect local times. I think some active peer has local time +35 days in the future, and sometimes I also see records with a date several years in the future.

2. This node spreads incorrect information to its peers.

3. On receiving health information from the remote peer, `PopularityCommunity` [calls `MetadataStore.process_torrent_health` method](https://github.com/Tribler/tribler/blob/8b9c3eff19c951faf560ccf59b6143a5235d2219/src/tribler/core/components/popularity/community/popularity_community.py#L91):
```python
    @db_session
    def process_torrents_health(self, torrent_healths):
        infohashes_to_resolve = set()
        for infohash, seeders, leechers, last_check in torrent_healths:
            added = self.mds.process_torrent_health(infohash, seeders, leechers, last_check)  # <-- here
            if added:
                infohashes_to_resolve.add(infohash)
        return infohashes_to_resolve
```
and this method stores information in the DB with the only check that the [health's last check date should be later than the last check date in the DB](https://github.com/Tribler/tribler/blob/6abf27f9461813749c4648a22a88a1c8a2a17cd2/src/tribler/core/components/metadata_store/db/store.py#L485).
```python
    def process_torrent_health(self, infohash: bytes, seeders: int, leechers: int, last_check: int) -> bool:
        health = self.TorrentState.get_for_update(infohash=infohash)
        if health and last_check > health.last_check:
            health.set(seeders=seeders, leechers=leechers, last_check=last_check)  # <-- here
            ...
```
What is important, during this update, the state of the `TorrentState.self_check` flag does not alter. That means if a peer checked the health of this torrent at least once in the past, then the **health record is considered a locally checked record, even if the info from the remote peer overrode it**.

[...]

PR #7322, opened 2023-03-16
A DB upgrade that cleans erroneous health info from the DB

kozlovsky:
This DB upgrade finishes #7295 by cleaning incorrect records from the DB


src/tribler/core/upgrade/upgrade.py
```diff
@@ -128,6 +130,16 @@ def upgrade_tags_to_knowledge(self):
         migration = MigrationTagsToKnowledge(self.state_dir, self.secondary_key)
         migration.run()
 
+    def upgrade_pony_db_14to15(self):
+        mds_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
+
+        mds = MetadataStore(mds_path, self.channels_dir, self.primary_key, disable_sync=True,
+                            check_tables=False, db_version=13) if mds_path.exists() else None
```
xoriole, line 137: Should the db_version be 14 here?

src/tribler/core/upgrade/tests/test_upgrader.py
```diff
@@ -194,6 +196,41 @@ def _exists(db, table, column):
...
+    with db_session:
+        mds.TorrentState(infohash=unhexlify(b'0001'), seeders=0, leechers=0, self_checked=0, has_data=0, last_check=0)
+        mds.TorrentState(infohash=unhexlify(b'0002'), seeders=1, leechers=2, self_checked=0, has_data=1,
+                         last_check=now - 1000)
+        mds.TorrentState(infohash=unhexlify(b'0003'), seeders=1, leechers=2, self_checked=1, has_data=1,
+                         last_check=now - 1000)
+        mds.TorrentState(infohash=unhexlify(b'0004'), seeders=1, leechers=2, self_checked=0, has_data=1,
+                         last_check=now + 1000)
+        mds.TorrentState(infohash=unhexlify(b'0005'), seeders=1, leechers=2, self_checked=1, has_data=1,
+                         last_check=now + 1000)
```
drew2a, line 214: It is not clear which parameters are important for the test and which are not.
I assume there are two important parameters:
* self_checked
* last_check

Therefore I suggest changing this code part to the following:
```suggestion
    now = int(time.time())
    in_the_past = now - 1000
    in_the_future = now + 1000

    mds = MetadataStore(mds_path, channels_dir, trustchain_keypair, check_tables=False)

    @db_session
    def _add_torrent_state(self_checked=1, last_check=0):
        mds.TorrentState(infohash=random.randbytes(20), seeders=0, leechers=0, self_checked=self_checked, has_data=0,
                         last_check=last_check)

    _add_torrent_state(self_checked=0, last_check=0)
    _add_torrent_state(self_checked=0, last_check=in_the_past)
    _add_torrent_state(self_checked=1, last_check=in_the_past)
    _add_torrent_state(self_checked=0, last_check=in_the_future)
    _add_torrent_state(self_checked=1, last_check=in_the_future)
```

src/tribler/core/upgrade/tests/test_upgrader.py
```diff
@@ -194,6 +196,41 @@ def _exists(db, table, column):
...
+    mds.shutdown()
+
+    upgrader.upgrade_pony_db_14to15()
+
+    mds = MetadataStore(mds_path, channels_dir, trustchain_keypair, check_tables=False)
+    with db_session:
+        assert mds.get_value('db_version') == '15'
+        assert mds.db.execute('select count(*) from TorrentState').fetchone()[0] == 5
+        assert mds.db.execute('select count(*) from TorrentState where self_checked > 0').fetchone()[0] == 0
+        assert mds.db.execute(f'select count(*) from TorrentState where last_check > {now}').fetchone()[0] == 0
```
drew2a, line 229: NIT: These asserts are hard to read. Consider simplifying them:

```suggestion
    @db_session
    def _execute(query: str):
        return mds.db.execute(query).fetchone()[0]

    with db_session:
        assert mds.get_value('db_version') == '15'
        assert _execute('select count(*) from TorrentState') == 5
        assert _execute('select count(*) from TorrentState where self_checked > 0') == 0
        assert _execute(f'select count(*) from TorrentState where last_check > {now}') == 0
```

drew2a (changes requested):

src/tribler/core/upgrade/tests/test_upgrader.py
kozlovsky: Agreed, changed the test this way (with minor differences)

src/tribler/core/upgrade/tests/test_upgrader.py
kozlovsky: Good point, done

src/tribler/core/upgrade/tests/test_upgrader.py
```diff
@@ -194,6 +196,41 @@ def _exists(db, table, column):
...
+        mds.TorrentState(infohash=unhexlify(b'0003'), seeders=1, leechers=2, self_checked=1, has_data=1,
+                         last_check=now - 1000)
+        mds.TorrentState(infohash=unhexlify(b'0004'), seeders=1, leechers=2, self_checked=0, has_data=1,
+                         last_check=now + 1000)
+        mds.TorrentState(infohash=unhexlify(b'0005'), seeders=1, leechers=2, self_checked=1, has_data=1,
+                         last_check=now + 1000)
+    with db_session:
+        assert mds.db.execute('select count(*) from TorrentState').fetchone()[0] == 5
+        assert mds.db.execute('select count(*) from TorrentState where self_checked > 0').fetchone()[0] == 2
+        assert mds.db.execute(f'select count(*) from TorrentState where last_check > {now}').fetchone()[0] == 2
```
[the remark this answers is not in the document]
kozlovsky: It may be possible that an SQL query contains a subtle bug, and as a result, the query always returns the same result, `0`, even if the upgrade could not fix the problem and the table still contains undesirable rows.

By checking the same query before and after the upgrade, I can test that the upgrade is indeed changed the rows in the table, and now the same query returns a different result, while before the upgrade, it was able to find undesirable rows in the table.

drew2a (approved):

PR #7327, opened 2023-03-20
Add `local/` to ignored folders

drew2a:
This PR adds the `local/` folder to `.gitignore`. 
The developers will benefit from it by using the local folder as a place to store their "experimental/throwaway/dont-want-to-commit" code that is dependent on Tribler.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ _build/
 .Tribler/
 
 .mailmap
+
+# this folder is dedicated to storing local experimental code that don't want to be committed
+local/


kozlovsky (changes requested):
Good Idea! I think the folder purpose should be described somewhere in the doc.

I suggest adding a file `local/README` with a description of the `local` directory purpose. While the folder itself is added to `.gitignore`, it does not preclude explicitly adding a README file from this folder to Git. The presence of this file will also ensure that the local folder will be created in a new repository clone. Otherwise, Git will not create the empty local directory in a new repo clone, as it does not track empty directories.

xoriole (approved):
I'm fine with just ignoring the local/ folder. To me, it is descriptive enough by the name, and would not like to see it in the repo.

kozlovsky (approved):

PR #7330, opened 2023-03-22
Fix medium issues reported by Codacy

### LINKED ISSUE #7303: Addressing identified code quality issues?
Hi, all.

Is anyone working on the issues identified, here:

https://app.codacy.com/gh/Tribler/tribler/issues

This would probably be if higher priority than implementing feature requests. I know it's difficult when resources are stretched.

Thanks for your work/ contributions,

Chris
chris_debian
UK

drew2a:
This PR closes #7303

### DIFF
--- a/src/tribler/core/components/libtorrent/torrentdef.py
+++ b/src/tribler/core/components/libtorrent/torrentdef.py
@@ -63,7 +63,7 @@ class TorrentDef:
                 try:
                     lt.torrent_info(metainfo)
                 except RuntimeError as exc:
-                    raise ValueError(str(exc))
+                    raise ValueError from exc
             self.metainfo = metainfo
             self.infohash = sha1(lt.bencode(self.metainfo[b'info'])).digest()
             self.copy_metainfo_to_torrent_parameters()
@@ -489,7 +489,7 @@ class TorrentDefNoMetainfo:
     def get_infohash(self):
         return self.infohash
 
-    def get_length(self, selectedfiles=None):
+    def get_length(self, selectedfiles=None):  # pylint: disable=unused-argument
         return 0
 
     def get_metainfo(self):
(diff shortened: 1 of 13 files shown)


src/tribler/core/components/libtorrent/torrentdef.py
```diff
@@ -489,7 +489,7 @@ def get_name(self):
     def get_infohash(self):
         return self.infohash
 
-    def get_length(self, selectedfiles=None):
+    def get_length(self, _=None):
```
kozlovsky, line 492: I can't say I'm completely happy with replacing unused argument names with underscores. Even if the name is not used, it still carries potentially useful information about the meaning of the argument, which can be helpful in further refactorings.

The PyLint documentation suggests a way to deal with arguments that are not deliberately used: [explicitly deleting the argument in the function body](http://pylint-messages.wikidot.com/messages:w0613):

```python
def myfunc(a1, a2):
    del a2 # unused
    return a1*2
```

It looks a little unusual, but it keeps the name and avoids the warning.

I also think that the W0613 Unused argument warning can be turned off entirely because this is a false alarm in most cases in Tribler code. Besides, in the IDE, an unused argument is grayed out anyway, so it's hard to make a mistake by forgetting to use an argument.

kozlovsky (approved):
Looks good! I think we can consider removing Pylint W0613 "Unused argument" warning

src/tribler/core/components/libtorrent/torrentdef.py
xoriole: I'm in favor of using underscore if the argument is not used. 

If the argument is not used then I consider it is not much relevant for the understanding of the code/logic inside the function. If it appears really meaningful to keep an unused named argument, then we can make a pylint exception on that particular line.

drew2a: I also was in favor of using underscore everywhere, but now I'm not so sure about it.

The link that @kozlovsky  shared earlier illustrates a valid case. I guess that replacing unused arguments with `_` could potentially result in breaking the Tribler logic. Mainly if a class inheritance is used and no (or poorly) written tests exist.

---
> Other solutions such as renaming the argument to 'unused_a2' or the like are not recommended, because a caller could pass the argument by name:

```python
def myfunc(a1, unused_a2):
    return a1*2

def func2():
    return myfunc(42, a2=16) # now fails!
```
---
So, for this particular case, I would use pylint `pylint: disable=unused-argument`.

PR #7334, opened 2023-03-24
On GUI side, handle Core REST API responses without data

### LINKED ISSUE #7333: KeyError in register_request callback in SearchResultsWidget.search
```
[
tribler.gui.utilities.CreationTraceback: , 
  File "run_tribler.py", line 103, in <module>, 
  File "tribler/gui/start_gui.py", line 84, in run_gui, 
  File "tribler/gui/utilities.py", line 413, in trackback_wrapper, 
  File "tribler/gui/tribler_window.py", line 1083, in on_top_search_bar_return_pressed, 
  File "tribler/gui/widgets/searchresultswidget.py", line 120, in search, 
  File "tribler/gui/network/request_manager.py", line 80, in put, 
  File "tribler/gui/network/request.py", line 64, in __init__, 
  File "tribler/gui/utilities.py", line 416, in trackback_wrapper, 
  File "tribler/gui/utilities.py", line 413, in trackback_wrapper, 
  File "tribler/gui/widgets/searchresultswidget.py", line 114, in register_request, 
KeyError: 'peers'
]
```
[Here](https://github.com/Tribler/tribler/blob/f5d6aeb2e4befc55c1385ce6596393b95aa4d7c1/src/tribler/gui/widgets/searchresultswidget.py#L114):
```python
# searchresultswidget.py
class SearchResultsWidget(AddBreadcrumbOnShowMixin, widget_form, widget_class):
    ...
    def search(self, query: Query) -> bool:
        ...
        def register_request(response):
            peers = set(response["peers"])  # <-- here
            self.search_request = SearchRequest(response["request_uuid"], query, peers)
            self.search_progress_bar.set_remote_total(len(peers))

        params = {'txt_filter': fts_query, 'hide_xxx': self.hide_xxx, 'tags': list(query.tags),
                  'metadata_type': REGULAR_TORRENT, 'exclude_deleted': True}
        request_manager.put('remote_query', register_request, url_params=params)

        return True
```

The reason for the error is we got an empty dict in the callback. The empty dict is created [here](https://github.com/Tribler/tribler/blob/93927779051ddfe2412731067b5e3d4c14ad162d/src/tribler/gui/network/request.py#L107):

```python
class Request(QObject):
    ...
    def on_finished(self):
        ...
        try:
            ...
            data = bytes(self.reply.readAll())
            ...

            if not data:
                self.on_finished_signal.emit({})  # <-- here
                return
```

The code for returning an empty dict is added in #7298 to fix #7297. Apparently, when no data is returned in the reply, it is better not to call the callback at all, as the callback usually expects some keys in the resulting dict.

Also, it looks like the `on_finish` callback is better to rename to `on_success`, as these callback functions are expected to be called only on successful responses with correct results.

The actual reason why it is possible to receive an empty response from the reply is still unknown, I'll describe it in a separate issue #7335, so it will be possible to discuss it later.

kozlovsky:
This PR fixes #7333:
1. It adds the handling of `QNeworkReply` errors mentioned in #7335 by checking the QNetworkReply error code before reading the response data;
2. If there is an error or if the response data is empty, now Request does not call the `on_success` callback (renamed from `on_finish`);
3. In case of an error, it displays QNetworkReply errors in debug window as a request status code.

### DIFF
--- a/src/tribler/gui/network/request.py
+++ b/src/tribler/gui/network/request.py
@@ -17,6 +17,19 @@ if TYPE_CHECKING:
 DATA_TYPE = Optional[Union[bytes, str, Dict, List]]
 
 
+def make_reply_errors_map() -> Dict[int, str]:
+    errors_map = {}
+    for attr_name in dir(QNetworkReply):
+        if attr_name[0].isupper() and attr_name.endswith('Error'):  # SomeError, but not the `setError` method
+            error_code = getattr(QNetworkReply, attr_name)
+            if isinstance(error_code, int):  # an additional safety check, just for case
+                errors_map[error_code] = attr_name
+    return errors_map
+
+
+reply_errors = make_reply_errors_map()
+
+
 class Request(QObject):
     GET = 'GET'
     POST = 'POST'
@@ -93,6 +106,13 @@ class Request(QObject):
 
         self.logger.info(f'Finished: {self}')
         try:
+            error_code = self.reply.error()
+            if error_code != QNetworkReply.NoError:
+                error_name = reply_errors.get(error_code, '<unknown error>')
+                self.logger.warning(f'Request {self} finished with error: {error_code} ({error_name})')
+                self.update_status(-1)
+                return
+
             if status_code := self.reply.attribute(QNetworkRequest.HttpStatusCodeAttribute):
                 self.update_status(status_code)
 
--- a/src/tribler/gui/network/request.py
+++ b/src/tribler/gui/network/request.py
@@ -82,6 +82,7 @@ class Request(QObject):
 
         self.time = time()
         self.status_code = 0
+        self.status_text = "unknown"
         self.cancellable = True
 
     def set_manager(self, manager: RequestManager):
@@ -99,6 +100,10 @@ class Request(QObject):
     def update_status(self, status_code: int):
         self.logger.debug(f'Update {self}: {status_code}')
         self.status_code = status_code
+        if status_code > 0:  # positive codes are HTTP response codes
+            self.status_text = str(status_code)
+        else:  # negative codes represent QNetworkReply error codes
+            self.status_text = f'{status_code}: {reply_errors.get(-status_code, "<unknown error code>")}'
 
     def on_finished(self):
         if not self.reply or not self.manager:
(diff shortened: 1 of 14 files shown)


src/tribler/gui/network/tests/test_request.py
```diff
@@ -40,12 +40,12 @@ def test_str_data_constructor():
 
 def test_on_finished():
     # Test that if 'request.reply' is empty, the `on_finish` method is called with an empty dict.
```
xoriole, line 42: This comment also needs to be updated

src/tribler/gui/network/request.py
```diff
@@ -17,6 +17,19 @@
 DATA_TYPE = Optional[Union[bytes, str, Dict, List]]
 
 
+def make_reply_errors_map() -> Dict[int, str]:
```
xoriole, line 20: Interesting

src/tribler/gui/network/tests/test_request.py
kozlovsky: Thanks, fixed!

xoriole (approved):

xoriole:
Related to https://github.com/Tribler/tribler/issues/7558

By checking the QNetworkReply error code before reading the response data, I noticed that it is masking the HTTP error response like 413 HTTPRequestEntityTooLarge because the error code according to QT is not zero in such case. It returns error code: 299 (UnknownContentError).

Therefore, in my opinion, it would be best to check for the HTTP status code first and process the response (including empty and error responses). If there is no HTTP status code present then the QT error code be processed.

kozlovsky:
@xoriole, in my opinion, we should check the QNetworkReply error code before the HTTP error code because it covers a wider range of possible errors, like `QNetworkReply::ConnectionRefusedError` or `QNetworkReply::UnknownNetworkError`. It does not look correct to check the HTTP code if no HTTP response was received.

But we should allow further processing of responses with `QNetworkReply::UnknownContentError`, as in this case, we successfully got the response from the server, and the error is content-related and most probably explained by the HTTP error code and JSON error message.

So I think we should change
```
if error_code != QNetworkReply.NoError:
    ...
```
to
```
if error_code not in (QNetworkReply.NoError, QNetworkReply::UnknownContentError):
    ...
```
and allow processing of responses with `QNetworkReply::UnknownContentError`.

PR #7340, opened 2023-03-29
Rename the column with a torrent's creation date from "Updated" to "Created" to avoid confusion

kozlovsky:
This PR renames the column with the torrent's creation date from "Updated" to "Created" to avoid user confusion in the UI and adds the corresponding translation to all languages.

The corresponding data item field is still named "updated" in the application code. It is not changed to avoid accidental bugs in application logic and can be renamed later.


drew2a (approved):

drew2a:
> The corresponding data item field is still named "updated" in the application code. It is not changed to avoid accidental bugs in application logic and can be renamed later.

Could you please create a corresponding issue? Overwise there is a risk to forget.

kozlovsky:
I was able to rename it from "updated" to "created" in all necessary places, it turned out to be easier than I expected :)

drew2a (approved):

PR #7341, opened 2023-03-30
Add an option to disable `check_already_running`

drew2a:
This PR adds an option to disable `check_already_running` as it is useful for some experiments to run multiple instances simultaneously.

### DIFF
--- a/src/tribler/core/utilities/tiny_tribler_service.py
+++ b/src/tribler/core/utilities/tiny_tribler_service.py
@@ -40,11 +40,21 @@ class TinyTriblerService:
         It is good place to add a custom code.
         """
 
-    def run(self, fragile: bool = False):
+    def run(self, fragile: bool = False, check_already_running: bool = True):
+        """ Run the service
+
+        Args:
+            fragile: if True, the loop will be made fragile (fail on a first exception)
+            check_already_running: if True, verifies no other Tribler instance is running to prevent parallel
+                instances from writing to the same state directory or log files. It is necessary for components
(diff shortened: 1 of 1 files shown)


kozlovsky (changes requested):
As I understand it, it should already be possible to use multiple instances if each uses its own root state directory. Running multiple instances in the same directory can lead to data corruption and unpredictable behavior. So, it looks like this PR is incorrect and should be closed.

src/tribler/core/utilities/tiny_tribler_service.py
```diff
@@ -40,11 +40,21 @@ async def on_tribler_started(self):
...
 
-    def run(self, fragile: bool = False):
+    def run(self, fragile: bool = False, check_already_running: bool = True):
+        """ Run the service
+
+        Args:
+            fragile: if True, the loop will be made fragile (fail on a first exception)
+            check_already_running: if True, the check for already running Tribler instance will be performed. It is
+                necessary for some components like MetadataStoreComponent and KnowledgeComponent to be sure that there
+                is no another instance of Tribler running.
```
kozlovsky, line 50: Can we also mention the reason for the restriction (preventing the writes to the same directory or log files)? Something like:
```suggestion
            check_already_running: if True, verifies no other Tribler instance is running to prevent parallel
                instances from writing to the same state directory or log files. It is necessary for components
                such as MetadataStoreComponent and KnowledgeComponent.
```

kozlovsky (approved):

PR #7347, opened 2023-04-03
Add support for HTTPS tracker URL

xoriole:
Currently, our tracker utility that parses tracker URLs does not support HTTPS. This PR adds that support. Many trackers like Ubuntu tracker have moved to use HTTPS already so it is time we support them.

### DIFF
--- a/src/tribler/core/utilities/tests/test_tracker_utils.py
+++ b/src/tribler/core/utilities/tests/test_tracker_utils.py
@@ -6,137 +6,74 @@ from tribler.core.utilities.tracker_utils import (
     parse_tracker_url,
 )
 
+TRACKER_HOST = 'tracker.example.com'
 
-def test_uniform_scheme_correct_udp():
-    result = get_uniformed_tracker_url("udp://tracker.openbittorrent.com:80")
-    assert result == "udp://tracker.openbittorrent.com:80"
+EXPECTED_UNIFORM_URLS = [
+    (f'udp://{TRACKER_HOST}:80', f'udp://{TRACKER_HOST}:80'),  # Proper UDP URL
+    (f'udp://{TRACKER_HOST}', None),  # UDP no port
+    (f'udp://{TRACKER_HOST}:6969/announce', f'udp://{TRACKER_HOST}:6969'),  # UDP remove path
+    (f'udp://{TRACKER_HOST}:80\x00', f'udp://{TRACKER_HOST}:80'),  # Trailing zero hex
+    (f'udp://{TRACKER_HOST}:80\xff', None),  # Trailing non-zero hex
 
+    (f'http://{TRACKER_HOST}:6969/announce', f'http://{TRACKER_HOST}:6969/announce'),  # Correct HTTP URL
+    (f'http://{TRACKER_HOST}:6969/announce/', f'http://{TRACKER_HOST}:6969/announce'),  # Trailing /
+    (f'http://{TRACKER_HOST}', None),  # HTTP no path
+    (f'http://{TRACKER_HOST}\xa0', None),  # HTTP no path nbsp
+    (f'http://{TRACKER_HOST}/announce', f'http://{TRACKER_HOST}/announce'),  # HTTP default port
+    (f'http://{TRACKER_HOST}:80/announce', f'http://{TRACKER_HOST}/announce'),  # HTTP default port given
+    (f'http://{TRACKER_HOST}/?do=upload', None),  # Bad URL encoding
+    (f'http://{TRACKER_HOST}:80/anno...', None),  # Truncated URL
 
-def test_uniform_scheme_correct_http():
-    result = get_uniformed_tracker_url("http://torrent.ubuntu.com:6969/announce")
-    assert result == "http://torrent.ubuntu.com:6969/announce"
+    (f'https://{TRACKER_HOST}/announce', f'https://{TRACKER_HOST}/announce'),  # HTTPS default port
+    (f'https://{TRACKER_HOST}:443/announce', f'https://{TRACKER_HOST}/announce'),  # HTTPS default port given
 
-
-def test_uniform_scheme_correct_http_training_slash():
-    result = get_uniformed_tracker_url("http://torrent.ubuntu.com:6969/announce/")
-    assert result == "http://torrent.ubuntu.com:6969/announce"
-
-
-def test_uniform_scheme_unknown():
-    result = get_uniformed_tracker_url("unknown://tracker.openbittorrent.com/announce")
-    assert not result
-
-
-def test_uniform_http_no_path():
-    result = get_uniformed_tracker_url("http://tracker.openbittorrent.com")
-    assert not result
-
-
-def test_uniform_udp_no_port():
-    result = get_uniformed_tracker_url("udp://tracker.openbittorrent.com")
-    assert not result
+    (f'unknown://{TRACKER_HOST}/announce', None),  # Unknown scheme
+    (f'wss://{TRACKER_HOST}:80/announce', None),  # Wrong URL scheme
+    ('ftp://tracker.examp\xffle.org:80/announce', None),  # Wrong URL scheme
+    (';', None),  # Split error
+    ('', None),  # Empty URL
+]
 
 
-def test_uniform_udp_remove_path():
-    result = get_uniformed_tracker_url("udp://tracker.openbittorrent.com:6969/announce")
-    assert result == "udp://tracker.openbittorrent.com:6969"
+@pytest.mark.parametrize("given_url, expected_uniform_url", EXPECTED_UNIFORM_URLS)
+def test_get_uniformed_tracker_url(given_url, expected_uniform_url):
+    uniform_url = get_uniformed_tracker_url(given_url)
+    assert uniform_url == expected_uniform_url
 
 
-def test_uniform_http_no_path_nbsp():
-    result = get_uniformed_tracker_url("http://google.nl\xa0")
-    assert not result
+PARSED_TRACKER_URLS = [
+    # UDP with port
+    (f'udp://{TRACKER_HOST}:80', ("udp", (f"{TRACKER_HOST}", 80), '')),
+    # HTTP(S) with port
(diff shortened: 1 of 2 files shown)


src/tribler/core/utilities/tracker_utils.py
```diff
@@ -62,7 +62,7 @@ def get_uniformed_tracker_url(tracker_url):
         # accessing urlparse attributes may throw UnicodeError's or ValueError's
         try:
             # scheme must be either UDP or HTTP
-            if url.scheme == 'udp' or url.scheme == 'http':
+            if url.scheme == 'udp' or url.scheme == 'http' or url.scheme == 'https':
```
kozlovsky, line 65: **nitpick:** I'd simplify it to:
```suggestion
            if url.scheme in ('udp', 'http', 'https'):
```

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -76,6 +76,9 @@ def get_uniformed_tracker_url(tracker_url):
                 # HTTP trackers default to port HTTP_PORT
                 elif url.scheme == 'http':
                     uniformed_port = HTTP_PORT
+                # HTTPS trackers default to port HTTPS_PORT
+                elif url.scheme == 'https':
+                    uniformed_port = HTTPS_PORT
```
kozlovsky, line 81: **nitpick:** Just for extra safety, can we add
```python
else:
    continue
```
?

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -62,7 +62,7 @@ def get_uniformed_tracker_url(tracker_url):
         # accessing urlparse attributes may throw UnicodeError's or ValueError's
         try:
             # scheme must be either UDP or HTTP
-            if url.scheme == 'udp' or url.scheme == 'http':
+            if url.scheme == 'udp' or url.scheme == 'http' or url.scheme == 'https':
                 uniformed_scheme = url.scheme
```
kozlovsky, line 66: **nitpick:** `uniformed_scheme` variable looks unnecessary; we can just use `url.scheme` everywhere

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -116,11 +121,14 @@ def parse_tracker_url(tracker_url):
     if tracker_url.lower().startswith('http://') and ':80/' in tracker_url:
         tracker_url = tracker_url.replace(':80/', '/', 1)
 
+    if tracker_url.lower().startswith('https://') and ':443/' in tracker_url:
+        tracker_url = tracker_url.replace(':443/', '/', 1)
```
kozlovsky, line 125: Why is this code necessary? Can you provide an example of the incorrect URL that this code fixes? (I suspect the fixed URL is not entirely correct and can include an excess `/`, but may be wrong on that)

kozlovsky (approved):
Looks good, I just have a few minor comments

src/tribler/core/utilities/tracker_utils.py
drew2a: If I understood correctly, tuples in Python have O(n) complexity for checking an item's presence in the list.
Sets, however, have O(1) complexity.

So, I propose to use a set here:
```python
            if url.scheme in {'udp', 'http', 'https'}:
```

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -90,6 +93,8 @@ def get_uniformed_tracker_url(tracker_url):
 
             if url.scheme == 'http' and uniformed_port == HTTP_PORT:
                 uniformed_url = f'{uniformed_scheme}://{uniformed_hostname}{uniformed_path}'
+            elif url.scheme == 'https' and uniformed_port == HTTPS_PORT:
```
drew2a, line 96: NIT: The string `https` is repeated 6 times in this file. Please, consider extracting it to the const.

The same is true for the string `http`.

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -90,6 +93,8 @@ def get_uniformed_tracker_url(tracker_url):
...
+                uniformed_url = f'{uniformed_scheme}://{uniformed_hostname}{uniformed_path}'
```
drew2a, line 97: The line `uniformed_url = f'{uniformed_scheme}://{uniformed_hostname}{uniformed_path}'` is duplicated here. It could lead to an error in the future when a developer will change one line and forget to change another.

Consider to change it to something like:

```suggestion
            if (url.scheme == 'http' and uniformed_port == HTTP_PORT) or \
                    (url.scheme == 'https' and uniformed_port == HTTPS_PORT):
                uniformed_url = f'{uniformed_scheme}://{uniformed_hostname}{uniformed_path}'
```

src/tribler/core/utilities/tracker_utils.py
drew2a, line 125: The numbers `443` and `80` are hardcoded here. Also, there are duplications in strings (`':80/'` and `':443/'`)

Consider replacing them with something like:
```suggestion
    url = tracker_url.lower()
    http_port = f':{HTTP_PORT}/'
    https_port = f':{HTTPS_PORT}/'
    if url.startswith(f'http://') and http_port in url:
        tracker_url = tracker_url.replace(http_port, '/', 1)

    if url.startswith('https://') and https_port in url:
        tracker_url = tracker_url.replace(https_port, '/', 1)
```

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -116,11 +121,14 @@ def parse_tracker_url(tracker_url):
...
 
+    if tracker_url.lower().startswith('https://') and ':443/' in tracker_url:
+        tracker_url = tracker_url.replace(':443/', '/', 1)
+
     if tracker_url != get_uniformed_tracker_url(tracker_url):
         raise MalformedTrackerURLException(f'Could not sanitize url ({tracker_url}).')
 
     url = urlparse(tracker_url)
-    if not (url.scheme == 'udp' or url.scheme == 'http'):
+    if url.scheme not in ('udp', 'http', 'https'):
```
drew2a, line 131: NIT: I believe it is better to use a set here (see the explanation above).

src/tribler/core/utilities/tracker_utils.py
```diff
@@ -132,6 +140,12 @@ def parse_tracker_url(tracker_url):
     if url.scheme == 'http' and not url.path:
         raise MalformedTrackerURLException(f'Missing announce path for HTTP tracker url ({tracker_url}).')
 
+    if url.scheme == 'https' and not url.port:
+        return url.scheme, (url.hostname, 443), url.path
```
drew2a, line 144: The number 443 is hardcoded here. Consider changing it to:
```suggestion
        return url.scheme, (url.hostname, HTTPS_PORT), url.path
```

drew2a (changes requested):

src/tribler/core/utilities/tracker_utils.py
kozlovsky: I used to use tuples because some time ago, Python was able to cache and re-use immutable tuple literals but re-constructed mutable set literals each time the line of the code was executed (and that was much slower), but now Python can properly cache mutable set literals as well, and using set literals is better now.

xoriole: Good point! addressed.

src/tribler/core/utilities/tracker_utils.py
xoriole: Refactored it differently now.

src/tribler/core/utilities/tests/test_tracker_utils.py
```diff
@@ -6,137 +6,74 @@
...
+    (f'http://{TRACKER_HOST}:6969/announce/', f'http://{TRACKER_HOST}:6969/announce'),  # Trailing /
+    (f'http://{TRACKER_HOST}', None),  # HTTP no path
+    (f'http://{TRACKER_HOST}\xa0', None),  # HTTP no path nbsp
+    (f'http://{TRACKER_HOST}/announce', f'http://{TRACKER_HOST}/announce'),  # HTTP default port
+    (f'http://{TRACKER_HOST}:80/announce', f'http://{TRACKER_HOST}/announce'),  # HTTP default port given
+    (f'http://{TRACKER_HOST}/?do=upload', None),  # Bad URL encoding
+    (f'http://{TRACKER_HOST}:80/anno...', None),  # Truncated URL
+
+    (f'https://{TRACKER_HOST}/announce', f'https://{TRACKER_HOST}/announce'),  # HTTPS default port
+    (f'https://{TRACKER_HOST}:443/announce', f'https://{TRACKER_HOST}/announce'),  # HTTPS default port given
```
xoriole, line 28: > Why is this code necessary? Can you provide an example of the incorrect URL that this code fixes? (I suspect the fixed URL is not entirely correct and can include an excess /, but may be wrong on that)

@kozlovsky As in this test, if the default port is given, then it is removed from the expected uniform URL.

src/tribler/core/utilities/tracker_utils.py
xoriole: Addressed

src/tribler/core/utilities/tracker_utils.py
xoriole: Refactored it differently now.

src/tribler/core/utilities/tracker_utils.py
xoriole: Addressed

src/tribler/core/utilities/tracker_utils.py
xoriole: Addressed

src/tribler/core/utilities/tracker_utils.py
xoriole: HTTPS_PORT number is extracted

drew2a (approved):
Great 🤝 !

src/tribler/core/utilities/tests/test_tracker_utils.py
kozlovsky: Ah, ok, I misread it

kozlovsky (approved):

PR #7370, opened 2023-04-17
Fix KeyError for translations

### LINKED ISSUE #7363: [7.13.0-RC1]  KeyError: 'semanas'
Sentry Issue: [TRIBLER-11V](https://sentry.tribler.org/organizations/tribler/issues/1917/?referrer=github_integration)

```python
CreationTraceback: 
  File "run_tribler.py", line 103, in <module>
  File "tribler\gui\start_gui.py", line 84, in run_gui
  File "tribler\gui\utilities.py", line 413, in trackback_wrapper
  File "tribler\gui\widgets\downloadspage.py", line 142, in load_downloads
  File "tribler\gui\network\request_manager.py", line 50, in get
  File "tribler\gui\network\request.py", line 77, in __init__

KeyError: 'semanas'
(1 additional frame(s) were not displayed)
...
  File "tribler\gui\utilities.py", line 413, in trackback_wrapper
  File "tribler\gui\widgets\downloadspage.py", line 190, in on_received_downloads
  File "tribler\gui\widgets\downloadwidgetitem.py", line 69, in update_with_download
  File "tribler\gui\widgets\downloadwidgetitem.py", line 104, in update_item
  File "tribler\gui\utilities.py", line 161, in duration_to_string
```

drew2a:
This PR fixes #7363 by fixing the root cause of the issue — incorrect translation files.

Also, it modifies the `tr`  function in a way that it is safe from this type of error:

```python
class TranslatedString(str):
    """ This class is used to wrap translated strings to be able to log untranslated strings in case of errors.
        Thanks to this class no `KeyError` exceptions are raised when a translation is missing.
    """

    def __new__(cls, translation, original_string):  # pylint: disable=unused-argument
        return super().__new__(cls, translation)

    def __init__(self, translation: str, original_string: str):  # pylint: disable=unused-argument
        super().__init__()
        self.original_string = original_string

    def __mod__(self, other):
        try:
            return str.__mod__(self, other)
        except KeyError as e:
            msg = f'No value provided for {e} in translation "{self}", original string: "{self.original_string}"'
            logger.warning(f'{type(e).__name__}: {msg}')
            return self.original_string % other


def tr(key):
    return f"{QCoreApplication.translate('@default', key)}"
    translated_string = QCoreApplication.translate('@default', key)
    return TranslatedString(translated_string, original_string=key)
```


kozlovsky (changes requested):
A nice update of the `tr` function!
It is also necessary to convert .ts files to .qm files by executing the `update_translations.sh` script and commit the updated .qm files.

drew2a:
@kozlovsky it gives me the following error:
```shell
➜ ./update_translations.sh
./update_translations.sh: line 3: lrelease: command not found
```

kozlovsky (changes requested):
I tested Tribler with ES_ES locale, and it still crashes on
```
    <message>
        <location filename="../utilities.py" line="166"/>
        <source>%(seconds)is</source>
        <translation>%(segundos)is</translation>
    </message>
```

kozlovsky:
It looks like intercepting `KeyError` inside the `tr` function doesn't work because the actual error happens outside, for example:

```python
return tr("%(seconds)is") % data
```
Here, the `tr` function returns an incorrect translation with a wrong key, and then the `%` operation is applied to this returned translation string.

The `QCoreApplication.translate` call inside the `tr` function doesn't raise any `KeyError`, so it is better to remove try/except here.

If we want to prevent this type of `KeyError` exception, we can return a special subclass of a string with overridden `%` method that catches `KeyError` and returns the result of `%` applied to the original string.

kozlovsky:
For example, we can use the following `TranslatedString` class to catch KeyError exceptions and properly log warnings:
```python
class TranslatedString(str):
    __slots__ = 'original_string'

    def __new__(cls, translation, original_string):
        return super().__new__(cls, translation)

    def __init__(self, translation, original_string):
        super().__init__()
        self.original_string = original_string

    def __mod__(self, other):
        try:
            return str.__mod__(self, other)
        except KeyError as e:
            msg = f'No value provided for {e} in translation "{self}", original string: ({self.original_string})'
            logger.warning(f'{type(e).__name__}: {msg}')
            return self.original_string % other


def tr(key):
    translated_string = QCoreApplication.translate('@default', key)
    return TranslatedString(translated_string, original_string=key)
```

Tests:
```python
def test_correct_translation():
    original_string = 'original %(key1)s string'
    translated_string = 'translated %(key1)s string'
    s = TranslatedString(translated_string, original_string)
    assert s % {'key1': 'VALUE'} == 'translated VALUE string'


@patch('tribler.gui.utilities.logger.warning')
def test_missed_key_in_translated_string(warning: Mock):
    original_string = 'original %(key1)s string'
    translated_string = 'translated %(key2)s string'
    s = TranslatedString(translated_string, original_string)

    assert s % {'key1': 'VALUE'} == 'original VALUE string'

    warning.assert_called_once_with(
        'KeyError: No value provided for \'key2\' in translation "translated %(key2)s string", '
        'original string: (original %(key1)s string)')


@patch('tribler.gui.utilities.logger.warning')
def test_missed_key_in_translated_and_original_string(warning: Mock):
    original_string = 'original %(key1)s string'
    translated_string = 'translated %(key2)s string'
    s = TranslatedString(translated_string, original_string)

    with pytest.raises(KeyError, match=r"^'key1'$"):
        s % {'key3': 'value'}

    warning.assert_called_once_with(
        'KeyError: No value provided for \'key2\' in translation "translated %(key2)s string", '
        'original string: (original %(key1)s string)')
```

kozlovsky (changes requested):
This part of the .ts file still needs to be fixed (with recompilation of the .qm file):
```
    <message>
        <location filename="../utilities.py" line="166"/>
        <source>%(seconds)is</source>
        <translation>%(segundos)is</translation>
    </message>
```

src/tribler/gui/i18n/es_ES.ts
```diff
@@ -1123,32 +1123,32 @@ Si no está seguro, pulse &quot;No&quot;. Más adelante podrá eliminar esas car
...
         <location filename="../utilities.py" line="157"/>
         <source>%(years)iy %(weeks)iw</source>
-        <translation>%(años)iy %(semanas)iw</translation>
+        <translation>%(years)ia %(weeks)is</translation>
     </message>
     <message>
         <location filename="../utilities.py" line="159"/>
         <source>%(weeks)iw %(days)id</source>
-        <translation>%(semanas)iw %(días)id</translation>
+        <translation>%(weeks)is %(days)id</translation>
```
kozlovsky, line 1131: **nitpick:** Currently, both weeks (semanas) and seconds (segundos) are abbreviated as `s`. I suspect it can cause some confusion. [According to the Wikipedia](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Spain), in Spain, "the letters "m" and "s" are used to denote minutes and seconds". With this, I'd prefer to leave "weeks" without an abbreviation:
```suggestion
        <translation>%(weeks)i semanas %(days)i días</translation>
```
But we can leave it as is as well.

Issue #7360, opened 2023-04-12
10x speedup of all downloads in Tribler when switching to Search tab and performing a search


kozlovsky:
We have a user report https://github.com/Tribler/dispersy/issues/576 from @casbtosint; according to it, all downloads are approx 10x faster when switching to the Search tab during the download and performing a search. It looks like we have some bug in Tribler, and opened Downloads page causes some drastic slowdown.

You can see the demonstration of the effect in the following video, starting from 3rd minute:
https://peervideo.club/w/sznmzZvyd4dLm817xgwQwT

We should investigate the reason for the slowdown and fix it; it should have a huge effect on Tribler's performance.

### Steps to reproduce:

1. Start a download in Tribler.
2. Observe the download speed while on the Downloads page.
3. Switch to the Search tab and perform a search.
4. Observe the increase of the network traffic and download speed while the Search tab is open

drew2a:
It could be related to the following behavior of Tribler: if the downloads tab is opened, it produces a lot of `downloads?get_pieces=1` requests to the core:

```python
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
INFO - RequestManager(128) - Request: GET http://localhost:20100/downloads?get_pieces=1
```

ichorid:
[...]

PR #7377, opened 2023-04-19
Decrease the frequency of the `downloads` request

drew2a:
This PR is related to #7360 and should decrease the load by decreasing the frequency of the `downloads` request.

`Background refresh interval` has been changed from `1s` to `5s`. 
`Refresh now interval` has been changed from `0s` to `0.01s` (https://doc.qt.io/qt-5/qtimer.html#start).

This PR also offers an alternative solution for the problem addressed here: https://github.com/Tribler/tribler/pull/7372

### DIFF
--- a/src/tribler/gui/widgets/downloadspage.py
+++ b/src/tribler/gui/widgets/downloadspage.py
@@ -27,6 +27,13 @@ from tribler.gui.utilities import compose_magnetlink, connect, format_speed, tr
 from tribler.gui.widgets.downloadwidgetitem import DownloadWidgetItem, LoadingDownloadWidgetItem
 from tribler.gui.widgets.loading_list_item import LoadingListItem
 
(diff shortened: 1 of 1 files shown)


src/tribler/gui/widgets/downloadspage.py
```diff
@@ -109,13 +118,16 @@ def start_loading_downloads(self):
         self.loading_list_item = LoadingListItem(self.window().downloads_list)
         self.window().downloads_list.addTopLevelItem(self.loading_message_widget)
         self.window().downloads_list.setItemWidget(self.loading_message_widget, 2, self.loading_list_item)
-        self.schedule_downloads_timer(now=True)
```
drew2a, line 112: It is not necessary to refresh the downloads on the Tribler start because they will be refreshed when the page is shown.

xoriole (approved):

PR #7383, opened 2023-04-24
Check on ancient version

### LINKED ISSUE #7369: UpgradeError: OperationalError: no such table: MiscData
```python
  File "tribler\gui\upgrade_manager.py", line 45, in run, 
  File "tribler\gui\upgrade_manager.py", line 80, in upgrade_state_dir, 
  File "tribler\core\upgrade\upgrade.py", line 97, in run, 
  File "tribler\core\upgrade\upgrade.py", line 371, in upgrade_pony_db_8to10, 
  File "tribler\core\upgrade\db8_to_db10.py", line 247, in get_db_version, 
sqlite3.OperationalError: no such table: MiscData, 
  File "tribler\gui\upgrade_manager.py", line 209, in on_worker_finished, 
tribler.gui.exceptions.UpgradeError: OperationalError: no such table: MiscData
```

Note that `upgrade_pony_db_8to10` is the very first step of the current upgrade procedure:
```python
    def run(self):
        """
        Run the upgrader if it is enabled in the config.
        """
        self._logger.info('Run')

        self.upgrade_pony_db_8to10()
        self.upgrade_pony_db_10to11()
        ...
```

The previous upgrade steps from versions of Tribler < 7.5 were removed in #6558.

So it seems like a user is trying to upgrade from the very old Tribler version. In that case, we can detect the missing `MiscData` table and add a popup window that suggests the user first install some older version of Tribler (say, 7.10) and upgrade to it, and then install the latest Tribler version. The number of users who still use old Tribler versions is small, so it is not worth supporting all ancient versions directly, but an appropriate error message with specific suggestions to the user can be helpful.

drew2a:
This PR fixes #7369


src/tribler/core/upgrade/version_manager.py
```diff
@@ -72,14 +72,14 @@ class TriblerVersion:
...
+    def __init__(self, root_state_dir: Path, version_str: str, files_to_copy: List[str] = None,
                  last_launched_at: Optional[float] = None):
         if last_launched_at is None:
             last_launched_at = time.time()
         self.logger = logging.getLogger(self.__class__.__name__)
         self.version_str = version_str
-        self.version_tuple = tuple(LooseVersion(version_str).version)
-        self.major_minor = self.version_tuple[:2]
+        self.version = LooseVersion(version_str)
+        self.major_minor = tuple(self.version.version[:2])
```
kozlovsky, line 82: **nitpick:** `self.version.version[:2]` should already be a tuple, so it is not necessary to use `tuple(...)` here

```suggestion
        self.major_minor = self.version.version[:2]
```

src/tribler/gui/upgrade_manager.py
```diff
@@ -157,9 +162,9 @@ def should_cleanup_old_versions(self) -> List[TriblerVersion]:
             state_size = version.calc_state_size()
             claimable_storage += state_size
             storage_info += f"{version.version_str} \t {format_size(state_size)}\n"
-
+        self._logger.info(f'Storage info: {storage_info}')
         # Show a question to the user asking if the user wants to remove the old data.
-        title = "Delete state directories for old versions?"
+        title = tr("Delete state directories for old versions?")
```
kozlovsky, line 167: We need also to add a corresponding translations (not necessary in this PR)

kozlovsky (approved):

src/tribler/core/upgrade/version_manager.py
drew2a: `self.version.version` is a list:

```python
class LooseVersion (Version):
    component_re = re.compile(r'(\d+ | [a-z]+ | \.)', re.VERBOSE)

    def __init__ (self, vstring=None):
        if vstring:
            self.parse(vstring)


    def parse (self, vstring):
        self.vstring = vstring
        components = [x for x in self.component_re.split(vstring)
                              if x and x != '.']
        for i, obj in enumerate(components):
            try:
                components[i] = int(obj)
            except ValueError:
                pass

        self.version = components
```

PR #7401, opened 2023-05-02
Fix `FileNotFoundError`

### LINKED ISSUE #7396: [7.12.1] FileNotFoundError: [WinError 3] Системе не удается найти указанный путь
Sentry Issue: [TRIBLER-162](https://sentry.tribler.org/organizations/tribler/issues/2088/?referrer=github_integration)

```
CreationTraceback: 
  File "run_tribler.py", line 100, in <module>
  File "tribler\gui\start_gui.py", line 77, in run_gui
  File "tribler\gui\tribler_window.py", line 793, in on_settings_button_click
  File "tribler\gui\widgets\settingspage.py", line 344, in load_settings
  File "tribler\gui\tribler_request_manager.py", line 181, in __init__

FileNotFoundError: [WinError 3] Системе не удается найти указанный путь: 'C:\\Users\\<user>\\AppData\\Roaming\\.Tribler\\7.12\\channels\\6ccd7a08817feb2f1971382a2945de390c92cdda3d09c355'
(6 additional frame(s) were not displayed)
...
  File "tribler\gui\widgets\settingspage.py", line 247, in _version_dir_checkbox
  File "tribler\gui\widgets\settingspage.py", line 247, in <genexpr>
  File "pathlib.py", line 1136, in glob
  File "pathlib.py", line 583, in _select_from
  File "pathlib.py", line 530, in _select_from
```

drew2a:
This PR fixes #7396

### DIFF
--- a/src/tribler/core/utilities/path_util.py
+++ b/src/tribler/core/utilities/path_util.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import itertools
+import os
 import pathlib
 import sys
 import tempfile
@@ -36,8 +38,26 @@ class Path(type(pathlib.Path())):
 
         return self
 
-    def size(self) -> int:
-        return self.stat().st_size
+    def size(self, include_dir_sizes: bool = True) -> int:
+        """ Return the size of this file or directory (recursively).
+
+        Args:
+            include_dir_sizes: If True, return the size of files and directories, not the size of files only.
+
+        Returns: The size of this file or directory.
+        """
+        if not self.exists():
+            return 0
+
+        if self.is_file():
+            return self.stat().st_size
+
+        size = os.path.getsize(self.absolute()) if include_dir_sizes else 0  # get root dir size
+        for root, dir_names, file_names in os.walk(self):
+            names = itertools.chain(dir_names, file_names) if include_dir_sizes else file_names
+            paths = (os.path.join(root, name) for name in names)
+            size += sum(os.path.getsize(p) for p in paths if os.path.exists(p))
+        return size
 
     def startswith(self, text: str) -> bool:
         return self.match(f"{text}*")
--- a/src/tribler/core/utilities/tests/test_path_utils.py
+++ b/src/tribler/core/utilities/tests/test_path_utils.py
@@ -3,6 +3,12 @@ import pytest
 from tribler.core.utilities.path_util import Path, tail
 
 
+# pylint: disable=redefined-outer-name
+@pytest.fixture
+def tribler_tmp_path(tmp_path):
+    return Path(tmp_path)
+
+
 def test_put_path_relative(tmpdir):
     assert Path(tmpdir).normalize_to(None) == Path(tmpdir)
     assert Path(tmpdir).normalize_to('') == Path(tmpdir)
@@ -23,16 +29,16 @@ def test_tail_no_file():
         tail('missed_file.txt')
 
 
-def test_tail_small_file(tmpdir: Path):
+def test_tail_small_file(tribler_tmp_path: Path):
     """Test that tail works correct with a small file """
-    log_file = tmpdir / 'log.txt'
+    log_file = tribler_tmp_path / 'log.txt'
     log_file.write_text('text', 'utf-8')
     assert tail(log_file) == 'text'
 
 
-def test_tail_count(tmpdir: Path):
+def test_tail_count(tribler_tmp_path: Path):
     """Test that tail returns desired count of lines"""
-    log_file = tmpdir / 'log.txt'
+    log_file = tribler_tmp_path / 'log.txt'
 
     # add 100 lines
     content = '\n'.join(f'{i}' for i in range(100))
@@ -47,7 +53,7 @@ def test_tail_count(tmpdir: Path):
         tail(log_file, -1)
 
 
-def test_tail_encodings(tmpdir: Path):
+def test_tail_encodings(tribler_tmp_path: Path):
     """Test that the `tail` function can read logs with "utf-8", "ascii", "latin-1" encodings """
     encodings = ["utf-8", "ascii", "latin-1"]
     log_files = []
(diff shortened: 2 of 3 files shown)


xoriole (commented):
It could be that one or more files are unavailable. So, I would suggest iterating over the files and checking the filesize (and adding them) if exists instead of setting the full size to zero. Zero as size would likely not be a representative state dir size representation if it is only a single channel file that is not found.

drew2a:
The exception occurs during the iteration over the iterator returned by `Path.glob()`:

```python
    def glob(self, pattern):
        """Iterate over this subtree and yield all existing files (of any
        kind, including directories) matching the given relative pattern.
        """
        sys.audit("pathlib.Path.glob", self, pattern)
        if not pattern:
            raise ValueError("Unacceptable pattern: {!r}".format(pattern))
        drv, root, pattern_parts = self._flavour.parse_parts((pattern,))
        if drv or root:
            raise NotImplementedError("Non-relative patterns are unsupported")
        selector = _make_selector(tuple(pattern_parts), self._flavour)
        for p in selector.select_from(self): # <- here
            yield p
```

The `FileNotFoundError` exception raised by this iterator must be handled outside of the iterator, and the iteration cannot continue. @kozlovsky, am I correct?

If we want a more accurate size (than zero) of the folder when there are bad files inside, we should change the approach to calculating the size.

The current approach is:
```python
sum(f.lstat().st_size for f in state_dir.glob('**/*')
```

drew2a:
As an option, slightly changed the current approach (will stop iterating on the first bad file/folder):

```python
    def _dir_size(self, state_dir):
        size = 0
        try:
            for p in state_dir.glob('**/*'):
                if p.exists():
                    size += p.lstat().st_size
        except FileNotFoundError as e:
            self.logger.exception(e)

        return size
```

`os.walk`:

```python
    def _dir_size(self, state_dir):
    size = 0
    for root, dirs, files in os.walk(directory):
        joined_files = (os.path.join(root, f) for f in files)
        size += sum(os.path.getsize(f) for f in joined_files if os.path.exists(f))
    return size
```

kozlovsky:
> The FileNotFoundError exception raised by this iterator must be handled outside of the iterator, and the iteration cannot continue, am I correct?

Yes, you are right; in most iterators, the iteration cannot be continued after an exception. Using `os.walk` (that you already use now) looks like the correct approach, as it suppresses errors by default and also allows the application code to handle them (for logging purposes, for example) by [specifying an `onerror` handler](https://docs.python.org/3/library/os.html#os.walk):

> By default, errors from the [scandir()](https://docs.python.org/3/library/os.html#os.scandir) call are ignored. If optional argument onerror is specified, it should be a function; it will be called with one argument, an [OSError](https://docs.python.org/3/library/exceptions.html#OSError) instance. It can report the error to continue with the walk, or raise the exception to abort the walk. Note that the filename is available as the filename attribute of the exception object.

src/tribler/core/utilities/path_util.py
```diff
@@ -36,8 +38,23 @@ def normalize_to(self, base: str = None) -> Path:
 
         return self
 
-    def size(self) -> int:
-        return self.stat().st_size
+    def size(self, only_files: bool = False) -> int:
+        """ Return the size of this file or directory (recursively).
+
+        Args:
+            only_files: If True, only return the size of files in this directory, not the size of files+directories.
```
kozlovsky, line 45: The `only_files` option looks a bit confusing. At the first read, I interpreted it as "calculate only the size of files in the top-level-directory and not in subdirectories" and thought it was better to name it `include_subdirectories=True` instead of `only_files=False`.

But then I realized that it means the size of directory entries that are often missed from the total size.

I like the idea of taking into account the size of directory entries, but I suspect that for other developers in the future, it also may not be totally obvious what this option is doing. Can we rename it to something like `include_directory_entry_sizes=True` or `include_dir_sizes=True`?

xoriole (approved):

src/tribler/core/utilities/path_util.py
```diff
@@ -36,8 +38,23 @@ def normalize_to(self, base: str = None) -> Path:
...
+        Returns: The size of this file or directory.
+        """
+        if self.is_file():
+            return self.stat().st_size
+
+        size = 0
+        for root, dir_names, file_names in os.walk(self):
+            names = file_names if only_files else itertools.chain(dir_names, file_names)
+            paths = (os.path.join(root, name) for name in names)
+            size += sum(os.path.getsize(p) for p in paths if os.path.exists(p))
```
kozlovsky, line 56: A very small nitpick, the function does not take into account the size of the top-level root directory entry. To be absolutely correct, it should be something like:
```suggestion
        for root, dir_names, file_names in os.walk(self):
            if not only_files and os.path.exists(root):
                size += os.path.getsize(root)  # add the size of the directory entry
            for file_name in file_names:
                file_path = os.path.join(root, file_name)
                if os.path.exists(file_path):
                    size += os.path.getsize(file_path)
```

kozlovsky (approved):

src/tribler/core/utilities/path_util.py
drew2a: Sure

src/tribler/core/utilities/path_util.py
drew2a: What about:

```python
    def size(self, only_files: bool = False) -> int:
        """ Return the size of this file or directory (recursively).

        Args:
            only_files: If True, only return the size of files in this directory, not the size of files+directories.

        Returns: The size of this file or directory.
        """
        if not self.exists():
            return 0
        
        if self.is_file():
            return self.stat().st_size

        size = os.path.getsize(self.absolute())
        for root, dir_names, file_names in os.walk(self):
            names = file_names if only_files else itertools.chain(dir_names, file_names)
            paths = (os.path.join(root, name) for name in names)
            size += sum(os.path.getsize(p) for p in paths if os.path.exists(p))
        return size
```

src/tribler/core/utilities/path_util.py
drew2a: Done

src/tribler/core/utilities/path_util.py
drew2a: Done

PR #7404, opened 2023-05-03
Suppress `ConnectionResetError`

### LINKED ISSUE #7345: [7.13.0-RC1] ConnectionResetError: Cannot write to closing transport
Sentry Issue: [TRIBLER-14H](https://sentry.tribler.org/organizations/tribler/issues/2027/?referrer=github_integration)

```python
ConnectionResetError: Cannot write to closing transport
  File "tribler\core\components\restapi\rest\events_endpoint.py", line 117, in write_data
  File "aiohttp\web_response.py", line 514, in write
  File "aiohttp\http_writer.py", line 116, in write
  File "aiohttp\http_writer.py", line 76, in _write
```

drew2a:
This PR fixes #7345 by suppressing `ConnectionResetError`.

`ConnectionResetError` can be raised in three cases:
1. there is no aiohttp connection
2. aiohttp transport is None
3. aiohttp transport is closing

https://github.com/aio-libs/aiohttp/blob/29b6f302ad6736f71ef3fb029d4c5bac786a2fb7/aiohttp/http_writer.py#L69-L76

```python
    def _write(self, chunk: bytes) -> None:
        size = len(chunk)
        self.buffer_size += size
        self.output_size += size
        transport = self.transport
        if not self._protocol.connected or transport is None or transport.is_closing():
            raise ConnectionResetError("Cannot write to closing transport")
        transport.write(chunk)
```

In all cases described above, we cannot do anything from the core side. All reconnect logic is on the GUI side.

The most probable root cause of the error is the situation in which the GUI goes through the shutdown procedure and closes the connection with the core. For this particular case, `ConnectionResetError` should be ignored on the core side anyway.

There is another way to fix the issue (@kozlovsky suggestion). 

>To fix it, we need:
> - Cancel all requests to Core only after the response to the shutdown request is received, not earlier;
> - Suppress ConnectionResetError in Sentry if Core is in the shutdown state.

However, this fix requires much more work, increases the complexity of the GUI-core interactions, and involves Sentry for suppressing the specific error.

Since this is not a critical issue to address, it is better to fix it in the most straightforward and simple way.

### DIFF
--- a/src/tribler/core/components/restapi/rest/events_endpoint.py
+++ b/src/tribler/core/components/restapi/rest/events_endpoint.py
@@ -37,7 +37,6 @@ topics_to_send_to_gui = [
     notifications.report_config_error,
 ]
 
-
 MessageDict = Dict[str, Any]
 
 
@@ -143,8 +142,17 @@ class EventsEndpoint(RESTEndpoint):
             self._logger.exception(e)
             return
 
+        processed_responses = []
         for response in self.events_responses:
-            await response.write(message_bytes)
+            try:
+                await response.write(message_bytes)
+                # by creating the list with processed responses we want to remove responses with
+                # ConnectionResetError from `self.events_responses`:
+                processed_responses.append(response)
+            except ConnectionResetError as e:
+                # The connection was closed by GUI
+                self._logger.warning(e, exc_info=True)
+        self.events_responses = processed_responses
 
     # An exception has occurred in Tribler. The event includes a readable
     # string of the error and a Sentry event.
--- a/src/tribler/core/components/restapi/rest/tests/test_events_endpoint.py
+++ b/src/tribler/core/components/restapi/rest/tests/test_events_endpoint.py
@@ -1,10 +1,12 @@
 import asyncio
 import json
+import logging
 from asyncio import CancelledError, Event, create_task
 from contextlib import suppress
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


src/tribler/core/components/restapi/rest/events_endpoint.py
```diff
@@ -144,7 +143,11 @@ async def _write_data(self, message: MessageDict):
             return
 
         for response in self.events_responses:
-            await response.write(message_bytes)
+            try:
+                await response.write(message_bytes)
+            except ConnectionResetError as e:
+                # The connection was closed by GUI
+                self._logger.warning(e, exc_info=True)
```
kozlovsky, line 150: I think we should also remove the `response` object with closed connection from `self.event_responses`. Note that we should not remove an item from a list during the iteration. To handle this, we can copy the list of event_responses before the loop:
```python
        for response in self.events_responses.copy():
            try:
                await response.write(message_bytes)
            except ConnectionResetError as e:
                # The connection was closed by GUI
                self._logger.warning(e, exc_info=True)
                self.events_responses.remove(response)
```

kozlovsky (changes requested):

src/tribler/core/components/restapi/rest/events_endpoint.py
drew2a: Yes, you are right! I'll implement the suggestion a bit later.

kozlovsky (approved):

Issue #7406, opened 2023-05-08
UnicodeDecodeError: unexpected end of data in `on_tracker_error_alert`


kozlovsky:
[Here](https://github.com/Tribler/tribler/blob/504eb18640a0f39eb76bb614aae24080f35603aa/src/tribler/core/components/libtorrent/download_manager/download.py#L315):

```python
    def on_tracker_error_alert(self, alert: lt.tracker_error_alert):
        self._logger.error(f'On tracker error alert: {alert}')  # <-- here

        # try-except block here is a workaround and has been added to solve
        # the issue: "UnicodeDecodeError: invalid continuation byte"
        try:
            ...
```

It is a regression of #5740: f-string formatting of the `logging.error` output should also be placed inside the following try/except block to have protection from UnicodeDecodeError.

As an additional protection, we can add a helper function like `alert_to_str` and use it in all places where we have `{alert}` formatting to be sure that all alerts can be safely converted to str

PR #7410, opened 2023-05-08
Add `_safe_log`

drew2a:
This PR fixes https://github.com/Tribler/tribler/issues/7406 i by introducing the `_safe_log` function.

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/download.py
+++ b/src/tribler/core/components/libtorrent/download_manager/download.py
@@ -184,7 +184,7 @@ class Download(TaskManager):
         return atp
 
     def on_add_torrent_alert(self, alert: lt.add_torrent_alert):
-        self._logger.info(f'On add torrent alert: {alert}')
+        self._safe_log(logging.INFO, 'On add torrent alert', alert)
 
         if hasattr(alert, 'error') and alert.error.value():
             self._logger.error("Failed to add torrent (%s)", self.tdef.get_name_as_unicode())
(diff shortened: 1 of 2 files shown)


kozlovsky (changes requested):
The PR solves the problem by introducing a new method of the `Download` class that should be used instead of standard logging methods. It catches possible exceptions that arise during the log message formatting.

Although the proposed change solves the problem, the chosen solution has some drawbacks:
* The original essence of the problem has nothing to do with logging. It is related to the fact that converting an alert to a string can lead to an error. At the same time, the proposed solution is explicitly tied to logging, replacing standard logging methods with a new non-standard method.
* The proposed new logging method assumes a rigid message structure when the only possible format is given inside a helper function. It looks inflexible and not versatile enough.

I would suggest considering alternative solutions:

Instead of the `_safe_log` method of the `Download` class, we could use the more generic `safe_repr` function, applying it to alert objects. With this approach, the developers:
* will be able to continue to use the usual methods for logging
* will not be limited in the structure of the message used when logging
* will be able to apply `safe_repr` to alert objects in other situations not related to logging.
* the meaning of the code will look clearer for readers since the reader will see that the original problem is not in logging as such but in the string display of the alert.
* The `safe_repr` function can be made generic and applicable not only to alerts but also to other types of objects - by catching all exceptions thrown when calling `repr(obj)`, it can output some safe alternative result, for example, including the object type and its id.

drew2a:
I think the solution proposed in the PR is good enough. It solves the problem. It is simple. It is straightforward. The format with `f'{message}: {alert}'` is flexible enough to cover all calls in the whole almost 1k file.

xoriole:
In my opinion, it is possible that the error `tracker_error_alert` is only raising the exception.
Here is an old ticket (also fixed): https://github.com/arvidn/libtorrent/issues/143
Also, the error is reported for Windows only. So, it would be great if the issue could be actually reproduced.

PS. The link on the issue ticket is wrong. It should be #7406 instead of #7404

xoriole (changes requested):
I'd agree with @kozlovsky's idea of `safe_repr` applying to alert objects instead of `_safe_log`.

PR #7414, opened 2023-05-09
Detect slow coroutines and freezes in the asyncio loop

kozlovsky:
This PR adds functionality to detect slow coroutine step execution in the asyncio loop. If a coroutine step requires too much time (more than 1 second by default), the coroutine is reported to the error log.

A separate thread watches whether the main line is currently inside the coroutine for too long and reports a possible freeze.
 
It is possible to specify an environment variable `SLOW_CORO_STACK_TRACING=1`; in that case, the coroutine stack is also tracked and reported in case of a possible freeze. With stack tracing enabled, the Core works slower, so this option is intended to be run not on user machines (although this is possible).

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -11,7 +11,9 @@ import encodings.idna  # pylint: disable=unused-import
 
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter, SentryStrategy
 from tribler.core.sentry_reporter.sentry_scrubber import SentryScrubber
+from tribler.core.utilities.slow_coro_detection.main_thread_stack_tracking import start_main_thread_stack_tracing
 from tribler.core.utilities.osutils import get_root_state_directory
+from tribler.core.utilities.utilities import is_frozen
 from tribler.core.version import version_id
 
 logger = logging.getLogger(__name__)
@@ -94,6 +96,13 @@ if __name__ == "__main__":
         from tribler.core.components.reporter.exception_handler import default_core_exception_handler
 
         init_sentry_reporter(default_core_exception_handler.sentry_reporter)
+
+        slow_coro_stack_tracking = os.environ.get('SLOW_CORO_STACK_TRACING', '0' if is_frozen() else '1')
+        # By default, the stack tracking of slow coroutines is enabled when running the Tribler from sources
+        # and disabled in the compiled version, as it makes the Python code of Core work slower.
+        if slow_coro_stack_tracking == '1':
+            start_main_thread_stack_tracing()
+
         run_core(api_port, api_key, root_state_dir, parsed_args)
     else:  # GUI
         from tribler.gui.start_gui import run_gui
--- a/src/tribler/core/start_core.py
+++ b/src/tribler/core/start_core.py
@@ -39,6 +39,7 @@ from tribler.core.config.tribler_config import TriblerConfig
 from tribler.core.logger.logger import load_logger_config
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter, SentryStrategy
 from tribler.core.upgrade.version_manager import VersionHistory
+from tribler.core.utilities import slow_coro_detection
 from tribler.core.utilities.process_manager import ProcessKind, ProcessManager, TriblerProcess, \
     set_global_process_manager
 
@@ -132,6 +133,9 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     logger.info(f'Start tribler core. API port: "{api_port}". '
                 f'API key: "{api_key}". State dir: "{state_dir}". '
                 f'Core test mode: "{gui_test_mode}"')
+
+    slow_coro_detection.patch_asyncio()  # Track the current coroutine handled by asyncio
+    slow_coro_detection.start_watching_thread()  # Run a separate thread to watch for the main thread asyncio freezes
 
     config = TriblerConfig.load(state_dir=state_dir, reset_config_on_error=True)
     config.gui_test_mode = gui_test_mode
--- a/src/tribler/core/components/session.py
+++ b/src/tribler/core/components/session.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import os
 import sys
+import time
 from asyncio import Event, create_task, gather, get_event_loop
 from pathlib import Path
 from typing import Dict, List, Optional, Type, TypeVar
@@ -79,7 +80,8 @@ class Session:
         component.session = self
 
     async def start_components(self):
-        self.logger.info('Start components...')
+        t = time.time()
+        self.logger.info('Starting components...')
         self.logger.info(f'State directory: "{self.config.state_dir}"')
         create_state_directory_structure(self.config.state_dir)
         patch_crypto_be_discovery()
@@ -90,8 +92,12 @@ class Session:
 
         coros = [comp.start() for comp in self.components.values()]
         await gather(*coros, return_exceptions=not self.failfast)
-        if self._startup_exception:
+        duration = time.time() - t
+        if e := self._startup_exception:
+            self.logger.warning(f'Components started in {duration:.3f} seconds with exception: {type(e).__name__}: {e}')
             self._reraise_startup_exception_in_separate_task()
+        else:
+            self.logger.info(f'All components started in {duration:.3f} seconds')
 
     def _reraise_startup_exception_in_separate_task(self):
         self.logger.info('Reraise startup exception in separate task')
--- a/src/tribler/core/utilities/slow_coro_detection/__init__.py
+++ b/src/tribler/core/utilities/slow_coro_detection/__init__.py
@@ -1,0 +1,7 @@
+# pylint: disable=wrong-import-position
+
+import logging
+logger = logging.getLogger(__name__)
+
+from .patch import patch_asyncio
+from .watching_thread import start_watching_thread
--- a/src/tribler/core/utilities/slow_coro_detection/main_thread_stack_tracking.py
+++ b/src/tribler/core/utilities/slow_coro_detection/main_thread_stack_tracking.py
@@ -1,0 +1,95 @@
+import linecache
+import sys
+from types import FrameType
+from typing import Callable, List, Optional, Tuple
+
+from tribler.core.utilities.slow_coro_detection import logger
+
+# The default interval Python uses to switch threads is 0.005 seconds. When obtaining the main stack,
+# the _get_main_thread_stack_info() function temporarily switches it to a much bigger value of 0.1 seconds
+# to prevent a thread switch at that moment, so the debug thread can copy the main_thread_stack list content
+# without interruption.
+SWITCH_INTERVAL = 0.1
+
+_main_thread_stack: List[FrameType] = []
+
+
+_main_stack_tracking_activated: bool = False
+
+
+def main_stack_tracking_is_activated() -> bool:
+    return _main_stack_tracking_activated
+
+
+def main_thread_profile(frame: FrameType, event: str, _):
+    """
+    A hook that calls before and after a function call in the main thread if the stack tracking is activated
+    """
+    if event == 'call':
+        _main_thread_stack.append(frame)
+    elif event == 'return' and _main_thread_stack:
+        # Usually, 'call' and 'return' are always paired, so 'return' removes the frame added by the previous 'call'.
+        # But at the very beginning, when the `start_main_thread_stack_tracing` function is called, we are already
+        # inside the function, so we get unpaired `return` when we exit from the function. At that moment, the stack
+        # is empty. By checking that the stack can be empty, we handle this situation by ignoring the unpaired 'return'.
+        _main_thread_stack.pop()
+    return main_thread_profile
+
+
+def start_main_thread_stack_tracing() -> Callable:
+    """
+    Activates the profiler hook in the main thread. Note that it makes Python functions about two times slower.
+    The compiled code is run as fast, so libtorrent calls and database queries should be as efficient as before.
+
+    Returns the profiler function (for testing purpose)
+    """
+    logger.info('Start tracing of coroutine stack to show stack for slow coroutines (makes code execution slower)')
+    global _main_stack_tracking_activated  # pylint: disable=global-statement
+    _main_stack_tracking_activated = True
+    sys.setprofile(main_thread_profile)
+    return main_thread_profile
+
+
+def stop_main_thread_stack_tracing() -> Callable:
+    """
+    Deactivates the profiler hook in the main thread.
+    Returns the previous profiler function (for testing purpose)
+    """
+    previous_profiler = sys.getprofile()
+    sys.setprofile(None)
+    global _main_stack_tracking_activated  # pylint: disable=global-statement
+    _main_stack_tracking_activated = False
+    return previous_profiler
+
+
+def _get_main_thread_stack_info() -> List[Tuple[str, str, Optional[int]]]:
+    """
+    Quickly copies necessary information from the main thread stack, so it is possible later to format a usual
+    traceback in a separate thread.
+
+    The function temporarily changes the interpreter’s thread switch interval to prevent thread switch during
+    the stack copying. It is a lighter analogue of holding the GIL (Global Interpreter Lock).
+    """
+    previous_switch_interval = sys.getswitchinterval()
+    sys.setswitchinterval(SWITCH_INTERVAL)
+    try:
+        stack_info = [(frame.f_code.co_name, frame.f_code.co_filename, frame.f_lineno)
+                      for frame in _main_thread_stack]
+    finally:
+        sys.setswitchinterval(previous_switch_interval)
+    return stack_info
+
+
+def get_main_thread_stack() -> str:
+    """
+    Obtains the main thread stack and format it in a usual way.
+    """
+    lines = ['Traceback (most recent call last):']
+    stack_info = _get_main_thread_stack_info()
+    for func_name, file_name, line_number in stack_info:
+        line = f'  File "{file_name}", line {line_number or "?"}, in {func_name}'
+        lines.append(line)
+        if line_number:
+            source_line = linecache.getline(file_name, line_number)
+            lines.append('    ' + (source_line.strip() or '?'))
+    return '\n'.join(lines) + '\n'
--- a/src/tribler/core/utilities/slow_coro_detection/patch.py
+++ b/src/tribler/core/utilities/slow_coro_detection/patch.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+
+import time
+from asyncio import Handle
+
+from tribler.core.utilities.slow_coro_detection import logger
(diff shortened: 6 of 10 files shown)


drew2a:
> the Core works slower, so this option is intended to be run not on user machines (although this is possible).

How much slower?

drew2a:
> the Core works slower, so this option is intended to be run not on user machines (although this is possible).

How much slower?

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
+import linecache
+import sys
+from types import FrameType
+from typing import List, Optional, Tuple
+
+SWITCH_INTERVAL = 0.1
```
drew2a, line 6: NIT: Could you please consider adding a comment to this constant? It's not clear why the value is 0.1 and how it affects the logic of the module.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+from tribler.core.utilities.main_thread_stack_tracking import get_main_thread_stack
+
+from tribler.core.utilities.main_thread_stack_tracking import main_stack_tracking_is_activated
+
+# pylint: disable=protected-access
+
+logger = logging.getLogger(__name__)
+
+SLOW_CORO_DURATION_THRESHOLD = 1.0
+WATCHING_THREAD_INTERVAL = 1.0
```
drew2a, line 18: The relationship between these constants is unclear. It might be better (though I'm not certain) to have just a single constant, as they could easily be misconfigured.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class DebugInfo:
+    def __init__(self):
+        self.handle = None
+        self.start_time = None
```
drew2a, line 24: NIT: Types could be added here.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+current = DebugInfo()
+lock = Lock()
```
drew2a, line 28: In this file, there are several global variables scattered throughout. It would be easier to read the file if they were all located in one place, preferably at the beginning of the file.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class SlowCoroWatchingThread(Thread):
+    """
+    A thread that detects and reports slow coroutines.
+    """
+    def run(self):
+        prev_reported_handle = None  # to detect in the loop when we have the second report of the same slow coroutine
```
drew2a, line 36: NIT is the same as above (could be a matter of preference). 
```suggestion
        previous_handle = None  # to detect in the loop when we have the second report of the same slow coroutine
```

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+        while True:
+            time.sleep(WATCHING_THREAD_INTERVAL)
+            with lock:
+                handle, start_time = current.handle, current.start_time
+            if handle is not None:
```
drew2a, line 41: The level of indentation could be reduced with the use of an inverted 'if' condition:

```python
            if handle is None:
                continue
```

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def report_freeze(handle: Handle, duration: float, first_report: bool):
+    if not first_report:
+        info_str = format_info(handle, include_stack=False)
+        logger.error(f"Still executing {info_str}")
+        return
+
+    info_str = format_info(handle, include_stack=True)
+    logger.error(f'Slow coroutine is occupying the loop for {duration:.3f} seconds already: {info_str}')
```
drew2a, line 88: This function would be easier to read if the blocks were rearranged, with the first report coming first, followed by all others.

```suggestion
def report_freeze(handle: Handle, duration: float, first_report: bool):
    if first_report:
        info_str = format_info(handle, include_stack=True)
        logger.error(f'Slow coroutine is occupying the loop for {duration:.3f} seconds already: {info_str}')
        return

    info_str = format_info(handle, include_stack=False)
    logger.error(f"Still executing {info_str}")
```

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+        original_handle_run(self)
+    finally:
+        with lock:
+            assert current.handle is self, "asyncio loop should never run one handle inside another one"
+            current.handle = current.start_time = None
+
+        duration = time.time() - start_time
+        if duration > SLOW_CORO_DURATION_THRESHOLD:
+            # The coroutine step is finished successfully (without freezing), but the execution time was too long
+            report_long_duration(self, duration)
```
drew2a, line 68: This appears to be somewhat duplicative, as this report seems to replicate (I'm not completely sure) the previous logic `SlowCoroWatchingThread.run()` as they both should be triggered in the case of a slow coroutine.
`SlowCoroWatchingThread.run()` will be triggered first and `patched_handle_run()` second.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+    to access it from the separate thread and detect slow coroutines.
+    """
+    start_time = time.time()
+    with lock:
+        current.handle, current.start_time = self, start_time
+    try:
+        original_handle_run(self)
+    finally:
+        with lock:
+            assert current.handle is self, "asyncio loop should never run one handle inside another one"
```
drew2a, line 62: Is this assert is really necessary? What is the purpose of it?

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+            # The coroutine step is finished successfully (without freezing), but the execution time was too long
+            report_long_duration(self, duration)
+
+        self = None  # Needed to break cycles when an exception occurs (copied from the original Handle._run method)
+
+
+original_handle_run = Handle._run
+
+
+def report_long_duration(handle: Handle, duration: float):
```
drew2a, line 76: It looks like a private function:
```suggestion
def _report_long_duration(handle: Handle, duration: float):
```

src/tribler/core/start_core.py
```diff
@@ -133,6 +134,9 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
                 f'API key: "{api_key}". State dir: "{state_dir}". '
                 f'Core test mode: "{gui_test_mode}"')
 
+    slow_coro_detection.patch_asyncio()  # Track the current coroutine handled by asyncio
```
xoriole, line 137: Since we have environment variables, we can do a conditional call to patch asyncio and to start watching thread. Same for setting loop.set_debug(True)

src/tribler/core/start_core.py
```diff
@@ -161,6 +165,7 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     loop = asyncio.get_event_loop()
     exception_handler = default_core_exception_handler
     loop.set_exception_handler(exception_handler.unhandled_error_observer)
+    # loop.set_debug(True)  # uncomment this line to see asyncio-generated warnings on slow coroutines
```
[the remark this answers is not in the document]
kozlovsky: I removed the comment, as asyncio-duplicate warnings are basically the same warnings but with fewer details and with a separately specified limit for slow callback duration.

Theoretically, it sometimes may be useful to set the asyncio loop to debug mode, as it performs some additional checks in this mode, so the comment makes it easier to find the place where the asyncio debug mode can be enabled, but it is probably not necessary in most situations

src/tribler/core/utilities/main_thread_stack_tracking.py
kozlovsky: I added the comment

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def main_thread_profile(frame: FrameType, event: str, _):
+    """
+    A hook that calls before and after a function call in the main thread if the stack tracking is activated
+    """
+    if event == 'call':
+        main_thread_stack.append(frame)
+    elif event == 'return' and main_thread_stack:
+        main_thread_stack.pop()
```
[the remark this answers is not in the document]
kozlovsky: I think the code is correct. I added the comment:
```python
# Usually, 'call' and 'return' are always paired, so 'return' removes the frame added by the previous 'call'.
# But at the very beginning, when the `start_main_thread_stack_tracing` function is called, we are already
# inside the function, so we get unpaired `return` when we exit from the function. At that moment, the stack
# is empty. By checking that the stack can be empty, we handle this situation by ignoring the unpaired 'return'.
```

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
...
+    return main_thread_profile
+
+
+def start_main_thread_stack_tracing():
+    """
+    Activates the profiler hook in the main thread. Note that it makes Python functions about two times slower.
+    The compiled code is run as fast, so libtorrent calls and database queries should be as efficient as before.
+    """
+    global _main_stack_tracking_activated
```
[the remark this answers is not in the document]
kozlovsky: Done!

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
...
+    sys.setprofile(main_thread_profile)
+
+
+def _get_main_thread_stack_info() -> List[Tuple[str, str, Optional[int]]]:
+    """
+    Quickly copies necessary information from the main thread stack, so it is possible later to format a usual
+    traceback in a separate thread.
+
+    The function temporarily changes the interpreter’s thread switch interval to prevent thread switch during
+    the stack copying. It is a lighter analogue of holding the GIL.
```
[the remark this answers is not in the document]
kozlovsky: Agreed!

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    prev_switch_interval = sys.getswitchinterval()
+    sys.setswitchinterval(SWITCH_INTERVAL)
+    try:
+        stack_info = [(frame.f_code.co_name, frame.f_code.co_filename, frame.f_lineno)
+                      for frame in main_thread_stack]
+    finally:
+        sys.setswitchinterval(prev_switch_interval)
+    return stack_info
```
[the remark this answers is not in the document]
kozlovsky: Done!

src/tribler/core/utilities/main_thread_stack_tracking.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    lines = ['Traceback (most recent call last):']
+    stack_info = _get_main_thread_stack_info()
+    for func_name, file_name, line_number in stack_info:
+        line = f'  File "{file_name}", line {line_number or "?"}, in {func_name}'
+        lines.append(line)
+        if line_number:
+            source_line = linecache.getline(file_name, line_number)
+            lines.append('    ' + (source_line.strip() or '?'))
+    return '\n'.join(lines) + '\n'
```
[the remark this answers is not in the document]
kozlovsky: I think it is relatively easy to test it; I'd prefer to keep obtaining the thread stack as an implementation detail of the function

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: I think they have different meanings, so I better add comments to them

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Done

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: I can't move it to the very beginning of the file, as I need the definition of `DebugInfo`; I grouped them in this place.

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Done

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Changed

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: The difference is in the following:

`patched_handle_run()` triggers after the coroutine step is finished if the step execution takes too much time. It always triggers for slow coroutine step and shows its execution time.

`SlowCoroWatchingThread.run()` checks periodically that we are not currently in the coroutine step that already took too much time but is not finished yet. This way, we can detect a freezer coroutine that never finished or a coroutine that freezes the loop until the GUI process decides to kill the Core process. In contrast to `patched_handle_run()`, it is not guaranteed that `SlowCoroWatchingThread.run()` is triggered for each coroutine, a coroutine that was just slightly longer than the limit may be missed by `SlowCoroWatchingThread.run()`.

Also, `SlowCoroWatchingThread.run()` displays the current main stack (as we are inside a running coroutine) while `patched_handle_run()` does not display the stack (as the coroutine step is already finished).

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Asserts are handy for declaring non-trivial invariants that should always be satisfied if the code does not contain a bug. They can serve both as executable comments "the code always satisfies this condition" and runtime checks that allows the detection of rare bugs if an invariant is not satisfied.

I removed it as it is not strictly necessary to have it.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+    task: Task = getattr(func, '__self__', None)
+    if not isinstance(task, Task) or not include_stack:
+        return repr(func)
+
+    if main_stack_tracking_is_activated():
+        stack = get_main_thread_stack()
+    else:
+        stream = io.StringIO()
+        task.print_stack(limit=3, file=stream)
+        stack = stream.getvalue()
```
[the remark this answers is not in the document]
kozlovsky: I usually prefer to avoid "if not" as it makes code a bit simpler.

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+    return f"{task}\n{stack}"
+
+
+_thread: Optional[SlowCoroWatchingThread] = None
+
+
+def patch_asyncio():
```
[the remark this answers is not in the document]
kozlovsky: Agreed, moved to the top

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Renamed

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+    if not first_report:
+        info_str = format_info(handle, include_stack=False)
+        logger.error(f"Still executing {info_str}")
+        return
+
+    info_str = format_info(handle, include_stack=True)
+    logger.error(f'Slow coroutine is occupying the loop for {duration:.3f} seconds already: {info_str}')
+
+
+def format_info(handle: Handle, include_stack: bool = False) -> str:
```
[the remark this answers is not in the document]
kozlovsky: Renamed

src/tribler/core/utilities/slow_coro_detection.py
```diff
@@ -0,0 +1,136 @@
...
+
+original_handle_run = Handle._run
+
+
+def report_long_duration(handle: Handle, duration: float):
+    info_str = format_info(handle)
+    logger.error(f'Slow coroutine step execution (duration={duration:.3f} seconds): {info_str}')
+
+
+def report_freeze(handle: Handle, duration: float, first_report: bool):
```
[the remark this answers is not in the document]
kozlovsky: Renamed

src/tribler/core/utilities/slow_coro_detection.py
drew2a, line 1: Possible bug: the empty file.

src/run_tribler.py
```diff
@@ -94,6 +95,10 @@ def init_boot_logger():
         from tribler.core.components.reporter.exception_handler import default_core_exception_handler
 
         init_sentry_reporter(default_core_exception_handler.sentry_reporter)
+
+        if os.environ.get('SLOW_CORO_STACK_TRACING') == '1':
```
drew2a, line 99: Perhaps it is handier to add an option to always trace a stack for developers.
Something like:
```suggestion
        if os.environ.get('SLOW_CORO_STACK_TRACING') == '1' or is_run_from_source():
```

drew2a (changes requested):
All previous comments have been addressed. It seems that the PR is almost ready to be merged.

I left one comment about a possibly forgotten empty file, and another about enabling stack tracing by default for developers.

src/tribler/core/utilities/slow_coro_detection.py
kozlovsky: Thanks, fixed

src/run_tribler.py
kozlovsky: It is possible, but I'm not sure we really want this, as it slows doun the code and it may be harder to turn the option off when running from sources

drew2a: As a developer, I really want this.

drew2a: To me, it seems like a good strategy to enable all debugging tools on each developer's machine and use environment variables as an option to disable a tool.

For example, if you're developing 3 tools, I'm developing 2 tools, and Sandip is developing 3 tools, that means Marcel would need to:
1. Know about all 8 tools
1. Enable all of them by setting 8 environment variables

kozlovsky: Agreed. I changed the code to turn on stack tracking when running from sources by default and allow easier overriding
```python
slow_coro_stack_tracking = os.environ.get('SLOW_CORO_STACK_TRACING', '0' if is_frozen() else '1')
# By default, the stack tracking of slow coroutines is enabled when running the Tribler from sources
# and disabled in the compiled version, as it makes the Python code of Core work slower.
if slow_coro_stack_tracking == '1':
    start_main_thread_stack_tracing()
```

drew2a: Thank you 🙏

drew2a (approved):

PR #7415, opened 2023-05-09
Fix invalid handle when there is no torrent metadata

xoriole:
The PR updates the `require_handle` decorator to execute the function if TorrentDef has metadata.
- Fixes https://github.com/Tribler/tribler/issues/7412
- Fixes https://github.com/Tribler/tribler/issues/6915

### DIFF
--- a/src/tribler/core/components/libtorrent/utils/torrent_utils.py
+++ b/src/tribler/core/components/libtorrent/utils/torrent_utils.py
@@ -4,6 +4,7 @@ from contextlib import suppress
 from hashlib import sha1
 from typing import Any, Dict, Iterable, List, Optional
 
+from tribler.core.components.libtorrent import torrentdef
 from tribler.core.components.libtorrent.utils.libtorrent_helper import libtorrent as lt
 from tribler.core.utilities.path_util import Path
 
@@ -33,14 +34,18 @@ def require_handle(func):
     Invoke the function once the handle is available. Returns a future that will fire once the function has completed.
     Author(s): Egbert Bouman
     """
-
     def invoke_func(*args, **kwargs):
         result_future = Future()
 
         def done_cb(fut):
             with suppress(CancelledError):
                 handle = fut.result()
-            if not fut.cancelled() and not result_future.done() and handle == download.handle and handle.is_valid():
+
+            if not fut.cancelled() \
+                    and not result_future.done() \
+                    and handle == download.handle \
+                    and handle.is_valid() \
+                    and not isinstance(download.tdef, torrentdef.TorrentDefNoMetainfo):
                 result_future.set_result(func(*args, **kwargs))
 
         download = args[0]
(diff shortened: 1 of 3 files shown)


src/tribler/core/components/libtorrent/utils/torrent_utils.py
```diff
@@ -33,14 +34,18 @@ def require_handle(func):
...
         def done_cb(fut):
             with suppress(CancelledError):
                 handle = fut.result()
-            if not fut.cancelled() and not result_future.done() and handle == download.handle and handle.is_valid():
+
+            if not fut.cancelled() \
+                    and not result_future.done() \
+                    and handle == download.handle \
+                    and handle.is_valid() \
+                    and not isinstance(download.tdef, torrentdef.TorrentDefNoMetainfo):
```
kozlovsky, line 48: From issues, I see that the error arises in [`apply_ip_filter`](https://github.com/Tribler/tribler/issues/7412) and [`force_dht_announce`](https://github.com/Tribler/tribler/issues/6915). But the fix is applied to all the `@require_handle`-decorated `Download` methods. The proposed fix suppresses the call of the corresponding `handle` method when the download with only an infohash is just started.

The description of `TorrentDefNoMetainfo` says:

  > Instances of this class are used when working with a torrent def that contains no metainfo **(YET)**, for instance,
    when starting a download with only an infohash

That is, in my understanding, having no meta info is just a temporal state of some torrent.

With this, I have two questions:

1. Is it possible that for some other `Download` methods decorated with `@require_handle`, it is valid to have `TorrentDefNoMetainfo` as `download.tdef` and still call the corresponding `handle` method? Can the current PR accidentally break these other `@require_handle`-decorated methods when the downloading of a torrent with only an infohash is just started?
  
  2. The proposed fix skips calling the corresponding handle method. Is it safe? Is it possible that instead of just skipping, we should postpone calling these handle methods for `apply_ip_filter` etc., until the download is fully started and the handle is ready?

kozlovsky (changes requested):

src/tribler/core/components/libtorrent/utils/torrent_utils.py
xoriole: 1. For all methods where we require a valid handle, it is expected the torrent already has a metainfo indicating that it is a valid download. The case where there is no metainfo as mentioned in the docs is while doing a metainfo request to show the torrent details on the download dialog. At that point, the torrent is not a valid download where the methods requiring handle could be applied. In such a state, we shouldn't need to call any methods that should require a valid handle.

2. I think it is safe to skip for the same reason above. For the methods which explicitly require a valid handle, there should be a valid handle and metainfo already present. For methods like `apply_ip_filter` and `force_dht_announce`, it is expected that the handle is valid, if not trying to execute such action on download is not meaningful.

kozlovsky (approved):

PR #7417, opened 2023-05-11
Add queue for multithread processing

drew2a:
This is a continuation of the previous PR #7413 and the final fix for the deadlock problem.

It adds a thread-safe queue that is used for processing incoming MDS entities, which are created in a different thread.
Additionally, it removes unnecessary logic from Endpoints that perform writes to both databases (`KnowledgeDB` and `MetadataDB`).

As a drawback, this PR reduces the speed of tag extraction, as they are now not performed immediately upon entity addition to the `MetadataDB`. Given that tags aren't directly used in the current release and we plan to remove MDS in upcoming releases, this drawback is acceptable.

### DIFF
--- a/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
+++ b/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
@@ -119,7 +119,6 @@ class KnowledgeRulesProcessor(TaskManager):
         added = 0
         for torrent in batch:
             added += self.process_torrent_title(torrent.infohash, torrent.title)
-            torrent.tag_processor_version = self.version
             processed += 1
 
         self.set_last_processed_torrent_id(end)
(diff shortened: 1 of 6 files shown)


kozlovsky (approved):
It seems that with this change KnowledgeRulesProcessor performs better.

But `process_batch` still updates two databases - MetadataDB by updating `torrent.tag_processor_version` and KnowledgeDB by processing the torrent's title. In some next PR, it may be beneficial to modify `process_batch` so it starts adding titles to the queue instead of processing them directly; after that, updating both databases will be completely separated.

drew2a:
> But process_batch still updates two databases - MetadataDB by updating torrent.tag_processor_version and KnowledgeDB by processing the torrent's title.

Good catch, I overlooked that.
I've removed this line as it's not necessary. `torrent.tag_processor_version` should be set in all cases except `process_batch` (as it processes the MDB incrementally).

kozlovsky (approved):

PR #7418, opened 2023-05-12
Make `KnowledgeRulesProcessor` less aggressive

drew2a:
This is a continuation of the previous PR #7417 and the final PR from the `KnowledgeRulesProcessor` improvements series.

As demonstrated in #7414 by @kozlovsky, `process_batch` is one of the few "long-running" coroutines. This behavior could lead to a brief freeze of Tribler every 10 seconds. This happens on the core side and should go unnoticed by users. However, it's possible to reduce the time of synchronous `process_batch` execution by making `process_torrent_title` an async function.

I've tried to carefully isolate the use of `db_session` so it won't be affected by the switch to async. @kozlovsky, could you please confirm that I've done everything correctly?

### DIFF
--- a/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
+++ b/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
@@ -1,5 +1,6 @@
 import logging
 import queue
+import time
 from asyncio import Event
 from dataclasses import dataclass
 from typing import Optional, Set
@@ -14,13 +15,16 @@ from tribler.core.components.knowledge.rules.rules_general_tags import general_r
 from tribler.core.components.knowledge.rules.tag_rules_base import extract_only_valid_tags
 from tribler.core.components.metadata_store.db.serialization import REGULAR_TORRENT
 from tribler.core.components.metadata_store.db.store import MetadataStore
+from tribler.core.utilities.async_force_switch import force_switch
 from tribler.core.utilities.notifier import Notifier
 from tribler.core.utilities.unicode import hexlify
 
-DEFAULT_BATCH_INTERVAL = 10
-DEFAULT_BATCH_SIZE = 1000
+DEFAULT_BATCH_INTERVAL = 5
+DEFAULT_BATCH_SIZE = 50
 
-DEFAULT_QUEUE_INTERVAL = 10
+DEFAULT_QUEUE_INTERVAL = 5
+DEFAULT_QUEUE_BATCH_SIZE = 100
+DEFAULT_QUEUE_MAX_SIZE = 10000
 
 LAST_PROCESSED_TORRENT_ID = 'last_processed_torrent_id'
 RULES_PROCESSOR_VERSION = 'rules_processor_version'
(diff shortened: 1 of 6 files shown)


kozlovsky (approved):
The code looks correct. Now each `save_statements` call is processed as a separate transaction, and it should be slower, but it is more important to occupy the loop only for a small amount of time, and this code achieves it.

I'd like to see an extended log message, not just "Processed: N titles. Added M tags." but "Processed: N titles in S seconds (K titles / second). Added M tags (Q tags / second)."
It may be convenient to understand the throughput of the processor. The estimated numbers will not take into account parallel tasks, but they should allow an estimation of how long the processing of the entire database can take.

kozlovsky (approved):
The code looks correct. Now each `save_statements` call is processed as a separate transaction, and it should be slower, but it is more important to occupy the loop only for a small amount of time, and this code achieves it.

I'd like to see an extended log message, not just "Processed: N titles. Added M tags." but "Processed: N titles in S seconds (K titles / second). Added M tags (Q tags / second)."
It may be convenient to understand the throughput of the processor. The estimated numbers will not take into account parallel tasks, but they should allow an estimation of how long the processing of the entire database can take.

drew2a:
During the testing of this PR, we realized that async/await might not be used correctly in the Tribler codebase.
Async/await has quite a non-intuitive behavior. See https://stackoverflow.com/questions/59586879/does-await-in-python-yield-to-the-event-loop for more information (credits to @kozlovsky for finding this).

Below is a code example:
```python
import asyncio


async def a():
    async def a_print():
        print('a')

    while True:
        await a_print()


async def b():
    async def b_print():
        print('b')

    while True:
        await b_print()


async def main():
    tasks = {
        asyncio.create_task(a()),
        asyncio.create_task(b())
    }

    await asyncio.wait(tasks)


asyncio.run(main())
```

Intuitively the code should print

```
a
b
a
b
a
b
```

But instead of this the output is:

```
a
a
a
a
a
a

```
So, this PR also includes a decorator to actually interrupt a coroutine at the moment of awaiting.

src/tribler/core/utilities/async_interruption.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def interruption(func):
+    """Decorator to allow interruption of a coroutine.
+
+    For more information, see the example at the end of this file.
+     Also check this: https://stackoverflow.com/questions/59586879/does-await-in-python-yield-to-the-event-loop
+    """
+
+    async def wrapper(*args, **kwargs):
```
kozlovsky, line 11: Can you also please apply [`wraps`](https://stackoverflow.com/questions/308999/what-does-functools-wraps-do) decorator to the wrapper function?
```suggestion
    @functools.wraps(func)
    async def wrapper(*args, **kwargs):
```

src/tribler/core/utilities/tests/test_async_interuption.py
```diff
@@ -0,0 +1,47 @@
...
+import os
+import threading
+
+import pytest
+
+from tribler.core.utilities.async_interruption import interruption
+
+
+@pytest.mark.looptime(False)
+async def test_without_interruption():
```
kozlovsky, line 11: It is strange that the added test lines are not covered by tests; it looks like pytest does not run this test file

kozlovsky (changes requested):
Looks good! I have only a few minor comments.

src/tribler/core/utilities/tests/test_async_interuption.py
drew2a: I'll check it.

drew2a: Fixed

kozlovsky (approved):

PR #7440, opened 2023-05-25
Replace list comprehensions with generator expressions and similar functions

### LINKED ISSUE #7433: Remove list comprehension from `str.join` function calls
**Describe the bug**
Thanks to generator expressions and other lazy evaluation functions (such as `map`), there's no need to build a list when calling `str.join`. This saves memory during runtime since no list is built. Below each relevant line I wrote a substitute I offer.

https://github.com/Tribler/tribler/blob/8df8154193f404e20ba4ff9ba9d46f082820a132/src/tribler/core/components/libtorrent/torrentdef.py#L306

```py
                    return "".join(map(filter_character, name))
```

https://github.com/Tribler/tribler/blob/8df8154193f404e20ba4ff9ba9d46f082820a132/src/tribler/core/components/libtorrent/torrentdef.py#L386

```py
                            return "".join(map(filter_character, name))
```

https://github.com/Tribler/tribler/blob/8df8154193f404e20ba4ff9ba9d46f082820a132/src/tribler/core/components/metadata_store/remote_query_community/payload_checker.py#L104

```py
                " ".join(getattr(self.payload, attr) for attr in ("title", "tags", "text") if hasattr(self.payload, attr))
```

https://github.com/Tribler/tribler/blob/8df8154193f404e20ba4ff9ba9d46f082820a132/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py#L53

```py
        return ''.join(map(chr, ext_peer_info))
```

- Tribler's version [main branch]

Solomon1732:
Replaced generator expressions with similar functions, such as `map` and `filter`. In addition, replaced an indented pair of `for` loops and `if` statement with a creation of `set`. It uses a generator expression and a call to `filter` to create it in one go. This way it avoids repeatedly appending to a list. Fixes #7433 .

### DIFF
--- a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
@@ -50,7 +50,7 @@ def _safe_extended_peer_info(ext_peer_info):
         return ensure_unicode(ext_peer_info, "utf8")
     except UnicodeDecodeError:
         # We might have some special unicode characters in here
-        return ''.join([chr(c) for c in ext_peer_info])
+        return ''.join(map(chr, ext_peer_info))
 
 
 def get_extended_status(tunnel_community, download) -> DownloadStatus:
--- a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
@@ -44,13 +44,13 @@ def _safe_extended_peer_info(ext_peer_info):
     """
     # First see if we can use this as-is
     if not ext_peer_info:
-        return ''
+        return ""
 
     try:
         return ensure_unicode(ext_peer_info, "utf8")
     except UnicodeDecodeError:
         # We might have some special unicode characters in here
-        return ''.join(map(chr, ext_peer_info))
+        return "".join(map(chr, ext_peer_info))
 
 
 def get_extended_status(tunnel_community, download) -> DownloadStatus:
--- a/src/tribler/core/components/libtorrent/torrentdef.py
+++ b/src/tribler/core/components/libtorrent/torrentdef.py
@@ -15,6 +15,25 @@ from tribler.core.utilities.unicode import ensure_unicode
 from tribler.core.utilities.utilities import bdecode_compat, is_valid_url, parse_magnetlink
 
 
+def _filter_characters(name: str) -> str:
+    """Sanitize the names in path to unicode by replacing out all
+    characters that may -even remotely- cause problems with the '?'
+    character.
+
+    :param name: the name to sanitize
+    :type name: str
+    :return: the sanitized string
+    :rtype: str
+    """
+    def filter_character(char: str) -> str:
+        if 0 < char < 128:
+            return chr(char)
+        self._logger.debug("Bad character 0x%X", char)
+        return "?"
+
+    return "".join(map(filter_character, name))
+
+
 def escape_as_utf8(string, encoding='utf8'):
     """
     Make a string UTF-8 compliant, destroying characters if necessary.
@@ -296,16 +315,7 @@ class TorrentDef:
             # all characters that may -even remotely- cause problems
             # with the '?' character
             try:
-                def filter_characters(name):
-                    def filter_character(char):
-                        if 0 < char < 128:
-                            return chr(char)
-                        self._logger.debug("Bad character 0x%X", char)
-                        return "?"
-
-                    return "".join([filter_character(char) for char in name])
-
-                return filter_characters(self.metainfo[b"info"][b"name"])
+                return _filter_characters(self.metainfo[b"info"][b"name"])
             except UnicodeError:
                 pass
 
@@ -376,16 +386,7 @@ class TorrentDef:
                     # replacing out all characters that may -even
                     # remotely- cause problems with the '?' character
                     try:
-                        def filter_characters(name):
-                            def filter_character(char):
-                                if 0 < char < 128:
-                                    return chr(char)
-                                self._logger.debug("Bad character 0x%X", char)
-                                return "?"
-
-                            return "".join([filter_character(char) for char in name])
-
-                        yield (Path(*[filter_characters(element) for element in file_dict[b"path"]]),
+                        yield (Path(*[_filter_characters(element) for element in file_dict[b"path"]]),
                                file_dict[b"length"])
                         continue
                     except UnicodeError:
--- a/src/tribler/core/components/libtorrent/torrentdef.py
+++ b/src/tribler/core/components/libtorrent/torrentdef.py
@@ -198,17 +198,20 @@ class TorrentDef:
         """
         return self.torrent_parameters.get(b'announce-list', [])
 
-    def get_trackers_as_single_tuple(self):
+    def get_trackers(self) -> set:
         """
-        Returns a flat tuple of all known trackers.
+        Returns a flat set of all known trackers.
+
+        :return: all known trackers
+        :rtype: set
         """
         if self.get_tracker_hierarchy():
             trackers = itertools.chain.from_iterable(self.get_tracker_hierarchy())
-            return tuple(set(filter(None, trackers)))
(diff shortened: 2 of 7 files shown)


src/tribler/core/components/libtorrent/torrentdef.py
```diff
@@ -184,12 +203,8 @@ def get_trackers_as_single_tuple(self):
...
         """
         if self.get_tracker_hierarchy():
-            trackers = []
-            for level in self.get_tracker_hierarchy():
-                for tracker in level:
-                    if tracker and tracker not in trackers:
-                        trackers.append(tracker)
-            return tuple(trackers)
+            trackers = itertools.chain.from_iterable(self.get_tracker_hierarchy())
+            return tuple(set(filter(None, trackers)))
```
drew2a, line 207: This is a nice simplification!
This code is correct assuming that the order of the tuple isn't important, as the returned value now doesn't maintain the same tracker order as presented in `self.get_tracker_hierarchy()`.

However, this assumption is valid as all places that use `get_trackers_as_single_tuple()` convert it to a set anyway.

To prove the correctness of the code, please consider adding a test similar to the one presented below, which demonstrates that both versions of the code (original and new) work as expected:

```python
def test_get_trackers_as_single_tuple(tdef):
    # Test that `get_trackers_as_single_tuple` returns flat tuple of trackers
    tdef.get_tracker_hierarchy = Mock(return_value=[['t1', 't2'], ['t3'], ['t4']])
    t = tdef.get_trackers_as_single_tuple()
    assert set(t) == {'t1', 't2', 't3', 't4'}
```

If you'd like, you could further refine the refactoring process by reworking the entirety of the `get_trackers_as_single_tuple` function. It's now clear that the return type can be changed from a `tuple` to a `set`.

Solomon1732: Thank you! 😄 I'll add such a test. I'll check where I can change the return type.

src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
```diff
@@ -50,7 +50,7 @@ def _safe_extended_peer_info(ext_peer_info):
         return ensure_unicode(ext_peer_info, "utf8")
     except UnicodeDecodeError:
         # We might have some special unicode characters in here
-        return ''.join([chr(c) for c in ext_peer_info])
+        return ''.join(map(chr, ext_peer_info))
```
drew2a, line 53: It's not your fault that there aren't any tests for this line.

Feel free to add the following:

```python
@patch('tribler.core.components.libtorrent.restapi.downloads_endpoint.ensure_unicode',
       Mock(side_effect=UnicodeDecodeError('', b'', 0, 0, '')))
def test_safe_extended_peer_info():
    # Test that we return the string mapped by `chr` in the case of `UnicodeDecodeError`
    extended_peer_info = _safe_extended_peer_info(b'abcd')
    assert extended_peer_info == 'abcd'
```

src/tribler/core/components/libtorrent/torrentdef.py
```diff
@@ -135,6 +136,24 @@ async def load_from_url(url):
...
+        :return: the sanitized string
+        :rtype: str
+        """
+        def filter_character(char: int) -> str:
+            if 0 < char < 128:
+                return chr(char)
+            self._logger.debug("Bad character 0x%X", char)
+            return "?"
+
+        return "".join(map(filter_character, name))
```
drew2a, line 155: I see that this is a slightly modified version of the original code. Feel free to add a test for it if you have some spare time. If not, I'll handle it for this PR.

src/tribler/core/components/libtorrent/torrentdef.py
drew2a: You are absolutely correct :)

src/tribler/core/components/libtorrent/torrentdef.py
Solomon1732: The function was repeated twice. Because of this I extracted it to an underscore-prefixed method, along with the obvious change of removing the list comprehension from `str.join`. I'll try to add a test for it.

src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
Solomon1732: Thanks!

drew2a (changes requested):
Great work! As a team, we truly appreciate your contribution to improving the quality of our codebase. This PR requires some fine-tuning, such as the addition of tests. If I understand correctly, you might be experiencing difficulties with the Tribler development environment due to the M1(M2) chipset.

If you're unable or prefer not to do this further refinement, I can handle it. Otherwise, please feel free to complete the PR on your own.

Solomon1732:
You understand me correctly 🙂. I'll try to add tests on my own.

drew2a:
> How can I tell the test unit it's fine to access protected class members? Static code analysis complains about accesses to protected class members ([_safe_extended_peer_info](https://github.com/Tribler/tribler/pull/7440/files#annotation_11555381573) and [_filter_characters](https://github.com/Tribler/tribler/pull/7440/files#annotation_11555381577)). But I need to access them to test them.

You should add this comment at the beginning of the file:

https://github.com/Tribler/tribler/blob/13f8b886585d997901336c0a3e7865fd16a046c6/src/tribler/core/components/ipv8/eva/tests/test_exceptions.py#L9

> Also I noticed the Windows-latest pytest was canceled and didn't finish? It didn't happen before.

This situation can occasionally occur and the reasons are not always clear. You can try rerunning the check, which should resolve the issue.

Solomon1732:
> You can try rerunning the check, which should resolve the issue.

How can I rerun it? I'm unfamiliar with the bot commands 😅. Also I don't know if I can even run them.

Solomon1732:
I'm not sure why the mac build timed out. It's the second time in a row it happened.
https://github.com/Tribler/tribler/actions/runs/5091540598/jobs/9151698599?pr=7440

drew2a:
> How can I rerun it? I'm unfamiliar with the bot commands 😅. Also I don't know if I can even run them.

Essentially, you have two options:
1. Push a new commit.
1. Click the button (although I'm not certain you have the necessary permissions).
image


> I'm not sure why the mac build timed out. It's the second time in a row it happened.

This sometimes happens, and the reason is currently unknown. You can find more details on this issue at https://github.com/Tribler/tribler/issues/7134. For your pull requests, it's generally safe to ignore these timeouts.

Solomon1732:
Yes, I lack the necessary permissions.

Is there something else I need to fix in this PR? Or can I remove the "WIP"?

drew2a:
The PR looks good, except for one minor issue: these comments are somewhat confusing:

>   This test is needed to check a protected method. Therefore, this test
     needs to be disabled
     pylint: disable=protected-access

I would recommend refraining from explaining the pylint disable (as it should be self-explanatory).

Like from:

```python
def test_safe_extended_peer_info():
    # This test is needed to check a protected method. Therefore, this test
    # needs to be disabled
    # pylint: disable=protected-access
    """
    Test that we return the string mapped by `chr` in the case of `UnicodeDecodeError`
    """
    extended_peer_info = download_endpoint._safe_extended_peer_info(b"abcd")
    assert extended_peer_info == "abcd"
```

to 

```python
def test_safe_extended_peer_info():
    """
    Test that we return the string mapped by `chr` in the case of `UnicodeDecodeError`
    """
    # pylint: disable=protected-access
    extended_peer_info = download_endpoint._safe_extended_peer_info(b"abcd")
    assert extended_peer_info == "abcd"
```

or even to


```python
def test_safe_extended_peer_info():
    """
    Test that we return the string mapped by `chr` in the case of `UnicodeDecodeError`
    """
    extended_peer_info = download_endpoint._safe_extended_peer_info(b"abcd")  # pylint: disable=protected-access
    assert extended_peer_info == "abcd"
```

And yes, feel free to remove [WIP] from the PR title :)

drew2a (approved):
Thank you for your contribution and for the cooperation :)

Issue #7398, opened 2023-05-01
The big migration: from the Channels to the Knowledge Graph


drew2a:
This issue describes the migration plan for the Tribler's metadata from the Channels to the KnowledgeDatabase. See more information here: #6214

## The recent history

* **Release 7.11**: that was the first release for #6396
* **Release 7.12**: there we introduced #6708 and #6718 

## Current state

Currently, we have the [7.13 release candidate](https://github.com/Tribler/tribler/releases/tag/v7.13.0-RC1) that includes a migration from the `TagDatabase` to the `KnowledgeDatabase` (as it's a crucial step towards a global knowledge graph): #7070 
After this release is successfully launched, we will be ready for the next steps.

The direction for the next step could involve completing the replacement of Channels with the global Knowledge Graph.

## Future

Questions should be answered before the big migration:
* Spam prevention: Can we use [MeritRank](https://github.com/grimadas/meritrank) for this purpose? (The answer is "No")
*  What is the best way to disseminate knowledge? Is the current mechanism sufficient? Should we use the PopularityCommunity for this purpose, or continue using the KnowledgeCommunity? ~~Should we use Skip Graph?~~

`7.14`
* **Metadata migration**: write a migration for the database:`mds.db`->`knowledge.db`
* **Alternative remote search**: develop a remote search identical to the search by MDS, disabled by default.

Outcome: tribler network is ready for testing search by knowledge.

`7.14-experimental` 
* Enable remote search by Knowledge by default (as a full replacement for the current mechanism).

Outcome: Search quality evaluation based on users' feedback and our own experience.

`7.15`
* Implement `MeritRank` for the `KnowledgeCommunity`

Outcome:
* The Tribler network is ready for testing 'search by knowledge' with spam prevention.
* Search quality evaluation based on users' feedback and our own experience 

`8.0`
* Remove channels from GUI
* Remove channels from core

Outcome: Tribler network migrated from the Channels to the KnowledgeDatabase.

xoriole:
Channels are a way of organizing content on Tribler Network. As it is created by a human (and not autogenerated), the way the channels are organized into a hierarchy of directories and torrents are placed on different directories, makes sense at least to the creator of the channel if not for more people who subscribe to the channel. This involves the channel creator's human intelligence in the organization of content which likely required some time and effort.

While moving away from channels to knowledge database, we should try to incorporate this content organization done by the user and not simply erase it.

drew2a:
[...]

PR #7445, opened 2023-05-30
Add `CONTENT_ITEM` relation

drew2a:
This PR splits two relations: `TITLE` and `CONTENT_ITEM` as they will be used for different purposes.

* `CONTENT_ITEM` to combine items into groups (#6214)
* `TITLE` to store the original title of the torrent after the big migration (#7398)

Also, it fixes inconsistency with statement size limitations.


src/tribler/core/components/knowledge/knowledge_constants.py
```diff
@@ -1,2 +1,2 @@
-MIN_RESOURCE_LENGTH = 3
+MIN_RESOURCE_LENGTH = 2
```
drew2a, line 1: This change is necessary for the language statements transfer (nl, ru, en)

src/tribler/core/components/knowledge/restapi/knowledge_endpoint.py
```diff
@@ -65,24 +65,21 @@ async def update_knowledge_entries(self, request):
             return error_response
 
         # Validate whether the size of the tag is within the allowed range and filter out duplicate tags.
-        tags = set()
```
drew2a, line 68: This `set()` is unnecessary here because statements anyway will be converted to sets later.

kozlovsky (approved):

PR #7454, opened 2023-06-01
Fix KeyError on description_text

### LINKED ISSUE #7453: KeyError: 'description_text'
Sentry Issue: [TRIBLER-17M](https://sentry.tribler.org/organizations/tribler/issues/2161/?referrer=github_integration)

```
CreationTraceback: 
  File "run_tribler.py", line 112, in <module>
  File "tribler/gui/start_gui.py", line 78, in run_gui
  File "tribler/gui/utilities.py", line 435, in trackback_wrapper
  File "tribler/gui/network/request.py", line 147, in on_finished
  File "tribler/gui/utilities.py", line 435, in trackback_wrapper
  File "tribler/gui/widgets/tablecontentmodel.py", line 402, in on_query_results
  File "tribler/gui/utilities.py", line 435, in trackback_wrapper
  File "tribler/gui/widgets/channelcontentswidget.py", line 2...
KeyError: 'description_text'
  File "tribler/gui/utilities.py", line 438, in trackback_wrapper
  File "tribler/gui/utilities.py", line 435, in trackback_wrapper
  File "tribler/gui/widgets/channeldescriptionwidget.py", line 228, in _on_description_received
```

xoriole:
Fix #7453

### DIFF
--- a/src/tribler/gui/widgets/channeldescriptionwidget.py
+++ b/src/tribler/gui/widgets/channeldescriptionwidget.py
@@ -224,12 +224,8 @@ class ChannelDescriptionWidget(AddBreadcrumbOnShowMixin, widget_form, widget_cla
         self.description_page.setHidden(False)
 
     def _on_description_received(self, result):
-        if result:
-            self.description_text = result["description_text"]
-            self.description_text_preview.setMarkdown(self.description_text)
-        else:
-            self.description_text = None
-            self.description_text_preview.setMarkdown("")
+        self.description_text = result.get("description_text") if result else None
+        self.description_text_preview.setMarkdown(self.description_text or "")
 
         request_manager.get(f'channels/{self.channel_pk}/{self.channel_id}/thumbnail',
                             on_success=self._on_thumbnail_received,


src/tribler/gui/widgets/channeldescriptionwidget.py
```diff
@@ -225,7 +225,7 @@ def show_description_page(self):
 
     def _on_description_received(self, result):
         if result:
-            self.description_text = result["description_text"]
+            self.description_text = result.get("description_text", None)
```
kozlovsky, line 228: From the previous version of the code, it looks like the code assumes that the `description_text` value can be `None`, but the `description_text_preview` value is always a string. After the fix, we set `description_text_preview` to `None` as well. I don't know whether it is correct or not, but apparently, it works.

I think we can simplify the code a bit.

First, the `get` method of the dictionary uses `None` as the default value for the second argument, so it is unnecessary to specify it.

Second, the two branches of the `if` statement look similar, so instead of
```python
     if result:
            self.description_text = result.get("description_text")
            self.description_text_preview.setMarkdown(self.description_text)
        else:
            self.description_text = None
            self.description_text_preview.setMarkdown("")
```
we can write something like
```python
        self.description_text = result.get("description_text") if result else None
        self.description_text_preview.setMarkdown(self.description_text or "")
```

kozlovsky (changes requested):

kozlovsky (approved):

PR #7463, opened 2023-06-07
Fix/unminimizing when clicking on file

### LINKED ISSUE #7462: [7.13.0-RC2][Windows] Tribler window does not unminimized when clicking on a torrent file
When Tribler is running, and a user double-clicks on a torrent file, it is expected that Tribler opens a dialog window suggesting to download the file.

As it turns out, on Windows, when the Tribler window is minimized, and a user double-clicks on a torrent file, the Tribler window is not restored. If the user then manually clicks the Tribler icon at the taskbar notification area, the Tribler window is successfully restored, and the user can see the opened dialog that suggests downloading the file.

The correct behavior is to unminimize the Tribler window automatically.

The code investigation shows that currently, we have two different TriblerWindow methods that restore the minimized window. The first method, `restore_from_minimized`, is used when the Tribler application receives a new file:
```python
    def restore_from_minimised(self):
        self.setWindowState(self.windowState() & ~Qt.WindowMinimized)
        self.raise_()
```

The second method, `raise_window`, is used when the user manually clicks the Tribler icon in the taskbar notification area:
```python
    def raise_window(self):
        self.setWindowState(self.windowState() & ~Qt.WindowMinimized | Qt.WindowActive)
        self.show()
        self.raise_()
        self.activateWindow()
```

The `raise_window` method works, and the `restore_from_minimised` method doesn't. It looks like the same method, `raise_window`, should be used in both cases, and this change fixes the problem.

kozlovsky:
This PR fixes #7462 by using the same method, `TriblerWindow.raise_window()` in both cases: when a user clicks on the Tribler icon in the taskbar notification area and when the user double-clicks the torrent file.

### DIFF
--- a/src/tribler/gui/single_application.py
+++ b/src/tribler/gui/single_application.py
@@ -89,7 +89,7 @@ class QtSingleApplication(QApplication):
         self._incoming_stream.setCodec('UTF-8')
         connect(self._incoming_connection.readyRead, self._on_ready_read)
         if self.tribler_window:
-            self.tribler_window.restore_from_minimised()
+            self.tribler_window.raise_window()
 
     def _on_ready_read(self):
         while True:
--- a/src/tribler/gui/tribler_window.py
+++ b/src/tribler/gui/tribler_window.py
@@ -1250,10 +1250,6 @@ class TriblerWindow(QMainWindow):
         if event.key() == Qt.Key_Escape:
             self.escape_pressed.emit()
 
-    def restore_from_minimised(self):
-        self.setWindowState(self.windowState() & ~Qt.WindowMinimized)
-        self.raise_()
-
     def handle_uri(self, uri):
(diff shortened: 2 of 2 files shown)


drew2a (approved):

src/tribler/gui/single_application.py
```diff
@@ -89,11 +89,12 @@ def _on_new_connection(self):
         self._incoming_stream.setCodec('UTF-8')
         connect(self._incoming_connection.readyRead, self._on_ready_read)
         if self.tribler_window:
-            self.tribler_window.restore_from_minimised()
+            self.tribler_window.raise_window()
```
xoriole, line 92: I assume this could be platform dependent. Does it work on three major platforms (Win, macOS, Linux)?

xoriole (approved):

PR #7470, opened 2023-06-09
Fix the space bug in the Sentry Scrubber

drew2a:
This PR fixes a bug in the Sentry Scrubber discovered by @kozlovsky 

See the example of the bug: https://sentry.tribler.org/organizations/tribler/issues/2105/


kozlovsky (approved):
I checked; it works properly now. The obfuscated version: https://sentry.tribler.org/organizations/tribler/issues/2105/?query=is%3Aunresolved&referrer=issue-stream&stream_index=0

PR #7484, opened 2023-06-12
Fix crash on Ubuntu while selecting File Chooser

xoriole:
Summary of the issue:
1. The binary built on Ubuntu 20.04 works on 20.04 but not Ubuntu 22.04 (and higher likely, not tested)
2. The binary built on Ubuntu 22.04 works on 22.04 but not lower.

The issue was related to `GdkPixbuf` internally required by PyQt5 File chooser on Linux. Turns out PyInstaller did not import required internal gtk libs within the build bundle. For the binaries built on the same distribution, the executable was able to find the required libs from the system but not on different distros.

A solution is presented on this PR. Here I add `PyGObject` as a build dependency for Linux which provides GTK libs for python, then add `gi` and `gi.repository` as hidden imports. These hidden imports make sure that `GdkPixbuf` is accessible and that PyQt Filechooser can work without a crash. 

I have tested the build with these fixes on Ubuntu 20.04 and 22.04 and can confirm to work. 

Here is a build to test.
https://jenkins-ci.tribler.org/job/Build-Tribler_release/job/Build-Ubuntu64/653/

### DIFF
--- a/.github/workflows/build_ubuntu.yml
+++ b/.github/workflows/build_ubuntu.yml
@@ -59,6 +59,13 @@ jobs:
           fetch-depth: 0
           ref: ${{inputs.ref || github.ref}}
 
+      - name: Install Ubuntu dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install debhelper devscripts
+          # for qt:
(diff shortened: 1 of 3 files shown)


drew2a (approved):
👍

kozlovsky (approved):

drew2a (approved):

Issue #7336, opened 2023-03-28
Refactoring: `BandwidthCommunity` may be removed.


drew2a:
We still have issues related to the `BandwidthCommunity` which causes the Tribler crashes.
At the same time, the `BandwidthCommunity` functionality has not been used for at least a couple of years.

It looks like a big piece of code that could be safely removed.

synctext:
can't find the code https://github.com/Tribler/tribler/search?q=BandwidthCommunity&type=code
this is different then the bartercast or ledgerzero community

drew2a:
The core part:
https://github.com/Tribler/tribler/tree/main/src/tribler/core/components/bandwidth_accounting

The GUI part:
https://github.com/Tribler/tribler/blob/main/src/tribler/gui/widgets/trustgraphpage.py

synctext:
This should be repaired for MeritRank. Please fix the stability issues. For long term lab progress we need a reputation graph.

drew2a:
The error continues to occur in `BandwidthCommunity`. See the latest: https://sentry.tribler.org/organizations/tribler/issues/2548/.

Should we continue trying to fix them, or should we remove the entire feature as it will eventually be replaced in favor of MeritRank?

@synctext

synctext:
Indeed, we might as well delete it all now. This commit from [3 weeks ago deleted a lot](https://github.com/Tribler/tribler/commit/7fe164612a112262b404692e86bb916ccb7e7563). 
Then @InvictusRMC and @grimadas can start with a somewhat clean start with MeritRank deployment.

Issue #5438, opened 2020-07-03
ValueError: bytes must be in range(0, 256)


ichorid:
**7.5.1 Windows 10**

from Reporter

```python
Traceback (most recent call last):
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 112, in on_read_data
RuntimeError: bytes must be in range(0, 256)
--LONG TEXT--
Traceback (most recent call last):
File "lib\asyncio\events.py", line 88, in _run
File "lib\asyncio\selector_events.py", line 965, in _read_ready
File "src\tribler-core\tribler_core\modules\tunnel\socks5\udp_connection.py", line 51, in datagram_received
File "src\tribler-core\tribler_core\modules\tunnel\community\dispatcher.py", line 91, in on_socks5_udp_data
File "ipv8\messaging\anonymization\community.py", line 485, in send_data
File "ipv8\messaging\anonymization\community.py", line 471, in send_cell
File "ipv8\messaging\anonymization\payload.py", line 77, in to_bin
File "ipv8\messaging\anonymization\payload.py", line 32, in encode_address
File "ipv8\util.py", line 24, in cast_to_bin
ValueError: bytes must be in range(0, 256)

--CONTEXT--
{'message': 'Exception in callback _SelectorDatagramTransport._read_ready()', 'exception': ValueError('bytes must be in range(0, 256)'), 'handle': <Handle _SelectorDatagramTransport._read_ready()>}

```

qstokkink:
@egbertbouman do you want to fix this a second time?

ichorid:
It was reported today, so I guess it is still a problem.

qstokkink:
That is odd, the line numbers do not match up. `encode_address` spans lines 20-27. This is reporting an error on line 32.

qstokkink:
The last time `encode_address` was even present on line 32 was in October of last year.

egbertbouman:
This is a different issue. The original one was caused by `inet_aton` raising a `ValueError` if the IP had embedded null characters (fixed in https://github.com/Tribler/py-ipv8/commit/82c43ed17618e0bd031b44283380898602f35427).

This time it's failing [here](https://github.com/Tribler/py-ipv8/blob/master/ipv8/messaging/anonymization/payload.py#L32).

qstokkink:
[...]

PR #7487, opened 2023-06-14
Consider HealthInfo with negative seeders and leechers as invalid

### LINKED ISSUE #7430: [7.12.1] PackError ipv8.messaging.serialization in pack_serializable
https://sentry.tribler.org/organizations/tribler/issues/2134/

```
Could not pack item: ('I', -2), 
error: argument out of range
```

```python
  File "ipv8\messaging\serialization.py", line 365, in pack_serializable, 
  File "ipv8\messaging\serialization.py", line 252, in pack, 
struct.error: argument out of range, 
  File "asyncio\events.py", line 81, in _run, 
  File "ipv8\taskmanager.py", line 131, in done_cb, 
  File "ipv8\taskmanager.py", line 18, in interval_runner, 
  File "ipv8\util.py", line 35, in call_async, 
  File "tribler\core\components\popularity\community\popularity_community.py", line 116, in gossip_random_torrents_health, 
  File "tribler\core\components\popularity\community\popularity_community.py", line 110, in _gossip_torrents_health, 
  File "ipv8\lazy_community.py", line 199, in ez_send, 
  File "ipv8\lazy_community.py", line 215, in _ez_senda, 
  File "ipv8\lazy_community.py", line 231, in ezr_pack, 
  File "ipv8\lazy_community.py", line 234, in _ez_pack, 
  File "ipv8\messaging\serialization.py", line 379, in pack_serializable_list, 
  File "ipv8\messaging\serialization.py", line 379, in <genexpr>, 
  File "ipv8\messaging\serialization.py", line 363, in pack_serializable, 
  File ", 
def to_pack_list(self):, 
    return [("I", self.random_torrents_length), ("I", self.torrents_checked_length), ("varlenI", self.fix_pack_random_torrents(self.random_torrents)), ("raw", self.fix_pack_torrents_checked(self.torrents_checked))], 
        ", line 3, in to_pack_list, 
  File "tribler\core\components\popularity\community\payload.py", line 28, in fix_pack_random_torrents, 
  File "tribler\core\components\popularity\community\payload.py", line 28, in <genexpr>, 
  File "ipv8\messaging\serialization.py", line 367, in pack_serializable, 
ipv8.messaging.serialization.PackError: Could not pack item: ('I', -2), 
error: argument out of range
```

drew2a:
This PR fixes #7430 by suppressing all errors in `PopularityCommunity.gossip_random_torrents_health()`.

In #7430 the root cause of the error is a packing error in the attempt to pack the signed number -2 into an unsigned 32-bit integer.

We have several strategies for addressing this bug:

1. Inspect all instances where the leecher count is created/modified to ensure it consistently remains positive.
1. Guarantee that num_leechers is positive before invoking Community.ez_send().
1. Suppress all errors within the community.

I've opted for the third approach, in line with the existing ipv8 policy of error suppression within communities to prevent Tribler from crashing due to, for instance, a packet of death.

This bug uncovers the inconsistency in ipv8 error handling methods: some suppress exceptions, while others do not. Ideally, ipv8 error handling would be made more consistent, but due to strict contribution rules, this isn't easily achievable (it's nearly impossible for Tribler developers to introduce changes to ipv8). Hence, we should address this issue on Tribler's end.

<del>
I've implemented an annotation 
This annotation can be appended to any method (synchronous or asynchronous), and it will suppress all exceptions that arise within the method. However, it's important to note that this isn't a comprehensive solution. The complete solution would involve incorporating the same logic into all fragile community methods. Implementing this fully is a significant change for Tribler, so it would be more prudent to reserve such a comprehensive overhaul for the next release.
</del>

### DIFF
--- a/src/tribler/core/components/popularity/community/popularity_community.py
+++ b/src/tribler/core/components/popularity/community/popularity_community.py
@@ -62,7 +62,8 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
             return []
 
         # Filter torrents that have seeders
-        return [health for health in self.torrent_checker.torrents_checked.values() if health.seeders > 0]
+        return [health for health in self.torrent_checker.torrents_checked.values() if
+                health.seeders > 0 and health.leechers >= 0]
 
     def gossip_random_torrents_health(self):
         """
--- a/src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
+++ b/src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
@@ -53,7 +53,7 @@ class HealthInfo:
         return hexlify(self.infohash)
 
     def is_valid(self) -> bool:
-        return self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
+        return self.seeders >= 0 and self.leechers >= 0 and self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
 
     def old(self) -> bool:
         now = int(time.time())
--- a/src/tribler/core/components/torrent_checker/torrent_checker/tests/test_health_info_should_update.py
+++ b/src/tribler/core/components/torrent_checker/torrent_checker/tests/test_health_info_should_update.py
@@ -27,6 +27,12 @@ def test_invalid_health():
     assert not health.should_replace(prev_health)
 
 
+def test_health_negative_seeders_or_leechers():
+    """ Test that health with negative seeders or leechers is considered invalid"""
+    assert not HealthInfo(INFOHASH, seeders=-1).is_valid()
+    assert not HealthInfo(INFOHASH, leechers=-1).is_valid()
+
+
 def test_self_checked_health_update_remote_health():
     prev_health = HealthInfo(INFOHASH)
     health = HealthInfo(INFOHASH, self_checked=True)


xoriole (changes requested):
I don't think we should suppress exceptions on the calls like `gossip_random_torrents_health()` that are triggered by the user itself.  This only affects the  user trying to send the message and not any other users receiving the message. If there are exceptions, we should resolve the errors instead. In this case, I would be in favor of handling the pack error.

Regarding suppression of exceptions, I would agree on suppressing exceptions on message handlers so they do not crash Tribler for an innocent user who received a death packet.

qstokkink:
I see IPv8. I consider myself summoned. 🧙 

> This bug uncovers the inconsistency in ipv8 error handling methods: some suppress exceptions, while others do not.

This is by design. I recall a heated debate with @egbertbouman in #5438. Long story short (though you can read the entire exchange in that issue if you want): I argued for the same "catch all" approach as you are arguing for right now. Back then I was convinced by @egbertbouman's following two sentences, which I hope will be of guidance to you too:

> Crashing the app gives us error reports, and revealed this bug. If this didn't happen we would never know about this issue.

Besides that, regarding _how_ to suppress errors (if you do end up doing this), I would like to point you to the `ignore` argument of `register_task()`. If you call `self.register_task("gossip_random_torrents", self.gossip_random_torrents_health, interval=PopularityCommunity.GOSSIP_INTERVAL_FOR_RANDOM_TORRENTS)` with `ignore=(Exception,)`, all exceptions are caught and logged for this particular task.

Other than this recounting of history and objective info, I won't offer any opinion on what approach to take.

drew2a:
@qstokkink thank you for sharing the historical context behind this design decision and for the tip regarding `ignore=(Exception,)`.

I agree with the statement:
> Crashing the app gives us error reports, and revealed this bug. If this didn't happen we would never know about this issue.

By allowing Tribler to be susceptible to such failures, we indeed increase the chance of uncovering bugs, assuming that some of our users will experience these issues and kindly send us reports. This is a valid point. However, I do have concerns about the cost associated with this approach. In this case, the cost is our users. I don't have exact statistics, but from my estimation, it might look like this: out of 10 users who experience a bug, 4 might uninstall Tribler as a result, and only one sends us a report.

I'd like to suggest a slightly modified approach:

1. Improve our test coverage by writing additional unit tests.
1. Expand the number of scenarios for the Application Tester.
1. Maintain Tribler's fragility to these tests.
1. Strengthen Tribler's resilience for the end users.

I recognize that this paradigm shift might be a departure from what we've been doing, but it's a topic worth discussing. I suggest we could include this in our agenda for next week's meeting.

Another potential strategy could be:

1. Enhance Tribler's stability for the end users.
1. Develop a new reporting system that anonymously reports all suppressed exceptions.

I look forward to further discussions on this matter. Thank you again for your insights.

egbertbouman:
Since I'm being mentioned I'll also respond.

The argument that I made against a catchall in tunnel-related logic was mostly due to the amount of logic that is behind these calls. Even with a proper stack trace debugging an error could potentially take days. Besides, it wasn't even needed in this specific case since the input was already checked. As the former maintainer of the tunnel community I was always keen on being informed of any serious errors. The tradeoff could be different for other (less vital) components.

Anyway, that's why I added the `ignore` kwarg to the `TaskManager`. That way developers can make these tradeoffs for themselves.

Note that if you `await` the task it will still throw the error. It's just meant to prevent the exception from ending up in the asyncio exception handler (the one that's registered using `add_exception_handler`).

kozlovsky (changes requested):
I have a distant analogy from another domain - the biological field. The DNA of a living organism can undergo mutations. Some of these mutations are genuinely dangerous, rendering the organism unviable. However, many mutations are mildly harmful. They do not lead to the immediate death of the organism, but they make it slightly less viable. With each mildly harmful mutation, an individual becomes a bit weaker, less intelligent, and less attractive, but the effect is negligible. For instance, a mildly harmful mutation might make the organism 0.01% less adapted.

Surprisingly, such mildly harmful mutations can be far more dangerous than truly harmful ones, which quickly kill the individual or make it an easy prey for predators. The effect of mildly harmful mutations is too insignificant to be filtered out by natural selection. If an animal is only 0.01% less intelligent than its counterparts, other random factors (such as luck) play a more significant role in survival.

As a result, natural selection cannot effectively filter out these mildly harmful mutations; they spread from one individual to the entire population, leading to a decrease in the fitness of the species. To combat these mutations, specialized mechanisms are required to work in conjunction with natural selection. These mechanisms detect combinations of several mildly harmful mutations and eliminate the embryo at early stages.

In a way, bugs in a software product resemble these mutations. Severely harmful mutations are akin to bugs that cause the application to crash, often leading to the permanent 'death' of the instance (i.e., the installed Tribler application). Mildly harmful mutations are akin to bugs that are not outwardly noticeable but degrade the application's performance and functionality without the user noticing.

I am concerned that by converting many bugs from the category of severely harmful mutations to mildly harmful mutations, these mildly harmful mutations will accumulate. Eventually, it will become apparent that Tribler is performing slower and worse, but no one will understand why, and will have no idea how to fix it.

Coming back to the current PR, in my opinion, it is more accurate to say that this fix doesn't correct bug #7430, but rather masks it. The root of the issue is that a negative number is being passed as an argument instead of a positive one, not that the user sees an error.

I agree that users might be frustrated by application crashes in case of non-critical errors. However, this means that we should develop an alternative way of anonymously reporting such errors without halting the application. I believe it would be sufficient to anonymously transmit the exception traceback, keeping the developers informed that there is an error in the code.

If the error report contains only the traceback, we could propagate it without requiring the user's consent, allowing us to more accurately assess the quantity of such errors.

Therefore, I believe that in the context of the current reporting system, it's more prudent to address and rectify the root issue (specifically, the negative number of seeders) rather than concealing the error.

kozlovsky:
Regarding

> 1. Improve our test coverage by writing additional unit tests.
> 2. Expand the number of scenarios for the Application Tester.
> 3. Maintain Tribler's fragility to these tests.
> 4. Strengthen Tribler's resilience for the end users.

I think unit tests are helpful in finding regressions of known bugs, but it may be hard to find a genuinely new bug with tests. The same with Application Tester - it is very useful and allows us to discover some important issues. Still, as it only follows pre-defined scenarios (even if they are combined in random order), it cannot detect important bugs when a user does something slightly unusual.

Realistically, we can't add and maintain hundreds of scenarios to Application Tester in the hope they allow us to find all future bugs, as we do not have enough resources for it: it is one thing to add a specific scenario to check a particular case and another thing to add all possible scenarios to find all possible bugs, not known in advance.

Unit tests are very helpful, as well as the Application Tester, but we can't expect we can find all critical bugs with them.

drew2a:
@kozlovsky I appreciate the use of analogies for better understanding, however, I would exercise caution when using these particular analogies as software often operates in varied environments and under distinct regulations.

Such comparisons might potentially lead to misunderstandings or give a false sense of comprehension of the issue. Here's how it might transpire: You begin with a valid example (in this case, evolution and DNA mutations), you create a wrong analogy (bugs functioning like DNA mutations), and then you draw a false conclusion (we should avoid artificially selecting certain bugs).

When discussing software products, it might be more beneficial to use terminology and concepts specific to this field, primarily because software does not function exactly like organisms in nature. I'd like to see a Tribbler who produces offspring :)
 
Again, I appreciate your input and look forward to more insightful discussions.

@egbertbouman thank you for your balanced view. You're absolutely correct, it is about carefully determining the trade-offs for each component of Tribler. In the case of critical sections like tunnels and libtorrent, allowing Tribler to crash might be acceptable. This is because there's a possibility, albeit uncertain, that we might receive a report from the user, enabling us to address the issue.

However, for other, less critical parts of Tribler, this might not be as significant. For instance, if there's a minor rare bug in the TorrentChecker or PopularityCommunity (or do I also mention` BandwidthCommunity` #7336 to summon @devos50 :)?), it might be acceptable to simply overlook it.

---

Just to clarify, I'm not advocating for completely ignoring errors, nor am I suggesting that every error should result in crashing the application and displaying the `FeedbackDialog`. My position favors a balanced approach. In the case of non-critical components of Tribler, errors could potentially be overlooked. However, for crucial parts, they absolutely should not be ignored. I appreciate your understanding in this matter.

drew2a:
@kozlovsky 
>Coming back to the current PR, in my opinion, it is more accurate to say that this fix doesn't correct bug https://github.com/Tribler/tribler/issues/7430, but rather masks it. The root of the issue is that a negative number is being passed as an argument instead of a positive one, not that the user sees an error.

The root cause of the appearance of negative numbers is still a mystery. The negative number could come to DB from some old version of Tribler or be returned by a tracker.

We see the error because of the inconsistency of these two data structures:
```python
@dataclass(order=True)
class HealthInfo:
    infohash: bytes = field(repr=False)
    seeders: int = 0
    leechers: int = 0
    last_check: int = field(default_factory=lambda: int(time.time()))
    self_checked: bool = False
```

 ```python
@vp_compile
class TorrentInfoFormat(VariablePayload):
    format_list = ['20s', 'I', 'I', 'Q']
    names = ['infohash', 'seeders', 'leechers', 'timestamp']
    length = 36
```

Given that we aim to retain the current payloads of PopularityCommunity as they are, one potential approach to 'fix the bug' would be to more rigorously define types in HealthInfo and conduct careful checks. However, this might necessitate substantial refactoring and I question whether the benefit would justify the effort.

A more straightforward, and perhaps more appropriate approach in this situation, might be to simply ignore this and all potential errors, as suggested in commit https://github.com/Tribler/tribler/pull/7487/commits/dcff75dc0acfc2c53be9da40a54be908e833867a

Alternatively, there's another option available in commit https://github.com/Tribler/tribler/pull/7487/commits/85fedc968334beb5f9edb8a2515566cf1788a4d3

Could you please let me know which one you prefer? Would you consider accepting it as a resolution for the bug?

**Comment by kozlovsky** [comment]
@drew2a after some investigation, I think I understand the root cause of negative leechers. It is indeed possible for some versions of BitTorrent trackers to return a negative number of leechers: https://github.com/webtorrent/bittorrent-tracker/issues/65

That means Tribler should be able to handle this properly.

The solutions in https://github.com/Tribler/tribler/pull/7487/commits/dcff75dc0acfc2c53be9da40a54be908e833867a and https://github.com/Tribler/tribler/pull/7487/commits/9b8b31eeb6c461ee0a4eff391dae7d30645bd4a2 are focused on preventing the gossiping of wrong health information to other peers. But it may be better not to store the invalid health information in the database as well, as it may somehow break the logic of some queries. For this reason, it is better to reject invalid health when receiving it from a tracker.

In Tribler 7.13, we now use a convenient class `HealthInfo`, that allows us to handle this check nicely

The place where we process the incoming health is this:
```python
    async def get_tracker_response(self, session: TrackerSession) -> TrackerResponse:
        ...
        with db_session:
            for health in result.torrent_health_list:
                self.update_torrent_health(health)
```

And inside the `update_torrent_health` method, we do the following:
```python
    def update_torrent_health(self, health: HealthInfo) -> bool:
        if not health.is_valid():
            self._logger.warning(f'Invalid health info ignored: {health}')
            return False
        ...
```
So, we can extend the `HealthInfo.is_valid` method to check the negative value of seeders and leechers. Instead
```python
    def is_valid(self) -> bool:
        return self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
```
we can do
```python
    def is_valid(self) -> bool:
        return self.seeders >= 0 and self.leechers >= 0 and \
                self.last_check < int(time.time()) + TOLERABLE_TIME_DRIFT
```

This way, we not only protect other peers from receiving incorrect information but also protect the local database and an in-memory `TorrentChecker._torrents_checked` dictionary

kozlovsky:
> And what should we do with the numbers that are already in the DB?

I looked at my local database and found no negative seeder or leecher. Considering that we have only a single Sentry report with this type of error, the typical user database contains very few incorrect health records.

It is possible to find all such records and set the number of leechers to zero with a query like
```
with db_session:
    for ts in TorrentState.select(lambda ts: ts.leechers < 0):
        ts.leechers = 0
```
Or with raw SQL query:
```
with db_session:
    db.execute("""
        update TorrentState set leechers = 0 where leechers < 0
    """)
```
This query can be part of a database upgrade. While the actual number of invalid records is probably low (maybe zero), the query is still potentially slow and can take 10-15 seconds on the user's machine.

But I checked database queries, and they should work fine for records with a negative number of leechers. So, another option is to leave the database as is and fix HealthInfo records after retrieving them from the database by resetting the negative number of leechers to zero. It can be implemented in the following way:

1. Add the handling of negative leechers to the `TorrentState.to_health()` method:
```python
    class TorrentState(db.Entity):
        ...
        def to_health(self) -> HealthInfo:
            leechers = max(0, self.leechers)  # fix the rare case of negative leechers in the DB
            return HealthInfo(self.infohash, self.seeders, leechers, self.last_check, self.self_checked)
```
And use the `to_health` method when constructing HealthInfo from the database. For this, we need to replace this:
```python
    @db_session
    def load_torrents_checked_from_db(self) -> Dict[bytes, HealthInfo]:
        ...
        for torrent in checked_torrents:
            result[torrent.infohash] = HealthInfo(torrent.infohash, torrent.seeders, torrent.leechers,
                                                  last_check=torrent.last_check, self_checked=True)
        return result
```
With this:
```python
    @db_session
    def load_torrents_checked_from_db(self) -> Dict[bytes, HealthInfo]:
        ...
        for torrent_state in checked_torrents:
            result[torrent.infohash] = torrent_state.to_health()
        return result
```

I think fixing `TorrentState.to_health` this way without fixing the database records should be enough

drew2a:
@kozlovsky 
>I looked at my local database and found no negative seeders or leechers. Considering that we have only a single Sentry report with this type of error, the typical user database contains very few incorrect health records.

Then perhaps the easiest correct solution is just to ignore errors that occur because of these records, right?

PR #7492, opened 2023-06-19
Suppress exceptions in `KnowledgeCommunity`

drew2a:
As shown in #7487 not all parts of an ipv8 community are equal in error handling. This PR fixes this inconsistency for `KnowledgeCommunity`.


drew2a:
Following our offline discussion, the majority has voted not to shift Tribler's error handling policy from a developer-centric approach to a user-centric one.

Issue #7489, opened 2023-06-15
Unable to close Tribler after subscribing to a channel.


milahu:
**Describe the bug**
on shutdown, the log is flooded with this message:

```
WARNING - KnowledgeRulesProcessor(177) - Queue is full
```

tribler seems to hang forever. when i kill tribler, the log says

```
Exception ignored in: <module 'threading' from '/nix/store/0pyymzxf7n0fzpaqnvwv92ab72v3jq8d-python3-3.10.9/lib/python3.10/threading.py'>
Traceback (most recent call last):
  File "/nix/store/0pyymzxf7n0fzpaqnvwv92ab72v3jq8d-python3-3.10.9/lib/python3.10/threading.py", line 1537, in _shutdown
[tribler-core PID:1205929] 2023-06-15 14:12:00,886 - WARNING - KnowledgeRulesProcessor(177) - Queue is full
atexit_call()
  File "/nix/store/0pyymzxf7n0fzpaqnvwv92ab72v3jq8d-python3-3.10.9/lib/python3.10/concurrent/futures/thread.py", line 31, in _python_exit
t.join()
  File "/nix/store/0pyymzxf7n0fzpaqnvwv92ab72v3jq8d-python3-3.10.9/lib/python3.10/threading.py", line 1096, in join
self._wait_for_tstate_lock()
  File "/nix/store/0pyymzxf7n0fzpaqnvwv92ab72v3jq8d-python3-3.10.9/lib/python3.10/threading.py", line 1116, in _wait_for_tstate_lock
if lock.acquire(block, timeout):
KeyboardInterrupt:
[tribler-core PID:1205929] 2023-06-15 14:12:00,908 - WARNING - KnowledgeRulesProcessor(177) - Queue is full
[tribler-gui PID:1205913] 2023-06-15 14:12:02,501 - INFO - CoreManager(284) - Core process finished
[tribler-gui PID:1205913] 2023-06-15 14:12:02,729 - INFO - tribler.gui.start_gui(87) - Shutting down Tribler
```

**To Reproduce**
close tribler.

**Expected behavior**
be less verbose.

**Environment**

- OS: linux
- Tribler version: 889eb2c85024cfdd9d80b7346fb8d7443df1215c

milahu:
activities: start tribler, subscribe to some channels, wait some minutes, close tribler (Ctrl-C).

logs: i will try to reproduce.

> subscribe to some channels

at the time of closing, tribler was either/or

[...]

PR #7493, opened 2023-06-19
Print `Queue is full` once per 5 seconds

### LINKED ISSUE #7490: Spam in logs "Queue is full" during a channel download
After subscribing to a channel Tribler's logs full of

```python
WARNING - KnowledgeRulesProcessor(177) - Queue is full
WARNING - KnowledgeRulesProcessor(177) - Queue is full
WARNING - KnowledgeRulesProcessor(177) - Queue is full
WARNING - KnowledgeRulesProcessor(177) - Queue is full
```

**To Reproduce**

1. Open Tribler
2. Subscribe to a channel
3. Wait until this channel starts to download

This issue was extracted from #7489 reported by @milahu

drew2a:
This PR fixes #7490

### DIFF
--- a/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
+++ b/src/tribler/core/components/knowledge/rules/knowledge_rules_processor.py
@@ -60,6 +60,8 @@ class KnowledgeRulesProcessor(TaskManager):
         self.queue_batch_size = queue_batch_size
         self.queue_max_size = queue_max_size
 
+        self._last_warning_time = 0
+
         # this queue is used to be able to process entities supplied from another thread.
         self.queue: queue.Queue[TorrentTitle] = queue.Queue(maxsize=self.queue_max_size)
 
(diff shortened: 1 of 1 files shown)


src/tribler/core/start_core.py
```diff
@@ -169,6 +169,7 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     try:
         exit_code = loop.run_until_complete(core_session(config, components=list(components_gen(config))))
     finally:
+        logger.info(f'Finished with code: {exit_code if exit_code else 0}')
```
drew2a, line 172: This is not a part of the printing logic and was added during the investigation of #7489.
I can extend this change to another PR if necessary.

kozlovsky: This line looks incorrect for two reasons:
1. In case of an exception in `loop.run_until_complete`, this line will lead to `UnboundLocalError: local variable 'exit_code' referenced before assignment`.
2. In case of exception in `loop.run_until_complete`, it is misleading to say that the process is finished with exit code 0, because 0 means that the process is finished correctly and without errors.

As this line does not relate to other changes in this PR, I think it can be removed. And to make the line correct, you may initialize the `exit_code` variable to None before the `try/finally` block and output `The asyncio loop is finished with exception` in case the `exit_code` is still None

kozlovsky (changes requested):

kozlovsky (approved):

PR #7494, opened 2023-06-19
Add Logging for Slow Database Sessions

kozlovsky:
Sometimes we can observe a very long execution of coroutine steps that entirely freeze the asyncio loop for huge periods ([like, 87.143 seconds](https://jenkins-ci.tribler.org/job/Deployment-Tests/job/Run_Tribler_Release/job/Run_Win10_64bit/1286/consoleFull))

The primary cause of these freezes is usually the waiting time to acquire an exclusive lock before a database transaction begins. This occurs when another database session, operating in a different thread, holds the database lock.

This PR introduces logging functionality to monitor slow database sessions across all threads in the Tribler application. Slow database sessions can sometimes cause the asyncio loop to freeze for extended periods, which impacts the performance and responsiveness of the application.

Key Features:

* This PR establishes a threshold of one second for database sessions. Any database session that exceeds this duration will be reported in the application logs.
* The logs will contain comprehensive information, including the session's thread, duration, wait time for the exclusive lock, and the time for which the exclusive lock was held.
* Additionally, the logs will present a stack trace of the database session and query statistics, including the total number of queries, execution time, and details of the slowest query.

This enhancement will assist in identifying performance bottlenecks and aid in optimizing database interactions within the Tribler application.

Example log entry:

```
[tribler-core PID:15612] 2023-06-19 14:10:55,918 - WARNING - tribler.core.utilities.pony_utils(147) - Long db_session detected.
Session info:
    Thread: 'asyncio_6'
    db_session duration: 0.055 seconds
    db_session waited for the exclusive lock for 0.051 seconds
    db_session held exclusive lock for 0.004 seconds
The db_session stack:
  File "C:\dev\tribler\src\tribler\core\components\metadata_store\db\store.py", line 473, in process_compressed_mdblob
    return self.process_squashed_mdblob(decompressed_data, health_info=health_info, **kwargs)
  File "C:\dev\tribler\src\tribler\core\components\metadata_store\db\store.py", line 536, in process_squashed_mdblob
    with db_session(immediate=True):

Queries statistics for the current db_session:
13 queries executed in a total of 0.002 seconds;
The slowest query (0.001 seconds) is:
    SELECT "g"."rowid", "g"."metadata_type", "g"."reserved_flags", "g"."origin_id", "g"."public_key", "g"."id_", "g"."timestamp", "g"."signature", "g"."added_on", "g"."status", "g"."title", "g"."tags", "g"."num_entries", "g"."infohash", "g"."size", "g"."torrent_date", "g"."tracker_info", "g"."xxx", "g"."health", "g"."tag_processor_version", "g"."start_timestamp", "g"."subscribed", "g"."share", "g"."votes", "g"."local_version"
    FROM "ChannelNode" "g"
    WHERE "g"."metadata_type" IN (400, 300)
      AND ("g"."infohash" = ? OR "g"."id_" = ? AND "g"."public_key" = ?)
    LIMIT 1

Queries statistics for the entire application:
19090 queries executed in a total of 3.182 seconds;
The slowest query (0.327 seconds) is:
    SELECT "health"."rowid", "health"."infohash", "health"."seeders", "health"."leechers", "health"."last_check", "health"."self_checked", "health"."has_data"
    FROM "TorrentState" "health"
    WHERE "health"."last_check" >= ?
      AND ("health"."seeders" > 0 OR "health"."leechers" > 0)
    ORDER BY "health"."seeders" DESC, "health"."leechers" DESC, "health"."last_check" DESC
    LIMIT 100
```


src/tribler/core/utilities/pony_utils.py
```diff
@@ -65,3 +85,139 @@ def wrapper():
...
+    def acquire_lock(self):
+        assert core.local.acquire_time is None
+        core.local.acquire_count += 1
+        t = time.time()
+        super().acquire_lock()
+        core.local.lock_wait_total_duration += time.time() - t
+        core.local.acquire_time = time.time()
+
+    def release_lock(self):
+        assert core.local.acquire_time is not None
```
drew2a, line 205: Are you sure asserts are necessary here?

drew2a (changes requested):

src/tribler/core/utilities/pony_utils.py
```diff
@@ -65,3 +87,139 @@ def wrapper():
...
+                           f"The db_session stack:\n{io.getvalue()}\n"
+                           f"Queries statistics for the current db_session:\n{local_summary}\n\n"
+                           f"Queries statistics for the entire application:\n{global_summary}")
+
+    @staticmethod
+    def merge_stats(stats_iter: Iterable[StatDict]) -> StatDict:
+        result: StatDict = {}
+        for stats in stats_iter:
+            for sql, stat in stats.items():
+                if sql not in result:
```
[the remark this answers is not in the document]
kozlovsky: I tried to use defaultdict, but it is impossible in this case, as `default_factory` does not receive the current dictionary key, and I need it to initialize a new `QueryStat` instance. The current code looks minimal (I don't think we should introduce a new [`keydefaultdict`](https://stackoverflow.com/questions/2912231/is-there-a-clever-way-to-pass-the-key-to-defaultdicts-default-factory) class just for this specific case).

src/tribler/core/utilities/tests/test_pony_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        @pony_utils.db_session
+        def perform_queries():
+            for i in range(1000):
+                Entity1(a=i)
+            commit()
+            rollback()
+            Entity1.select().fetch()
+
+        caplog.clear()
+        with patch('tribler.core.utilities.pony_utils.SLOW_DB_SESSION_DURATION_THRESHOLD', 0.0):
```
drew2a, line 65: This line illustrates that the pony_utils are difficult to customize, as you resort to a patch to alter the SLOW_DB_SESSION_DURATION_THRESHOLD value.

I would recommend revising how the utils can be customized.

src/tribler/core/utilities/pony_utils.py
```diff
@@ -65,3 +88,200 @@ def wrapper():
...
+
+Queries statistics for the entire application:
+{application_query_statistics}
+"""
+
+
+class TriblerDbSession(core.DBSessionContextManager):
+    def _enter(self):
+        is_top_level_db_session = core.local.db_session is None
+        super()._enter()
```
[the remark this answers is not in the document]
drew2a: If the sequence of these lines is intentional and must be maintained as such, please include a comment to clarify.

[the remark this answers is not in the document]
kozlovsky: Why do you think the lines should be swapped? It is not correct, as `super()._enter()` changes the state of `core.local.db_session`, and we should check that we do not have any active `db_session` before calling `_enter()`, not after

src/tribler/core/utilities/tests/test_pony_utils.py
kozlovsky: I added the keyword argument `db_session(duration_threshold=...)` to not patch the constant here

src/tribler/core/utilities/pony_utils.py
```diff
@@ -185,24 +192,37 @@ def _stop_tracking(self):
 
         start_time = info.start_time
         db_session_duration = time.time() - start_time
-        if db_session_duration > SLOW_DB_SESSION_DURATION_THRESHOLD:
-            logger.warning(self._format_warning(db_session_duration, info))
 
-    def _format_warning(self, db_session_duration: float, info: DbSessionInfo) -> str:
-        local_statistics = self._summarize_stat(db.local_stats for db in databases_to_track)
+        threshold = SLOW_DB_SESSION_DURATION_THRESHOLD if self.duration_threshold is None else self.duration_threshold
```
drew2a, line 196: Please consider the following option:

```python
class TriblerDbSession(core.DBSessionContextManager):
    def __init__(self, *args, duration_threshold: float = SLOW_DB_SESSION_DURATION_THRESHOLD, **kwargs):
        # `duration_threshold` specifies how long db_session should be to trigger the long db_session warning.
        # When `duration_threshold` is None, `SLOW_DB_SESSION_DURATION_THRESHOLD` value is used instead.
        self.duration_threshold = duration_threshold
```

this way it is more clear (and should keep the desired logic).

drew2a (changes requested):
I've just left one minor comment. The rest of the PR appears good to me.

xoriole (approved):

src/tribler/core/utilities/pony_utils.py
kozlovsky: Sometimes it may be useful to change the value of `SLOW_DB_SESSION_DURATION_THRESHOLD` dynamically (for example, in the `run_tribler.py` script). The current design allows it.

The alternate approach of setting `SLOW_DB_SESSION_DURATION_THRESHOLD` as a default parameter in the `db_session` constructor has a critical drawback. In Python, default values are evaluated at the point of function definition in the defining scope. Therefore, when `@db_session` is used as a decorator, it captures the value of `SLOW_DB_SESSION_DURATION_THRESHOLD` at the time of import. This makes it impossible to modify the `SLOW_DB_SESSION_DURATION_THRESHOLD` value post-import in a manner that affects the behavior of the `@db_session` decorator. Consequently, the sequence in which modules are imported gains significance, which can introduce unexpected behaviors and maintenance challenges.

In contrast, the current design ensures that the `db_session` retrieves the value of `SLOW_DB_SESSION_DURATION_THRESHOLD` each time it is invoked. This live retrieval accommodates dynamic changes and guarantees that the session consistently uses the most up-to-date threshold, thus circumventing the abovementioned issues.

In conclusion, the present architecture not only affords greater flexibility but also prevents potential complications arising from import order dependencies.

drew2a: The proposed modification enables dynamic alteration of the `duration_threshold` value. This is evidenced by the `test_patched_db_session` passing after the said changes were applied to the current code:

```python
diff --git a/src/tribler/core/utilities/pony_utils.py b/src/tribler/core/utilities/pony_utils.py
index 70f0859..6142082 100644
--- a/src/tribler/core/utilities/pony_utils.py
+++ b/src/tribler/core/utilities/pony_utils.py
@@ -15,8 +15,8 @@ from weakref import WeakSet
 
 from pony import orm
 from pony.orm import core
-from pony.orm.dbproviders import sqlite
 from pony.orm.core import Database, select
+from pony.orm.dbproviders import sqlite
 from pony.utils import localbase
 
 SLOW_DB_SESSION_DURATION_THRESHOLD = 1.0
@@ -132,7 +132,7 @@ Queries statistics for the entire application:
 
 
 class TriblerDbSession(core.DBSessionContextManager):
-    def __init__(self, *args, duration_threshold: Optional[float] = None, **kwargs):
+    def __init__(self, *args, duration_threshold: float = SLOW_DB_SESSION_DURATION_THRESHOLD, **kwargs):
         # `duration_threshold` specifies how long db_session should be to trigger the long db_session warning.
         # When `duration_threshold` is None, `SLOW_DB_SESSION_DURATION_THRESHOLD` value is used instead.
@@ -193,8 +193,7 @@ class TriblerDbSession(core.DBSessionContextManager):
         start_time = info.start_time
         db_session_duration = time.time() - start_time
 
-        threshold = SLOW_DB_SESSION_DURATION_THRESHOLD if self.duration_threshold is None else self.duration_threshold
-        if db_session_duration > threshold:
+        if db_session_duration > self.duration_threshold:
             self._log_warning(db_session_duration, info)
 
     def _log_warning(self, db_session_duration: float, info: DbSessionInfo):
```

kozlovsky: > The proposed modification enables dynamic alteration of the duration_threshold value. This is evidenced by the test_patched_db_session passing after the said changes were applied to the current code.

No, it doesn't. After the suggested change, any db_session applied as a decorator in the module scope will use the value of `SLOW_DB_SESSION_DURATION_THRESHOLD` at the moment of the module import.

There are two reasons why the test is passed successfully after your changes:
1. The `db_session` in the test uses an explicitly specified argument `duration_threshold=0.0`, and in the current discussion, we are talking about the case when the duration threshold is not explicitly specified for a db_session (that is, for all db_sessions used in Tribler's actual code)
2. Even if we remove the explicit `duration_threshold=0.0` specification for the `db_session` in the test, the `db_session` is applied to the `_perform_queries` inner function at the test function execution time and not at the module import time, so its successful execution does not mean the same will work for db_sessions applied to module-level functions.
  
I added a new `test_patched_db_session_default_duration_threshold` test that checks the `db_session` defined at the module level, and the new test fails with your changes. In the test, I'm again dynamically changing the value `SLOW_DB_SESSION_DURATION_THRESHOLD` (as I did originally) because this is the case I want to test here, that the dynamic change of `SLOW_DB_SESSION_DURATION_THRESHOLD` is handled correctly.

drew2a: Your points are valid.

However, I don't think it's necessary or beneficial to customize the 'default value' of the threshold. Could you provide a plausible example where this feature would be used?

The change I proposed, allows for the dynamic alteration of the actual threshold value, which you even confirmed by acknowledging that the tests passed due to the explicitly specified threshold.

> In the current discussion, we are discussing the case when the duration threshold is not explicitly specified.

I perceive our current discussion differently.

We both agreed on the point: "Sometimes it may be useful to dynamically change the threshold value". Both our code versions allow this.

I don't see any advantages to allowing the `default value` to be changed. However, I do see the benefits in a simplified version of the code without this feature.

kozlovsky: I'm not sure I understand all your points correctly; probably, we sometimes name things differently, and it leads to some confusion.

> However, I don't think it's necessary or beneficial to customize the default value of the threshold. Could you provide a plausible example where this feature would be used?

What do you mean by "customizing the default value of the threshold"? The ability to pass a specific `duration_threshold` value to a specific `db_session` instance? If so, I agree that it is unnecessary. In the original version of the code, there was no such a possibility, as I expected the duration threshold to be changed globally by modifying the value of `SLOW_DB_SESSION_DURATION_THRESHOLD`. I also think the `duration_threshold` argument of `db_session` is unnecessary. The only reason I added it was your review comment:

> ```python
> with patch('tribler.core.utilities.pony_utils.SLOW_DB_SESSION_DURATION_THRESHOLD', 0.0):
> ```
> This line illustrates that the pony_utils are difficult to customize, as you resort to a patch to alter the SLOW_DB_SESSION_DURATION_THRESHOLD value.
> I would recommend revising how the utils can be customized.

I didn't agree with your opinion that changing `SLOW_DB_SESSION_DURATION_THRESHOLD` in the test is bad but I added an alternative way to customize the `db_session` behavior as you suggested.

> The change I proposed, allows for the dynamic alteration of the actual threshold value, which you even confirmed by acknowledging that the tests passed due to the explicitly specified threshold.

Can you please explain how your change allows the dynamic alteration of the current threshold value? I do not see any dynamic element here. And as I explained, the test is passed for unrelated reasons and does not check that it is possible to change the threshold value dynamically.

> We both agreed on the point: "Sometimes it may be useful to dynamically change the threshold value".

That's true!

> Both our code versions allow this.

Can you explain in more detail how your version allows this?

> I don't see any advantages to allowing the default value to be changed.

I'm not sure I understand correctly what exactly you are speaking about here. If you mean "it is not necessary for db_session to have a `duration_threshold` argument to customize the threshold on a per-db_session basis", I agree with this; it was added only because of your previous review comment I mentioned earlier.

> However, I do see the benefits in a simplified version of the code without this feature.

I agree with this! But the result of simplification should be an error-free code that doesn't demonstrate any surprising behavior.

drew2a: > What do you mean by 'customizing the default value of the threshold'?

The default threshold duration is specified as follows:

```python
SLOW_DB_SESSION_DURATION_THRESHOLD = 1.0
```

And this is how the actual threshold duration is determined:

```python
    def _stop_tracking(self):
        ...
        threshold = SLOW_DB_SESSION_DURATION_THRESHOLD if self.duration_threshold is None else self.duration_threshold
        ...
```

The direct method to alter the actual threshold duration has been demonstrated in the tests:

```python
def test_patched_db_session(tmp_path):
    ...

    @pony_utils.db_session(duration_threshold=0.0)
    def _perform_queries():
        ...
```

This can be done dynamically by passing a value to the decorator:

```python
    for duration in range(10):
        @pony_utils.db_session(duration_threshold=duration)
        def _perform_queries():
            ...
```

Thus, both the original and the proposed versions have this capability.

---

> What do you mean by "customizing the default value of the threshold"?

Regarding your code, you have the same understanding of the default value, please take a look on the `test_patched_db_session_default_duration_threshold` test:

```python
def test_patched_db_session_default_duration_threshold(tmp_path):
    # The test checks that db_session uses the current dynamic value of SLOW_DB_SESSION_DURATION_THRESHOLD
    # if no duration_threshold was explicitly specified for db_session

    with patch('pony.orm.dbproviders.sqlite.provider_cls', pony_utils.PatchedSQLiteProvider):
        db = pony_utils.TriblerDatabase()
        db.bind('sqlite', str(tmp_path / 'db.sqlite'), create_db=True)

        class Entity1(db.Entity):
            a = Required(int)

        db.generate_mapping(create_tables=True)

        # We change the value of SLOW_DB_SESSION_DURATION_THRESHOLD, and the current value should be used by db_session
        with patch('tribler.core.utilities.pony_utils.SLOW_DB_SESSION_DURATION_THRESHOLD', 0.0):
            with patch.object(pony_utils.TriblerDbSession, '_format_warning',
                              return_value='<warning text>') as format_warning_mock:
                perform_queries(db, Entity1)

        format_warning_mock.assert_called()
```

kozlovsky: As I see it, we understand the phrase "change the value dynamically" differently.

In your example, you create a new function dynamically and specify a duration threshold for it:
```python
    for duration in range(10):
        @pony_utils.db_session(duration_threshold=duration)
        def _perform_queries():
            ...
```
Here, while the function itself is created dynamically, the duration is specified statically. With your version of the code, you cannot change the duration threshold value later to the same `@db_session`-decorated function; you can only create a new `@db_session`-decorated function with a new duration threshold value.

When I speak about changing the duration value dynamically, I mean a completely different thing. Consider a `@db_session`-decorated function defined at the module level, a usual Tribler function that uses db_session decorator:
```python
        @db_session
        def consolidate_channel_torrent(self):
            ...
```
It is just a normal Tribler function. I want to be able to change the duration threshold for this function, as well as all other `@db_session` decorated functions. In most cases, it is not necessary for me to specify a duration threshold on a per-function basis; I usually want to specify it globally for all `@db_session`-decorated functions. I can do it "statically" by specifying a new version of `SLOW_DB_SESSION_DURATION_THRESHOLD` right inside the `tribler.core.utilities.pony_utils` module. But sometimes, it may be convenient to change the value at runtime in a script. For example, I can change it inside the `run_tribler.py`

```python
# run_tribler.py
import X
from tribler.core.utilities import pony_utils

...
pony_utils.SLOW_DB_SESSION_DURATION_THRESHOLD = <new_value>

import Y
```

In that case, your version of the code will not work for all `db_session` in all modules that were imported before the `SLOW_DB_SESSION_DURATION_THRESHOLD` value was changed (like, `import X`). Only `import Y` will work correctly, and only if it is not imported also from inside the module `X`. It may be confusing for some developers.

The current version of the code in the PR should work for all modules; it is not important whether they are imported before or after the value of `SLOW_DB_SESSION_DURATION_THRESHOLD ` was changed. In my opinion, it is much easier to handle.

drew2a: I agree with your point, but I question the necessity of this feature. We don't need to specify this value per module. Our only requirement is to specify this value per function or per code block (in tests).

kozlovsky: > I question the necessity of this feature

Sorry, I don't understand what feature you are talking about

> We don't need to specify this value per module

I agree with this, and the current PR code has nothing with it. We never discussed specifying the value per module (until your last comment) and have no code for this unnecessary feature.

> Our only requirement is to specify this value per function or code block (in tests)

Yes, I implemented this feature as you requested. It allows specifying the duration for a specific function or code block. Note that it is only useful in tests, as in actual Tribler code, we only want to specify the duration globally.

What I want is the ability to change the value of `SLOW_DB_SESSION_DURATION_THRESHOLD` while the Tribler application is running and have this change reflected by all db_sessions that do not have an explicit duration threshold specified. The PR code allows this.

If I understand correctly, you want to make a change that unnecessarily breaks this possibility and does not allow us to change the `SLOW_DB_SESSION_DURATION_THRESHOLD` value at runtime. The change you are suggesting to do makes the code unnecessarily brittle, as after this change, some `db_session` will respect the modified `SLOW_DB_SESSION_DURATION_THRESHOLD` value while others don't. The only reason for the change you suggest is the "simplicity" of the resulting code, as it saves one line of the code. But this change comes with a high cost, as the result of changing the `SLOW_DB_SESSION_DURATION_THRESHOLD` value at runtime becomes buggy and unpredictable, and this fact may be non-obvious for other developers. I do not see the reason for this change, and in my opinion, it makes the code worse.

So, I'd prefer to keep the possibility of changing the `SLOW_DB_SESSION_DURATION_THRESHOLD` value at runtime and be able to handle it by our code correctly.

drew2a: > So, I'd prefer to keep the possibility of changing the SLOW_DB_SESSION_DURATION_THRESHOLD value at runtime and be able to handle it by our code correctly.

We don't need it. It's unnecessary. During this discussion, you've argued that we need this feature and that my code makes this feature "buggy and unpredictable". However, my version of the code doesn't include this feature at all. And there's no real evidence that we need it.

Again, we only need two options to change it:

Globally (for all of Tribler). This is possible in both versions.
Locally, for a specific function or code block. This is also possible in both versions."

kozlovsky: There is a third scenario to consider: altering the threshold value in startup scripts like `run_tribler.py`, `run_exit_node.py`, `run_bandwidth_crawler.py`, `seeder.py`, etc.

The utility of this flexibility might be subject to debate, but it can be invaluable in certain contexts. I acknowledge your perspective, but I must emphasize that a future developer could potentially find this option beneficial.

The current PR seamlessly accommodates this scenario. The modification you suggest, while simpler, inadvertently cripples this functionality. This not only restricts the versatility but can also lead to cryptic and unintended behavior.

It's crucial to recognize that if a future developer adjusts the threshold through a script, the absence of any warning or error will leave them unaware of the underlying issue, leading to unpredictable outcomes.

I contend that retaining this flexibility is in the best interest of code robustness and future-proofing. The trade-off of sparing a single line of code is not worth the potential issues and limitations we might inadvertently introduce.

Therefore, I respectfully stand by the initial implementation as it ensures versatility, reliability, and clarity for future developers.

drew2a: >The modification you suggest, while simpler, inadvertently cripples this functionality. This not only restricts the versatility but can also lead to cryptic and unintended behavior.

Yes, it cripples unnecessary functionality that nobody is currently using or requesting.

Your claim that it 'can also lead to cryptic and unintended behavior'. The behavior becomes simpler with my version of the code because fewer scenarios are possible. Less code is executed.
Can you provide evidence to support this claim?

You refer to a "future developer" with a very specific behavior "adjusts the threshold through a script" which is pure imaginary situation. By doing this you can justify literally any change you want to make. For example:

> I acknowledge your perspective, but I must emphasize that a future developer could potentially find the option of sending emails about test results to our mail beneficial.

This sentence makes sense but it couldn't serve as an excuse to have unnecessary functionality.

kozlovsky: The point is not about functionality. It is about surprising code behavior. The ability to adjust configuration at runtime through scripts is a common practice in software development. What I'm advocating for is not an exotic feature but a standard flexibility that could be expected by developers who are used to configuring applications dynamically. 

Suppose a future developer attempts to modify the threshold at runtime, assuming it will affect all instances. In that case, they might be perplexed when only a subset of instances are affected due to the import order, as I mentioned earlier. The behavior would be inconsistent and might lead to a lot of time spent debugging.

Your analogy with sending emails about test results is not directly comparable. Email notifications are an entirely different feature, whereas the capability I am defending is an extension of configuration flexibility, which is a fundamental aspect of software customization.

Considering that the extra line of code does not introduce significant complexity and offers better flexibility, it is reasonable to keep it.

kozlovsky: To present a plausible use case, imagine running a pytest suite where there's a need to adjust the `db_session` duration threshold across all functions under test. This is a legitimate and practical scenario. The implementation in my PR caters to this requirement perfectly, but the modification you propose would make it infeasible.

kozlovsky: > Any code written by someone else can lead to unexpected behavior. This is why it should be as simple as possible and provide the most straightforward API to the user (future developer) as possible.

Absolutely! And one such unexpected behavior in Python relates to default function arguments, a pitfall that's [well-documented](https://medium.com/python-features/how-to-avoid-classic-pitfall-while-passing-default-values-in-python-7002c0dc4c7c) and being addressed via [PEP 671](https://peps.python.org/pep-0671/).

My approach aims to use the late binding of `SLOW_DB_SESSION_DURATION_THRESHOLD`, keeping the logic straightforward. However, the proposed change uses default arguments, which bind early and can lead to the aforementioned pitfall.

Let me clarify:
1. The proposed change affects an important PR that is essential for the Tribler release.
2. It disrupts use cases that I believe are valid:
   - Adjusting the configuration when a specific script starts
   - Altering the `db_session` duration threshold for testing purposes.
3. It introduces a potential complexity and error-prone behavior by relying on Python's default arguments.
4. The rationale for the proposed change seems to be primarily centered on simplicity, but I contend that it is at the cost of flexibility and falling into known Python pitfalls.

> Reduced complexity results in fewer bugs. Even a small step in this direction is worth taking.

I concur, and evading the default argument issue in Python is a step toward reducing complexity.

> We can imagine so many features that "could potentially be beneficial", but it doesn't mean we should implement them.

I agree, but in this instance, the flexibility is a natural outcome of the implementation in the PR, not an additional feature. It is streamlined, and the scenarios are accommodated by a single line of code.

In conclusion, while I appreciate the insights and value our collaborative efforts, I firmly believe that the current implementation is in the best interest of flexibility and reliability.

drew2a: The proposed change you're referring to is not as consequential as you suggest:

1. As we discussed yesterday, it is not essential for the Tribler release. All the necessary features are already in the release branch.
1. Any perceived blockage is only on your side as we already have one approval. This means a majority of developers are in agreement with the change.

Your argument about disrupting valid use cases is speculative. We do not have any real necessity for supporting these scenarios. 

Your assertion that the proposed change introduces potential complexity and error-prone behavior contradicts the arguments I wrote above (also, I asked you to prove your earlier claim, but you were not able to prove it).

drew2a (approved):

src/tribler/core/utilities/pony_utils.py
kozlovsky: > Any perceived blockage is only on your side as we already have one approval. This means a majority of developers are in agreement with the change.

I appreciate the clarification regarding the approval process. I was under the impression that an explicit rejection would necessitate addressing it before proceeding. Thank you for clearing that up.

As for the suggested change, I respect the opinions and the discussion we had, but I must stand by my conviction that the proposed modification could introduce design issues. With the understanding that the PR has received majority approval, I would like to proceed without incorporating this specific change.

Thank you for the engagement and valuable feedback throughout this review.

Issue #7500, opened 2023-06-26
GUI Tests does not finish properly


xoriole:
GUI Tests finish with non-zero exit code with logs like:
```bash
QProcess: Destroyed while process ("/home/runner/.virtualenvs/.venv/bin/python") is still running.
Error: Process completed with exit code 1.
```

This is because the core process initiated by the GUI is not terminated properly.

xoriole:
Closed by https://github.com/Tribler/tribler/pull/7501

PR #7501, opened 2023-06-26
Fix GUI Test non-zero exit code

xoriole:
The reason for the non-zero exit code is the core process not being terminated correctly.
With this PR, we wait a max 10 seconds to let the core process terminate properly.

- Fixes https://github.com/Tribler/tribler/issues/7500

### DIFF
--- a/src/tribler/gui/tests/test_gui.py
+++ b/src/tribler/gui/tests/test_gui.py
@@ -76,6 +76,11 @@ def fixture_window(tmp_path_factory):
 
     window.close_tribler()
     screenshot(window, name="tribler_closing")
+
+    # Before quitting the application, wait for max 10 seconds until the core process has successfully finished
+    # Otherwise, process exits with non-zero exit code.
+    # See: https://github.com/Tribler/tribler/issues/7500
+    wait_for_signal(window.core_manager.core_process.finished, timeout=10)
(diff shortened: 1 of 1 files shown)


src/tribler/gui/tests/test_gui.py
```diff
@@ -76,6 +76,11 @@ def fixture_window(tmp_path_factory):
 
     window.close_tribler()
     screenshot(window, name="tribler_closing")
+
+    # Before quitting the application, wait for max 10 seconds until the core process has successfully finished
+    # Otherwise, process exits with non-zero exit code.
+    # See: https://github.com/Tribler/tribler/issues/7500
+    wait_for_signal(window.core_manager.core_process.finished, timeout=10)
```
kozlovsky, line 83: Question: are ten seconds always enough here? Something like 30 seconds looks safer to me

drew2a (approved):

src/tribler/gui/tests/test_gui.py
xoriole: In my opinion, 10 seconds is sufficient since the test is running with the test state directory which shouldn't be big. If the test is timing out for 10 seconds timeout value, then I think we should first investigate and then decide whether to increase the timeout.

PR #7502, opened 2023-06-26
Fix calling `await app.shutdown()` in REST API tests

kozlovsky:
As it turns out, many REST API-related test fixtures have an `app.shutdown()` call at the end instead of `await app.shutdown()`, and it can be the reason for many problems with test stability.

In this PR, I add a `web_app` fixture that provides an instance of `aiohttp.web_app.Application` and correctly calls `await app.shutdown()` at the end.

The fixture is named `web_app` and not `app` to avoid PyCharm confusion with the `sphinx.testing.fixtures.app` fixture it tries to use when the new fixture is named `app`.

### DIFF
--- a/src/tribler/core/components/conftest.py
+++ b/src/tribler/core/components/conftest.py
@@ -1,0 +1,11 @@
+import pytest
+from aiohttp.web_app import Application
+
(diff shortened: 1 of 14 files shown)


xoriole (approved):
Nice!

Issue #7495, opened 2023-06-20
Flaky core tests


drew2a:
I'm trying to find flaky tests in the Tribler test suite by using this tool: https://github.com/dropbox/pytest-flakefinder

Here is the description of the experiment: Each test was run 50 times (no parallel), using macOS as the operating system and Python version 3.9.6.

The founded flaky tests are the following:
 - [x] core/components/metadata_store (test_channel_update_and_download)
 - [x] core/components/libtorrent (test_move_to_non_existing_dir, test_seeding) FREEZE
 - [x] core/components/restapi (no running event loop)
 - [x] core/components/restapi (test_write_data)
 - [x] core/components/restapi (test_maybe_add)
 - [x] core/components/restapi (test_aiohttp_assertion_patched)
 - [ ] core/components/tunnel (test_payouts_e2e)
 - [x] ~~core/tests (test_enable_fault_handler)~~ false positive
 - [x] ~~core/utilities (test_show_system_popup_win)~~ false positive
 - [x] core/utilities (test_create_torrent_two_files) FREEZE

From real PRs:
 - [ ] test_consolidate_channel_torrent https://github.com/Tribler/tribler/actions/runs/5543081002/jobs/10118622017?pr=7542#step:6:1236 (FREEZE), can not reproduce
 - [ ] test_commit_channel_torrent https://github.com/Tribler/tribler/actions/runs/5543923045/jobs/10121257517?pr=7543#step:6:1272 (FREEZE) can not reproduce

### metadata_store

```console
➜ pytest src/tribler/core/components/metadata_store --flake-finder

==================================== short test summary info =====================================
FAILED src/tribler/core/components/metadata_store/tests/test_channel_download.py::test_channel_update_and_download[2] - Failed: Timeout >60.0s
================== 2 failed, 10848 passed, 26001 warnings in 1952.47s (0:32:32) ==================
```

Another issue with the same test:

```console
src/tribler/core/components/metadata_store/tests/test_channel_download.py:58 (test_channel_update_and_download[26])
1565621688018 != 1565621688015

Expected :1565621688015
Actual   :1565621688018
```

### libtorrent

```console
➜ pytest src/tribler/core/components/libtorrent --flake-finder
```

A freeze was detected during the execution of `test_move_to_non_existing_dir` in `test_downloads_endpoint.py.` The test appears to freeze indefinitely:
[...]

PR #7503, opened 2023-06-26
Use TestError instead of AssertionError in test to avoid confusion

kozlovsky:
Some tests (of the Core test suite) deliberately raise an AssertionError and expect it to be in `error_handler._handled_exceptions`. It is better to raise a specific `TestError` instead of a generic `AsserionError` not to be confused about whether the error was planned or unexpected. In this PR, I change `AssertionError` to `TestError` to avoid possible confusion.

The reason I want to fix it is that I had observed seen some strange cases when the string `"AssertionError"` was in `error_handler._handled_exceptions` while it was expected to be empty (unfortunately can't provide an example right now), and I want to eliminate possible cases of it.

### DIFF
--- a/src/tribler/gui/tests/test_error_handler.py
+++ b/src/tribler/gui/tests/test_error_handler.py
@@ -22,11 +22,15 @@ def reported_error():
     return ReportedError(type='Exception', text='text', event={})
 
 
+class TestError(Exception):
+    pass
+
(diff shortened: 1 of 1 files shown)


drew2a (approved):

drew2a:
Related to #7495

PR #7504, opened 2023-06-26
The `get_circuit_slots` method of `DebugEndpoint` should handle the case when `tunnel_community` is `None`

kozlovsky:
The `run` method of `RESTComponent` has [the following lines](https://github.com/Tribler/tribler/blob/55d02e606d16edb1d09100a486da0deaf0385cea/src/tribler/core/components/restapi/restapi_component.py#L92):

```python
torrent_checker = None if config.gui_test_mode else torrent_checker_component.torrent_checker
tunnel_community = None if config.gui_test_mode else tunnel_component.community
gigachannel_manager = None if config.gui_test_mode else gigachannel_manager_component.gigachannel_manager
```

That is, `tunnel_community` is `None` when Tribler is started for GUI tests.

At the same time, the `DebugEndpoint` in the `RESTComponent.run` method is initialized [in the following way](https://github.com/Tribler/tribler/blob/55d02e606d16edb1d09100a486da0deaf0385cea/src/tribler/core/components/restapi/restapi_component.py#L100):
```python
self.maybe_add('/debug', DebugEndpoint, config.state_dir, log_dir,
               tunnel_community=tunnel_community,
               resource_monitor=resource_monitor_component.resource_monitor,
               core_exception_handler=self._core_exception_handler)
```
Here, `tunnel_community` is passed as `None` when Tribler is in `gui_test_mode`.

The `maybe_add` method does not affect this situation, as it only handles special `NoneComponent` arguments and not normal `None` values.

Therefore, the code of `DebugEndpoint` should expect to handle the situation gracefully when the `tunnel_community` is None. But in the `get_circuit_slots` method, it currently expects a non-None instance of `tunnel_community`.

This PR fixes the problem by handling the case when `self.tunnel_community` is None in `DebugEndpoint`

### DIFF
--- a/src/tribler/core/components/restapi/rest/debug_endpoint.py
+++ b/src/tribler/core/components/restapi/rest/debug_endpoint.py
@@ -2,18 +2,19 @@ import datetime
 import logging
 import sys
 from io import StringIO
+from typing import Optional
 
 import psutil
 from aiohttp import web
 from aiohttp.abc import BaseRequest
 from aiohttp_apispec import docs
 from ipv8.REST.schema import schema
-from ipv8.messaging.anonymization.community import TunnelCommunity
 from marshmallow.fields import Boolean, Float, Integer, String
 
 from tribler.core.components.reporter.exception_handler import CoreExceptionHandler
 from tribler.core.components.resource_monitor.implementation.base import ResourceMonitor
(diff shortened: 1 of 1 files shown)


drew2a (approved):

drew2a:
Related to #7495

xoriole (approved):

Issue #7498, opened 2023-06-21
Flaky GUI tests


drew2a:
I'm trying to find flaky tests in the Tribler test suite by using this tool: https://github.com/dropbox/pytest-flakefinder

Here is the description of the experiment: Each test was run 50 times (no parallel), using macOS as the operating system and Python version 3.9.6.

### no UI

```console
➜ pytest src/tribler/gui --flake-finder

======================================================== short test summary info ========================================================
FAILED src/tribler/gui/tests/test_core_manager.py::test_on_core_read_ready_os_error_suppressed[41] - assert 4 == 2
===================================== 49 failed, 2901 passed, 950 skipped, 7802 warnings in 33.59s ======================================
```

### UI

```console
➜ pytest src/tribler/gui --flake-finder

======================================================== short test summary info ========================================================
FAILED src/tribler/gui/tests/test_gui.py::test_add_download_url[6] - tribler.gui.tests.test_gui.TimeoutException: Variable downloads_page.downloads within 10 seconds
FAILED src/tribler/gui/tests/test_gui.py::test_discovered_page[42] - tribler.gui.tests.test_gui.TimeoutException: The list was not populated within 10 seconds
FAILED src/tribler/gui/tests/test_gui.py::test_add_deeptorrent[31] - AssertionError: assert not {'AttributeError'}
FAILED src/tribler/gui/tests/test_gui.py::test_tags_dialog[43] - assert []
====================================== 144 failed, 806 passed, 1903 warnings in 3192.56s (0:53:12) ======================================

```

This issue is related to https://github.com/Tribler/tribler/issues/7132 and #7134

drew2a:
I'm closing this issue as it appears that everything fixable has already been addressed. For any new detections of frozen tests, please refer to the registry at [#7706](https://github.com/link-to-issue-7706). For new flaky tests, feel free to open a separate issue.

PR #7505, opened 2023-06-26
Fix the `get_torrent_health` endpoint in GUI tests

kozlovsky:
Currently, the `RESTComponent.run` method has the following code lines:

```python
        torrent_checker = None if config.gui_test_mode else torrent_checker_component.torrent_checker
        tunnel_community = None if config.gui_test_mode else tunnel_component.community
        gigachannel_manager = None if config.gui_test_mode else gigachannel_manager_component.gigachannel_manager
```
and
```python
        self.maybe_add('/metadata', MetadataEndpoint, torrent_checker, metadata_store_component.mds,
                       knowledge_db=knowledge_component.knowledge_db,
                       tag_rules_processor=knowledge_component.rules_processor)
```

That is, under GUI tests, the `torrent_checker` argument of `MetadataEndpoint`  can be `None`. The `maybe_add` call does not handle this situation, as it can only handle `NoneComponent` arguments and not the normal `None` values.

But, the `MetadataEndpoint.get_torrent_health` methods expect `self.torrent_checker` to be not-None.

For some reason, during GUI tests, the `get_torrent_health` endpoint can be randomly called in different GUI tests, provoking a Core exception and a test failure.

This PR fixes the `get_torrent_health` endpoint by handling the case when `self.torrent_checker` is `None`.

### DIFF
--- a/src/tribler/core/components/metadata_store/restapi/metadata_endpoint.py
+++ b/src/tribler/core/components/metadata_store/restapi/metadata_endpoint.py
@@ -1,10 +1,11 @@
 from binascii import unhexlify
+from typing import Optional
 
 from aiohttp import ContentTypeError, web
 from aiohttp_apispec import docs
 from ipv8.REST.base_endpoint import HTTP_BAD_REQUEST, HTTP_NOT_FOUND
 from ipv8.REST.schema import schema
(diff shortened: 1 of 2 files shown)


drew2a:
Related to #7498

PR #7507, opened 2023-06-27
Add `ClientOSError` as an expected exception

drew2a:
This PR fixes the flakiness of `test_aiohttp_assertion_patched` from #7495
@kozlovsky please confirm that this is a proper fix.


kozlovsky (approved):
@drew2a it looks like a correct fix

PR #7517, opened 2023-06-28
Rendezvous Certificates

InvictusRMC:
This PR adds rendezvous certificates to Tribler. This serves as a method for determining the online time of fellow peers. The initial certificate piggybacks on the introduction message of the `PopularityCommunity`. Subsequent rendezvous pings are served through separate payloads. The implementation does not use a separate community in order to reduce overhead.
The design works as follows: Peer `A` sends a ping message including a nonce serving as a challenge. Peer `B` returns the signed nonce. Peer `A` then increases its counter for Peer `B`. Online time can be estimated by multiplying the number of pings by the interval between pings.

### DIFF
--- a/src/tribler/core/components/popularity/community/popularity_community.py
+++ b/src/tribler/core/components/popularity/community/popularity_community.py
@@ -43,7 +43,7 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
     GOSSIP_RANDOM_TORRENT_COUNT = 10
 
     PING_INTERVAL_RENDEZVOUS = 60  # seconds
-    DB_NAME = 'rendezvous.db'
+    RENDEZVOUS_DB_NAME = 'rendezvous.db'
 
     community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1649')
 
--- a/src/tribler/core/components/popularity/community/popularity_community.py
+++ b/src/tribler/core/components/popularity/community/popularity_community.py
@@ -43,9 +43,8 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
     GOSSIP_RANDOM_TORRENT_COUNT = 10
 
     PING_INTERVAL_RENDEZVOUS = 60  # seconds
-    RENDEZVOUS_DB_NAME = 'rendezvous.db'
 
-    community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1649')
+    community_id = unhexlify('9aca62f878969c437da9844cba29a134917e1648')
 
     def __init__(self, *args, torrent_checker=None, rendezvous_db=None, **kwargs):
         # Creating a separate instance of Network for this community to find more peers
@@ -62,9 +61,9 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
 
         self.logger.info('Popularity Community initialized (peer mid %s)', hexlify(self.my_peer.mid))
         self.register_task("gossip_random_torrents", self.gossip_random_torrents_health,
-                           interval=PopularityCommunity.GOSSIP_INTERVAL_FOR_RANDOM_TORRENTS)
+                           interval=self.GOSSIP_INTERVAL_FOR_RANDOM_TORRENTS)
         self.register_task("ping_rendezvous", self.ping_rendezvous,
-                           interval=PopularityCommunity.PING_INTERVAL_RENDEZVOUS)
+                           interval=self.PING_INTERVAL_RENDEZVOUS)
 
         # Init version community message handlers
         self.init_version_community()
@@ -255,14 +254,13 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
         if not self.crypto.is_valid_signature(peer.key, raw_payload.challenge, signature.signature):
             self.logger.warning(f"Received invalid signature from {peer.mid}")
             return
-        else:
-            # This nonce has been burned.
-            self.rendezvous_cache.clear_peer_challenge(peer)
+
+        # This nonce has been burned.
+        self.rendezvous_cache.clear_peer_challenge(peer)
 
         self.logger.debug(f"Received valid rendezvous response from {peer.mid}")
-        with db_session:
+        with db_session(immediate=True):
             certificate = self.rdb.Certificate.get(public_key=peer.mid)
             if not certificate:
                 certificate = self.rdb.Certificate(public_key=peer.mid, counter=0)
             certificate.counter += 1
-        return
--- a/src/tribler/core/components/popularity/community/popularity_community.py
+++ b/src/tribler/core/components/popularity/community/popularity_community.py
@@ -1,15 +1,24 @@
 from __future__ import annotations
 
 import random
+import secrets
 from binascii import unhexlify
 from typing import List, TYPE_CHECKING
 
+from cryptography.exceptions import InvalidSignature
 from ipv8.lazy_community import lazy_wrapper
+from ipv8.messaging.interfaces.udp.endpoint import UDPv4Address, UDPv4LANAddress
+from ipv8.messaging.serialization import PackError
 from pony.orm import db_session
 
 from tribler.core.components.metadata_store.remote_query_community.remote_query_community import RemoteQueryCommunity
 from tribler.core.components.popularity.community.payload import PopularTorrentsRequest, TorrentsHealthPayload
 from tribler.core.components.popularity.community.version_community_mixin import VersionCommunityMixin
+from tribler.core.components.popularity.rendezvous.rendezvous import RendezvousRequestPayload, \
+    RendezvousResponsePayload, RawRendezvousResponsePayload, \
+    RendezvousChallenge, RendezvousSignature
+from tribler.core.components.popularity.rendezvous.db.database import RendezvousDatabase
+from tribler.core.components.popularity.rendezvous.rendezvous_cache import RendezvousCache
 from tribler.core.components.torrent_checker.torrent_checker.dataclasses import HealthInfo
 from tribler.core.utilities.pony_utils import run_threaded
 from tribler.core.utilities.unicode import hexlify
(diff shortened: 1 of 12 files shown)


src/tribler/core/components/popularity/popularity_component.py
```diff
@@ -22,13 +24,16 @@ async def run(self):
         metadata_store_component = await self.require_component(MetadataStoreComponent)
         torrent_checker_component = await self.require_component(TorrentCheckerComponent)
 
+        rendezvous_db = RendezvousDatabase(db_path=self.session.config.state_dir / STATEDIR_DB_DIR / PopularityCommunity.RENDEZVOUS_DB_NAME)
```
InvictusRMC, line 27: Does this require a separate `Component` instead?

drew2a: No, it's perfectly acceptable to create a database here. However, don't forget to close the database if it's necessary.

https://github.com/Tribler/tribler/blob/3b760abf3f5d5d29a84d8d0c27df543f07fc7084/src/tribler/core/components/knowledge/knowledge_component.py#L57-L58

drew2a (changes requested):
First of all, congratulations on your first contribution to the Tribler codebase!

Regarding the PR — it looks great. I've only reviewed the community part so far and we've already discussed some points offline. I assume that addressing these points could lead to changes in the code, so I will return to the review process later (please request a review when you are ready).

Besides that, I have a comment about defining the database. You've used the old "metadata"-style approach to define separate entities, which is as follows:

```python
        self.MiscData = misc.define_binding(self.database)
        self.Certificate = certificate.define_binding(self.database)
```

I could be mistaken, but it seems clear that this approach may not be convenient for the end-developer. We have transitioned to a different approach, as utilized in KnowledgeDB. Please see the following:
https://github.com/Tribler/tribler/blob/3b760abf3f5d5d29a84d8d0c27df543f07fc7084/src/tribler/core/components/knowledge/db/knowledge_db.py#L69

@kozlovsky am I right regarding the DB definition?

synctext:
First of all, congratulations also from my side on one of those rare phd code contributions to Tribler! Much appreciated :rocket: :1st_place_medal: :rocket: 

Goals is said to be: ```determining the online time of fellow peers```. Reading through the code I got inspiration for an "algorithm 1" type of innovation we require for solid publications. [```return RendezvousCertificate.get(public_key == pk).count()```](https://github.com/Tribler/tribler/pull/7517/files#diff-4326f20048dbe8f9e60a55b038217a4d7cdb1376c96e26cafe84c22234654d60R12) this code calculates the count of rendezvous certificates. In future, use an "algorithm 1" type of approach to calculate the probability of this identity being a Sybil, given the volume, age, and IPv4 diversity of the "rendezvous DAG". This goes a bit beyond MeritRank or is equal to meritRank? Should be N log N complexity.

synctext (approved):
First of all, congratulations also from my side on one of those rare phd code contributions to Tribler! Much appreciated :rocket: :1st_place_medal: :rocket: 

Goals is said to be: ```determining the online time of fellow peers```. Reading through the code I got inspiration for an "algorithm 1" type of innovation we require for solid publications. [```return RendezvousCertificate.get(public_key == pk).count()```](https://github.com/Tribler/tribler/pull/7517/files#diff-4326f20048dbe8f9e60a55b038217a4d7cdb1376c96e26cafe84c22234654d60R12) this code calculates the count of rendezvous certificates. In future, use an "algorithm 1" type of approach to calculate the probability of this identity being a Sybil, given the volume, age, and IPv4 diversity of the "rendezvous DAG". This goes a bit beyond MeritRank or is equal to meritRank? Should be N log N complexity.

InvictusRMC:
Great idea! This first version is just datacollection. This scoring will go beyond MeritRank as it will require multiple dimensions. It's not extreme complex, as we can introduce weights to individual MeritRank scores after having run MeritRank on the relevant metric.

kozlovsky:
> Besides that, I have a comment about defining the database. You've used the old "metadata"-style approach to define separate entities, which is as follows:
>
> ```python
>         self.MiscData = misc.define_binding(self.database)
>         self.Certificate = certificate.define_binding(self.database)
> ```
>
> I could be mistaken, but it seems clear that this approach may not be convenient for the end-developer. We have transitioned > to a different approach, as utilized in KnowledgeDB. Please see the following:
https://github.com/Tribler/tribler/blob/3b760abf3f5d5d29a84d8d0c27df543f07fc7084/src/tribler/core/components/knowledge/db/knowledge_db.py#L69
>
> @kozlovsky am I right regarding the DB definition?

I think the current approach used in this PR has some benefits; it allows PyCharm IDE to understand the type of expressions like `self.Certificate`.

In the `KnowledgeDatabase`, all entities are defined inside a single `define_binding` method. This way it is possible to use less number of files, as all entities are defined in a single file. The drawback is that PyCharm can't deduce types of expressions like `self.instance.StatementOp`.

If you want to combine the benefits of both approaches, you can define all entities in a single `RendezvousDatabase` method called from the `__init__`, return a tuple of entity classes as a result value, and assign them as fields of the `RendezvousDatabase`:

```
class RendezvousDatabase:
    def __init__(self, db_path: Union[Path, type(MEMORY_DB)]):
        self.database = Database()
        self.Certificate, self.MiscData = self.define_binding()
        ...
    def define_binding(self):
        class Certificate(self.database.Entity):
            ...
            
        class MiscData(self.database.Entity):
            ...
            
        return Certificate, MiscData
```

Then, all entities can be defined in a single file (if it is considered beneficial), and PyCharm understands the types of expressions like `rdb.Certificate`

src/tribler/core/components/popularity/rendezvous/db/orm_bindings/certificate.py
```diff
@@ -0,0 +1,14 @@
...
+
+def define_binding(db):
+    class RendezvousCertificate(db.Entity):
+        public_key = Required(bytes, index=True)
+        counter = Required(int)
+
+        @classmethod
+        @db_session
+        def get_count(cls, pk: bytes) -> int:
+            return RendezvousCertificate.get(public_key == pk).count()
```
kozlovsky, line 12: It looks like the method is currently not used. It can probably be deleted.

If you want to keep the method, it should be fixed, as the code is incorrect: `RendezvousCertificate.get(...)` call returns a single certificate object, and a single certificate object does not have the `count()` method.

The correct code (if it is necessary) should probably looks like:

```python
        def get_count(cls, pk: bytes) -> int:
            certificate = RendezvousCertificate.get(public_key=pk)
            return 0 if certificate is None else certificate.counter
```

src/tribler/core/components/popularity/community/popularity_community.py
```diff
@@ -140,3 +232,37 @@ def get_random_torrents(self) -> List[HealthInfo]:
...
+        if not self.crypto.is_valid_signature(peer.key, raw_payload.challenge, signature.signature):
+            self.logger.warning(f"Received invalid signature from {peer.mid}")
+            return
+        else:
+            # This nonce has been burned.
+            self.rendezvous_cache.clear_peer_challenge(peer)
+
+        self.logger.debug(f"Received valid rendezvous response from {peer.mid}")
+        with db_session:
+            certificate = self.rdb.Certificate.get(public_key=peer.mid)
```
kozlovsky, line 264: To avoid possible database locking, it is better to either use here `db_session(immediate=True)` or `Certificate.get_for_update(public_key=peer.mid)` (the result is the same)

kozlovsky (changes requested):

synctext:
@InvictusRMC Can you address the comments and requested changes? {you will find out I guess that "production code polishing" is like eating oatmeal, brushing teeth, etc. :rofl: } 
Please inquire about the release scheduling with @drew2a. We're stabilising for many months for a Tribler release. Your code would be the exclusive focus of the next-release possibly.

synctext:
btw Advise of @qstokkink is to test the impact on performance using minimal test Gumby experiment. A single UDP message triggering a database write/commit is scary :fearful: Something we had in [2013 Dispersy times](https://github.com/Tribler/dispersy/pull/14)

InvictusRMC:
> @InvictusRMC Can you address the comments and requested changes? {you will find out I guess that "production code polishing" is like eating oatmeal, brushing teeth, etc. 🤣 }
> Please inquire about the release scheduling with @drew2a. We're stabilising for many months for a Tribler release. Your code would be the exclusive focus of the next-release possibly.

Comments addressed! Thank you for the reminder.

InvictusRMC:
> btw Advise of @qstokkink is to test the impact on performance using minimal test Gumby experiment. A single UDP message triggering a database write/commit is scary 😨 Something we had in [2013 Dispersy times](https://github.com/Tribler/dispersy/pull/14)

Discussed with @qstokkink and @kozlovsky  and they advised me to implement batching logic, as a large number of transactions would slow things down considerably.

src/tribler/core/components/popularity/community/popularity_community.py
```diff
@@ -35,22 +42,98 @@ class PopularityCommunity(RemoteQueryCommunity, VersionCommunityMixin):
...
         self.register_task("gossip_random_torrents", self.gossip_random_torrents_health,
                            interval=PopularityCommunity.GOSSIP_INTERVAL_FOR_RANDOM_TORRENTS)
+        self.register_task("ping_rendezvous", self.ping_rendezvous,
+                           interval=PopularityCommunity.PING_INTERVAL_RENDEZVOUS)
 
         # Init version community message handlers
         self.init_version_community()
+        self.rendezvous_cache = RendezvousCache()
+
+    def send_introduction_request(self, peer):
```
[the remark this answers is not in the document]
InvictusRMC: The current approach is still compatible! Older peers will just ignore the extra bytes. I could drop this entire extra logic though. We can get it to work through only separate payloads.

qstokkink:
After getting a small lecture about MeritRank, my advice is as follows.

1. Hook up a `PeerObserver` to `ipv8` before starting it (line 63) here: https://github.com/Tribler/tribler/blob/737a7e7dae84284be9f1d061e972dc14cb54f75f/src/tribler/core/components/ipv8/ipv8_component.py#L60-L63
2. Inside your new observer's `remove_peer(self, peer: Peer)` callback, store `time.time() - peer.creation_time` for the `peer.public_key` in a database. This database is probably best managed inside of the `Ipv8Component` itself.
3. Future work? Make a new community (or hook into MeritRank code) to share these entries.

synctext:
This task is now taking over 4 months. @qstokkink indicated it is possible to re-factor the ```introduction-request``` and ```introduction-response``` with the _ping_ features required for Sybil attack protection. 

Related work. This would make Tribler the first academically self-organising system with Sybil protection. See [IPFS attack in a USENIX paper](usenix.org/conference/usenixsecurity22/presentation/prunster), [DHT repair blog](https://blog.ipfs.tech/2020-10-30-dht-hardening/) and [DHT health reporting](https://github.com/plprobelab/network-measurements/blob/542d9c4df0882fdad87c7f7a316e06bafd1639cf/results/rfm17-provider-record-liveness.md)

synctext:
@qstokkink as you pointed out today: no ORM in IPv8 :x: No database storage. Can you comment here a possible new API which would provide the signed certificates to the IPv8 community. Thus how can we request the rendezvous certificates (Pub-key-Them,Pub-key-OURS,nonce,signature-Them) from IPv8? 
We are not trying to slow the network, so the default ```RATE_RENDEZVOUS_CERTIFICATES == 10 seconds``` inside IPv8. Meaning maximum 1 new _random_ certificate per 10 seconds (or some other simple rate limit mechanism).

InvictusRMC:
@qstokkink rebased this into the ipv8 module of Tribler: #7630. Thank you for picking up the slack 🙏. Closing for now.

drew2a:
Please be aware that #7630 is not a rebase of #7517; rather, it's an entirely distinct PR.

qstokkink:
The commit https://github.com/Tribler/tribler/pull/7630/commits/e99dba89d7c488aed88dd5bc3209bfd5b84d2737, which is part of #7630, is the rebase of #7517.

PR #7522, opened 2023-06-29
Make `rest_api` fixture async

drew2a:
This PR fixes `rest_api` by making it async.

Previously the common pattern was:

```python

@pytest.fixture
def rest_api(event_loop, aiohttp_client, knowledge_endpoint):
    app = Application()
    app.add_subapp('/knowledge', knowledge_endpoint.app)
    yield event_loop.run_until_complete(aiohttp_client(app))
    app.shutdown()
```

which is incorrect as `app.shutdown()` is async call.

Additionally, this PR includes some minor refactoring.

Related to #7495

PR #7526, opened 2023-07-05
The `rest_api` fixture deduplication

drew2a:
This PR removes `rest_api` duplicates by introducing the base `rest_api` fixture

```python
@pytest.fixture
async def rest_api(aiohttp_client, endpoint: RESTEndpoint):
    app = Application(middlewares=[error_middleware])
    app.add_subapp(endpoint.path, endpoint.app)

    yield await aiohttp_client(app)

    await endpoint.shutdown()
    await app.shutdown()
```

In each test file that requires the use of this fixture, the endpoint fixture needs to be specified.

Example:
```python
@pytest.fixture
def endpoint(tribler_config):
    return SettingsEndpoint(tribler_config)
```

Thus, a specialized instance of the endpoint will be used in each test file.

This new logic has been made possible by the introduction of the path property to the `RESTEndpoint` class.

Related to #7522

### DIFF
--- a/src/tribler/core/components/bandwidth_accounting/restapi/bandwidth_endpoint.py
+++ b/src/tribler/core/components/bandwidth_accounting/restapi/bandwidth_endpoint.py
@@ -15,6 +15,7 @@ class BandwidthEndpoint(RESTEndpoint):
     """
     This endpoint is responsible for handing requests for bandwidth accounting data.
     """
+    path = '/bandwidth'
 
     def __init__(self, bandwidth_community: BandwidthAccountingCommunity):
         super().__init__()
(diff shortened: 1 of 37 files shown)


kozlovsky:
Looks pretty nice!

It may be convenient to remove the `path` argument of the `maybe_add` method and take the path from the endpoint itself inside the method.

In that case, for the `/collections` endpoint, we can have a simple subclass of `ChannelsEndpoint` with a different path.

(Actually, it seems to me that the `/collections` entry may be unnecessary, as it is handled by the identical copy of the `/channels` endpoint, initialized with the same arguments, that does not check the URL path of the request)

drew2a:
@kozlovsky I'll try to remove the path argument, indeed. Also as far as I see and understand from your message, the `/collections` endpoint binding could be safely removed.

kozlovsky (approved):

PR #7534, opened 2023-07-10
Remove unused core dependencies

drew2a:
This PR cleans up the `requirements-core.txt` by removing unused dependencies (`pyasn1`, `service-identity`)

Also, it contains refactoring of the `instrumentation.py` to make it possible to remove the `decorator` (library) dependency.


kozlovsky (changes requested):
I'd like to provide feedback on multiple aspects:

(first, TLDR version:)

1. I believe that the `@synchronized` decorator isn't an optimal abstraction since it conceals lock handling, which, ideally, should be explicitly managed.
2. There is a newly introduced bug within the `WatchDog.run` method due to the current refactoring.
3. This bug doesn't surface since the `WatchDog.run` method isn't invoked.
4. The `WatchDog` class, originally designed when Tribler was based on threads and not asyncio, is now essentially obsolete except for one function: `WatchDog.get_threads_info`.
5. The `WatchDog.get_threads_info` function is flawed as it doesn't account for potential thread context switches during its execution.

Therefore, I recommend the complete removal of the `tribler.core.utilities.instrumentation` module. We could rework `WatchDog.get_threads_info` as a standalone function after addressing its concurrency issue, while the rest of the `tribler.core.utilities.instrumentation` module can be discarded.

Let's delve into specifics:

The `@synchronized` decorator may not be necessary in our context.

Its purpose is to abstract away locks. However, such abstraction can inadvertently result in deadlocks, given that locks are non-composable and the order of their usage should be explicit. If we're employing the `@synchronized` decorator on various functions, we need to ensure they don't call each other to avoid deadlock scenarios. This might be challenging when the lock usage is obscured by the decorator.

In our codebase, the `@synchronized decorator` is used only on three functions within the same class:

```python
class WatchDog(Thread):
    def __init__(self):
        ...

    @synchronized
    def _reset_state(self):
        ...
        
    @synchronized
    def register_event(self, event, name, timeout=10):
        ...

    @synchronized
    def unregister_event(self, name):
        ...

    def run(self):
        events_to_unregister = []
        while not self.should_stop:
            sleep(0.2)
            with self._synchronized_lock:
                ...
```

Note that the run method uses `self._synchronized_lock`, which was injected by the previous version of the `@synchronized` decorator. The new decorator version doesn't inject this lock, causing the `run` method to encounter an uninitialized attribute. However, since the `WatchDog` thread is not currently initiated in Tribler, this error in the `run` method doesn't trigger an exception.

We could rewrite the class to function without the `@synchronized` decorator. Here's an example that seems clearer:

```python
class WatchDog(Thread):
    def __init__(self):
        ...
        self._lock = Lock()

    def _reset_state(self):
        with self._lock:
            ...
        
    def register_event(self, event, name, timeout=10):
        with self._lock:
            ...

    def unregister_event(self, name):
        with self._lock:
            ...

    def run(self):
        events_to_unregister = []
        while not self.should_stop:
            sleep(0.2)
            with self._lock:
                ...
```

However, we're not using this class currently, aside from the single `WatchDog.get_threads_info` method. Thus, we could discard the `WatchDog` class entirely, barring this one function.

`WatchDog.get_threads_info` fails to account for the possibility of a thread context switch during its execution. This can be remedied by temporarily adjusting `sys.setswitchinterval` during the function's execution. However, caution is advised since we're already modifying `sys.setswitchinterval` elsewhere in the codebase.

drew2a:
I've rewritten thread frames formatting:
image

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
...
 
     return new_arr
+
+
+def shorten(s: Optional[str], width: int = 50, placeholder='[...]', cut_at_the_end: bool = True):
+    """ Shorten a string to a given width.
+        Example: shorten('hello world', 5) -> 'hello[...]'
+    """
+    if s and len(s) > width:
+        return f'{s[:width]}{placeholder}' if cut_at_the_end else f'{placeholder}{s[len(s) - width:]}'
```
kozlovsky, line 83: **nitpick:** instead of `s[len(s) - width:]` I think it is possible to just use `s[-width:]`

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
             new_arr.append(item)
 
     return new_arr
+
+
+def shorten(s: Optional[str], width: int = 50, placeholder='[...]', cut_at_the_end: bool = True):
```
kozlovsky, line 78: I'd expect `width` to be the total width *include* placeholder, not that placeholder is added to that width.

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
...
+    """
+    if s and len(s) > width:
+        return f'{s[:width]}{placeholder}' if cut_at_the_end else f'{placeholder}{s[len(s) - width:]}'
+    return s
+
+
+def format_frames(frame: Optional[FrameType], file_width: int = 50, value_width: int = 100) -> Iterator[str]:
+    """ Format a stack trace."""
+    while frame:
+        filename = shorten(frame.f_code.co_filename, width=file_width, cut_at_the_end=False)
```
kozlovsky, line 90: **nitpick:** It may be a bit easier to read the function call with the `True` argument value:
```python
filename = shorten(frame.f_code.co_filename, width=file_width, cut_at_the_beginning=True)
```
And in the reverse case don't need to specify the `False` value explicitly, as it is the default value.

But it is a tradeoff between the readability of the function signature and the actual function call.

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
...
+        header = f"{filename}:{frame.f_lineno} {frame.f_code.co_name}"
+        local = ''
+        for key, value in list(frame.f_locals.items()):
+            value = shorten(repr(value), width=value_width)
```
kozlovsky, line 94: Taking the `repr` of an unknown value can trigger an exception; I think we should add a try/except block here. Something like:
```python
try:
    value = repr(value)
except Exception as e:
    value = f'<exception when taking repr: {e.__class__.__name__}: {e}>'
value = shorten(value, width=value_width)    
```

src/tribler/core/components/restapi/rest/utils.py
kozlovsky: **nitpick** The `'[...]'` default placeholder looks a bit heavyweight to me; maybe we can just use `'...'`

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
...
+    if s and len(s) > width:
+        return f'{s[:width]}{placeholder}' if cut_at_the_end else f'{placeholder}{s[len(s) - width:]}'
+    return s
+
+
+def format_frames(frame: Optional[FrameType], file_width: int = 50, value_width: int = 100) -> Iterator[str]:
+    """ Format a stack trace."""
+    while frame:
+        filename = shorten(frame.f_code.co_filename, width=file_width, cut_at_the_end=False)
+        header = f"{filename}:{frame.f_lineno} {frame.f_code.co_name}"
```
kozlovsky, line 91: **nitpick:** It may be convenient to also see the actual code line text as in the usual traceback

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +73,42 @@ def fix_unicode_array(arr):
...
+        local = ''
+        for key, value in list(frame.f_locals.items()):
+            value = shorten(repr(value), width=value_width)
+            local += f'\n\t{key} = {value}'
+        frame = frame.f_back
+        yield header + local
+
+
+def get_threads_info() -> List[Dict]:
```
kozlovsky, line 100: I'm a bit nervous about that function, as it is not thread-safe. The function call execution can potentially be very slow, especially because we calculate the repr of some random values, and some reprs can be huge, for example, for some JSON structures.

That means that the thread switch can happen during this calculation, and it is possible that the other thread will change the value, the repr of which we are calculating in the current thread, and it may not be thread-safe.

Even without repr rendering, the context switch can happen at any time when we traverse `frame = frame.f_back`, and it probably can lead to some surprising results.

So, it is important to avoid context switches while we are calculating the result. We can do it by setting `sys.setswitchinterval()` to a reasonably high value and then restoring it at the function end. But we should avoid the race condition with `main_thread_stack_tracking._get_main_thread_stack_info` that also calls `sys.setswitchinterval()` and can be executed in a parallel thread.

To avoid the race condition, we can introduce a context manager like:

```python
switch_interval_lock = Lock()

@contextmanager
def switch_interval(value):
    with switch_interval_lock:
        prev_value = sys.getswitchinterval(value)
        sys.setswitchinterval(value)
        try:
            yield
        finally:
            sys.setswitchinterval(prev_value)
```

And use it in all functions that currently call `sys.setswitchinterval`, like:

```python
def get_threads_info() -> List[Dict]:
    with switch_interval(ONE_SECOND):
        ...
```

kozlovsky (changes requested):

src/tribler/core/components/restapi/rest/utils.py
drew2a: 👍

src/tribler/core/components/restapi/rest/utils.py
drew2a: It is just your personal preference. https://docs.python.org/3/library/textwrap.html

src/tribler/core/components/restapi/rest/utils.py
drew2a: Could you provide an example of a dangerous call?

src/tribler/core/components/restapi/rest/utils.py
drew2a: What do you mean?

src/tribler/core/components/restapi/rest/utils.py
drew2a: Isn't it too complicated for a minor feature in the debug tab?

src/tribler/core/components/restapi/rest/utils.py
drew2a: > I'd expect width to be the total width including the placeholder, not that the placeholder is added to that width.

Yes, that version is more accurate. However, I would argue that we don't necessarily need it. Please keep in mind that this correct version introduces significant complexity compared to the original one.

drew2a: It would be something like this:
```python
def shorten(s: Optional[str], width: int = 50, placeholder='[...]', cut_at_the_end: bool = True):
    """ Shorten a string to a given width.
        Example: shorten('hello world', 5) -> 'hello[...]'
    """
    max_length = width - len(placeholder)
    if max_length < 0:
        max_length = width
        placeholder = ''

    if s and len(s) > max_length:
        return f'{s[:max_length]}{placeholder}' if cut_at_the_end else f'{placeholder}{s[-max_length:]}'
    return s
```

src/tribler/core/components/restapi/rest/utils.py
drew2a: After some thinking, I agree that this improvement should be implemented. Thank you.

src/tribler/core/components/restapi/rest/utils.py
kozlovsky: It is expected for the `__repr__` method of any object to return a string and don't raise any exception. With this, a `__repr__` method is just a usual method and can raise an exception due to a bug in its code.
```python
class AClassWithABadRepr1:
    def __repr__(self):
        1/0
```

In Tribler, we had a case of incorrect implementation of the `alert.__str__()` method that raises `UnicodeDecodeError` (#5740). In this case, the `alert.__repr__()` implementation does not raise an error, but it is possible to discover a similar problem later with an incorrect `__repr__` implementation of some class.

As another example, historically, some `repr` methods of container classes could not handle situations when the container itself was added as its item (or as an item of another item, etc.) and raised `RecursionError`. Now Python standard structures like lists and dicts handle this case correctly, but it is possible that some user-defined classes still have this problem.

[Some blog posts](https://blog.ionelmc.ro/2020/01/20/is-there-anything-safe-in-python/) describe the problems with the repr of arbitrary objects.

Python issue tracker has an issue for a case like ours: [traceback.FrameSummary does not handle exceptions from `repr()`](https://bugs.python.org/issue39228). They decided not to catch these exceptions because Python follows the ["fail fast"](https://en.wikipedia.org/wiki/Fail-fast) approach:
> Bugs should be reported, not masked. I don't think that FrameSummary should be an exception to this.

But they agree that the case is real, and the exception is possible. We don't follow the "fail fast" approach in Tribler, so we don't want Tribler to crash because a user opened a debug window.

Also, `unittest.util` has a `safe_repr` function, and probably for a reason.

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -1,8 +1,30 @@
...
+_switch_interval_lock = threading.Lock()
+
+
+@contextmanager
+def switch_interval(value):
+    """ Temporarily change the sys.setswitchinterval value."""
+    with _switch_interval_lock:
+        previous_value = sys.getswitchinterval()
+        try:
+            sys.setswitchinterval(value)
```
kozlovsky, line 23: **nitpick:** Usually the call like this is placed before the `try` block:
```python
change_something()
try:
   ...
finally:
    change_something_back()
```
In the example above, if `change_something()` behaves atomically and raises an exception, that means that "something" hasn't changed, and restoring it is unnecessary.

In our case, `sys.setswitchinterval(value)` also behaves atomically. It can throw an exception if the value has an incorrect type, but in that case, the internal variable wasn't changed, and restoring anything is unnecessary.

src/tribler/core/components/restapi/rest/utils.py
kozlovsky: Also, it is possible to throw an exception:
```
if len(placeholder) > width:
    raise ValueError(f"Placeholder {placeholder!r} is too big for the specified width {width}")
max_length = width - len(placeholder)
```

But I agree it is not very important in our case not to exceed the specified width, so we can skip it.

src/tribler/core/components/restapi/rest/utils.py
```diff
@@ -66,3 +88,43 @@ def fix_unicode_array(arr):
...
+        yield header + local
+
+
+def get_threads_info() -> List[Dict]:
+    """
+    Return information about available threads.
+    """
+
+    result = []
+    with switch_interval(ONE_SECOND):
```
kozlovsky, line 121: It is important to also use `with switch_interval` in `tribler.coro.utilities.slow_coro_detection.main_thread_stack_tracking._get_main_thread_stack_info`; otherwise, it does not protect from the race condition.

src/tribler/core/components/restapi/rest/utils.py
drew2a: I'm not entirely sure about your statement that 'we don't follow the "fail fast" approach in Tribler'. In the previous dev meeting, the majority of the developers voted to maintain the current approach, which is a form of "fail fast".

However, I do agree with your argument and appreciate the clear examples you've provided.

I have now implemented your suggestion.

drew2a:
The formatted threads info after all improvements:
image

src/tribler/core/components/restapi/rest/utils.py
drew2a: Added 🙏

src/tribler/core/components/restapi/rest/utils.py
drew2a: Like this?

```python

def _get_main_thread_stack_info() -> List[StackFrameInfo]:
    """
    Quickly copies necessary information from the main thread stack, so it is possible later to format a usual
    traceback in a separate thread.

    The function temporarily changes the interpreter’s thread switch interval to prevent thread switch during
    the stack copying. It is a lighter analogue of holding the GIL (Global Interpreter Lock).
    """
    with switch_interval(SWITCH_INTERVAL):
        stack_info = []
        for frame, start_time in _main_thread_stack:
            func_name = frame.f_code.co_name

            if func_name == 'profile_wrapper':
                # We have profiler enabled for this function. Do not show nested frames (if present). Instead of
                # the name `profile_wrapper`, let's show the original function name to make the traceback clearer
                original_func: FunctionType = frame.f_locals.get('func')
                if original_func is not None:
                    stack_info.append(StackFrameInfo(
                        func_name=original_func.__name__,
                        file_name=original_func.__code__.co_filename,
                        start_time=start_time,
                        is_under_profiling=True
                    ))
                    break

            stack_info.append(StackFrameInfo(
                func_name=func_name,
                file_name=frame.f_code.co_filename,
                line_number=frame.f_lineno,
                start_time=start_time
            ))
            
    return stack_info
```

kozlovsky: Exactly!

kozlovsky (approved):

PR #7541, opened 2023-07-11
Make the `libtorrent` tests more responsive

drew2a:
This PR doesn't completely solve the issue causing the flakiness of the libtorrent tests. However, the changes have made the tests more responsive and should, at the very least, prevent the main thread from freezing.

Besides minor refactoring, there are two significant changes in this PR:
1. An asyncio switch was added to the `wait_for_status` function due to the potential of blocking the main thread during pytest execution. To understand the possibility, consider this example: https://github.com/Tribler/tribler/blob/018cd5fce25a0d9759d0c0d22b41b158c60428cd/src/tribler/core/utilities/tests/test_async_force_switch.py#L10-L34

2. The shutdown logic has been rewritten to be slightly more accurate. While this change may not directly influence test stability, it could contribute slightly to it.

Related to #7495

Refs:
* https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel
* https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/download.py
+++ b/src/tribler/core/components/libtorrent/download_manager/download.py
@@ -34,7 +34,7 @@ from tribler.core.utilities.simpledefs import DOWNLOAD, DownloadStatus
 from tribler.core.utilities.unicode import ensure_unicode, hexlify
 from tribler.core.utilities.utilities import bdecode_compat
 
-Getter = Optional[Callable[[Any], Any]]
+Getter = Callable[[Any], Any]
 
 
 class Download(TaskManager):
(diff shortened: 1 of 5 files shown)


src/tribler/core/components/libtorrent/download_manager/download.py
```diff
@@ -123,8 +128,8 @@ def get_torrent_data(self) -> Optional[object]:
     def register_alert_handler(self, alert_type: str, handler: lt.torrent_handle):
         self.alert_handlers[alert_type].append(handler)
 
-    def wait_for_alert(self, success_type: str, success_getter: Optional[Callable[[Any], Any]] = None,
-                       fail_type: str = None, fail_getter: Optional[Callable[[Any], Any]] = None) -> Future:
+    def wait_for_alert(self, success_type: str, success_getter: Getter = None,
```
kozlovsky, line 131: **nitpick:** It may be easier to understand the function signature if, in the type declaration, `Getter` was declared as `Callable[[Any], Any]` (without `Optional`). Then, we could have `success_getter: Optional[Getter] = None` in the function signature.

kozlovsky (approved):

src/tribler/core/components/libtorrent/download_manager/download.py
drew2a: Changed

PR #7542, opened 2023-07-13
Add `ensure_gc` for each test call

drew2a:
This PR should stabilize the test suites. It has worked for my local runs, but doesn't resolve all problems—just many of them.

My hypothesis is that, to properly release all resources used by libtorrent, object destructors must be called. This happens during garbage collection. It appears that after some tests, the garbage collector doesn't have enough time to be invoked, leading to random, inexplicable bugs. However, I'm not completely certain about this explanation.

Related to #7495 #7498

Refs:
* https://docs.pytest.org/en/stable/explanation/fixtures.html#fixture-instantiation-order

### DIFF
--- a/src/tribler/core/conftest.py
+++ b/src/tribler/core/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+import gc
 import logging
 import platform
 import sys
@@ -62,6 +63,29 @@ def pytest_runtest_protocol(item: Function, log=True, nextitem=None):
         print(f' in {duration:.3f}s ({human_readable.time_delta(total)} in total)', end='')
     else:
         yield
+
+
+
+@pytest.fixture(autouse=True)
+def ensure_gc():
+    """ Ensure that the garbage collector runs after each test.
+    This is critical for test stability as we use Libtorrent and need to ensure all its destructors are called. """
+    # For this fixture, it is necessary for it to be called as late as possible within the current test's scope.
+    # Therefore it should be placed at the first place in the "function" scope.
+    # If there are two or more autouse fixtures within this scope, the order should be explicitly set through using
+    # this fixture as a dependency.
+    # See the discussion in https://github.com/Tribler/tribler/pull/7542 for more information.
+
+    yield
+    # Without "yield" the fixture triggers the garbage collection at the beginning of the (next) test.
+    # For that reason, the errors triggered during the garbage collection phase will take place not in the erroneous
+    # test but in the randomly scheduled next test. Usually, these errors are silently suppressed, as any exception in
+    # __del__ methods is silently suppressed, but they still can somehow affect the test.
+    #
+    # By adding the yield we move the garbage collection phase to the end of the current test, to not affect the next
+    # test.
+
+    gc.collect()
 
 
 @pytest.fixture
--- a/src/tribler/gui/tests/conftest.py
+++ b/src/tribler/gui/tests/conftest.py
@@ -1,3 +1,4 @@
+import gc
 import logging
 import time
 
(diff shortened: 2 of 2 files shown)


src/tribler/core/conftest.py
```diff
@@ -64,6 +65,13 @@ def pytest_runtest_protocol(item: Function, log=True, nextitem=None):
         yield
 
 
+@pytest.fixture(autouse=True)
+def ensure_gc():
+    """ Ensure that the garbage collector runs after each test.
+    This is critical for test stability as we use Libtorrent and need to ensure all its destructors are called. """
+    gc.collect()
```
kozlovsky, line 72: Currently, the fixture triggers the garbage collection at the beginning of the (next) test. For that reason, the errors triggered during the garbage collection phase will take place not in the erroneous test but in the randomly scheduled next test. Usually, these errors are silently suppressed, as any exception in `__del__` methods is silently suppressed, but they still can somehow affect the test, I think.

By adding the `yield` we can move the garbage collection phase to the end of the current test, to not affect the next test:

```suggestion
    yield
    gc.collect()
```

drew2a:
> Can we somehow ensure that this fixture is top-most and that all other fixtures process their shutdown code before it?

I think so. I'm going to do this.

kozlovsky:
As I see in the pytest documentation, all autouse fixtures are applied before non-autouse fixtures of the same scope. As the `ensure_gc` fixture is the only autouse function-scope fixture in Tribler, it should be applied outside of all other fixtures, so it should work fine.

https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order

drew2a:
@kozlovsky see also:
https://stackoverflow.com/a/28593102
> I was just having this problem with two function-scoped autouse fixtures. I wanted fixture b to run before fixture a, but every time, a ran first. I figured maybe it was alphabetical order, so I renamed a to c, and now b runs first. Pytest doesn't seem to have this documented. It was just a lucky guess. :-)

kozlovsky:
@drew2a this answer is from 2015, and as I see in [another answer](https://stackoverflow.com/a/64471427) from 2020, pytest advises not to rely on alphabetical order for fixtures. If two fixtures have the same scope and the same value of `autouse`, the order of applying them is indeterminate until one fixture explicitly uses another one as a dependency.

In our case, we have only one autouse function-scoped fixture, so currently the order is clear, right before all non-autouse function-scoped fixtures

drew2a:
@kozlovsky I'll add this rule as a comment.

src/tribler/core/conftest.py
drew2a: Applied.

src/tribler/core/conftest.py
```diff
@@ -64,6 +65,27 @@ def pytest_runtest_protocol(item: Function, log=True, nextitem=None):
         yield
 
 
+@pytest.fixture(autouse=True)
+def ensure_gc():
+    """ Ensure that the garbage collector runs after each test.
+    This is critical for test stability as we use Libtorrent and need to ensure all its destructors are called. """
+    # For this fixture, it is necessary for it to be called as late as possible within the current test's scope.
+    # Therefore it should be placed at the highest possible scope. If there are two or more autouse fixtures within
```
kozlovsky, line 73: > Therefore it should be placed at the highest possible scope

The wording looks a bit misleading to me: according to the Pytest documentation, the highest possible scope is "session", and we want to use the lowest possible scope, "function"

https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

But in "function" scope, we want to be on the first place. So, maybe we should write something like

> Therefore it should be placed at the first place in the "function" scope

drew2a: Done

kozlovsky (approved):

src/tribler/gui/tests/conftest.py
```diff
@@ -61,3 +62,25 @@ def pytest_runtest_protocol(item, log=True, nextitem=None):
     total = time.time() - pytest_start_time
     if enable_extended_logging:
         print(f' in {duration:.3f}s ({total:.1f}s in total)', end='')
+
+
+@pytest.fixture(autouse=True)
+def ensure_gc():
```
xoriole, line 68: Do we need this on the GUI test?

kozlovsky: Probably not, but it can help isolate GUI tests a tiny bit better

PR #7543, opened 2023-07-13
Fixed double-setting of Download fields

qstokkink:
These `Download` fields were being set twice.


qstokkink:
`pytest / run (windows-latest)` never seems to get close to the end of the tests before timing out. Please advise.

drew2a (approved):

drew2a:
@qstokkink, it's mysteriously working again for now. I've added the case to #7495 and will revisit it a bit later. Also, it might be looked into by someone with a Windows machine.

qstokkink:
@drew2a Thanks for the help. 👍

PR #7546, opened 2023-07-14
Ported Bandwidth/GigaChannel tests to modern TestBase

qstokkink:
I've been told nobody uses the advanced `TestBase` functionality because there is no example code to copy-paste from in the Tribler source. So, I modernized the tests for the `BandwidthCommunity` and the `GigaChannelCommunity`.

In passing, I've also changed up some test code that has test nodes be able to observe each other's private keys (by being sent each other's `my_peer`). Secondarily, it seems to me that `test_remote_select_channel_contents_happy_eyeballs` had a 50/50 chance of succeeding when not fixing random numbers, which I also changed into a 100% success guarantee.

The `TestGigaChannelUnits` could probably be cleaned up even further by extracting string literals, etc. However, that is not the goal of this PR. The goal of this PR is only to showcase advanced `TestBase` functionality.

### DIFF
--- a/src/tribler/core/components/bandwidth_accounting/tests/test_community.py
+++ b/src/tribler/core/components/bandwidth_accounting/tests/test_community.py
@@ -1,5 +1,3 @@
-from asyncio import sleep
-
 from ipv8.keyvault.crypto import default_eccrypto
 from ipv8.peer import Peer
 from ipv8.test.base import TestBase
@@ -13,6 +11,8 @@ from tribler.core.components.bandwidth_accounting.db.database import BandwidthDa
 from tribler.core.components.bandwidth_accounting.db.transaction import BandwidthTransactionData, EMPTY_SIGNATURE
 from tribler.core.components.bandwidth_accounting.settings import BandwidthAccountingSettings
 from tribler.core.utilities.utilities import MEMORY_DB
+
+ID1, ID2, ID3 = range(3)
 
 
 class TestBandwidthAccountingCommunity(TestBase):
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/bandwidth_accounting/tests/test_community.py
```diff
@@ -28,128 +30,121 @@ def create_node(self):
...
+
+    def add_cache(self, i, cache):
+        return self.overlay(i).request_cache.add(cache)
+
     async def test_single_transaction(self):
         """
         Test a simple transaction between two parties.
         """
-        await self.nodes[0].overlay.do_payout(self.nodes[1].overlay.my_peer, 1024)
+        await self.overlay(ID1).do_payout(self.overlay(ID2).my_peer, 1024)
```
qstokkink, line 43:
```suggestion
        await self.overlay(ID1).do_payout(self.peer(ID2), 1024)
```

I forgot to change the `my_peer` to the "public" `self.peer()` here.

src/tribler/core/components/bandwidth_accounting/tests/test_community.py
```diff
@@ -28,128 +30,121 @@ def create_node(self):
...
-        cache = self.nodes[0].overlay.request_cache.add(BandwidthTransactionSignCache(self.nodes[0].overlay, tx1))
-        self.nodes[0].overlay.send_transaction(tx2, self.nodes[1].my_peer.address, cache.number)
+        cache = self.add_cache(ID1, BandwidthTransactionSignCache(self.overlay(ID1), tx1))
+        self.overlay(ID1).send_transaction(tx2, self.address(ID2), cache.number)
         await self.deliver_messages()
 
         # This one should be ignored by node 1
-        cache = self.nodes[0].overlay.request_cache.add(BandwidthTransactionSignCache(self.nodes[0].overlay, tx1))
-        self.nodes[0].overlay.send_transaction(tx1, self.nodes[1].my_peer.address, cache.number)
+        cache = self.add_cache(ID1, BandwidthTransactionSignCache(self.overlay(0), tx1))
```
qstokkink, line 118:
```suggestion
        cache = self.add_cache(ID1, BandwidthTransactionSignCache(self.overlay(ID1), tx1))
```

I forgot to convert the node id `0` into its corresponding variable.

src/tribler/core/components/bandwidth_accounting/tests/test_community.py
```diff
@@ -28,128 +30,121 @@ def create_node(self):
...
         """
-        other_peer = self.nodes[1].my_peer
-        tx = self.nodes[0].overlay.construct_signed_transaction(other_peer, 300)
+        tx = self.overlay(ID1).construct_signed_transaction(self.peer(ID2), 300)
         tx.signature_a = b"invalid"
-        self.nodes[0].overlay.database.BandwidthTransaction.insert(tx)
-        cache = self.nodes[0].overlay.request_cache.add(BandwidthTransactionSignCache(self.nodes[0].overlay, tx))
-        self.nodes[0].overlay.send_transaction(tx, other_peer.address, cache.number)
+        self.database(ID1).BandwidthTransaction.insert(tx)
+        cache = self.add_cache(ID1, BandwidthTransactionSignCache(self.overlay(0), tx))
```
qstokkink, line 86:
```suggestion
        cache = self.add_cache(ID1, BandwidthTransactionSignCache(self.overlay(ID1), tx))
```

I forgot to convert the node id 0 into its corresponding variable.

src/tribler/core/components/gigachannel/community/tests/test_gigachannel_community.py
```diff
@@ -101,134 +120,111 @@ async def test_gigachannel_search(self):
...
-        self.nodes[1].overlay.introduction_response_callback(Peer(default_eccrypto.generate_key("low")), None, payload)
+        self.overlay(ID2).settings.queried_peers_limit = 2
+        with self.assertReceivedBy(ID3, [SelectResponsePayload], message_filter=[SelectResponsePayload]):
+            self.overlay(ID2).send_introduction_request(self.peer(ID3))
+            await self.deliver_messages()
+        self.assertEqual(len(self.overlay(ID2).queried_peers), 2)
+
+        self.add_node_to_experiment(self.create_node())
+        with self.assertReceivedBy(3, [SelectResponsePayload], message_filter=[SelectResponsePayload]):
+            self.overlay(ID2).send_introduction_request(self.peer(3))
```
qstokkink, line 166: I could've bound `ID4 = 3` here to make the tests more consistent.

src/tribler/core/components/bandwidth_accounting/tests/test_community.py
```diff
@@ -14,6 +12,10 @@
 from tribler.core.components.bandwidth_accounting.settings import BandwidthAccountingSettings
 from tribler.core.utilities.utilities import MEMORY_DB
 
+ID1 = 0
+ID2 = 1
+ID3 = 2
```
qstokkink, line 17:
```suggestion
ID1, ID2, ID3 = range(3)
```

This would be shorter but less explicit.

src/tribler/core/components/gigachannel/community/tests/test_gigachannel_community.py
```diff
@@ -12,31 +14,54 @@
...
+
+    # The following methods allow for use as a Mapping (i.e., the "**key"-syntax)
+    def __iter__(self):
+        return iter(asdict(self))
+
+    def __getitem__(self, item):
+        return getattr(self, item)
+
+    def __len__(self):
+        return len(fields(self))
```
kozlovsky, line 54: It looks like it is possible to use a [typed `NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) here instead of dataclass.
```suggestion
class ChannelKey(NamedTuple):
    channel_pk: bytes
    origin_id: int
```

Then the iteration of items will be available out of the box. With this change, `mapping.add(peer, *key.values())` becomes `mapping.add(peer, *key)`, and `mds.get_entries(**key)` becomes `mds.get_entries(**key._asdict())`. Despite underscore, `_asdict` is an [official public method](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict) for `namedtuple`.

src/tribler/core/components/bandwidth_accounting/tests/test_community.py
kozlovsky: I prefer the shorter form; it is clear enough. But both forms look good.

kozlovsky (approved):
Looks nice!

When testing, I caught a single failure of `test_gigachannel_search` but could not reproduce it again. The failure was caused by the fact that the search returned just `[U_CHANNEL]` instead of `[U_CHANNEL, U_TORRENT]` and that led to an assertion failure. I think this random test failure is not related to the current changes in this PR.

src/tribler/core/components/gigachannel/community/tests/test_gigachannel_community.py
qstokkink: Thanks for the suggestion: that was also my initial approach. The reason that I did not use `NamedTuple`, in the end, is that the context menu in de PyCharm IDE does not list `_asdict`:

contextmenu

If you absolutely insist on using `NamedTuple`, I'm also fine with using it though.

src/tribler/core/components/bandwidth_accounting/tests/test_community.py
qstokkink: Based on localized consensus (in the office), I'll adopt your short-form preference.

PR #7549, opened 2023-07-17
Actualize test dependencies

drew2a:
This PR updates versions of test dependencies and removes unused ones.

### DIFF
--- a/src/tribler/core/components/libtorrent/conftest.py
+++ b/src/tribler/core/components/libtorrent/conftest.py
@@ -58,8 +58,8 @@ async def test_download(mock_dlmgr, test_tdef):
 
 
 @pytest.fixture
-def mock_handle(mocker, test_download):
-    return mocker.patch.object(test_download, 'handle')
+def mock_handle(test_download):
+    test_download.handle = MagicMock()
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


src/tribler/core/components/libtorrent/tests/conftest.py
```diff
@@ -15,10 +17,10 @@ def tdef():
...
-def mock_download_config(mocker, test_download):
-    return mocker.patch.object(test_download, 'config')
+def mock_download_config(test_download):
+    test_download.config = MagicMock()
 
 
 @pytest.fixture
-def mock_download_state(mocker, test_download):
-    return mocker.patch.object(test_download, 'get_state')
+def mock_download_state(test_download):
```
kozlovsky, line 25: It looks like the `mock_download_state` fixture is not used; I think we can delete it.

src/tribler/core/components/libtorrent/tests/conftest.py
```diff
@@ -15,10 +17,10 @@ def tdef():
 
 
 @pytest.fixture
-def mock_download_config(mocker, test_download):
-    return mocker.patch.object(test_download, 'config')
+def mock_download_config(test_download):
+    test_download.config = MagicMock()
```
kozlovsky, line 21: It looks like this fixture is used in two tests only; can we inline this logic into the test itself and remove the fixture?

src/tribler/core/components/libtorrent/conftest.py
```diff
@@ -1,4 +1,4 @@
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
```
kozlovsky, line 1: > Unused patch imported from unittest.mock

Codacy is right :)

kozlovsky (changes requested):

PR #7551, opened 2023-07-18
Fix creating torrent with name containing escape characters

xoriole:
- Fixes https://github.com/Tribler/tribler/issues/7531

### DIFF
--- a/src/tribler/gui/utilities.py
+++ b/src/tribler/gui/utilities.py
@@ -4,6 +4,7 @@ import json
 import logging
 import math
 import os
+import re
 import sys
 import time
 import traceback
@@ -560,3 +561,9 @@ def get_objects_with_predicate(data_item: Dict, predicate: ResourceType) -> List
     Extract the objects that have a particular predicate from a particular data item.
     """
     return [stmt["object"] for stmt in data_item.get("statements", ()) if stmt["predicate"] == predicate]
+
+
+def sanitize_filename(filename: str):
+    """Removes some selected escape characters from the filename and returns the cleaned value."""
+    return re.sub(r'[\n\r\t]', '', filename)
+
--- a/src/tribler/gui/utilities.py
+++ b/src/tribler/gui/utilities.py
@@ -566,4 +566,3 @@ def get_objects_with_predicate(data_item: Dict, predicate: ResourceType) -> List
 def sanitize_filename(filename: str):
     """Removes some selected escape characters from the filename and returns the cleaned value."""
     return re.sub(r'[\n\r\t]', '', filename)
-
--- a/src/tribler/gui/utilities.py
+++ b/src/tribler/gui/utilities.py
@@ -563,6 +563,6 @@ def get_objects_with_predicate(data_item: Dict, predicate: ResourceType) -> List
     return [stmt["object"] for stmt in data_item.get("statements", ()) if stmt["predicate"] == predicate]
 
 
-def sanitize_filename(filename: str):
+def sanitize_filename(filename: str) -> str:
     """Removes some selected escape characters from the filename and returns the cleaned value."""
     return re.sub(r'[\n\r\t]', '', filename)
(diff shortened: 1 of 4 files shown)


drew2a:
@xoriole, do you have any idea how it's possible to add '\n' to the file name? Could it be due to some incomplete configuration of the QT widget? I believe these characters should be automatically filtered by QT.

xoriole:
@drew2a It is possible to have new line characters on the filename if the user has copied the text from the file content eg. copying multiple lines from a text file and pasting on the torrent name field causes the issue on Windows. It works fine on Ubuntu though. As such, I can reproduce the issue with the reported stack trace on Windows.

src/tribler/gui/utilities.py
```diff
@@ -547,3 +548,8 @@ def get_objects_with_predicate(data_item: Dict, predicate: ResourceType) -> List
     Extract the objects that have a particular predicate from a particular data item.
     """
     return [stmt["object"] for stmt in data_item.get("statements", ()) if stmt["predicate"] == predicate]
+
+
+def sanitize_filename(filename: str):
```
drew2a, line 553:
```suggestion
def sanitize_filename(filename: str) -> str:
```

drew2a (commented):
I'm torn about this PR.

While both the implementation and the intention are clear, I'm unsure if this change is beneficial to the codebase. Here's my line of thinking. The aim is to fix user-related mistakes, like an incorrect file name, by assuming that special characters should be replaced with ''. However, this covers only one of the many possible user errors. For instance, instead of the filename used in the tests, "This \nIs \r\nA \tTorrent Name.torrent", a user might replace spaces with special characters like "This\nIs\r\nA\tTorrent Name.torrent". This solution might not work for such cases, assuming the original filename was "This Is A Torrent Name.torrent".

In my view, it's better to let the user decide how to correct the path. We could display an error message stating "incorrect path", which wouldn't lead to a crash in Tribler or generate an error report. After seeing this error, the user should be able to identify the mistake and fix it.

drew2a:
@xoriole, thank you for the extended example. Indeed, there aren't many options for the user to correct the file name. This scenario highlights a potential issue with the UI widget (or its settings) as it should not allow the user to input multi-line text. Your example clarifies to me the root issue — the widget allowing multiline text input.
Regardless of the auto-correction algorithm, I would suggest fixing the core issue (multiline text input).

Regarding the algorithm, please note that in the above example, the expected filename would be `I'll Take Test(1982).torrent` as this is what the user sees. This illustrates the potential confusion that could arise with the proposed auto-correction algorithm.

I see two options here. One is to keep your auto-correction but apply it immediately after text is pasted. This way, the user will see the final version of the filename.

The other option is to modify the algorithm to something like
```python
s = 'string\nwith\nlines'
text_edit = s.split('\n', maxsplit=1)[0]
```

qstokkink:
You could also consider using a `QValidator` on your text field. You can hook this up to a widget to immediately check and fix the input text. [I have an example here.](https://github.com/qstokkink/ipv8-visual-community-creator/blob/main/nodes.py#L72-L88)

src/tribler/gui/utilities.py
```diff
@@ -560,3 +561,8 @@ def get_objects_with_predicate(data_item: Dict, predicate: ResourceType) -> List
     Extract the objects that have a particular predicate from a particular data item.
     """
     return [stmt["object"] for stmt in data_item.get("statements", ()) if stmt["predicate"] == predicate]
+
+
+def sanitize_filename(filename: str) -> str:
+    """Removes some selected escape characters from the filename and returns the cleaned value."""
+    return re.sub(r'[\n\r\t]', '', filename)
```
drew2a, line 568: As it is used only in one place (file createtorrentdialog.py) I would suggest moving this function to the `createtorrentdialog.py`.

drew2a (approved):

PR #7565, opened 2023-08-07
Avoid race conditions when starting Socks5 servers and REST API by allowing OS to choose a free port

### LINKED ISSUE #7564: [7.13.0-RC3] OSError in SocksServersComponent: [Errno 10048] error while attempting to bind on address
While initializing the Socks5 server in Tribler 7.13.0-RC3, an OSError occurs due to a potential race condition between port selection and its utilization:
```
  File "tribler\core\components\component.py", line 39, in start
  File "tribler\core\components\socks_servers\socks_servers_component.py", line 26, in run
  File "tribler\core\components\socks_servers\socks5\server.py", line 31, in start
  File "asyncio\base_events.py", line 1463, in create_server
OSError: [Errno 10048] error while attempting to bind on address ('127.0.0.1', 63843):
only one usage of each socket address (protocol/network address/port) is normally permitted
```

The reason for the error is a time gap between the moment Tribler selects a random free port and the moment it starts using the port ([in socks_server_component.py](https://github.com/Tribler/tribler/blob/6aceb9d728fd422fa6fe11d284eb0f8be01e9533/src/tribler/core/components/socks_servers/socks_servers_component.py#L24)):
```python
socks_server = Socks5Server(port=default_network_utils.get_random_free_port())
...
await socks_server.start()
```

When the gap is too wide, another process can grab the port before Socks5Server starts actually using it. When the gap is too small, I suspect it is possible that the operating system still believes sometimes that the port is occupied after it was temporarily taken inside the `get_random_free_port()` call.

The whole approach of using the `default_network_utils.get_random_free_port()` looks incorrect, as having a race condition with it is always possible.

The correct approach to grab a free port is to specify the value 0 for the port when starting the server. In this case, the operating system should automatically choose and assign the free port. Then we can retrieve the number of the given port and remember it for later usage.

As I understand it, the current logic with `default_network_utils.get_random_free_port()` was implemented this way because historically, Tribler consisted of parts with an order of initialization different from the order of dependencies, so it was necessary to have the port value before the server is actually started.

Now Tribler consists of components with clear dependencies, and their initialization order corresponds with dependency order. That allows us to retire the error-prone `default_network_utils.get_random_free_port()` usage and ask the OS for the port directly when starting the server.

We have similar logic in several places:
1. [When starting Socks5Server](https://github.com/Tribler/tribler/blob/6aceb9d728fd422fa6fe11d284eb0f8be01e9533/src/tribler/core/components/socks_servers/socks_servers_component.py#L24) (see #6149, #5437)
2. [When starting Libtorrent](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L261)
3. [When starting the REST API](https://github.com/Tribler/tribler/blob/7bd2b5cc0c3a0265c6675c54b844d96390c9401e/src/tribler/gui/tribler_window.py#L178) (see the issues #6767, #5547). In this latter case, the free port is first searched by Tribler GUI, and the found value then passed to the Tribler Core. After that, the Tribler Core [tries to grab the port in a range by performing a loop](https://github.com/Tribler/tribler/blob/ca4638f2e5ab74ca95f5dd271822fce5f5dab865/src/tribler/core/components/restapi/rest/rest_manager.py#L151). After that, the port value is reported back to Tribler GUI. All this complicated machinery is vulnerable to race conditions and is unnecessary now: it is possible for Tribler Core to request a free port from OS, start the REST API server and pass the port value to Tribler GUI via the processes database.

To do a correct fix we should stop rely on `default_network_utils.get_random_free_port()` and instead pass the port value zero when opening any listening port to allow OS to choose a free port without any race condition.

kozlovsky:
This PR fixes #7564. It stops using `default_network_utils.get_random_free_port()` and passes port value zero when starting a server to allow OS to choose a free port without race conditions.

The PR consists of several commits:
1. The first commit stops using `get_random_free_port()` when starting a new Socks5Server
3. The second commit renames `events_manager.connect(...)` to `events_manager.connect_to_core(...)`. The reason for this is the `connect` name is overused: the same name is used for subscribing to PyQt signals, for opening `UdpTrackerSession`, and for connecting to the sqlite3 database in ProcessManager, so it is better to have a distinct name to find all places when the specific method is used easily.
4. The third commit changes the logic of how REST API Manager gets the port. After the PR changes, Tribler GUI no longer tries to call  `get_random_free_port()` and detect the free port for Tribler Core's REST API usage. Instead, Tribler Core requests a free port directly from OS when starting the server and then passes the port value to Tribler GUI. Now it is unnecessary to have a loop over a range of ports on the Coree side and attempt to start a server with a possible retry, as the suitable port is available from the first attempt.
5. The fifth commit removes the REST API port from Tribler GUI settings. For most users, settings the specific REST API port value is unnecessary, as it is an internal details of Tribler's GUI/Core communication. If it is necessary for sone user to set a specific REST API port, he can do it by setting CORE_API_PORT environment variable (as specified in the documentation). In my opinion, UI settings should only display settings that can actually be important to many Tribler users, and it seems that REST API port value does not belong to this category.

As a result:
1. The logic of starting the servers was simplified and streamlined
2. The possible reasons for race conditions are eliminated
3. The retry logic when starting the REST API was removed as unnecessary
4. Tribler settings in GUI were simplified, and the unnecessary option was removed

Previously, the documentation said 20100 is a default port value for Swagger UI and REST API. It is not entirely true, as the port number was taken from a range. This part of the documentation is clarified: the user should specify the port value in the `CORE_API_PORT` environment variable to start the REST API and Swagger UI on a specific port.

### DIFF
--- a/doc/restapi/introduction.rst
+++ b/doc/restapi/introduction.rst
@@ -16,11 +16,11 @@ Some requests require one or more parameters. These parameters are passed using 
 
 .. code-block:: none
 
-    curl -X PUT -H "X-Api-Key: <YOUR API KEY>" http://localhost:20100/mychannel/rssfeeds/http%3A%2F%2Frssfeed.com%2Frss.xml
+    curl -X PUT -H "X-Api-Key: <YOUR API KEY>" http://localhost:<port>/mychannel/rssfeeds/http%3A%2F%2Frssfeed.com%2Frss.xml
 
-Alternatively, requests can be made using Swagger UI by starting Tribler and opening `http://localhost:20100/docs` in a browser.
+Alternatively, requests can be made using Swagger UI by starting Tribler and opening `http://localhost:<port>/docs` in a browser.
 
-Note: 20100 is a default port value. It can be changed by setting up "CORE_API_PORT" environment variable.
+The port can be specified by setting up "CORE_API_PORT" environment variable.
 
 Error handling
 ==============
--- a/src/tribler/core/start_core.py
+++ b/src/tribler/core/start_core.py
@@ -5,7 +5,7 @@ import os
 import signal
 import sys
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 from tribler.core import notifications
 from tribler.core.check_os import (
@@ -122,7 +122,8 @@ async def core_session(config: TriblerConfig, components: List[Component]) -> in
     return session.exit_code
 
 
-def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_test_mode: bool = False) -> int:
+def run_tribler_core_session(api_port: Optional[int], api_key: str,
+                             state_dir: Path, gui_test_mode: bool = False) -> int:
     """
     This method will start a new Tribler session.
     Note that there is no direct communication between the GUI process and the core: all communication is performed
@@ -130,7 +131,7 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
 
     Returns an exit code value, which is non-zero if the Tribler session finished with an error.
     """
-    logger.info(f'Start tribler core. API port: "{api_port}". '
+    logger.info(f'Start tribler core. API port: "{api_port or "<not specified>"}". '
                 f'API key: "{api_key}". State dir: "{state_dir}". '
                 f'Core test mode: "{gui_test_mode}"')
 
@@ -143,7 +144,10 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     if SentryReporter.is_in_test_mode():
         default_core_exception_handler.sentry_reporter.global_strategy = SentryStrategy.SEND_ALLOWED
 
-    config.api.http_port = api_port
+    # The -1 value is assigned for backward compatibility reasons when the port is not specified.
+    # When RESTManager actually uses the value, it converts -1 to zero.
+    # It is possible that later we can directly store zero to config.api.http_port, but I prefer to be safe now.
+    config.api.http_port = api_port or -1
     # If the API key is set to an empty string, it will remain disabled
     if config.api.key not in ('', api_key):
         config.api.key = api_key
@@ -179,7 +183,7 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     return exit_code
 
 
-def run_core(api_port, api_key, root_state_dir, parsed_args):
+def run_core(api_port: Optional[int], api_key: Optional[str], root_state_dir, parsed_args):
     logger.info(f"Running Core in {'gui_test_mode' if parsed_args.gui_test_mode else 'normal mode'}")
 
     gui_pid = GuiProcessWatcher.get_gui_pid()
@@ -193,11 +197,6 @@ def run_core(api_port, api_key, root_state_dir, parsed_args):
     if not current_process_is_primary:
         msg = 'Another Core process is already running'
         logger.warning(msg)
-        process_manager.sys_exit(1, msg)
-
-    if api_port is None:
-        msg = 'api_port is not specified for a core process'
-        logger.error(msg)
         process_manager.sys_exit(1, msg)
 
     version_history = VersionHistory(root_state_dir)
--- a/src/tribler/gui/network/request_manager.py
+++ b/src/tribler/gui/network/request_manager.py
@@ -9,7 +9,7 @@ from typing import Callable, Dict, Optional, Set
 from PyQt5.QtCore import QBuffer, QIODevice, QUrl
 from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkRequest
 
-from tribler.gui.defs import BUTTON_TYPE_NORMAL, DEFAULT_API_HOST, DEFAULT_API_PORT, DEFAULT_API_PROTOCOL
+from tribler.gui.defs import BUTTON_TYPE_NORMAL, DEFAULT_API_HOST, DEFAULT_API_PROTOCOL
 from tribler.gui.dialogs.confirmationdialog import ConfirmationDialog
 from tribler.gui.network.request import DATA_TYPE, Request
 from tribler.gui.utilities import connect
@@ -35,11 +35,17 @@ class RequestManager(QNetworkAccessManager):
 
         self.protocol = DEFAULT_API_PROTOCOL
         self.host = DEFAULT_API_HOST
-        self.port = DEFAULT_API_PORT
+        self.port: Optional[int] = None
         self.key = ''
         self.limit = limit
         self.timeout_interval = timeout_interval
         self.last_request_id = 0
+
+    def set_api_key(self, key: str):
+        self.key = key
+
+    def set_api_port(self, api_port: int):
+        self.port = api_port
 
     def get(self,
             endpoint: str,
@@ -167,6 +173,8 @@ class RequestManager(QNetworkAccessManager):
         return text
 
     def get_base_url(self) -> str:
+        if not self.port:
+            raise RuntimeError("API port is not set")
         return f'{self.protocol}://{self.host}:{self.port}/'
 
     @staticmethod
--- a/src/tribler/gui/network/tests/test_request_manager.py
+++ b/src/tribler/gui/network/tests/test_request_manager.py
@@ -2,6 +2,7 @@ from unittest.mock import Mock, patch
 
 import pytest
 
+from tribler.core.utilities.network_utils import default_network_utils
 from tribler.gui.network.request_manager import RequestManager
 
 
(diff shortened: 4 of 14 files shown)


src/tribler/core/components/libtorrent/download_manager/download_manager.py
```diff
@@ -258,17 +257,16 @@ def create_session(self, hops=0, store_listen_port=True):
...
             settings['user_agent'] = 'Tribler/' + version_id
             enable_utp = self.config.utp
             settings['enable_outgoing_utp'] = enable_utp
             settings['enable_incoming_utp'] = enable_utp
 
             settings['prefer_rc4'] = True
-            settings["listen_interfaces"] = "0.0.0.0:%d" % libtorrent_port
+
+            # We do not specify the port in settings dict as we allow libtorrent to automatically choose the port
+            # settings["listen_interfaces"] = "0.0.0.0:%d" % libtorrent_port
```
qstokkink, line 269: Allowing users to specify a port for torrent traffic is [critical functionality for some users](https://www.downloadprivacy.com/how-to-torrent/port-forwarding). The other ports probably don't matter as much but I'd leave the libtorrent port config option as-is.

kozlovsky: Thank you for this link!

Thinking about this, the current Tribler logic regarding the libtorrent port number handling does not look correct to me. This is how, in my understanding, the current logic of `DownloadManager.create_session()` works:

0. Initially, [in LibtorrentSettings](https://github.com/Tribler/tribler/blob/24d189c09a88bd91379034f8f66659ec9a6a0676/src/tribler/core/components/libtorrent/settings.py#L25), the port value is `None` by default (and a user can set a specific value):
```python
class LibtorrentSettings(TriblerConfigSection):
    ...
    port: Optional[int] = None
    ...
```
1. The value is read from config, and [an empty value is replaced with a random free port](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L261):
```python
libtorrent_port = self.config.port or default_network_utils.get_random_free_port()
```
2. The resulting port value [is passed to libtorrent](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L289):
```python
settings["listen_interfaces"] = "0.0.0.0:%d" % libtorrent_port
...
self.set_session_settings(ltsession, settings)
```
3. The libtorrent [is requested to listen on a port in a range](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L303), starting from a specified port value:
```python
ltsession.listen_on(libtorrent_port, libtorrent_port + 10)
```
4. Then, if the actual port used by libtorrent is different from the port that was initially specified, [the actual port value is stored in the config](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L305):
```python
if libtorrent_port != ltsession.listen_port() and store_listen_port:
    self.config.port = ltsession.listen_port()
```
(the `store_listen_port` parameter value is True by default and [is not specified](https://github.com/Tribler/tribler/blob/090ae756e76fd3649da553cd1ab5bab9671d3f2f/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L340) when the `DownloadManager.create_session` method is called).

From this, my conclusion is the following: even when a user specifies the libtorrent port value in settings, it is possible that the value will be changed to a different value (at step 3) if the original port is already occupied. Then the value provided by a user will be overridden in settings by a new value without any confirmation by a user.

From what I currently understand, the correct logic should work in the following way:
1. If the port value is explicitly specified by a user, it should be used. If the specified port is already occupied, Tribler should not use a different port silently. Instead, it should notify the user that the specified port is unavailable and stop.
2. If the value is not specified, Tribler should allow libtorrent to choose the port automatically. In that case, Tribler should not store the randomly assigned port value in `LibtorrentSettings`, as it is not guaranteed that the same port will be available next time, and it is not important to use the same port next time. Only the port value explicitly specified by a user should go to LibtorrentSettings.

@qstokkink what do you think?

qstokkink: Sounds good 👍 That's what I would expect from my torrent client.

kozlovsky: After the offline discussion with @xoriole we decided to extract libtorrent port handling changes to a separate PR

doc/restapi/introduction.rst
```diff
@@ -18,9 +18,9 @@ Some requests require one or more parameters. These parameters are passed using
 
     curl -X PUT -H "X-Api-Key: <YOUR API KEY>" http://localhost:20100/mychannel/rssfeeds/http%3A%2F%2Frssfeed.com%2Frss.xml
```
xoriole, line 19: There is a reference to `20100` here. To maintain consistency, it would be good to update this as well.

src/tribler/core/start_core.py
```diff
@@ -179,7 +183,7 @@ def run_tribler_core_session(api_port: int, api_key: str, state_dir: Path, gui_t
     return exit_code
 
 
-def run_core(api_port, api_key, root_state_dir, parsed_args):
+def run_core(api_port: Optional[int], api_key: Optional[str], root_state_dir, parsed_args):
```
xoriole, line 186: Here, `api_key` does not feel optional. If it is None, the process is killed.

src/tribler/gui/network/request_manager.py
```diff
@@ -35,12 +35,18 @@ def __init__(self, limit: int = 50, timeout_interval: int = 15):
...
         self.protocol = DEFAULT_API_PROTOCOL
         self.host = DEFAULT_API_HOST
-        self.port = DEFAULT_API_PORT
+        self.port: Optional[int] = None
         self.key = ''
         self.limit = limit
         self.timeout_interval = timeout_interval
         self.last_request_id = 0
 
+    def set_api_key(self, key):
```
xoriole, line 44: NIT: type

src/tribler/gui/network/tests/test_request_manager.py
```diff
@@ -8,13 +8,18 @@
 # pylint: disable=protected-access, redefined-outer-name
 
 
+DEFAULT_PORT = 2100
```
xoriole, line 11: I think you meant `20100`.

doc/restapi/introduction.rst
kozlovsky: Thanks, I changed it to `<port>`

src/tribler/gui/network/request_manager.py
kozlovsky: Added

src/tribler/gui/network/tests/test_request_manager.py
kozlovsky: I added a `free_port` fixture to GUI tests instead of a constant value, as it looks safer. A similar fixture was already defined in core tests, but not in GUI tests

src/tribler/core/start_core.py
kozlovsky: It is an interesting topic.
* On one side, I agree that the `None` value always leads to a process termination with an exit code
* But I want to save the error code and message to the process state in the process database
* Currently, the `run_core` function we discuss here creates an object that represents the process instance. For the Core process, the sequence is the following:
    1. First, the `run_tribler.py` script started. It is a simple script that reads process arguments and environment variables and then runs either the `run_gui` function or the `run_core` function.
    2. `run_core` function creates the `current_process` object and checks that the core process can be started (that is, it checks that there is no other core process already running and that core-related arguments are correct).
    3. If the core process should and can be started, `run_tribler_core_session` is called that run a session with all core components.
    
    That is, `run_core` is an intermediate layer between the `run_tribler` script and the `run_tribler_core_session` function that handles conditions on whether the core process can be started. It accepts `None` as a possible `api_key` vale and correctly stops the current process with an appropriate error message.
    
We can make the `api_key` argument non-optional, but then we should handle the error outside the `run_core` function, that is, directly in the `run_tribler` script. In this case, we need to instantiate the `current_process` object right inside the `run_tribler` script, and in this case, the `run_tribler` will not be a simple script anymore that just read arguments and environment variables and calls either `run_gui` or `run_core`. The logic of the `run_tribler` script will be much more complicated in this case.

For that reason, I think it is a reasonable approach to allow the intermediate `run_core` function to handle the case of missing the `api_key` value, and in that case, the argument should be optional.

xoriole: In the current main, core still works if the api_key is None so I would like to keep that behavior intact. Since this PR is targeted towards allowing a free port, I would consider changes on api_key behavior out of scope and would warrant its own PR and separate discussion on whether or not to change the default behavior.

xoriole (changes requested):

src/tribler/core/start_core.py
kozlovsky: I agree. I removed the check for `api_key`, and now it is possible to run core without specifying a CORE_API_KEY value again.

xoriole (approved):

PR #7568, opened 2023-08-11
Increase MacOS build timeout to 15 minutes

xoriole:
Mac builds sometimes fail because of timeout. Looking at the previous builds, it is taking between 8 to 10 minutes. So, I assume it is safe to increase it to 15 minutes.

### DIFF
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -68,7 +68,7 @@ jobs:
           python ./build/update_version.py -r .
 
       - name: Run build script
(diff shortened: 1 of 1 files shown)


xoriole:
What a coincidence! The `PR / mac / build` timed out for this PR. After merging this PR, a failure like this one should likely not happen.
https://github.com/Tribler/tribler/actions/runs/5833766691/job/15821897982?pr=7568
screenshot-github com-2023 08 11-16_45_13

kozlovsky (approved):

PR #7571, opened 2023-08-14
WIP: Added a ruff configuration and autofixed the codebase

qstokkink:
This PR was created for dev discussion.

These fixes were generated using the supplied `.ruff.toml` using `ruff check . --fix`.

Some things to change:

 - `target-version = "py37"` should be pointed to Python 3.8 for Tribler.
 - Some things need manual fixing, apparently.


src/tribler/core/utilities/process_manager/__init__.py
```diff
@@ -1,3 +0,0 @@
-from tribler.core.utilities.process_manager.manager import ProcessManager, get_global_process_manager, \
-    set_global_process_manager
-from tribler.core.utilities.process_manager.process import ProcessKind, TriblerProcess
```
qstokkink, line 3: Observation: these unused imports were externally used in `core/components/reporter/exception_handler`, causing the tests to fail after removal.

drew2a:
Interesting experiment! Essentially, Ruff seems like a new linter, just significantly faster. Currently, all our checks are performed in the cloud, on Codacy's side, and are set up with the basic code formatting rules from PyCharm. As of now, there's no option to enable Ruff in Codacy (though such a feature might be introduced in the future). However, there's a fundamental possibility to add it, as it can be run within a GitHub Action. Given that it doesn't offer advantages over other linters (apart from speed) and its integration would require changes to the workflow for both developers and PR checks, I'd recommend refraining from merging this PR.

image

qstokkink:
@drew2a Absolutely correct: the only thing you're winning is speed. There is no other advantage. However, having linting done within a single second + GitHub Actions overhead, instead of the 20 minutes it takes Codacy to report results, may make this more appealing to streamline the dev process.

PR #7575, opened 2023-08-15
Remove dependency on pytest-freezegun plugin

xoriole:
In this PR, I removed pytest-freezegun plugin used for freezing the date/time on the tests. This plugin was causing performance hits as reported on sentry. This plugin is now removed and replaced with an utility providing similar functionality. The test performance has improved as a result.


xoriole:
Sentry does not seem to block anymore with mocking datetime separately from freezegun, however, the tests still sometime timeout in 10 mins. If the Python environment needs to be freshly created, then it is likely that the tests will timeout.

qstokkink (approved):

PR #7577, opened 2023-08-19
Support true/false for boolean url query arguments.

### LINKED ISSUE #7562: improved openapi/swagger docs & support for true/false input to boolean arguments
**Is your feature request related to a problem? Please describe.**
Tribler is missing a proper web-ui but it does have the openapi /docs endpoint which can be used instead.

**Describe the solution you'd like**
The /docs endpoint that allow setting one of the boolean flags to true or false, currently the code only allows 0 & 1.

**Describe alternatives you've considered**
Calling the endpoints from postman.

**Additional context**
I'm happy to help improve the docs and openapi.

jutoft:
Allows using swagger ui & enable/disable boolean flags. Fixes #7562

remote_query have been removed from openapi since it does not appear to be used.
Split definition for search and remote query as they do not take the same args.

### DIFF
--- a/src/tribler/core/components/metadata_store/restapi/metadata_endpoint_base.py
+++ b/src/tribler/core/components/metadata_store/restapi/metadata_endpoint_base.py
@@ -10,6 +10,8 @@ from tribler.core.components.metadata_store.db.serialization import CHANNEL_TORR
 from tribler.core.components.metadata_store.db.store import MetadataStore
 from tribler.core.components.restapi.rest.rest_endpoint import RESTEndpoint
 
+from tribler.core.utilities.utilities import parse_bool
+
 # This dict is used to translate JSON fields into the columns used in Pony for _sorting_.
 # id_ is not in the list because there is not index on it, so we never really want to sort on it.
 
@@ -55,16 +57,16 @@ class MetadataEndpointBase(RESTEndpoint):
             "first": int(parameters.get('first', 1)),
             "last": int(parameters.get('last', 50)),
             "sort_by": json2pony_columns.get(parameters.get('sort_by')),
-            "sort_desc": bool(int(parameters.get('sort_desc', 1)) > 0),
+            "sort_desc": parse_bool(parameters.get('sort_desc', True)),
             "txt_filter": parameters.get('txt_filter'),
-            "hide_xxx": bool(int(parameters.get('hide_xxx', 0)) > 0),
+            "hide_xxx": parse_bool(parameters.get('hide_xxx', False)),
             "category": parameters.get('category'),
-            "exclude_deleted": bool(int(parameters.get('exclude_deleted', 0)) > 0),
+            "exclude_deleted": parse_bool(parameters.get('exclude_deleted', False)),
         }
         if 'tags' in parameters:
             sanitized['tags'] = parameters.getall('tags')
         if "remote" in parameters:
-            sanitized["remote"] = (bool(int(parameters.get('remote', 0)) > 0),)
+            sanitized["remote"] = (parse_bool(parameters.get('remote', False)),)
         if 'metadata_type' in parameters:
             mtypes = []
             for arg in parameters.getall('metadata_type'):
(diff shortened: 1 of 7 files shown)


src/tribler/core/utilities/tests/test_utilities.py
```diff
@@ -187,6 +187,24 @@ def test_parse_query():
     assert actual == expected
 
 
+def test_parse_bool():
+    assert not parse_bool('')
+    assert not parse_bool('false')
+    assert not parse_bool('False')
+    assert not parse_bool('0')
+    assert not parse_bool('0.0')
+    assert not parse_bool('0.0e2')
```
kozlovsky, line 196: Is it necessary to support float numbers here? What is the use case for that?

src/tribler/core/components/metadata_store/restapi/metadata_endpoint_base.py
```diff
@@ -55,16 +57,16 @@ def sanitize_parameters(cls, parameters):
             "first": int(parameters.get('first', 1)),
             "last": int(parameters.get('last', 50)),
             "sort_by": json2pony_columns.get(parameters.get('sort_by')),
-            "sort_desc": bool(int(parameters.get('sort_desc', 1)) > 0),
+            "sort_desc": parse_bool(parameters.get('sort_desc', 1)),
```
kozlovsky, line 60: With this new `parse_bool` function, let's use `True` and `False` as default values instead of `1` and `0`, as it looks clearer this way.

src/tribler/core/components/metadata_store/restapi/metadata_schema.py
```diff
@@ -11,13 +11,18 @@ class MetadataParameters(Schema):
     hide_xxx = Boolean(default=False, description='Toggles xxx filter')
     category = String()
     exclude_deleted = Boolean(default=False)
-    remote_query = Boolean(default=False)
     metadata_type = List(String(description='Limits query to certain metadata types (e.g. "torrent" or "channel")'))
 
 
+class SearchMetadataParameters(MetadataParameters):
+    include_total = Boolean(default=False, description='Include total amount found in query response, expensive if there is many rows')
```
kozlovsky, line 18: Please long lines to satisfy the linter

src/tribler/core/utilities/utilities.py
```diff
@@ -272,6 +272,27 @@ def to_fts_query(text):
...
+    Parse input to boolean True or False
+    Allow parsing text 'false', 'true' '1', '0' to boolean
+
+    :param obj: Object to parse
+    """
+    if isinstance(obj, str):
+        if obj.lower() == "false":
+            return False
+        
+        if obj.isnumeric():
```
kozlovsky, line 286: According to [Python documentation for `str.isnumeric()`](https://docs.python.org/3/library/stdtypes.html#str.isnumeric),
> "Numeric characters include digit characters, and all characters that have the Unicode numeric value property, e.g., U+2155, VULGAR FRACTION ONE FIFTH"

That means that even if `obj.isnumeric()` evaluates to `True`, it is possible to get `ValueError` when calling `int(obj)`.

src/tribler/core/utilities/utilities.py
```diff
@@ -272,6 +272,27 @@ def to_fts_query(text):
...
+            return bool(int(obj))
+        
+        try :
+            return bool(float(obj))
+        except:
```
kozlovsky, line 291: Can you add `# pylint: disable=...` comments to suppress linter warnings?

kozlovsky (changes requested):
Thank you for your contribution! Looks nice; there are some minor issues that I commented on.

src/tribler/core/utilities/tests/test_utilities.py
jutoft: it was mostly to be conformant with what bool() would do for a float value.
But i will remove it.

kozlovsky (approved):

Issue #7581, opened 2023-08-22
pytest-timeout is not working as expected when pytest-rerunfailures is enabled


xoriole:
`pytest-timeout` is not working well with `pytest-rerunfailures`.
See: https://github.com/pytest-dev/pytest-timeout/issues/18

Because of this, the default timeout of 60 seconds is not applied on the tests and sometimes some tests get stuck for several minutes resulting in GitHub action's execution timeout of 10 minutes.

One of the such tests is `test_check_and_regen_personal_channel_torrent_sleep` which was stuck for approx 8 minutes.
Logs here: https://github.com/Tribler/tribler/actions/runs/5937813183/job/16103116115?pr=7575

Link might not be available for long so here is the relevant logs snippet:
```python
...
Tue, 22 Aug 2023 12:08:22 GMT
12:08:22.908 Starting "test_check_channels_updates"...
Tue, 22 Aug 2023 12:08:23 GMT
src/tribler/core/components/gigachannel_manager/tests/test_gigachannel_manager.py::test_check_channels_updates PASSED [ 26%] in 0.204s (a minute in total)
Tue, 22 Aug 2023 12:08:23 GMT
12:08:23.112 Starting "test_check_and_regen_personal_channel_torrent_sleep"...
Tue, 22 Aug 2023 12:16:12 GMT
src/tribler/core/components/gigachannel_manager/tests/test_gigachannel_manager.py::test_check_and_regen_personal_channel_torrent_sleep 
Tue, 22 Aug 2023 12:16:12 GMT
Error: The operation was canceled.
```
---
So, I would be in favor of disabling `pytest-rerunfailures` so we can find out the tests which are getting stuck and fix them properly.

drew2a:
The example provided in the [issue](https://github.com/pytest-dev/pytest-timeout/issues/18) works well with `reruns == 0`

```python
@pytest.mark.timeout(2)
@pytest.mark.flaky(0)
def test_new():
    time.sleep(10)
    assert True
```

Perhaps if we run our tests with this argument, `pytest-sentry` will work as expected (this statement needs verification by doing an experiment).

drew2a:
Seems that we still have tests that could not be stopped by `pytest-timeout` although the pytest-rerunfailures is disabled.

PR #7582, opened 2023-08-22
Remove reruns and reruns-delay from pytest arguments

xoriole:
Addresses https://github.com/Tribler/tribler/issues/7581


kozlovsky (approved):

drew2a:
This PR does not look correct.

To complete the desired change you should remove the following line:
https://github.com/Tribler/tribler/blob/20fb22453c8c5349c6de4b8e679a5b590792b7df/requirements-test.txt#L8

Otherwise we leave unused artifacts in our code.

drew2a:
Workflows that use `secrets.PYTEST_SENTRY_DSN` have to be updated as well :)

drew2a:
Additionally (apologies for these incremental updates), the logic in https://github.com/Tribler/tribler/pull/7482 might need to be removed. It was initially developed to detect frozen tests, which later turned out to be an issue with `pytest-rerunfailures`.

PR #7583, opened 2023-08-23
Optimize the large torrent creation test

xoriole:
`test_create_torrent_endpoint.py::test_create_torrent_of_large_size` is consistently taking 20-30 seconds to finish the tests which is one of the contributor to timing out the full test suite.

This PR optimizes the test to reduce the test time to 100-200 ms (hopefully).


xoriole:
The test now executes in 5ms as compared to the previous run which took 27 seconds.
screenshot-sentry tribler org-2023 08 23-11_11_58

Previous
screenshot-sentry tribler org-2023 08 23-11_12_47

Issue #7422, opened 2023-05-15
 FeedbackDialog: replace its content with the Sentry event.


drew2a:
`FeedbackDialog` is outdated and should be updated. Ideally, all content should be replaced with the actual `Sentry` event.

A Sentry event is a dictionary, so one option is to present it to the user as a formatted string. Another option is to display it as a table.

<img width="361" alt="image" src="https://github.com/Tribler/tribler/assets/13798583/7ed3cd35-aece-46d7-a7e6-37ffc33abaf9">

drew2a:
The simplest option would be to display the Sentry event as a JSON dump:

<img width="600" alt="image" src="https://github.com/Tribler/tribler/assets/13798583/0a9cdc9c-ceb9-4944-9a8f-f1a7373246f0">

PR #7593, opened 2023-09-06
Refactor the FeedbackDialog

drew2a:
This PR implements #7422

FeedbackDialog:
image

The Sentry event (should be the same as before):
https://sentry.tribler.org/organizations/tribler/issues/2321


egbertbouman:
I like the idea that we show the exact event that we're about to send, altough I would have preferred to see the traceback itself properly formatted. But since that probably doesn't mean anything to the user anyway, this seems fine too.

One problem that I noticed is that my username doesn't get scrubbed anymore.

drew2a:
> I like the idea that we show the exact event that we're about to send, altough I would have preferred to see the traceback itself properly formatted. But since that probably doesn't mean anything to the user anyway, this seems fine too.

I'll display the formatted traceback at the top of the EditBox for easier readability.

> One problem that I noticed is that my username doesn't get scrubbed anymore.

I'll reach out to you to gather more information.

drew2a:
image

egbertbouman (approved):

PR #7594, opened 2023-09-07
Catch `AccessDenied` exception

### LINKED ISSUE #7408: [7.12.1] AccessDenied in `set_process_priority`
[Here](https://github.com/Tribler/tribler/blob/cda7cc5fc7261427e89920216275fbec0cfc4efe/src/tribler/core/check_os.py#L84):
```python
def set_process_priority(pid=None, priority_order=1):
    ...
    process = psutil.Process(pid if pid else os.getpid())

    if sys.platform == 'win32':
        ...
    elif sys.platform == 'darwin' or sys.platform == 'linux2':
        # On Unix, priority can be -20 to 20, but usually not allowed to set below 0, we set our classes somewhat in
        # that range.
        priority_classes = [5, 4, 3, 2, 1, 0]
        process.nice(priority_classes[priority_order])  # <-- here
```
```python
run_tribler.py in <module> at line 94
tribler/core/start_core.py in run_core at line 185
tribler/core/start_core.py in run_tribler_core_session at line 148
tribler/core/check_os.py in set_process_priority at line 84
__init__.py in nice at line 724
```

It looks like we should wrap the `process.nice` call with try/except block, catch the `psutil.AccessDenied` exception, and log it (probably put try/except at the top level of the function to catch possible Windows errors as well).

drew2a:
This PR fixes #7408

### DIFF
--- a/src/tribler/core/check_os.py
+++ b/src/tribler/core/check_os.py
@@ -66,22 +66,28 @@ def set_process_priority(pid=None, priority_order=1):
     """
     if priority_order < 0 or priority_order > 5:
         return
-
-    process = psutil.Process(pid if pid else os.getpid())
+    if sys.platform not in {'win32', 'darwin', 'linux'}:
+        return
 
     if sys.platform == 'win32':
-        priority_classes = [psutil.IDLE_PRIORITY_CLASS,
-                            psutil.BELOW_NORMAL_PRIORITY_CLASS,
-                            psutil.NORMAL_PRIORITY_CLASS,
-                            psutil.ABOVE_NORMAL_PRIORITY_CLASS,
-                            psutil.HIGH_PRIORITY_CLASS,
-                            psutil.REALTIME_PRIORITY_CLASS]
-        process.nice(priority_classes[priority_order])
-    elif sys.platform == 'darwin' or sys.platform == 'linux2':
+        priority_classes = [
+            psutil.IDLE_PRIORITY_CLASS,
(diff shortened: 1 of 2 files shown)


src/tribler/core/check_os.py
```diff
@@ -66,22 +66,28 @@ def set_process_priority(pid=None, priority_order=1):
     """
     if priority_order < 0 or priority_order > 5:
         return
-
-    process = psutil.Process(pid if pid else os.getpid())
+    if sys.platform not in {'win32', 'darwin', 'linux2', 'linux'}:
```
kozlovsky, line 69: **Nitpick:** It looks like the "linux2" value is only possible to get from `sys.platform` on Linux machines when using Python <= 3.3: https://stackoverflow.com/questions/446209/possible-values-from-sys-platform

As we use Python 3.8, we can only get "linux" value from `sys.platform` on Linux machines. So we can remove `linux2` from the list.

That also means that the previous code, if I understand correctly, never calls `process.nice` on Linux after upgrading Tribler from Python 3.3 to a more recent Python version.

kozlovsky (approved):

src/tribler/core/check_os.py
drew2a: Removed.

> That also means that the previous code, if I understand correctly, never calls process.nice on Linux after upgrading Tribler from Python 3.3 to a more recent Python version.

Indeed, I share your understanding.

kozlovsky (approved):

PR #7606, opened 2023-09-26
Migrate `knowledge.db` to `tribler.db`

drew2a:
Related to #7398

Since this `KnowledgeDatabase` will not only store knowledge but also information about trackers and torrent health, it needs to be renamed to something more general. Additionally, it should be extracted from the `KnowledgeComponent` to a separate component.


drew2a:
I plan to address the Codacy issues in a separate PR, as they are not related to the changes made in this one.

xoriole (approved):

PR #7608, opened 2023-09-29
`tribler.db` structure for the `TorrentHealth`

drew2a:
This PR introduces the `TorrentHealth` DB scheme.

While the new DB structure is intended to be the 'final' version, the data migration process itself will be defined in upcoming PRs

Related to #7398


src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -15,13 +16,23 @@
...
+
+
 @dataclass(order=True)
 class HealthInfo:
     infohash: bytes = field(repr=False)
     seeders: int = 0
     leechers: int = 0
     last_check: int = field(default_factory=lambda: int(time.time()))
     self_checked: bool = False
+    source: Source = Source.NONE
```
kozlovsky, line 35: I think we can add a separate `tracker` string field with a tracker's URL.

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -15,13 +16,23 @@
 HEALTH_FRESHNESS_SECONDS = 4 * HOUR  # Number of seconds before a torrent health is considered stale. Default: 4 hours
 
 
+class Source(IntEnum):
+    """ Source of the Torrent Health information.
+    """
+    NONE = 0
+    TRACKER = 1
+    DHT = 2
```
kozlovsky, line 24: I think this enum should not include `TRACKER` and `DHT` items. They represent orthogonal information. The source enum values can be something like `Uknown`, `Local`, `OldPopularityCommunity`, `PopularityCommunityRandomTorrent`, or `PopularityCommunityPopularTorrent`, and we can extend this list in the future.

src/tribler/core/components/database/db/layers/health_data_access_level.py
```diff
@@ -0,0 +1,56 @@
...
+class HealthDataAccessLayer:
+    def __init__(self, knowledge_layer: KnowledgeDataAccessLayer):
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.instance = knowledge_layer.instance
+        self.Resource = knowledge_layer.Resource
+        self.HealthInfo = self.define_binding(self.instance)
+
+    @staticmethod
+    def define_binding(db):
+        class HealthInfo(db.Entity):
```
kozlovsky, line 24: I think the `Tracker` entity should also be added to the database. The `HealthInfo` entity should also have a `tracker` field that refers to the `Tracker` entity.

kozlovsky (changes requested):
It looks good; I think the `HealthDataAccessLayer` should also define the `Tracker` entity.

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
```diff
@@ -15,13 +16,23 @@
 HEALTH_FRESHNESS_SECONDS = 4 * HOUR  # Number of seconds before a torrent health is considered stale. Default: 4 hours
 
 
+class Source(IntEnum):
+    """ Source of the Torrent Health information.
+    """
+    NONE = 0
```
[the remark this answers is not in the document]
drew2a: Renamed

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
drew2a: Added, but we need more @xoriole input here.

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
drew2a: Added

src/tribler/core/components/database/db/layers/health_data_access_level.py
drew2a: Done

src/tribler/core/components/database/db/layers/health_data_access_level.py
xoriole, line 1: The name of the file consistent with other files might be`health_data_access_layer.py`

src/tribler/core/components/database/db/layers/tests/test_health_data_access_level.py
xoriole, line 1: The name of the file consistent with other files might be`test_health_data_access_layer.py`

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
xoriole: I agree with the current changes: NONE, TRACKER, DHT, TRACKER, POPULARITY_COMMUNITY

src/tribler/core/components/database/db/layers/health_data_access_level.py
drew2a: Fixed

src/tribler/core/components/database/db/layers/tests/test_health_data_access_level.py
drew2a: Fixed

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
kozlovsky: @xoriole let me explain my thoughts in more detail.

Right now, when PopularityCommunity gossips HealthInfo, it does not pass information about the origin of that info - whether it is obtained from DHT or a tracker. Sometimes, the gossiped health contains a surprising number of seeders that do not correspond with the results of local checks.

One possible explanation is that the results of local checks and the obtained from PopularityCommunity were obtained from different sources. For example, local torrent metadata for a specific torrent hash can include trackers A and B, both dead. At the same time, on a different peer, the same torrent hash can be associated with a different metainfo that includes a tracker C with many seeders for that torrent.

For that reason, it may be beneficial in a future version of PopularityCommunity to gossip about the source of the information, including the tracker URL. It can help to understand the difference in results of checks and also can allow spreading the information about the actual live trackers.

Also, it may be beneficial to know WHY PopularityCommunity decided to gossip about some health information: was it because the torrent was randomly selected, because it is popular, or for some other reason. It allows us to see the proportion of torrents gossiped for different reasons and experiment with various algorithms for choosing which torrent to gossip about.

For that reason, it may be beneficial to separate three orthogonal types of information in the HealthInfo structure:
* **How** the torrent health was calculated - via DHT, BEP33, or some specific tracker
* **Who** calculated the torrent health - the local instance or a remote peer
* **Why** the health was sent to us - as a random torrent, as a popular torrent, or for a different reason

Currently, we do not store all this information in the database. But it does not harm to add these fields now when we create a new database structure, as it may be harder to change the database structure in the future. It does not require changes in the existing PopularityCommunity, as it can use just a subset of possible combinations of fields.

So, I don't think we should mix enum items DHT/TRACKER/POPULARITY_COMMUNITY in the same field. Yes, right now, we don't know whether the information gossiped by the popularity community was obtained from DHT or from some tracker. But in the future, we can extend the format of the gossip message and include this information, and in my opinion, it is better to put orthogonal information into different fields from the beginning.

kozlovsky (changes requested):

src/tribler/core/components/database/db/layers/health_data_access_layer.py
```diff
@@ -0,0 +1,68 @@
...
+    def define_binding(db):
+        class HealthInfo(db.Entity):
+            id = orm.PrimaryKey(int, auto=True)
+
+            torrent = orm.Required(lambda: db.Resource, index=True)
+
+            seeders = orm.Required(int, default=0)
+            leechers = orm.Required(int, default=0)
+            source = orm.Required(int, default=0)  # Source enum
+            trackers = orm.Set(lambda: Tracker)
```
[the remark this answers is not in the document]
drew2a: By saying, 'it is not necessary to import old health information from the metadata store,' do you mean that we can disregard all current health information and rely only on the new data?

src/tribler/core/components/database/db/layers/health_data_access_layer.py
```diff
@@ -0,0 +1,68 @@
...
+
+class HealthDataAccessLayer:
+    def __init__(self, knowledge_layer: KnowledgeDataAccessLayer):
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.instance = knowledge_layer.instance
+        self.Resource = knowledge_layer.Resource
+        self.HealthInfo, self.Tracker = self.define_binding(self.instance)
+
+    @staticmethod
+    def define_binding(db):
```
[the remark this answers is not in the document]
drew2a: Should we add it to a separate DAL?

src/tribler/core/components/database/db/layers/health_data_access_layer.py
```diff
@@ -0,0 +1,68 @@
...
+        class HealthInfo(db.Entity):
```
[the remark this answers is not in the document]
drew2a: Fixed

src/tribler/core/components/database/db/layers/health_data_access_layer.py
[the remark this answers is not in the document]
kozlovsky: Yes, we can add a separate DAL for the search. I think adding this DAL is outside the scope of the current PR

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
kozlovsky: So, mixing DHT / TRACKER / POPULARITY_COMMUNITY should work now but can restrict us later. On the other side, keeping them in separate fields will allow us to send tracker info via an extended version of PopularityCommunity, and it can be beneficial for us while being fully backward-compatible and supporting the current protocol of PopularityCommunity.

kozlovsky (approved):

src/tribler/core/components/torrent_checker/torrent_checker/dataclasses.py
xoriole: In my opinion, source as DHT / TRACKER / POPULARITY_COMMUNITY is sufficient for now.

While knowing tracker details could be useful we cannot trust this information as received via Popularity Community. I would consider only health info obtained from self-check to be somewhat reliable (not entirely). So, I do not see a clear picture of how having tracker info could be useful.

Regarding the questions:
- **How the torrent health was calculated - via DHT, BEP33, or some specific tracker:** This is specified by the external peer so might not be reliable what value it has.
- **Who calculated the torrent health - the local instance or a remote peer:** If the source is DHT or Tracker, I'd consider it as local. If it is Popularity community then it is remote.
- **Why the health was sent to us - as a random torrent, as a popular torrent, or for a different reason** - Since only health checked torrent info is shared, there might not be much difference here.

xoriole (approved):

PR #7616, opened 2023-10-02
Fixes #7598: Use safe_repr function to display alert reprs; catch UnicodeDecodeError in all alert handlers

### LINKED ISSUE #7598: UnicodeDecodeError in on_tracker_reply_alert
[Here](https://github.com/Tribler/tribler/blob/e21f9cf688d286184dc6ff27bd8c414d6514e5a4/src/tribler/core/components/libtorrent/download_manager/download.py#L309):
```python
    def on_tracker_reply_alert(self, alert: lt.tracker_reply_alert):
        self._logger.info(f'On tracker reply alert: {alert}')
        ...
```
Related to #7406.

In my opinion, the correct way to fix this class of errors is to add a `safe_repr` function and use it for getting alert repr:
```python
    def on_tracker_reply_alert(self, alert: lt.tracker_reply_alert):
        self._logger.info(f'On tracker reply alert: {safe_repr(alert)}')
        ...
```

kozlovsky:
This PR fixes #7598. Now the UnicodeDecode error is handled not only in `on_tracker_error_alert`, but in all other handlers as well.


src/tribler/core/components/libtorrent/download_manager/download.py
```diff
@@ -318,26 +322,23 @@ def on_tracker_error_alert(self, alert: lt.tracker_error_alert):
...
-                status = 'HTTP status code %d' % alert.status_code
-            elif alert.status_code == 0:
-                status = 'Timeout'
-            else:
-                status = 'Not working'
+        self._logger.error(f'On tracker error alert: {safe_repr(alert)}')
+        url = alert.url
+
+        if alert.msg:
+            status = 'Error: ' + alert.msg
```
drew2a, line 329: I believe this line should be tested since some portions of this function have already been covered by tests.

drew2a (changes requested):

src/tribler/core/components/libtorrent/download_manager/download.py
kozlovsky: I added more tests to cover all branches of this function

drew2a (approved):

PR #7620, opened 2023-10-04
Updated the IPv8 pointer

qstokkink:
This PR upgrades Tribler to the latest IPv8 version.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -9,5 +9,5 @@ sphinxcontrib-openapi==0.7.0
 configobj==5.0.6
 mistune==0.8.4  # sphinxcontrib-openapi==0.7.0 cannot work with the latest mistune version (2.0.0)
 MarkupSafe==2.0.1  # used by jinja2; 2.1.0 version removes soft_unicode and breaks jinja2-2.11.3
-pyipv8==2.8.0
+pyipv8==2.12.0
 setuptools>=65.5.1 # not directly required, pinned by Snyk to avoid a vulnerability
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -23,3 +23,4 @@ libtorrent==1.2.19
 file-read-backwards==3.0.0
 Brotli==1.0.9 # to prevent AttributeError on macOs: module 'brotli' has no attribute 'error' (in urllib3.response)
 human-readable==1.3.2
+typing_extensions==4.7.1
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -18,7 +18,7 @@ sentry-sdk==1.31.0
 yappi==1.4.0
 yarl==1.9.2 # keep this dependency higher than 1.6.3. See: https://github.com/aio-libs/yarl/issues/517
 bitarray==2.7.6
-pyipv8==2.11.0
+pyipv8==2.12.0
 libtorrent==1.2.19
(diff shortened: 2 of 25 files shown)


qstokkink:
I'm not sure what is wrong with `pytest / run (windows-latest)`.

qstokkink:
>I'm not sure that introducing the CommunitySettings is an improvement. What is the necessity of this intermediate class, and what benefits does it provide over simply and directly passing arguments through the constructor?

This was introduced due to pylint/ruff's `PLR0913` violation triggering on the IPv8 codebase. Ruff's documentation includes a small example of the benefits: https://docs.astral.sh/ruff/rules/too-many-arguments/ If you want a deeper Software Engineering argument, there is a more elaborate write-up here: https://matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/

The necessity comes from the fact that Tribler will not work anymore with the latest version of IPv8 without this change. However, you can always hotpatch in some dirty construction later if you really want to. For this PR, I'll stick to the recommended approach.

> Additionally, this PR introduces a new dependency (typing_extension). Could you elaborate on the advantages of using it and explain why native typing is insufficient?

`typing_extensions` is a collection of backports for native typing constructs but from future Python versions. So, in a sense, native typing IS sufficient BUT we are running a Python version that is too old. IPv8 uses `typing_extensions` for the backports of `Self` and `Protocol`.

drew2a:
> Settings

Too many arguments in a function can sometimes be a signal, or 'smell,' of suboptimal code or design. While this might be true in some instances, it's not universally applicable. The number of function parameters should not serve as an absolute indicator necessitating refactoring.

In Tribler, some communities do indeed have numerous parameters, while others do not. Therefore, only a few communities in Tribler appear to be potential candidates for refactoring, not all of them.

In this context, I'd argue that the necessity for a specialized `Setting` object feels a bit like 'dictatorship' from the ipv8 side. Speaking as a developer who has crafted communities in Tribler and will continue to do so in the future, I find this approach unsavory as it introduces an unnecessary level of abstraction, will likely clutter the code, and doesn’t present any significant benefits.

Please consider amending ipv8 such that this feature becomes optional instead of mandatory.

> typing_extensions

I understand. If 'IPv8 uses typing_extensions for the backports of Self and Protocol,' then IPv8 should install this dependency during its installation. It should not depend on the target application (Tribler) to install all required libraries.

qstokkink:
>In Tribler, some communities do indeed have numerous parameters, while others do not. Therefore, only a few communities in Tribler appear to be potential candidates for refactoring, not all of them.

`Community` has 5 arguments to `__init__`, `TriblerCommunity` adds a 6th one. ALL communities are over the Pylint recommended max of 5 arguments before they even start adding anything.

> I'd argue that the necessity for a specialized Setting object feels a bit like 'dictatorship' from the ipv8 side. 

Calling an API a dictatorship seems a bit overly dramatic but I guess I can agree to being a dictator: I don't think it's a bad thing to have an API.

> IPv8 should install this dependency during its installation. It should not depend on the target application (Tribler) to install all required libraries.

It does. However, Tribler redefines exact versions of dependencies that only appear in IPv8. I added this to be consistent.

---

I get the overall feeling that you are not favor of this PR, which is fine: I'm not here to force you to use anything and I'll just close it.

Issue #7603, opened 2023-09-18
[Mysterious error][7.13.0] CoreConnectionError: "The connection to the Tribler Core was lost" in event_request_manager.on_error


kozlovsky:
https://sentry.tribler.org/organizations/tribler/issues/2003/events/de644ea355b040ba95b14ae230256a88/

The error looks strange; I spent quite some time trying to figure it out. The following is how I understand what is going on:

1. A user tries to close the Tribler GUI.
2. `TriblerWindow.close_tribler()` is called.
3. `TriblerWindow.close_tribler()` method [calls `self.core_manager.stop()`](https://github.com/Tribler/tribler/blob/e68b01462016df3a330ee6cc771f9bdd191db483/src/tribler/gui/tribler_window.py#L1111):

    ```python
    class CoreManager(QObject):
        ...
        def stop(self, quit_app_on_core_finished=True):
            if self.shutting_down:
                return

            self.shutting_down = True
            self._logger.info("Stopping Core manager")  # 4

            if self.core_process and not self.core_finished:
                if not self.core_connected:
                    ...
                self.events_manager.shutting_down = True  # 5

                def send_shutdown_request(initial=False):
                    if initial:
                        self._logger.info(f"Sending {SHUTDOWN_ENDPOINT} request to Tribler Core")  # 6
                    else:
                        self._logger.warning(f"Re-sending {SHUTDOWN_ENDPOINT} request to Tribler Core")
                    ...
                send_shutdown_request(initial=True)
        ```
4. You can see the 'Stopping Core manager' message in the log, so the line marked as `#4` is executed.
5. The line `#6` that sets the `events_manager.shutting_down` flag to `True` should also be executed because...
6. ...in the log, you can see the message from the line `#6`, `"Sending shutdown request to Tribler Core"`.

So, based on log messages, I'm sure that the flag `events_manager.shutting_down` should be set to `True` before the shutdown request to Core is performed.

But then, the actual error happens in the [`on_error` method of the event manager](https://github.com/Tribler/tribler/blob/e68b01462016df3a330ee6cc771f9bdd191db483/src/tribler/gui/event_request_manager.py#L137):
```python
class EventRequestManager(QNetworkAccessManager):
    ...
    def on_error(self, error: int, reschedule_on_err: bool):
        if self.shutting_down:
            return

        if self.receiving_data:
            raise CoreConnectionError('The connection to the Tribler Core was lost')    
```

[...]

PR #7621, opened 2023-10-06
Fix multiple instances of Tribler on Flatpak

xoriole:
~This PR fixes the multiple instances of Tribler running on Flatpak. If the app is connected to the previous instance via QLocalSocket then there is another instance running. In this case, Tribler closes now by passing its arguments to the running process via the QT local socket.~

~In addition to that, in the PR, for the core process, it is now checked if the port is known whether or not REST API is accessible. If the API is not accessible, then the core process is considered not running. Checking on the pid is not sufficient on the Flatpak environment because the pid is not unique anymore on the sandbox environment.~

This PR addresses the issue of multiple instances of Flatpak application running in parallel and crashing Tribler when one of the instances is closed. It is addressed in the following way.

If the last primary process from the database has a PID that either matches the current process PID or can be confirmed to be running, then the current process instance is terminated. This prevents the second instance of Tribler from running.

It could be argued that the PID could be reused and could lead to Tribler terminating itself and no instances running at all. In such a case, the likelihood of the last Tribler process having the same PID as the current process and the database having the process record with a primary set to 1, in my opinion, is negligible. Furthermore, if such a case happens, Tribler will likely start on the next run as the PID would most likely change from the previous run which in my opinion is acceptable.

Fixes https://github.com/Tribler/tribler/issues/7626
Fixes partially https://github.com/Tribler/tribler/issues/7603

### DIFF
--- a/src/tribler/core/utilities/process_manager/process.py
+++ b/src/tribler/core/utilities/process_manager/process.py
@@ -140,7 +140,7 @@ class TriblerProcess:
 
     def is_current_process(self) -> bool:
         """Returns True if the object represents the current process"""
-        return self.pid == os.getpid() and self.is_running()
+        return self.pid == os.getpid()
 
     @with_retry
     def become_primary(self) -> bool:
--- a/src/tribler/core/utilities/process_manager/manager.py
+++ b/src/tribler/core/utilities/process_manager/manager.py
@@ -119,7 +119,7 @@ class ProcessManager:
             row = cursor.fetchone()
             if row is not None:
                 process = TriblerProcess.from_row(self, row)
-                if process.is_running():
+                if process.is_current_process() or process.is_running():
                     return process.rowid
 
                 # Process is not running anymore; mark it as not primary
--- a/src/tribler/core/utilities/process_manager/tests/test_manager.py
+++ b/src/tribler/core/utilities/process_manager/tests/test_manager.py
@@ -11,8 +11,8 @@ from tribler.core.utilities.process_manager.process import ProcessKind, TriblerP
 # pylint: disable=protected-access
 
 
-@patch('os.getpid', return_value=1)
-def test_become_primary(os_getpid: Mock, process_manager: ProcessManager):
+@patch('os.getpid', new=lambda: 1)
+def test_become_primary(process_manager: ProcessManager):
     # Initially process manager fixture creates a primary current process that is a single process in DB
     p1 = process_manager.current_process
     assert p1.primary
--- a/src/tribler/core/utilities/process_manager/tests/test_manager.py
+++ b/src/tribler/core/utilities/process_manager/tests/test_manager.py
@@ -11,7 +11,8 @@ from tribler.core.utilities.process_manager.process import ProcessKind, TriblerP
 # pylint: disable=protected-access
 
 
-def test_become_primary(process_manager: ProcessManager):
+@patch('os.getpid', return_value=1)
+def test_become_primary(os_getpid: Mock, process_manager: ProcessManager):
     # Initially process manager fixture creates a primary current process that is a single process in DB
     p1 = process_manager.current_process
     assert p1.primary


kozlovsky:
Thank you for your efforts in addressing this issue. After looking it over, I have some concerns about the correctness of the suggested approach. I'm diving deeper into the details and will provide a more comprehensive review in about half an hour.

src/tribler/gui/start_gui.py
```diff
@@ -64,7 +64,7 @@ def run_gui(api_port: Optional[int], api_key: Optional[str], root_state_dir, par
         translator = get_translator(settings.value('translation', None))
         app.installTranslator(translator)
 
-        if not current_process_is_primary:
+        if not current_process_is_primary or app.connected_to_previous_instance:
```
kozlovsky, line 67: This PR contains two fixes for the GUI and Core processes. Let's first look at the fix for the GUI process, which is implemented as an extended check at line 67 that adds the "app.connected_to_previous_instance" condition.

It looks like this PR can indeed solve the problem on Flatpak for the case when the first (primary) Flatpack-based Tribler instance is already running and the second (secondary) Flatpack-based Tribler instance is starting. Before your fix, the second GUI process incorrectly considers itself a primary instance because it has the same PID as the first GUI process. With your changes, the second GUI process is regarded as a primary process but finishes immediately because it can connect to the first GUI process via the local socket.

However, the solution proposed in this PR creates a race condition for the case when two instances of the Tribler application are starting at the same moment. This situation is very common for some obscure reason and needs to be handled properly. Unfortunately, with the simultaneous start of two GUI processes, it is possible to have a race condition when the first process claims itself primary in the `processes.sqlite` lock file, while the second process was the first to grab the local socket. In that case, it can be possible that the first (primary) process is connected to the local socket of the secondary process. It can lead to a situation when both processes stop simultaneously, the second process because it is not primary and the first because it was able to connect to the local socket of the secondary process.

In other words, this change fixes the Flatpak issue but re-creates a race condition for the non-Flatpak case. Previously, the race condition manifests itself as two GUI processes running in parallel, and now it appears as a Tribler application that refuses to start at all.

src/tribler/core/utilities/process_manager/process.py
```diff
@@ -177,8 +180,29 @@ def is_running(self):
...
+
+        This is done by checking the Docs endpoint which is always available
+        and does not require API_KEY to access.
+        """
+        if not self.api_port:
+            return False
+
+        try:
+            docs_url = f"http://localhost:{self.api_port}/docs"
+            _ = requests.get(docs_url, timeout=API_CHECK_TIMEOUT_IN_SECONDS)
```
kozlovsky, line 201: The second fix proposed in this PR is an HTTP request to the Tribler REST API added to the `TriblerProcess.is_running` method.

Note that the `is_running` method is called, for example, when `TriblerProcess.__str__()` is called. It looks a bit extreme to me to issue an HTTP request whenever someone tries to print an instance of a core process.

Second, it is not clear why it is helpful to have this check for the Core process for two reasons:
1. If the logic for GUI processes is correct, the second GUI instance should understand that it is secondary and should not start the second Core at all. So, the check for the Core is less important than for the GUI if the GUI check is implemented correctly.
2. It is hard to say what this check allows us to achieve. Initially, when the Core process is just started, its `api_port` is always None. Not that with the current implementation, the GUI process does not pass the API port number to the Core process. The Core process opens the REST API on any available port and sets its value to the database so the GUI process can look up the port value. So when the Core process is just starting, its `api_port` is always None, and this check will not be triggered. When two Core processes are already running by mistake, their `api_port` values will differ. Each Core process has a different value of its API port and can connect to it (by the way, can the code create a deadlock when the Core process connects to its port? The call to `request.get` is synchronous, so it looks like it should block the asyncio loop). It is hard to say why this check of the HTTP endpoint availability is useful and how correctly it works in all different use cases.

kozlovsky (changes requested):
In my opinion, it is an interesting attempt to solve the issue, but unfortunately, I think it can bring new significant issues and make the code logic more complicated. I think we need to find a different approach to this issue.

src/tribler/gui/start_gui.py
xoriole: Thank you for your feedback. Here's a more detailed explanation:

1. The `app.connected_to_previous_instance` was introduced to enhance the current check determining if the process is primary. Although the existing check for the primary process should ideally address the scenario where multiple instances of Tribler exist, it doesn't perform as expected in environments like Flatpak. The added condition is supplemental (an OR condition), ensuring it doesn't disrupt the standard execution flow.

2. The value of `app.connected_to_previous_instance` is set to True only when the new GUI process successfully connects to the local socket. If a race condition were to occur, this value would be set to False, retaining the original execution behavior, in my opinion.

I haven't encountered or been able to reproduce a race condition scenario in the current setup. However, I'm keen to understand your perspective and would welcome any suggestions on potential improvements or alternative solutions.

Issue #7382, opened 2023-04-21
Current upgrader's implementation is a time bomb


drew2a:
The current upgrader procedure is the following.

No matter what it runs all available migrations on each Tribler's run:

```python
TriblerUpgrader(377) - Upgrading GigaChannel DB from version 8 to 10
TriblerUpgrader(197) - Upgrade Pony DB 10 to 11
tribler.core.upgrade.config_converter(12) - Upgrade config to 7.6
TriblerUpgrader(213) - Upgrade bandwidth accounting DB 8 to 9
TriblerUpgrader(181) - Upgrade Pony DB 11 to 12
TriblerUpgrader(166) - Upgrade Pony DB 12 to 13
TriblerUpgrader(146) - Upgrade Pony DB from version 13 to version 14
TriblerUpgrader(130) - Upgrade tags to knowledge
TriblerUpgrader(109) - Remove old logs
TriblerUpgrader(135) - Upgrade Pony DB from version 14 to version 15
```

This is a historically established approach. We, as developers, follow this approach with the contemporary Tribler developments and just add more and more migrations to the existing ones.

However, this approach is subject to at least two problems. The first problem is easier to understand and avoid therefore I first start with it.

## The problem: a developer should be quite careful with the conditions inside his migration. 

There are no guidelines or even tips or hints on how the migration should be written. From the perspective of the code, it is just a function without any prerequisites. 

There is an example of the two very first migrations

```python
class TriblerUpgrader:
    ...
    def run(self):
        self.upgrade_pony_db_8to10()
        self.upgrade_pony_db_10to11()

    def upgrade_pony_db_8to10(self):
        database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
        if not database_path.exists() or get_db_version(database_path) >= 10:
            return
        ... # code of the migration

   def upgrade_pony_db_10to11(self):
        database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
        if not database_path.exists():
            return
        ... # code of the migration
```

From the code above, the conditions for a migration execution are 

#### upgrade_pony_db_8to10:
```python
        if not database_path.exists() or get_db_version(database_path) >= 10:
            return
```

#### upgrade_pony_db_10to11
```python
        database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
        if not database_path.exists():
            return
```

As these migrations execute each Tribler's run it is possible that false positives may occur if the conditions are not carefully chosen.

[...]

PR #7622, opened 2023-10-09
Add `TriblerDatabaseMigrationChain`

drew2a:
This PR introduces a more strict and formal way of DB migrations.

The main idea behind this change is to standardize the migration procedure into a series of functions that resemble:

### first migration
```python
@migration(execute_only_if_version=0)
def scheme_migration_0(db: TriblerDatabase, **kwargs):  # pylint: disable=unused-argument
    db.execute('ALTER TABLE "TorrentState" ADD "has_data" BOOLEAN DEFAULT 0')
    db.execute('UPDATE "TorrentState" SET "has_data" = 1 WHERE last_check > 0')
```

### second migration
```python
@migration(execute_only_if_version=1)
def scheme_migration_1(db: TriblerDatabase, **kwargs):  # pylint: disable=unused-argument
    db.execute('ALTER TABLE "Tracker" ADD "updated" INT DEFAULT 0')
```

This PR should fix some of the problems described in #7382
Related to #7398

### DIFF
--- a/src/tribler/core/components/database/db/tribler_database.py
+++ b/src/tribler/core/components/database/db/tribler_database.py
@@ -1,14 +1,20 @@
 import logging
+import os
 from typing import Any, Optional
 
 from pony import orm
 
 from tribler.core.components.database.db.layers.health_data_access_layer import HealthDataAccessLayer
 from tribler.core.components.database.db.layers.knowledge_data_access_layer import KnowledgeDataAccessLayer
-from tribler.core.utilities.pony_utils import TrackedDatabase, get_or_create
+from tribler.core.utilities.pony_utils import TrackedDatabase, db_session, get_or_create
+
+MEMORY = ':memory:'
 
 
 class TriblerDatabase:
+    CURRENT_VERSION = 1
+    _SCHEME_VERSION_KEY = 'scheme_version'
+
     def __init__(self, filename: Optional[str] = None, *, create_tables: bool = True, **generate_mapping_kwargs):
         self.instance = TrackedDatabase()
 
@@ -25,20 +31,30 @@ class TriblerDatabase:
         self.TorrentHealth = self.health.TorrentHealth
         self.Tracker = self.health.Tracker
 
-        self.instance.bind('sqlite', filename or ':memory:', create_db=True)
+        filename = filename or MEMORY
+        db_does_not_exist = filename == MEMORY or not os.path.isfile(filename)
+
+        self.instance.bind('sqlite', filename, create_db=db_does_not_exist)
         generate_mapping_kwargs['create_tables'] = create_tables
         self.instance.generate_mapping(**generate_mapping_kwargs)
         self.logger = logging.getLogger(self.__class__.__name__)
 
+        if db_does_not_exist:
+            self.fill_default_data()
+
     @staticmethod
     def define_binding(db):
         """ Define common bindings"""
-
-        class Misc(db.Entity):  # pylint: disable=unused-variable
+        class Misc(db.Entity):
             name = orm.PrimaryKey(str)
             value = orm.Optional(str)
 
         return Misc
+
+    @db_session
+    def fill_default_data(self):
+        self.logger.info('Filling the DB with the default data')
+        self.set_misc(self._SCHEME_VERSION_KEY, self.CURRENT_VERSION)
 
     def get_misc(self, key: str, default: Optional[str] = None) -> Optional[str]:
         data = self.Misc.get(name=key)
@@ -47,6 +63,19 @@ class TriblerDatabase:
     def set_misc(self, key: str, value: Any):
         key_value = get_or_create(self.Misc, name=key)
         key_value.value = str(value)
+
+    @property
+    def version(self) -> int:
+        """ Get the database version"""
+        return int(self.get_misc(key=self._SCHEME_VERSION_KEY, default=0))
+
+    @version.setter
+    def version(self, value: int):
+        """ Set the database version"""
+        if not isinstance(value, int):
+            raise TypeError('DB version should be integer')
+
+        self.set_misc(key=self._SCHEME_VERSION_KEY, value=value)
 
     def shutdown(self) -> None:
         self.instance.disconnect()
--- a/src/tribler/core/upgrade/tribler_db/decorator.py
+++ b/src/tribler/core/upgrade/tribler_db/decorator.py
@@ -1,0 +1,53 @@
+import functools
+import logging
+from typing import Callable, Optional
+
+from tribler.core.components.database.db.tribler_database import TriblerDatabase
+from tribler.core.utilities.pony_utils import db_session
+
+MIGRATION_METADATA = "_tribler_db_migration"
+
+logger = logging.getLogger('Migration (TriblerDB)')
+
+
+def migration(execute_only_if_version: int, set_after_successful_execution_version: Optional[int] = None):
+    """ Decorator for migration functions.
+    The migration executes in the single transaction. If the migration fails, the transaction is rolled back.
+    The decorator also sets the metadata attribute to the decorated function. It could be checked by
+    calling the `has_migration_metadata` function.
+    Args:
+        execute_only_if_version: Execute the migration only if the current db version is equal to this value.
+        set_after_successful_execution_version: Set the db version to this value after the migration is executed.
+            If it is not specified, then `set_after_successful_execution_version = execute_only_if_version + 1`
+    """
+
+    def decorator(func):
+        @functools.wraps(func)
+        @db_session
+        def wrapper(db: TriblerDatabase, **kwargs):
+            target_version = execute_only_if_version
+            if target_version != db.version:
+                logger.info(
+                    f"Function {func.__name__} is not executed because DB version is not equal to {target_version}. "
+                    f"The current db version is {db.version}"
+                )
+                return None
+
+            result = func(db, **kwargs)
+
+            next_version = set_after_successful_execution_version
+            if next_version is None:
+                next_version = target_version + 1
+            db.version = next_version
+
+            return result
+
+        setattr(wrapper, MIGRATION_METADATA, {})
+        return wrapper
+
+    return decorator
+
+
+def has_migration_metadata(f: Callable):
+    """ Check if the function has migration metadata."""
+    return hasattr(f, MIGRATION_METADATA)
--- a/src/tribler/core/upgrade/tribler_db/migration_chain.py
+++ b/src/tribler/core/upgrade/tribler_db/migration_chain.py
@@ -1,0 +1,48 @@
+import logging
+from typing import Callable, List, Optional
+
+from tribler.core.components.database.db.tribler_database import TriblerDatabase
+from tribler.core.upgrade.tribler_db.decorator import has_migration_metadata
+from tribler.core.upgrade.tribler_db.scheme_migrations.scheme_migration_0 import scheme_migration_0
+from tribler.core.utilities.path_util import Path
+from tribler.core.utilities.simpledefs import STATEDIR_DB_DIR
+
+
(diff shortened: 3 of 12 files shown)


src/tribler/core/upgrade/tribler_db/migration_chain.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Returns: True if all migrations were executed successfully, False otherwise.
+        An exception in any of the migrations will halt the execution chain and be re-raised.
+        """
+
+        if not self.db:
+            return False
+
+        for _ in self.steps():
+            ...
```
kozlovsky, line 44: Is it intended to be this way in the current PR?

src/tribler/core/upgrade/upgrade.py
```diff
@@ -107,6 +108,9 @@ def run(self):
         self.upgrade_pony_db_14to15()
         self.upgrade_knowledge_to_tribler_db()
 
+        migration = TriblerDatabaseMigrationChain(self.state_dir)
+        migration.execute()
```
kozlovsky, line 112: IMO, it is better to rename the variable to `migrations` or `migration_chain` here. Otherwise, we have confusing terminology and use the same terms with different meanings in different parts of the code.

It is better to have the following definitions:
* a migration chain consists of individual migrations
* an individual migration can possibly consist of several steps (usually only one)

src/tribler/core/upgrade/tribler_db/decorator.py
```diff
@@ -0,0 +1,51 @@
...
+        @functools.wraps(func)
+        @db_session
+        def wrapper(db: TriblerDatabase, **kwargs):
+            target_version = execute_only_if_version
+            if target_version != db.version:
+                logger.info(
+                    f"Function {func.__name__} is not executed because DB version is not equal to {target_version}. "
+                    f"The current db version is {db.version}"
+                )
+                return None
```
kozlovsky, line 36: It looks dangerous to silently skip the migration in case of a version mismatch and can lead to incorrect migrations.

As a possible example:
* We have migration chain `[a, b, c]`, and after the migration `c`, the database has version 3
* Two developers make two independent PRs with migrations `d` and `e`, both require db version 3
* After both PRs were successfully merged, the merge conflict in the migration chain was resolved to `[a, b, c, d, e]`, but both `d` and `e` still mistakenly require version 3 of the database.
* The execution of the migration chain silently skips the migration `e`

For that reason, it is better to raise an error instead of skipping the migration so the database will not be left in a surprising state.

src/tribler/core/upgrade/tribler_db/migration_chain.py
```diff
@@ -0,0 +1,53 @@
...
+
+        return True
+
+    def steps(self) -> Iterator:
```
kozlovsky, line 48: Regarding the terminology: I'd say we have a migration chain that consists of migrations, and a single migration can possibly consist of several steps. Here, in the `for` loop, we iterate over migrations and not over the steps of an individual migration. So, it is better to rename it from `steps` to `migrations` to avoid confusion.

src/tribler/core/upgrade/tribler_db/decorator.py
```diff
@@ -0,0 +1,51 @@
...
+
+from tribler.core.components.database.db.tribler_database import TriblerDatabase
+from tribler.core.utilities.pony_utils import db_session
+
+MIGRATION_METADATA = "_tribler_db_migration"
+
+logger = logging.getLogger('Migration (TriblerDB)')
+
+
+def migration(execute_only_if_version: int):
```
kozlovsky, line 13: **nitpick:** It looks correct, but it may be more convenient to specify the resulting version after the migration and assume the schema version before the migration should be one less. In that case, `@migration(schema_version=10)` would mean that the version after the migration is executed successfully should be 10, and the version before the migration should be 9. In that case, it will be a bit more obvious that after applying `@migration(schema_version=10)`, the resulting schema has version 10. But it is a matter of personal preferences.

src/tribler/core/components/database/db/tribler_database.py
```diff
@@ -1,14 +1,20 @@
...
 from tribler.core.components.database.db.layers.knowledge_data_access_layer import KnowledgeDataAccessLayer
-from tribler.core.utilities.pony_utils import TrackedDatabase, get_or_create
+from tribler.core.utilities.pony_utils import TrackedDatabase, db_session, get_or_create
+
+MEMORY = ':memory:'
 
 
 class TriblerDatabase:
+    CURRENT_VERSION = 1
+    _SCHEME_VERSION = 'scheme_version'
```
kozlovsky, line 16: **nitpick:** The `_version` suffix is a bit confusing here, as the variable does not contain the version value. Can we rename it to something like `version_key`?

kozlovsky (changes requested):

src/tribler/core/upgrade/tribler_db/migration_chain.py
drew2a: Yes, this executes each migration step-by-step. Do you have a better approach to achieve the same?

src/tribler/core/upgrade/upgrade.py
drew2a: Agree, renamed to the `migration_chain`

src/tribler/core/upgrade/tribler_db/decorator.py
drew2a: If we were to raise an exception as you suggested, then almost all runs would result in unsuccessful execution.

Example:

1. We have a migration chain [a, b, c] with the corresponding versions [1, 2, 3]. Let's represent them as [a(1), b(2), c(3)].
2. Our DB currently has a version set to 2.
3. We begin executing the migration chain starting with migration a(1).
4. This results in an exception because the DB version is not equal to 1.

src/tribler/core/upgrade/tribler_db/migration_chain.py
drew2a: We already have a property named `migrations` that stores all the migrations for this chain.

`def steps(self)` represents the execution steps for the migration chain.

src/tribler/core/upgrade/tribler_db/decorator.py
drew2a: I agree that this is "a matter of personal preferences". Let's ask @xoriole for a second opinion.

src/tribler/core/components/database/db/tribler_database.py
drew2a: Renamed

src/tribler/core/upgrade/tribler_db/migration_chain.py
kozlovsky: Currently, the migration is executed as a side effect of the iteration. It looks non-intuitive. I think it is better to split it into two different phases: the selection of suitable migrations and the execution of selected migrations:
```python
for migration in self.find_migrations_to_apply(current_version=self.db.version):
    migration(self.db, state_dir=self.state_dir)
```

This way, it becomes possible to filter out migrations that were already applied and only apply new migrations. It also allows to separately test the selection of suitable migrations and the execution of a migration.

src/tribler/core/upgrade/tribler_db/decorator.py
kozlovsky: We can first select suitable migrations, as described in the previous comment, and execute only them. This way, the migration decorator can throw an exception if the migration was performed at an improper moment. It looks safer than silently skipping migrations with incorrect db version numbers.

Also, consider the migrations [a, b, c, d, e] that mistakenly expect the schema version [1, 3, 2, 4, 5]. With silent skipping of the incorrect migrations, only migrations 1 and 2 will be executed. It is better to prevent such errors.

src/tribler/core/upgrade/tribler_db/migration_chain.py
kozlovsky: I mean, it is pretty common to say that an individual migration may consist of several steps. For that reason, using the word "steps" when referring to migrations looks confusing. Maybe we can rename `steps` to something like `find_suitable_migrations(current_schema_version)`.

src/tribler/core/upgrade/tribler_db/migration_chain.py
drew2a: Indeed it is just another approach to implementing the same feature. The current version performs the "selection" procedure inside the decorator. It is already tested in `test_decorator.py`

> This way, it becomes possible to filter out migrations that were already applied and only apply new migrations. It also allows to separately test the selection of suitable migrations and the execution of a migration.

The current approach already contains these properties.

drew2a: > Currently, the migration is executed as a side effect of the iteration

That was the intention. This way we can test the execution step-by-step, just iterating the steps.

But I'm also with rewriting it to something like:

```python
    def execute(self) -> bool:
        """ Execute all migrations in the chain.

        Returns: True if all migrations were executed successfully, False otherwise.
        An exception in any of the migrations will halt the execution chain and be re-raised.
        """

        if not self.db:
            return False

        for m in self.select():
            m(self.db, state_dir=self.state_dir)

        return True

    def select(self) -> Iterator:
        for m in self.migrations:
            if not has_migration_metadata(m):
                raise NotImplementedError(f'The migration {m} should have `migration` decorator')
            yield m
```

src/tribler/core/upgrade/tribler_db/decorator.py
drew2a: Actually, I'm not sure that it is a good idea to automatically fix **potential merge conflict errors**.

Also, I'm not sure that the Migration Chain should have a monopoly on changing the version value.

src/tribler/core/upgrade/tribler_db/migration_chain.py
drew2a: I've removed the `steps` property. It's simpler and more straightforward now, but we've lost the ability to execute steps one by one and check the results of each execution.

The last version:

```pyhon
    def execute(self) -> bool:
        """ Execute all migrations in the chain.

        Returns: True if all migrations were executed successfully, False otherwise.
        An exception in any of the migrations will halt the execution chain and be re-raised.
        """

        if not self.db:
            return False

        for m in self.migrations:
            if not has_migration_metadata(m):
                raise NotImplementedError(f'The migration {m} should have `migration` decorator')
            m(self.db, state_dir=self.state_dir)

        return True
```

synctext:
>  a more strict and formal way of DB migrations.

A single place where the new format is documented would be great. Now lots of stuff is scattered across multiple issues. Fine for now, but please don't invest too much time into a generic migration tool with sub-migrations and concurrent developer conflict resolution fanciness. Please give me a new stable Tribler release in 2 weeks, with the merged Randevous certificates!

kozlovsky:
> the PR has been pending for about a week. I haven't received any feedback from the development team. But before I can merge, I need approval.

But it was marked as ready for review just today. So, it looks reasonable that after the PR is ready for review, it can receive some review comments and not be accepted immediately.

src/tribler/core/upgrade/tribler_db/decorator.py
kozlovsky: > Actually, I'm not sure that it is a good idea to automatically fix potential merge conflict errors.

I agree, and what I like in the current approach is that the merge conflicts for migrations always require explicit conflict resolution, as migrations are explicitly listed in the same list.

What I'm concerned about is that the current PR approach allows silent skipping of migration if the schema number specified in the decorator does not agree with the order of migrations in the explicit list. Raising an error looks like a safer approach.

kozlovsky (approved):
After additional discussion with @drew2a, we decided that we can handle version conflicts between different migrations in the future, and it is not necessary to handle it right now, as the chance of this error is low.

Issue #5252, opened 2020-03-27
Database disk image is malformed


devos50:
From the error reporter (7.5.0-RC2):

```python
Traceback (most recent call last):
File "src\tribler-gui\tribler_gui\core_manager.py", line 106, in on_received_state
RuntimeError: database disk image is malformed
--LONG TEXT--
Traceback (most recent call last):
File "src\run_tribler.py", line 95, in start_tribler
File "src\tribler-core\tribler_core\session.py", line 440, in start
File "src\tribler-core\tribler_core\session.py", line 170, in load_ipv8_overlays
File "ipv8\attestation\trustchain\community.py", line 64, in __init__
File "ipv8\attestation\trustchain\database.py", line 43, in __init__
File "ipv8\attestation\trustchain\database.py", line 419, in open
File "ipv8\database.py", line 127, in open
File "ipv8\database.py", line 162, in _initial_statements
sqlite3.DatabaseError: database disk image is malformed
--CONTEXT--
{'message': 'Task exception was never retrieved', 'exception': DatabaseError('database disk image is malformed'), 'future': <Task finished coro=<start_tribler_core.<locals>.start_tribler() done, defined at src\run_tribler.py:64> exception=DatabaseError('database disk image is malformed')>}
```

Related to #1993

qstokkink:
What should we do when the database is corrupt beyond repair? Remove it?

ichorid:
In this case, we should mark the database corrupted and schedule a DB check on the next startup. Ideally, we should employ DB's own mechanisms for that.

devos50:
@ichorid do we want to fix this issue in 7.5?

ichorid:
There were only 2 bugreports with this kind of bug in the last few months. One should put some _real_ effort to garble their SQLite database to produce this kind of error. It is alson not trivial to recover from it later. I would mark it a low-priority and moved to backlog.

devos50:
On our (Proxmoxed) Mac, I now have a corrupt `metadata.db`, which is crashing consistently a few seconds after start-up. It errors with different kind of errors actually, depending on what the interactions with the database are. For example:

```
Traceback (most recent call last):
  File "ipv8/taskmanager.py", line 131, in done_cb
  File "tribler_core/modules/metadata_store/community/remote_query_community.py", line 149, in on_remote_select_response
  File "tribler_core/modules/metadata_store/store.py", line 287, in process_compressed_mdblob_threaded
[...]

Issue #1993, opened 2016-02-29
Tribler should be able to recover in case of corrupted files/databases


whirm:
To save our users situations like #1992

whirm:
Another example:

```
ERROR   1460769681.46        dispersy:1724  exception during handle_callback for torrent-collect-request
Traceback (most recent call last):
  File "Tribler\dispersy\dispersy.pyo", line 1719, in _update
  File "Tribler\community\search\community.pyo", line 466, in on_torrent_collect_request
  File "Tribler\community\search\community.pyo", line 512, in _create_pingpong
  File "Tribler\community\search\community.pyo", line 537, in __get_torrents
  File "Tribler\Core\CacheDB\SqliteCacheDBHandler.pyo", line 1370, in getRecentlyCollectedSwiftHashes
  File "Tribler\dispersy\util.pyo", line 43, in helper
  File "Tribler\Core\CacheDB\sqlitecachedb.pyo", line 2296, in fetchall
  File "Tribler\Core\CacheDB\sqlitecachedb.pyo", line 574, in fetchall
  File "Tribler\Core\CacheDB\sqlitecachedb.pyo", line 485, in execute_read
  File "Tribler\dispersy\util.pyo", line 43, in helper
  File "Tribler\Core\CacheDB\sqlitecachedb.pyo", line 2317, in _execute
CorruptError: CorruptError: database disk image is malformed
ERROR   1460769684.46   sqlitecachedb:2316  ===Twisted===
SQL Type: <type 'str'>
-----

            SELECT CT.swift_torrent_hash, CT.infohash, CT.num_seeders, CT.num_leechers, T.last_tracker_check, CT.insert_time
             FROM Torrent T, CollectedTorrent CT
             WHERE CT.torrent_id = T.torrent_id
             AND CT.swift_torrent_hash IS NOT NULL AND CT.swift_torrent_hash <> ''
             AND T.secret is not 1 ORDER BY CT.insert_time DESC LIMIT ?
             
-----
```

kozlovsky:
Fixed in #7628

PR #7628, opened 2023-10-09
Handle the "DatabaseError: database disk image is malformed" error

### LINKED ISSUE #7623: [7.13.0] TypeError: 'NoneType' object is not iterable
Sentry Issue: [TRIBLER-1D8](https://sentry.tribler.org/organizations/tribler/issues/2390/?referrer=github_integration)

```
CoreConnectTimeoutError: Could not connect with the Tribler Core within 120 seconds: ConnectionRefusedError (code 1)
TypeError: 'NoneType' object is not iterable[tribler-core PID:22] 2023-10-02 10:56:53,684 - ERROR <download:322> Download.on_tracker_error_alert(): On tracker error alert: Wallpaper desktop HD (2133x1200) part 12 (udp://tracker.bittor.pw:1337/announce)[<IP>:6881] tracker sent a failure message "" (3)
```

kozlovsky:
Fixes #7623, #7037, #5252, and #1993 (the latter is the pre-ORM version of the same problem)

This PR adds handling to the error "DatabaseError: database disk image is malformed". The error is handled in all databases that Tribler has: "metadata.db", "knowledge.db", "bandwidth.db", and "tags.db" (if it presents during the migration). The database "processes.sqlite" is already handled separately.

The main goal of the PR is to restore normal Tribler functionality by deleting the corrupted database file and allowing Tribler to create a new, fresh database file. Instead of deleting the file, we can rename it as an alternative way to handle the issue. Still, I'm not sure it is worth it, as it will take some valuable disk space, and not many users will try to restore information from the corrupted database file. It is possible to extract *some* data from the corrupted file, but I doubt it is worth it for us to provide such recovery tools.

What the PR does:
1. It adds tracking of the corresponding error to most places where the error can arise
2. In case of an error, the file "<db_filename>.is_corrupted" is created near the corresponding database file.
3. The notification message box about the database corruption fixing is displayed to the user.
4. The Tribler Core is automatically restarting under the hood, so the user does not need to restart Tibler.

### DIFF
--- a/src/tribler/core/components/component.py
+++ b/src/tribler/core/components/component.py
@@ -9,9 +9,9 @@ from typing import Optional, Set, TYPE_CHECKING, Type, Union
 from tribler.core.components.exceptions import ComponentStartupException, MissedDependency, NoneComponent
 from tribler.core.components.reporter.exception_handler import default_core_exception_handler
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.exit_codes import EXITCODE_DATABASE_IS_CORRUPTED
 from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
 from tribler.core.utilities.process_manager import get_global_process_manager
-from tribler.gui.defs import EXITCODE_DATABASE_IS_CORRUPTED
 
 if TYPE_CHECKING:
     from tribler.core.components.session import Session, T
--- a/src/tribler/core/components/component.py
+++ b/src/tribler/core/components/component.py
@@ -9,6 +9,9 @@ from typing import Optional, Set, TYPE_CHECKING, Type, Union
 from tribler.core.components.exceptions import ComponentStartupException, MissedDependency, NoneComponent
 from tribler.core.components.reporter.exception_handler import default_core_exception_handler
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
+from tribler.core.utilities.process_manager import get_global_process_manager
+from tribler.gui.defs import EXITCODE_DATABASE_IS_CORRUPTED
 
 if TYPE_CHECKING:
     from tribler.core.components.session import Session, T
@@ -47,8 +50,14 @@ class Component:
             self._set_component_status(msg, logging.ERROR, exc_info=exc_info)
             self.failed = True
             self.started_event.set()
+
+            if isinstance(e, DatabaseIsCorrupted):
+                process_manager = get_global_process_manager()
+                process_manager.sys_exit(EXITCODE_DATABASE_IS_CORRUPTED, e)
+
             if self.session.failfast:
                 raise e
+
             self.session.set_startup_exception(ComponentStartupException(self, e))
         self.started_event.set()
 
--- a/src/tribler/core/components/component.py
+++ b/src/tribler/core/components/component.py
@@ -9,8 +9,8 @@ from typing import Optional, Set, TYPE_CHECKING, Type, Union
 from tribler.core.components.exceptions import ComponentStartupException, MissedDependency, NoneComponent
 from tribler.core.components.reporter.exception_handler import default_core_exception_handler
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.db_corruption_handling.base import DatabaseIsCorrupted
 from tribler.core.utilities.exit_codes import EXITCODE_DATABASE_IS_CORRUPTED
-from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
 from tribler.core.utilities.process_manager import get_global_process_manager
 
 if TYPE_CHECKING:
@@ -52,8 +52,20 @@ class Component:
             self.started_event.set()
 
             if isinstance(e, DatabaseIsCorrupted):
+                # When the database corruption is detected, we should stop the process immediately.
+                # Tribler GUI will restart the process and the database will be recreated.
+
+                # Usually we wrap an exception into ComponentStartupException, and allow
+                # CoreExceptionHandler.unhandled_error_observer to handle it after all components are started,
+                # but in this case we don't do it. The reason is that handling ComponentStartupException
+                # starts the shutting down of Tribler, and due to some obscure reasons it is not possible to
+                # raise any exception, even SystemExit, from CoreExceptionHandler.unhandled_error_observer when
+                # Tribler is shutting down. It looks like in this case unhandled_error_observer is called from
+                # Task.__del__ method and all exceptions that are raised from __del__ are ignored.
+                # See https://bugs.python.org/issue25489 for similar case.
                 process_manager = get_global_process_manager()
                 process_manager.sys_exit(EXITCODE_DATABASE_IS_CORRUPTED, e)
+                return  # Added for clarity; actually, the code raised SystemExit on the previous line
 
             if self.session.failfast:
                 raise e
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -10,8 +10,8 @@ from typing import Callable, Dict, Optional, Set, Tuple, Type
 from tribler.core.components.exceptions import ComponentStartupException
 from tribler.core.components.reporter.reported_error import ReportedError
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.db_corruption_handling.base import DatabaseIsCorrupted
 from tribler.core.utilities.exit_codes import EXITCODE_DATABASE_IS_CORRUPTED
-from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
 from tribler.core.utilities.process_manager import get_global_process_manager
 
 # There are some errors that we are ignoring.
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -10,7 +10,9 @@ from typing import Callable, Dict, Optional, Set, Tuple, Type
 from tribler.core.components.exceptions import ComponentStartupException
 from tribler.core.components.reporter.reported_error import ReportedError
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
 from tribler.core.utilities.process_manager import get_global_process_manager
+from tribler.gui.defs import EXITCODE_DATABASE_IS_CORRUPTED
 
 # There are some errors that we are ignoring.
 
@@ -93,12 +95,18 @@ class CoreExceptionHandler:
             should_stop = context.pop('should_stop', True)
             message = context.pop('message', 'no message')
             exception = context.pop('exception', None) or self._create_exception_from(message)
-            # Exception
-            text = str(exception)
+
+            self.logger.exception(f'{exception.__class__.__name__}: {exception}', exc_info=exception)
+
+            if isinstance(exception, DatabaseIsCorrupted):
+                process_manager.sys_exit(EXITCODE_DATABASE_IS_CORRUPTED, exception)
+                return  # Added for clarity; actually, the code raised SystemExit on the previous line
+
             if isinstance(exception, ComponentStartupException):
                 self.logger.info('The exception is ComponentStartupException')
                 should_stop = exception.component.tribler_should_stop_on_component_error
                 exception = exception.__cause__
+
             if isinstance(exception, NoCrashException):
                 self.logger.info('The exception is NoCrashException')
                 should_stop = False
@@ -113,7 +121,7 @@ class CoreExceptionHandler:
 
             reported_error = ReportedError(
                 type=exception.__class__.__name__,
-                text=text,
+                text=str(exception),
                 long_text=long_text,
                 context=str(context),
                 event=self.sentry_reporter.event_from_exception(exception) or {},
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -10,9 +10,9 @@ from typing import Callable, Dict, Optional, Set, Tuple, Type
 from tribler.core.components.exceptions import ComponentStartupException
 from tribler.core.components.reporter.reported_error import ReportedError
 from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+from tribler.core.utilities.exit_codes import EXITCODE_DATABASE_IS_CORRUPTED
 from tribler.core.utilities.pony_utils import DatabaseIsCorrupted
 from tribler.core.utilities.process_manager import get_global_process_manager
-from tribler.gui.defs import EXITCODE_DATABASE_IS_CORRUPTED
 
 # There are some errors that we are ignoring.
 
--- a/src/tribler/core/upgrade/upgrade.py
+++ b/src/tribler/core/upgrade/upgrade.py
@@ -12,7 +12,7 @@ from pony.orm import db_session, delete
 from tribler.core.components.bandwidth_accounting.db.database import BandwidthDatabase
 from tribler.core.components.metadata_store.db.orm_bindings.channel_metadata import CHANNEL_DIR_NAME_LENGTH
 from tribler.core.components.metadata_store.db.store import (
-    MetadataStore,
+    CURRENT_DB_VERSION, MetadataStore,
     sql_create_partial_index_channelnode_metadata_type,
     sql_create_partial_index_channelnode_subscribed,
     sql_create_partial_index_torrentstate_last_check,
@@ -134,6 +134,9 @@ class TriblerUpgrader:
     def upgrade_pony_db_14to15(self):
         self._logger.info('Upgrade Pony DB from version 14 to version 15')
         mds_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
+        if not mds_path.exists() or get_db_version(mds_path, CURRENT_DB_VERSION) > 14:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
 
         mds = MetadataStore(mds_path, self.channels_dir, self.primary_key, disable_sync=True,
                             check_tables=False, db_version=14) if mds_path.exists() else None
@@ -146,6 +149,10 @@ class TriblerUpgrader:
         self._logger.info('Upgrade Pony DB from version 13 to version 14')
         mds_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
         tagdb_path = self.state_dir / STATEDIR_DB_DIR / 'tags.db'
+
+        if not mds_path.exists() or get_db_version(mds_path, CURRENT_DB_VERSION) > 13:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
 
         mds = MetadataStore(mds_path, self.channels_dir, self.primary_key, disable_sync=True,
                             check_tables=False, db_version=13) if mds_path.exists() else None
@@ -166,11 +173,14 @@ class TriblerUpgrader:
         self._logger.info('Upgrade Pony DB 12 to 13')
         # We have to create the Metadata Store object because Session-managed Store has not been started yet
         database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
-        if database_path.exists():
-            mds = MetadataStore(database_path, self.channels_dir, self.primary_key,
-                                disable_sync=True, check_tables=False, db_version=12)
-            self.do_upgrade_pony_db_12to13(mds)
-            mds.shutdown()
+        if not database_path.exists() or get_db_version(database_path, CURRENT_DB_VERSION) > 12:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
+
+        mds = MetadataStore(database_path, self.channels_dir, self.primary_key,
+                            disable_sync=True, check_tables=False, db_version=12)
+        self.do_upgrade_pony_db_12to13(mds)
+        mds.shutdown()
 
     def upgrade_pony_db_11to12(self):
         """
@@ -181,8 +191,10 @@ class TriblerUpgrader:
         self._logger.info('Upgrade Pony DB 11 to 12')
         # We have to create the Metadata Store object because Session-managed Store has not been started yet
         database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
-        if not database_path.exists():
-            return
+        if not database_path.exists() or get_db_version(database_path, CURRENT_DB_VERSION) > 11:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
+
         mds = MetadataStore(database_path, self.channels_dir, self.primary_key,
                             disable_sync=True, check_tables=False, db_version=11)
         self.do_upgrade_pony_db_11to12(mds)
@@ -197,8 +209,10 @@ class TriblerUpgrader:
         self._logger.info('Upgrade Pony DB 10 to 11')
         # We have to create the Metadata Store object because Session-managed Store has not been started yet
         database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
-        if not database_path.exists():
-            return
+        if not database_path.exists() or get_db_version(database_path, CURRENT_DB_VERSION) > 10:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
+
         # code of the migration
         mds = MetadataStore(database_path, self.channels_dir, self.primary_key,
                             disable_sync=True, check_tables=False, db_version=10)
@@ -215,8 +229,10 @@ class TriblerUpgrader:
         to_version = 9
 
         database_path = self.state_dir / STATEDIR_DB_DIR / 'bandwidth.db'
-        if not database_path.exists() or get_db_version(database_path) >= 9:
-            return  # No need to update if the database does not exist or is already updated
+        if not database_path.exists() or get_db_version(database_path, BandwidthDatabase.CURRENT_DB_VERSION) > 8:
+            # No need to update if the database does not exist or is already updated
+            return  # pragma: no cover
+
         self._logger.info('bw8->9')
         db = BandwidthDatabase(database_path, self.primary_key.key.pk)
 
@@ -378,7 +394,7 @@ class TriblerUpgrader:
         self._logger.info('Upgrading GigaChannel DB from version 8 to 10')
         database_path = self.state_dir / STATEDIR_DB_DIR / 'metadata.db'
 
-        if not database_path.exists() or get_db_version(database_path) >= 10:
+        if not database_path.exists() or get_db_version(database_path, CURRENT_DB_VERSION) >= 10:
             # Either no old db exists, or the old db version is up to date  - nothing to do
             return
 
--- a/src/tribler/core/upgrade/upgrade.py
+++ b/src/tribler/core/upgrade/upgrade.py
@@ -71,6 +71,19 @@ def cleanup_noncompliant_channel_torrents(state_dir):
 
 
 def catch_db_is_corrupted_exception(upgrader_method):
+    # This decorator applied for TriblerUpgrader methods. It suppresses and remembers the DatabaseIsCorrupted exception.
+    # As a result, if one upgrade method raises an exception, the following upgrade methods are still executed.
+
+    # The reason for this is the following: it is possible that one upgrade methods upgrades a database A, while
+    # the next upgrade method upgrades a database B. If a corruption detected in the database A, the database B still
+    # need to be upgraded. So we want to temporarily suppress DatabaseIsCorrupted exception until all upgrades are
+    # executed.
+
+    # If an upgrade found the database to be corrupted, the database is marked as corrupted. Then, the next upgrade
+    # will rename the corrupted database file (this is handled by the get_db_version call) and immediately return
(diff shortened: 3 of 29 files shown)


drew2a:
> The error is shown to the user in the usual way; it is possible to add a special handling of the error in the GUI.

Based on your description, this shouldn't be presented to the user in the typical manner. Instead, we should display an error to the user stating "Database is malformed," and provide additional details indicating that the database file will be renamed and removed later on. Furthermore, we should prevent these errors from being sent to us.

drew2a (changes requested):
I concur with @synctext's suggestion that we should resolve the issue for the user without needing to restart Tribler.

When a database corruption is detected, we should display a dialog to the user explaining the situation and assuring them that we will automatically address the problem as soon as they press the 'OK' button.

src/tribler/gui/tribler_window.py
```diff
@@ -1254,3 +1254,14 @@ def handle_uri(self, uri):
...
         if self.tribler_started and not self.start_download_dialog_active:
             self.process_uri_request()
+
+    def show_database_is_corrupted_dialog(self, db_file_path: Path):
+        self.dialog = ConfirmationDialog(
+            self,
+            tr("Database is corrupted: restart needed"),
+            tr("The database file found to be corrupted") + ":<br/>"
+            + f"<b>{db_file_path}</b><br/><br/>"
+            + tr("Please restart Tribler to re-create the database file"),
```
drew2a, line 1264: Tribler has to be restarted automatically or, at least closed automatically.

src/tribler/gui/defs.py
```diff
@@ -210,6 +210,11 @@ def get(cls, item, default=None):
 
 
 UPGRADE_CANCELLED_ERROR_TITLE = "Tribler Upgrade cancelled"
+
 NO_DISK_SPACE_ERROR_MESSAGE = "Not enough storage space available. \n" \
                               "Tribler requires at least %s space to continue. \n\n" \
                               "Please free up the required space and re-run Tribler. "
+
+DATABASE_IS_CORRUPTED_ERROR_MESSAGE = "The database file of the previous version found to be corrupted: \n\n" \
```
drew2a, line 218: It seems to be nearly identical with the strings https://github.com/Tribler/tribler/pull/7628/files#diff-df47295d04e11b939d3d1756c77296b153338b035e33de997902ec673ad37043R1261-R1264

Perhaps it is worth extracting it to a common place.

src/tribler/core/utilities/tests/test_pony_utils.py
```diff
@@ -120,3 +124,109 @@ def test_format_warning():
...
+    assert version == default_version
+
+    with sqlite3.connect(db_path) as connection:
+        assert not table_exists(connection.cursor(), 'MiscData')
+
+
+def test_mark_db_as_corrupted_file_does_not_exist(tmp_path: Path):
+    db_path = tmp_path / 'doesnotexist.db'
+    with pytest.raises(RuntimeError, match='^Corrupted database file not found: .*doesnotexist.db$'):
+        _mark_db_as_corrupted(db_path)
```
drew2a, line 232: Could you please add descriptions to the tests?

drew2a (changes requested):
I've reviewed the GUI components and the tests.

I'll proceed with reviewing the core once the Upgrader has been completed.

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +34,87 @@
...
+        with marking_corrupted_db(db_path):
+            with contextlib.closing(sqlite3.connect(db_path)) as connection:
+                with connection:
+                    cursor = connection.cursor()
+                    if not table_exists(cursor, 'MiscData'):
+                        version = None
+                    else:
+                        cursor.execute("SELECT value FROM MiscData WHERE name == 'db_version'")
+                        row = cursor.fetchone()
+                        version = int(row[0]) if row else None
```
drew2a, line 60: Can be simplified:
```suggestion
    handle_db_if_corrupted(db_path)
    version = None
    if db_path.exists():
        with marking_corrupted_db(db_path):
            with contextlib.closing(sqlite3.connect(db_path)) as connection:
                with connection:
                    cursor = connection.cursor()
                    if table_exists(cursor, 'MiscData'):
                        cursor.execute("SELECT value FROM MiscData WHERE name == 'db_version'")
                        row = cursor.fetchone()
                        version = int(row[0]) if row else None
```

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +34,87 @@
...
+
+def _handle_corrupted_db(db_path: Path):
+    if db_path.exists():
+        logger.warning(f'Database file was marked as corrupted, removing it: {db_path}')
+        db_path.unlink()
+
+    marker_path = _get_corrupted_db_marker_path(db_path)
+    if marker_path.exists():
+        logger.warning(f'Removing the corrupted database marker: {marker_path}')
+        marker_path.unlink()
```
drew2a, line 86: Should we handle potential `OSErrors` here, for instance, in cases where `unlink` raises an exception?

src/tribler/core/utilities/pony_utils.py
```diff
@@ -281,27 +378,71 @@ def acquire_lock(self):
...
+        Path(marker_filename).touch()
+
+    def get_pool(self, is_shared_memory_db, filename, create_db=False, **kwargs):
+        if not (is_shared_memory_db or filename == ':memory:'):
+            filename = absolutize_path(filename, frame_depth=cut_traceback_depth+5)  # see the base method for details
+            handle_db_if_corrupted(filename)
+        return TriblerPool(is_shared_memory_db, filename, create_db, **kwargs)
+
+
+class TriblerPool(SQLitePool):
```
drew2a, line 412: Please, add a description to the class 🙏

src/tribler/gui/tribler_window.py
kozlovsky: I completely reimplemented this logic, and now Tribler restarts automatically without requiring action from the user.

src/tribler/gui/defs.py
kozlovsky: Indeed, I extracted it to a single place. Now it is possible because the simplified message box never requires actions from the user.

src/tribler/core/utilities/tests/test_pony_utils.py
kozlovsky: I added descriptions to all tests

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +34,87 @@
...
+                        row = cursor.fetchone()
+                        version = int(row[0]) if row else None
+
+    if version is not None:
+        return version
+
+    if default is not None:
+        return default
+
+    raise RuntimeError(f'The version value is not found in database {db_path}')
```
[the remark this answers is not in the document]
kozlovsky: I'd like to keep the previous version of the code. Otherwise, it may be hard to read it because the reader needs to take into account, for example, that version can theoretically be 0 or None, as well as the default value and the exact logic of the expression `(value or default) is not None` is not obvious, you need to consider how `0 or None` or `None or 0` are calculated in Python, etc. The logic in the PR is simple and does not require any complicated analysis.

src/tribler/core/utilities/pony_utils.py
kozlovsky: Agreed

src/tribler/core/utilities/pony_utils.py
kozlovsky: In a normal situation, we should not have any `OSErrors` here. If we have them, it may be a sign that two processes run in parallel and create a race condition. In that case, we should crash with an error. Otherwise, two parallel Core processes can perform inconsistent changes in the database. If this type of error is possible, we should be brittle on it and not allow two Core processes to run in parallel.

src/tribler/core/utilities/pony_utils.py
```diff
@@ -144,16 +233,42 @@ def _enter(self):
...
+            # The tracking of slow db sessions was not enabled when the db session was started, so we skip analyzing it
+            return
+
+        start_time = info.start_time
+        db_session_duration = time.time() - start_time
+
+        threshold = SLOW_DB_SESSION_DURATION_THRESHOLD if self.duration_threshold is None else self.duration_threshold
+        if db_session_duration > threshold:
+            self._log_warning(db_session_duration, info)
+
```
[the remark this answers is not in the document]
kozlovsky: YEs, I rearranged the order of methods in a separate commit to make the class code more readable.

src/tribler/core/utilities/pony_utils.py
kozlovsky: Added

src/tribler/core/upgrade/upgrade.py
```diff
@@ -68,6 +70,22 @@ def cleanup_noncompliant_channel_torrents(state_dir):
                                  file_path)
 
 
+def catch_db_is_corrupted_exception(upgrader_method):
```
drew2a, line 73: Could you please add a description for this decorator with the notion that it could only apply to TriblerUpgradfer methods?

src/tribler/core/upgrade/upgrade.py
```diff
@@ -68,6 +70,22 @@ def cleanup_noncompliant_channel_torrents(state_dir):
...
+        try:
+            upgrader_method(*args, **kwargs)
+        except DatabaseIsCorrupted as exc:
+            self: TriblerUpgrader = args[0]
+            self._logger.exception(exc)
+
+            if not self._db_is_corrupted_exception:
+                self._db_is_corrupted_exception = exc
+
+    return new_method
```
drew2a, line 86: It is relatively easy to add tests and satisfy Codecov. I can help you with them if you want.

src/tribler/core/upgrade/upgrade.py
```diff
@@ -105,6 +124,9 @@ def run(self):
         self.remove_old_logs()
         self.upgrade_pony_db_14to15()
 
+        if self._db_is_corrupted_exception:
+            raise self._db_is_corrupted_exception  # pylint: disable=raising-bad-type
```
drew2a, line 128: Could you please clarify that this line is roughly equivalent to the `system.exit` (if it is correct).

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +35,86 @@
...
+def handle_db_if_corrupted(db_filename: Union[str, Path]):
+    db_path = Path(db_filename)
+    marker_path = _get_corrupted_db_marker_path(db_path)
+    if marker_path.exists():
+        _handle_corrupted_db(db_path)
+
+
+def _handle_corrupted_db(db_path: Path):
+    if db_path.exists():
+        logger.warning(f'Database file was marked as corrupted, removing it: {db_path}')
```
drew2a, line 79: It is not entirely correct. The following line just renames the DB but does not remove it.

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +35,86 @@
...
+        now = datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
+        db_path.rename(str(db_path) + f'-{now}.corrupted')
```
drew2a, line 81: Could you clarify why it is necessary to use {now} timestamp?

src/tribler/core/utilities/pony_utils.py
```diff
@@ -28,6 +35,86 @@
...
+        logger.warning(f'Removing the corrupted database marker: {marker_path}')
+        marker_path.unlink()
+
+
+def _get_corrupted_db_marker_path(db_filename: Path) -> Path:
+    return Path(str(db_filename) + '.is_corrupted')
+
+
+@contextmanager
+def marking_corrupted_db(db_filename: Union[str, Path]):
```
drew2a, line 94: I'm not fond of the name of this context manager because it might cause confusion:

```python
  with marking_corrupted_db(db_path):
            # do anything with db_path
            ...
```

To me, this block looks like it's 'marking the db as corrupted,' which isn't the case. It actually 'ensures' that the db file will be marked as corrupted in the event of an error.

 Here are five alternative names that might better convey its purpose:
* handle_db_corruption
* safe_db_operation
* error_safe_db_marking
* corruption_protect_db
* ensure_db_integrity

src/tribler/core/utilities/pony_utils.py
```diff
@@ -281,27 +378,78 @@ def acquire_lock(self):
...
             lock_hold_duration = time.time() - acquire_time
             info.lock_hold_total_duration += lock_hold_duration
 
+    def set_transaction_mode(self, connection, cache):
+        with marking_corrupted_db(self.pool.filename):
+            return super().set_transaction_mode(connection, cache)
+
+    def execute(self, cursor, sql, arguments=None, returning_id=False):
+        with marking_corrupted_db(self.pool.filename):
+            return super().execute(cursor, sql, arguments, returning_id)
```
drew2a, line 395: Could you please add a description explaining that for every transaction of `TriblerSQLiteProvider` (used for all Tribler databases), a corrupted database is handled appropriately?

src/tribler/core/utilities/pony_utils.py
```diff
@@ -281,27 +378,78 @@ def acquire_lock(self):
...
+
+    def mark_db_as_malformed(self):
+        filename = self.pool.filename
+        if not Path(filename).exists():
+            raise RuntimeError(f'Corrupted database file not found: {filename!r}')
+
+        marker_filename = filename + '.is_corrupted'
```
drew2a, line 402: I suggest consolidating the logic for the `.is_corrupted` extension (the pattern for the file path) into a single location. This would ensure that future developers only need to make changes in one place, reducing the risk of them modifying it in one location and forgetting to do so in another.

src/tribler/core/utilities/pony_utils.py
```diff
@@ -281,27 +378,78 @@ def acquire_lock(self):
...
+        Path(marker_filename).touch()
+
+    def get_pool(self, is_shared_memory_db, filename, create_db=False, **kwargs):
```
drew2a, line 405: Could you please add a description that clarifies that this method is called on each DB session and therefore the corrupted DB could be handled properly?

drew2a: And I'm curious why it is not enough to call `handle_db_if_corrupted` only in `bind` function?

drew2a (changes requested):

kozlovsky:
I discovered that the initially implemented in this PR corruption-handled logic was incomplete. It was implemented by handling the error in specific places of Tribler code, like opening the database connection or calling the `cursor.execute()` method. To implement this, a complicated code was added to the `TriblerDatabase`, `TriblerSQLiteProvider`, and `TriblerPool` classes that override the standard PonyORM classes for SQLite providers.

As it turns out, the corruption error is unpredictable and can randomly happen in multiple other places, like `cursor.fetchone()`, `cursor.__next__()`, `connection.commit()`, etc. It is unfeasible to handle all these places in the Tribler code explicitly. For that reason, I completely reimplemented the corruption-handling logic.

Now, I added a new `sqlite_utils` module that can replace the native `sqlite3` module completely. It provides improved versions of the `Connection` and `Cursor` classes that handle the database corruption error in all places where the error can arise.

Now, by using the `sqlite_utils` module instead of the base `sqlite3` module, it becomes possible to remove the error-handing logic from many places of Tribler code and make the code more straightforward, as the detection of the database corruption is concentrated in a single place, the `sqlite_util` module. Significant code was removed from the `pony_utils` module, and the `TriblerPool` class is no longer necessary. Now, using context manager in Tribler code to handle the database corruption error is not necessary as the low-level connection and cursor methods handle it.

Also, some Core-GUI interaction problems were fixed, and GUI now correctly re-connects to the restarted Core in all cases.

src/tribler/core/components/reporter/exception_handler.py
```diff
@@ -93,12 +95,18 @@ def unhandled_error_observer(self, _, context):
             should_stop = context.pop('should_stop', True)
             message = context.pop('message', 'no message')
             exception = context.pop('exception', None) or self._create_exception_from(message)
-            # Exception
-            text = str(exception)
+
+            self.logger.exception(f'{exception.__class__.__name__}: {exception}', exc_info=exception)
```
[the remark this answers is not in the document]
kozlovsky: I agree it is not necessary, but I'd prefer to have it this way, as it looks slightly more informative.

src/tribler/core/upgrade/upgrade.py
kozlovsky: I added a comment to it:

```python
    # This decorator applied for TriblerUpgrader methods. It suppresses and remembers the DatabaseIsCorrupted exception.
    # As a result, if one upgrade method raises an exception, the following upgrade methods are still executed.
    #
    # The reason for this is the following: it is possible that one upgrade method upgrades database A
    # while the following upgrade method upgrades database B. If a corruption is detected in the database A,
    # the database B still needs to be upgraded. So, we want to temporarily suppress the DatabaseIsCorrupted exception
    # until all upgrades are executed.
    #
    # If an upgrade finds the database to be corrupted, the database is marked as corrupted. Then, the next upgrade
    # will rename the corrupted database file (the get_db_version call handles this) and immediately return because
    # there is no database to upgrade. So, if one upgrade function detects database corruption, all the following
    # upgrade functions for this specific database will skip the actual upgrade. As a result, a new database with
    # the current DB version will be created on the Tribler Core start.
```

src/tribler/core/upgrade/upgrade.py
kozlovsky: Done

src/tribler/core/utilities/pony_utils.py
kozlovsky: Fixed, now the corrupted database is removed completely and not just renamed.

src/tribler/core/utilities/pony_utils.py
kozlovsky: The database us removed now and not just renamed, so this code was removed

src/tribler/core/utilities/pony_utils.py
kozlovsky: The logic was reworked entirely, and this context manager is no longer necessary and was removed.

src/tribler/core/utilities/pony_utils.py
kozlovsky: The logic was reworked entirely; these methods were no longer necessary and removed.

src/tribler/core/utilities/pony_utils.py
kozlovsky: Done (except tests)

src/tribler/core/utilities/pony_utils.py
kozlovsky: The logic was reworked entirely; the `get_pool` method was no longer necessary and is now removed.

kozlovsky: >  And I'm curious why it is not enough to call handle_db_if_corrupted only in bind function?

Now, the `handle_db_if_corrupted` function is not called in the `get_pool` method, as the method was removed.

`handle_db_if_corrupted` is still used in for places:
* in bind method to fix the corrupted database before connecting to it from ORM.
* in the `get_db_version` method that is called separately from ORM
* in `metadata_store/db/store.py` and in `bandwidth_accounting/db/store.py` because the current logic of these modules requires fixing the corrupted database before determining the value of the `create_db` flag.

src/tribler/gui/tribler_window.py
```diff
@@ -545,7 +545,6 @@ def tray_show_message(self, title, message):
     def on_core_connected(self, version):
         if self.core_connected:
             self._logger.warning("Received duplicate Tribler Core connected event")
-            return
```
kozlovsky, line 548: Removing `return` is necessary to successfully re-connect to the new Core when the crash of the previous Core happened right after the GUI was already connected to the Core but before the settings were successfully received.

src/tribler/core/components/component.py
```diff
@@ -47,8 +50,16 @@ async def start(self):
             self._set_component_status(msg, logging.ERROR, exc_info=exc_info)
             self.failed = True
             self.started_event.set()
+
+            if isinstance(e, DatabaseIsCorrupted):
+                # When the database corruption is detected, we should stop the process immediately.
+                # Tribler GUI will restart the process and the database will be recreated.
+                process_manager = get_global_process_manager()
+                process_manager.sys_exit(EXITCODE_DATABASE_IS_CORRUPTED, e)
```
drew2a, line 58: As it is essential logic of the PR it should be tested.

drew2a, line 58: Could be removed?

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DatabaseIsCorrupted(Exception):
+    pass
+
+
+@contextmanager
+def _handling_malformed_db_error(db_filepath: Path):
```
drew2a, line 36: Private methods should be placed after public ones.

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+        return sqlite3.connect(db_filename, **kwargs)
+
+
+class CursorBase(sqlite3.Cursor):
+    # This class simplifies testing of the Cursor class by allowing mocking of base class methods.
+    # Direct mocking of sqlite3.Cursor methods is not possible because they are C functions.
+    pass
+
+
+class Cursor(CursorBase):
```
drew2a, line 94: Could be spit to separate files.

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+        logger.warning(f'Database file was marked as corrupted, removing it: {db_path}')
+        db_path.unlink()
+
+    marker_path = get_corrupted_db_marker_path(db_path)
+    if marker_path.exists():
+        logger.warning(f'Removing the corrupted database marker: {marker_path}')
+        marker_path.unlink()
+
+
+def connect(db_filename: str, **kwargs) -> sqlite3.Connection:
```
drew2a, line 81: Up

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+    # Creates a new `*.is_corrupted` marker file alongside the database file
+    marker_path = get_corrupted_db_marker_path(db_filepath)
+    marker_path.touch()
+
+
+def get_corrupted_db_marker_path(db_filepath: Path) -> Path:
+    return Path(str(db_filepath) + '.is_corrupted')
+
+
+def handle_db_if_corrupted(db_filename: Union[str, Path]):
```
drew2a, line 61: Up

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+    # Handles the case when the database is corrupted in all relevant methods.
+    def __init__(self, db_filepath: str, *args, **kwargs):
+        super().__init__(db_filepath, *args, **kwargs)
+        self._db_filepath = Path(db_filepath)
+
+    def commit(self) -> None:
+        with _handling_malformed_db_error(self._db_filepath):
+            return super().commit()
+
+    def cursor(self, factory=Cursor) -> Cursor:  # pylint: disable=useless-parent-delegation
```
drew2a, line 159: Think about Cursor

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+        return super().cursor(factory)
+
+    def execute(self, *args, **kwargs) -> Cursor:
+        return self.cursor().execute(*args, **kwargs)
+
+    def executemany(self, *args, **kwargs) -> Cursor:
+        return self.cursor().executemany(*args, **kwargs)
+
+    def executescript(self, *args, **kwargs) -> Cursor:
+        return self.cursor().executescript(*args, **kwargs)
```
drew2a, line 169: Could be simplified

src/tribler/core/utilities/sqlite_utils.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def iterdump(self) -> Generator[str, None, None]:
+        # Not implemented because it is not used in Tribler.
+        # Can be added later with an iterator class that handles the malformed db error during the iteration
+        raise NotImplementedError
+
+    def backup(self, *args, **kwargs):
```
drew2a, line 176: Could be generalised

src/tribler/core/utilities/tests/test_sqlite.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def test_handling_malformed_db_error__malformed_error(db_filepath):
+    with pytest.raises(DatabaseIsCorrupted):
+        with _handling_malformed_db_error(db_filepath):
+            raise malformed_error
+    assert Path(str(db_filepath) + '.is_corrupted').exists()
+
+
+def test_handling_malformed_db_error__other_error(db_filepath):
```
drew2a, line 46: Comment

src/tribler/gui/event_request_manager.py
```diff
@@ -130,11 +130,10 @@ def on_error(self, error: int, reschedule_on_err: bool):
...
         if self.receiving_data:
-            # We are performing reconnect on the initial connection error only.
-            # In the future, if we consider it useful, we can immediately call here
-            # `self.reconnect(reschedule_on_err=False)`
-            # and raise an exception if it fails to reconnect
-            raise CoreConnectionError('The connection to the Tribler Core was lost')
+            # Most probably Core is crashed. If CoreManager decides to restart the core,
+            # it will also call event_manager.connect_to_core()
+            self._logger.error('The connection to the Tribler Core was lost')
+            return
```
drew2a, line 136: Check what is going on when core is killed by `kill`

drew2a (changes requested):

drew2a:
Discussed the requested changes with @kozlovsky in person.

src/tribler/core/components/component.py
kozlovsky: Added the test

kozlovsky: Currently, it is necessary. The actual reason is a bit complex to explain:

An error that is raised during the component's startup is wrapped in `ComponentStartupException`. Then, it is handled in a special way:

```python
    def _reraise_startup_exception_in_separate_task(self):
        self.logger.info('Reraise startup exception in separate task')

        async def exception_reraiser():
            self.logger.info('Exception reraiser')

            e = self._startup_exception
            if isinstance(e, ComponentStartupException) and e.component.tribler_should_stop_on_component_error:
                self.logger.info('Shutdown with exit code 1')
                self.exit_code = 1
                self.shutdown_event.set()

            # the exception should be intercepted by event loop exception handler
            self.logger.info(f'Reraise startup exception: {self._startup_exception}')
            raise self._startup_exception

        self.async_group.add_task(exception_reraiser())
```

This logic triggers a weird asyncio behavior described in https://togithub.com/python/cpython/issues/69675. In short, if `CoreExceptionHandler.unhandled_error_observer` is called during the Tribler Core shutdown, the `SystemExit` error raised from `unhandled_error_observer` can be suppressed and ignored by asyncio. Apparently, `unhandled_error_observer` can be called inside the `Task.__del__()` method that ignores all re-raised exceptions, including the `SystemExit` error. As a result, Tribler Core finishes with exit code 1 (instead of code 99 dedicated to the database corruption error), and Tribler GUI does not restart the core.

Fixing and refactoring the logic of `unhandled_error_observer` can be a tricky and complicated task that is outside of the current PR scope. On the other side, raising `SystemExit` directly from the component's method works well and provides the desired results.

Also, raising `SystemExit` in the component allows TriblerCore to restart without waiting first until all components start.

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: Done

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: I replaced `sqlite_utils` with `db_corruption_handling.base` and `db_corruption_handling.sqlite_replacement`. Now, the module `db_corruption_handling.sqlite_replacement` is sufficiently short to not split it further.

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: Done

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: Done

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: I removed it from the arguments to make the code a bit clearer.

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: Done

src/tribler/core/utilities/sqlite_utils.py
kozlovsky: Done

src/tribler/core/utilities/tests/test_sqlite.py
kozlovsky: Added

src/tribler/gui/event_request_manager.py
kozlovsky: The Tribler GUI correctly shows the error in this case

src/tribler/core/components/component.py
drew2a: Could you please put this explanation into the code?

drew2a (approved):
Looks good. Please, add a comment about necessity of `process_manager.sys_exit()` in component.py.

PR #7634, opened 2023-10-16
Remove pytest-rerunfailures and its dependents

xoriole:
This PR is the follow up for https://github.com/Tribler/tribler/pull/7582 which does not include the clean up the depdencies.

The dependencies are:
1. Removing the package from requirements
2. Removing the dependent package. In this case, `pytest-sentry` depends on it. This further implies removing `PYTEST_SENTRY_DSN` environment variable and secret used GitHub workflows.


xoriole:
With the merge of https://github.com/Tribler/tribler/pull/7708, this PR is redundant.

PR #7637, opened 2023-10-19
Fix the detection of running processes by including process UID to the check

### LINKED ISSUE #7603: [Mysterious error][7.13.0] CoreConnectionError: "The connection to the Tribler Core was lost" in event_request_manager.on_error

kozlovsky:
This PR should fix #7603 by associating a unique UID value with each process. It fixes the problem of non-unique PID values.
In addition, each process periodically (1 time per second) updates its `last_alive_at` time in the database. When the `last_alive_at` time is not updated long enough (for 5 seconds), the process is considered not running.

Now, the following situations should be handled correctly:
1. In Flatpak, when the new GUI process can't find a currently running primary GUI process, it understands that the `last_alive_at`  was recently updated and that the primary process is running in a separate virtual environment.
2. When a GUI process creates a Core process and awaits for the `api_port` value, it understands that the correct Core process should specify the correct UID of the current GUI process. That prevents connecting with the wrong Core process.
3. When Core and GUI processes display the recent process list, they are correctly linked to each other.

### DIFF
--- a/src/run_tribler_headless.py
+++ b/src/run_tribler_headless.py
@@ -81,6 +81,7 @@ class TriblerService:
         current_process = TriblerProcess.current_process(ProcessKind.Core)
         self.process_manager = ProcessManager(root_state_dir, current_process)
         set_global_process_manager(self.process_manager)
+        current_process.start_updating_thread()
 
         if not self.process_manager.current_process.become_primary():
             msg = 'Another Core process is already running'
--- a/src/tribler/core/components/gui_process_watcher/gui_process_watcher.py
+++ b/src/tribler/core/components/gui_process_watcher/gui_process_watcher.py
@@ -5,6 +5,7 @@ from typing import Callable, Optional
 import psutil
 from ipv8.taskmanager import TaskManager
 
+GUI_UID_ENV_KEY = 'TRIBLER_GUI_UID'
 GUI_PID_ENV_KEY = 'TRIBLER_GUI_PID'
 CHECK_INTERVAL = 10
 
@@ -50,6 +51,16 @@ class GuiProcessWatcher(TaskManager):
                 return int(pid)
             except ValueError:
                 logger.warning(f'Cannot parse {GUI_PID_ENV_KEY} environment variable: {pid}')
+        return None
+
+    @staticmethod
+    def get_gui_uid() -> Optional[int]:
+        uid = os.environ.get(GUI_UID_ENV_KEY, None)
+        if uid:
+            try:
+                return int(uid)
+            except ValueError:
+                logger.warning(f'Cannot parse {GUI_UID_ENV_KEY} environment variable: {uid}')
         return None
 
     @classmethod
--- a/src/tribler/core/components/gui_process_watcher/tests/test_gui_process_watcher.py
+++ b/src/tribler/core/components/gui_process_watcher/tests/test_gui_process_watcher.py
@@ -4,7 +4,8 @@ from unittest.mock import Mock, patch
 import psutil
 import pytest
 
-from tribler.core.components.gui_process_watcher.gui_process_watcher import GUI_PID_ENV_KEY, GuiProcessNotRunning, \
+from tribler.core.components.gui_process_watcher.gui_process_watcher import GUI_PID_ENV_KEY, GUI_UID_ENV_KEY, \
+    GuiProcessNotRunning, \
     GuiProcessWatcher
 
 
@@ -31,6 +32,19 @@ def test_get_gui_pid(caplog):
 
     with patch.dict(os.environ, {GUI_PID_ENV_KEY: '123'}):
         assert GuiProcessWatcher.get_gui_pid() == 123
+
+
+def test_get_gui_uid(caplog):
+    with patch.dict(os.environ, {GUI_UID_ENV_KEY: ''}):
+        assert GuiProcessWatcher.get_gui_pid() is None
+
+    with patch.dict(os.environ, {GUI_UID_ENV_KEY: 'abc'}):
+        caplog.clear()
+        assert GuiProcessWatcher.get_gui_uid() is None
+        assert caplog.records[-1].message == 'Cannot parse TRIBLER_GUI_UID environment variable: abc'
+
+    with patch.dict(os.environ, {GUI_UID_ENV_KEY: '123'}):
+        assert GuiProcessWatcher.get_gui_uid() == 123
 
 
 def test_get_gui_process():
--- a/src/tribler/core/start_core.py
+++ b/src/tribler/core/start_core.py
@@ -186,8 +186,9 @@ def run_tribler_core_session(api_port: Optional[int], api_key: str,
 def run_core(api_port: Optional[int], api_key: Optional[str], root_state_dir, parsed_args):
     logger.info(f"Running Core in {'gui_test_mode' if parsed_args.gui_test_mode else 'normal mode'}")
 
+    gui_uid = GuiProcessWatcher.get_gui_uid()
     gui_pid = GuiProcessWatcher.get_gui_pid()
-    current_process = TriblerProcess.current_process(ProcessKind.Core, creator_pid=gui_pid)
+    current_process = TriblerProcess.current_process(kind=ProcessKind.Core, creator_uid=gui_uid, creator_pid=gui_pid)
     process_manager = ProcessManager(root_state_dir, current_process)
     set_global_process_manager(process_manager)
     current_process_is_primary = process_manager.current_process.become_primary()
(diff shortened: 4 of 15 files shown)


drew2a:
I've been added as a reviewer for this PR. I noticed [another PR](https://github.com/Tribler/tribler/pull/7621) addresses the same issue. Have you (@kozlovsky, @xoriole) reached a consensus on which approach we should take?

drew2a:
Just a quick note: from my perspective, this PR is unlikely to resolve https://github.com/Tribler/tribler/issues/7603, as it addresses a very specific flatpak problem, which is just a small part of the original issue.

The platform distribution for the original issue is as follows:

image

kozlovsky:
@drew2a @xoriole 

First, I agree that this is not a complete fix for #7603. Essentially, this Sentry issue combines two unrelated bugs that look similar to Sentry.
1. A Flatpak-specific bug due to simultaneous processes having the same PID.
2. Another bug with an unclear origin.

Second, I disagree that the Flatpak issue is only a small part of the original issue. I manually traversed through all currently available events of this sentry issue and counted how many events are related to Flatpak: https://sentry.tribler.org/organizations/tribler/issues/2003/events/latest/?project=7&referrer=latest-event

In the available reports, I counted the following:
* 35 error reports from Flatpak environments;
* 49 error reports from non-Flatpak environments

So, the Flatpak-related issue corresponds to 41.7% of all recorded reports. It is a pretty significant number.

By addressing the Flatpak problem in 7.13.1 with this PR, we can demystify the reports, making the second bug easier to isolate and investigate.

Regarding another PR #7621 my opinion is:
* The "quick fix" in #7621 does not look correct to me, as it introduces a race condition similar to what we already encountered before. It fixes the Flatpack-related bug but introduces a different bug due to that race condition.
* It is theoretically possible to completely remove the lock file and rely only on QLocalSocket. But it is a pretty significant change that looks not like a simple fix but like a major rewriting that requires pretty significant efforts to implement. For that reason, it is hard to do it in a bugfix release. We should expect that this major rewrite can introduce completely new bugs and should have time to fix them in several RC releases of the following major release.

For that reason, I suggest the following plan:
* In 7.13.1, we can solve the Flatpack issue using the fix from the current PR
* It should allow us to investigate and fix the second (non-Flatpak) issue in the later release.
* In the next major release, we can try switching to a new purely QLocalSocket approach for process coordination and stop relying on file locks. However, implementing it is a non-trivial task.

drew2a:
I haven't done a deep dive into the PR yet, but my initial impression is one of concern due to its complexity. I believe it was already intricate to the point where, I suspect, no one besides @kozlovsky fully grasped its logic. Now, it appears even more complicated with the addition of a new thread and the `GUI_UID_ENV_KEY` environment variable. The underlying logic has also grown in complexity.

When I approved the [PR](https://github.com/Tribler/tribler/pull/7212), I assumed that despite its complexity, it would address the "two instance" problem, which remains unresolved. 

I'd suggest considering a rollback of the entire SQL lock implementation in favor of a slightly tweaked previous approach.

Below is a simplified example of a potential file lock solution. It's incredibly straightforward and would protect us from race conditions, which was the original intention behind implementing https://github.com/Tribler/tribler/pull/7212.

```python
import asyncio

from filelock import FileLock, Timeout

lock = FileLock("gui.lock", timeout=0)


async def run_tribler_gui():
    try:
        with lock:
            print('Tribler is running')
            await asyncio.sleep(10)  # simulation of a normal tribler run
    except Timeout:
        print("Tribler is already running. Closing the Tribler...")
        exit(1)


asyncio.run(run_tribler_gui())

```

Ref: https://py-filelock.readthedocs.io/en/latest/index.html

drew2a:
The simplified core part could be like:

```python
import asyncio

from filelock import FileLock, Timeout

lock = FileLock("core.lock", timeout=0)
port_file_name = 'core.port'


async def run_tribler_core():
    try:
        with lock:
            print('Tribler core is running')
            print('Creating REST server..')
            # assume it has been created on port 1234
            port = 1234
            with open(port_file_name, "w+") as f:
                f.write(str(port))

            await asyncio.sleep(10)  # simulation of a normal tribler run

    except Timeout:
        print("Tribler core is already running. Closing the Tribler core...")
        exit(1)


asyncio.run(run_tribler_core())
```

kozlovsky:
Yes, I agree that it is possible for us to use the [filelock](https://github.com/tox-dev/filelock) library. I considered it before implementing SQLite-based lock implementation, but at that moment, the `filelock` library [was considered unmaintained](https://togithub.com/tox-dev/filelock/issues/84) and had multiple open issues with different problems on different OSes. Since then, the library has been adopted by virtualenv maintainers, and multiple issues have been fixed. So it is probably safe now to use it.

With this, I doubt that the full solution will be as simple as in the suggested example:

1. SQLite is more widespread on different systems compared to `filelock`, and its lock implementation can be more robust.
2. Some possible bugs [are still present in the `filelock` library](https://github.com/tox-dev/filelock/issues/56), so by replacing SQLite-based locks with filelock locks, we can encounter new error cases.
3. One of the reasons for #7212 was that not only do Core processes require proper locking, but GUI processes need this as well. Originally, Tribler relied on QLocalSocket to ensure the proper "locking" of the GUI processes, but it turned out that QLocalSocket implementation has inherent race conditions that were impossible to solve without some additional mechanism. This means that with `filelock`, you most probably need not one, but several file locks, one for GUI processes and another for Core processes. You need to keep different data in lock-protected files for Core and GUI, and it can quickly become more complicated compared to keeping all information in a single file.
4. As the `filelock` user [reports](https://stackoverflow.com/questions/489861/locking-a-file-in-python), it is possible for a terminated process to leave the lock file on disk, and in that case, it can be necessary to manually delete the lock file:

    > As noted by a comment at the blog post, this solution isn't "perfect", in that it's possible for the program to terminate in such a way that the lock is left in place and you have to manually delete the lock before the file becomes accessible again.

    If in Tribler we encounter this type of problem with `filelock`, it can greatly complicate the startup logic.
    
5. The current implementation not only solves the lock problem but also maintains the list of currently running processes and the history of the last processes, which allows us to detect tricky bugs. It may be inconvenient to lose this type of valuable information.

synctext:
As discussed at length: lets try to keep it simple; use a file-lock and new dependency.

Always a good idea to see if we can keep it simple, as long as we are 90% certain it will work. So a 10% chance of introducing _new_ errors is fine, when we might keep-things-simple :thinking: 
@drew2a @kozlovsky @egbertbouman @xoriole Please review the upcoming code. Goal is 2 days of simple code.

kozlovsky:
Closed in favor of #7660

PR #7645, opened 2023-10-24
Updated IPv8 while keeping Community kwargs

qstokkink:
This PR is the second attempt to update the IPv8 version to `2.12` in Tribler.

The previous PR to update the version, #7620, was rejected because communities could not `__init__` with kwargs. In this PR, I introduce adapters to allow communities to `__init__` with `args` and `kwargs`, like in the previous IPv8 version (`2.11`).

 - The new class `TriblerMockIPv8` provides the old interface using the new `MockIPv8`.
 - The new class `TriblerTestBase` provides the old interface using the new `TestBase`.
 - An update to `TriblerCommunity` provides the old interface using the new `Community`.

_Disclaimer: I still prefer the implementation of #7620 over this._

### DIFF
--- a/src/tribler/core/components/ipv8/tribler_community.py
+++ b/src/tribler/core/components/ipv8/tribler_community.py
@@ -1,6 +1,14 @@
-from ipv8.community import Community
+from ipv8.community import Community, DEFAULT_MAX_PEERS
 
 from tribler.core.config.tribler_config_section import TriblerConfigSection
+
+def args_kwargs_to_community_settings(settings_class, args, kwargs):
+    return settings_class(my_peer=args[0] if len(args) > 0 else kwargs.pop("my_peer"),
+                          endpoint=args[1] if len(args) > 1 else kwargs.pop("endpoint"),
+                          network=args[2] if len(args) > 2 else kwargs.pop("network"),
+                          max_peers=args[4] if len(args) > 3 else kwargs.pop("max_peers", DEFAULT_MAX_PEERS),
+                          anonymize=args[5] if len(args) > 4 else kwargs.pop("anonymize", True),
+                          **kwargs)
 
 
 class TriblerCommunity(Community):
@@ -8,6 +16,7 @@ class TriblerCommunity(Community):
     """
 
     def __init__(self, *args, settings: TriblerConfigSection = None, **kwargs):
-        super().__init__(*args, **kwargs)
+        community_settings = args_kwargs_to_community_settings(self.settings_class, args, kwargs)
+        super().__init__(community_settings)
         self.settings = settings
         self.logger.info(f'Init. Settings: {settings}.')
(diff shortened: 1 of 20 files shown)


src/tribler/core/components/ipv8/tribler_community.py
```diff
@@ -1,13 +1,22 @@
...
 
 from tribler.core.config.tribler_config_section import TriblerConfigSection
 
+def args_kwargs_to_community_settings(settings_class, args, kwargs):
+    return settings_class(my_peer=args[0] if len(args) > 0 else kwargs.pop("my_peer"),
+                          endpoint=args[1] if len(args) > 1 else kwargs.pop("endpoint"),
+                          network=args[2] if len(args) > 2 else kwargs.pop("network"),
+                          max_peers=args[4] if len(args) > 3 else kwargs.pop("max_peers", DEFAULT_MAX_PEERS),
+                          anonymize=args[5] if len(args) > 4 else kwargs.pop("anonymize", True),
+                          **kwargs)
```
kozlovsky, line 11: I was wondering if it is possible to have a type-safe equivalent of `args_kwargs_to_community_settings` that specifies the types of its arguments, and came up with the following:

```python
community_class = TypeVar('community_class', bound=Community)

_missed = object()


def create_settings(cls: Type[community_class],
                    my_peer: Peer = _missed,
                    endpoint: Endpoint = _missed,
                    network: Network = _missed,
                    max_peers: int = _missed,
                    anonymize: bool = _missed,
                    **kwargs
                    ) -> community_class.settings_class:
    params = dict(my_peer=my_peer, endpoint=endpoint, network=network,
                  max_peers=max_peers, anonymize=anonymize)
    params.update(kwargs)
    params = {key: value for key, value in params.items() if value is not _missed}
    return cls.settings_class(**params)


def create_community(cls: Type[community_class],
                     my_peer: Peer = _missed,
                     endpoint: Endpoint = _missed,
                     network: Network = _missed,
                     max_peers: int = _missed,
                     anonymize: bool = _missed,
                     **kwargs
                     ) -> community_class:
    return cls(create_settings(cls, my_peer, endpoint, network, max_peers, anonymize, **kwargs))
```

With this, instead of
```python
community = DiscoveryCommunity(args_kwargs_to_community_settings(DiscoveryCommunity.settings_class,
                                                                 [self.peer, ipv8.endpoint, ipv8.network],
                                                                 {"max_peers": 100}))
```
It is possible to write
```python
community = create_community(DiscoveryCommunity, self.peer, ipv8.endpoint, ipv8.network, max_peers=100)
```
and argument types will be correctly checked.

But as `args_kwargs_to_community_settings` is used in only a few places, I'm not advocating for a type-safe version; the current PR version looks good enough to me

kozlovsky (approved):

src/tribler/core/components/ipv8/tribler_community.py
qstokkink: @kozlovsky That seems like a nice improvement. Would you be O.K. with opening a PR for your proposal after this PR has been merged?

kozlovsky: @qstokkink Sure!

qstokkink: @kozlovsky thanks! I'll merge this PR in its current state then.

PR #7650, opened 2023-10-26
Set the limit for int values of the libtorrent config

### LINKED ISSUE #5713: [7.12.1] OverflowError: Python int too large to convert to C long (in tribler_core.modules.libtorrent.download.on_add_torrent_alert)
Stack:
```python
[
  File "src\tribler-gui\tribler_gui\event_request_manager.py", line 116, in on_read_data, 
RuntimeError: Python int too large to convert to C long, 
--LONG TEXT--, 
  File "lib\asyncio\events.py", line 81, in _run, 
  File "ipv8\taskmanager.py", line 131, in done_cb, 
  File "ipv8\taskmanager.py", line 18, in interval_runner, 
  File "ipv8\util.py", line 55, in call_async, 
  File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 464, in _task_process_alerts, 
  File "src\tribler-core\tribler_core\modules\libtorrent\download_manager.py", line 350, in process_alert, 
  File "src\tribler-core\tribler_core\modules\libtorrent\download.py", line 220, in process_alert, 
  File "src\tribler-core\tribler_core\modules\libtorrent\download.py", line 181, in on_add_torrent_alert, 
OverflowError: Python int too large to convert to C long, 
--CONTEXT--, 
{'message': "Exception in callback TaskManager.register_task.<locals>.done_cb(<Task finishe...t to C long')>) at ipv8\\taskmanager.py:128", 'exception': OverflowError('Python int too large to convert to C long'), 'handle': <Handle TaskManager.register_task.<locals>.done_cb(<Task finishe...t to C long')>) at ipv8\taskmanager.py:128>}
]
```

Issue: http://131.180.27.167:9000/organizations/tribler/issues/41/?project=2&query=is%3Aunresolved+version%3A7.5.4&statsPeriod=14d

drew2a:
This PR fixes #5713 by setting the limit for int fields of the libtorrent config.

Ref: https://libtorrent.org/reference-Torrent_Handle.html#max-connections-set-max-connections

### DIFF
--- a/src/tribler/core/components/libtorrent/settings.py
+++ b/src/tribler/core/components/libtorrent/settings.py
@@ -8,7 +8,7 @@ from tribler.core.utilities.network_utils import NetworkUtils
 from tribler.core.utilities.osutils import get_home_dir
 from tribler.core.utilities.path_util import Path
 
-INT_MAX = 2147483647  # max int value for C
+INT32_MAX = 2147483647  # max int value for C
 
 TRIBLER_DOWNLOADS_DEFAULT = "TriblerDownloads"
 
@@ -48,7 +48,7 @@ class LibtorrentSettings(TriblerConfigSection):
     def __setattr__(self, key, value):
         """Override __setattr__ to limit the max int value to `INT_MAX` (the max int value for C)"""
         if isinstance(value, int):
-            value = min(INT_MAX, value)
+            value = min(INT32_MAX, value)
         super().__setattr__(key, value)
 
 
--- a/src/tribler/core/components/libtorrent/settings.py
+++ b/src/tribler/core/components/libtorrent/settings.py
@@ -8,6 +8,8 @@ from tribler.core.utilities.network_utils import NetworkUtils
 from tribler.core.utilities.osutils import get_home_dir
 from tribler.core.utilities.path_util import Path
 
+INT_MAX = 2147483647  # max int value for C
+
 TRIBLER_DOWNLOADS_DEFAULT = "TriblerDownloads"
 
 
@@ -42,6 +44,12 @@ class LibtorrentSettings(TriblerConfigSection):
     def validate_proxy_type(cls, v):
         assert v is None or 0 <= v <= 5, 'Proxy type must be in range [0..5]'
         return v
+
+    def __setattr__(self, key, value):
+        """Override __setattr__ to limit the max int value to `INT_MAX` (the max int value for C)"""
+        if isinstance(value, int):
+            value = min(INT_MAX, value)
+        super().__setattr__(key, value)
 
 
 class SeedingMode(str, Enum):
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/libtorrent/settings.py
```diff
@@ -8,6 +8,8 @@
 from tribler.core.utilities.osutils import get_home_dir
 from tribler.core.utilities.path_util import Path
 
+INT_MAX = 2147483647  # max int value for C
```
kozlovsky, line 11: For clarity and consistency, I suggest naming the constant `MAX_INT32`. This name explicitly indicates its role and the bit-depth, reducing potential ambiguities. Using the expression `2**31 - 1` instead of the direct numeric value further improves code readability and avoids the use of magic numbers.
```suggestion
MAX_INT32 = 2**31 - 1
```

Additionally, I noticed that `MAX_INT32` is [already defined](https://github.com/Tribler/tribler/blob/a2efecf9f957294ea74ebc5e388ed9be5d0f316a/src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py#L34) elsewhere in the Tribler code. We could consider centralizing it in the `utils` for consistency and reusability.

src/tribler/core/components/libtorrent/settings.py
```diff
@@ -43,6 +45,12 @@ def validate_proxy_type(cls, v):
         assert v is None or 0 <= v <= 5, 'Proxy type must be in range [0..5]'
         return v
 
+    def __setattr__(self, key, value):
+        """Override __setattr__ to limit the max int value to `INT_MAX` (the max int value for C)"""
+        if isinstance(value, int):
+            value = min(INT_MAX, value)
+        super().__setattr__(key, value)
```
kozlovsky, line 52: On one side, it may be good to have such "bullet-proof" protection from incorrect values that cannot be circumvented. However, it seems we conventionally handle such validations in `settingspage.py`. The validation for the maximum number of connections is [handled as shown here](https://github.com/Tribler/tribler/blob/e5b014e31e9e3f90d98c979d3e07cd9ed0653b61/src/tribler/gui/widgets/settingspage.py#L427):

```python
try:
    max_conn_download = int(self.window().max_connections_download_input.text())
except ValueError:
    ConfirmationDialog.show_error(
        self.window(),
        tr("Invalid number of connections"),
        tr(
            "You've entered an invalid format for the maximum number of connections. "
            "Please enter a whole number."
        ),
    )
    return
if max_conn_download == 0:
    max_conn_download = -1
settings_data['libtorrent']['max_connections_download'] = max_conn_download
```

We can extend the validation logic here to also check that the number of connections is in range.

By aligning with this existing pattern, we can maintain consistency and provide immediate feedback to users on invalid inputs rather than silently modifying the values. I'd suggest adding the new check in the same location for a cohesive experience.

kozlovsky (changes requested):

src/tribler/core/components/libtorrent/settings.py
drew2a: I believe the original solution is preferable over the suggested one, given that we have two potential sources for libtorrent settings:
1. Config
2. GUI

The initial approach encompasses both sources. Additionally, it seems more prudent to account for a broader range of cases with this check, rather than risking a Tribler crash due to improper input.

src/tribler/core/components/libtorrent/settings.py
drew2a: It's a matter of preference whether we use 2147483647 or 2**31 - 1 (both are essentially magic numbers).

Personally, I find 2147483647 more recognizable, as I've often seen it in various documentation.

kozlovsky (approved):

PR #7651, opened 2023-10-27
Fix MAX_INT32 value (the previous value was actually for MAX_UINT16)

kozlovsky:
Upon reviewing #7650, I noticed an inconsistency with the value of [`MAX_INT32`](https://github.com/Tribler/tribler/blob/a2efecf9f957294ea74ebc5e388ed9be5d0f316a/src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py#L34). Currently, it's defined as `2 ** 16 - 1`, which is appropriate for `MAX_UINT16`. The correct value for `MAX_INT32` should be `2 ** 31 - 1`. This [definition was introduced in 2015](https://github.com/Tribler/tribler/commit/4c90c9945b0aa6834bab0e55697058c7c551a9ae) and appears to have been a minor oversight.

In this PR, I've corrected the `MAX_INT32` value. However, to ensure there are no unintended side effects, it's important to verify the contexts where `MAX_INT32` is used. Currently, it's employed in:
1. Generating `transaction_id` in `UdpTrackerSession` [here](https://github.com/Tribler/tribler/blob/0382f8850a27f9c5ccd3077df3c66cc05cc92697/src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py#L263).
2. Generating `connection_id` in `UdpTrackerProtocol` [here](https://github.com/Tribler/tribler/blob/dd8a622f8b8f8be18e0bad24c430b16326e1d0f1/src/tribler/core/tests/tools/tracker/udp_tracker.py#L61).

Per the [UDP tracker protocol](https://www.rasterbar.com/products/libtorrent/udp_tracker_protocol.html), the `transaction_id` should be an `int32_t` value, and the `connection_id` should be an `int64_t` value. This aligns with our code, as seen [in the `!iiq` format string](https://github.com/Tribler/tribler/blob/dd8a622f8b8f8be18e0bad24c430b16326e1d0f1/src/tribler/core/tests/tools/tracker/udp_tracker.py#L62):

```python
response_msg = struct.pack('!iiq', TRACKER_ACTION_CONNECT, self.transaction_id, self.connection_id)
```

Thus, updating the range for `transaction_id` and `connection_id` from `UINT_16` to `INT_32` is expected to be safe and should not introduce any regressions.


src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
```diff
@@ -31,7 +31,7 @@
 TRACKER_ACTION_ANNOUNCE = 1
 TRACKER_ACTION_SCRAPE = 2
 
-MAX_INT32 = 2 ** 16 - 1
+MAX_INT32 = 2 ** 31 - 1
```
drew2a, line 34: NIT: The value `2147483647` is more immediately recognizable than `2 ** 31 - 1`, given its prevalence in documentation. Additionally, `2 ** 31 - 1` demands a bit more cognitive effort to understand or might require familiarity with the notation (which is not necessary as the name of the constant already states that it is related to the 32-bit version).

drew2a (approved):

src/tribler/core/components/torrent_checker/torrent_checker/torrentchecker_session.py
drew2a: As an aside, since this constant is related to C(am I correct on this?), the name `INT32_MAX` might be more appropriate.

kozlovsky: > The value 2147483647 is more immediately recognizable than 2 ** 31 - 1, given its prevalence in documentation.

It looks like your very personal preferences. When looking at `2 ** 31 - 1`, it is immediately clear how the value is constructed, and it is possible to understand that it is correct, intended for signed numbers and how many bits are in it, while for 2147483647, it is necessary to remember is it for INT or UNSIGNED INT, for example.

> As an aside, since this constant is related to C (am I correct on this?), the name INT32_MAX might be more appropriate.

It is not specifically related to C, just to signed binary arithmetic in general. I googled for it, and while both "MAX_INT32" and "INT32_MAX" are used, "INT32_MAX" returns more results, so I'll change it to "INT32_MAX".

synctext: Wow 64k hard limit on certain IDs, great catch! {btw in the future let's not discuss variable name cosmetics too much; correctness is obviously critical}

drew2a: @kozlovsky, indeed, it's a matter of preference as already mentioned.
@synctext, please note that the comment is labeled as **NIT** and the PR has received approval, so this discussion shouldn't block progress.

PR #7660, opened 2023-10-30
Use the filelock library to determine the primary GUI/Core process

kozlovsky:
This PR provides an alternative way (as compared to #7637) to determine the primary Tribler process. This PR uses the `filelock` library for that purpose:
* Now, when the primary Tribler GUI/Core process is running, the corresponding lock files `'tribler-gui.lock'`/`'tribler-core.lock'` are created and held until the related GUI/Core process is finished.
* ProcessManager still keeps all Tribler processes in the database but does not determine which process can be primary.
* The database of processes is still used to pass the opened REST API port from the Core process to the GUI process. The reason for this is that it is still possible that multiple processes of the same type can run in parallel (for example, one primary and several secondary processes), and it is safer to store port in a process-specific place inside the database instead of using a global variable; this way few errors caused by race conditions can happen.

The typical usage of the new API looks the following way:
1. First, the process tries to acquire the file lock:

    ```python
    process_lock = try_acquire_file_lock(lock_file_name)
    current_process_has_lock = bool(process_lock)
    ```
    
    If the attempt to grab the lock is successful, the process is considered a primary process; otherwise, it is a secondary process and should terminate after passing the magnet link to the primary process.
    
    The current process should hold the lock object returned from the `try_acquire_file_lock` call until the process finishes, as the lock releases if the lock object is garbage collected. The process can do `process_lock.release()` before it finishes, but it is not necessary, as (1) the object will be garbage collected after the program exits the scope where the lock was declared, and (2) even if the lock file was not removed from the disk, the lock is still considered as released after the current process is finished. On Linux, the lock file is not always deleted at the end because of the details of how Linux works with open file descriptors, but it is not a problem.
    
2. The `ProcessManager` is initialized so the current process (primary or secondary) can store information about itself in the database:

    ```python
    process_manager = ProcessManager(root_state_dir)
    process_manager.setup_current_process(kind=ProcessKind.Core, has_lock=current_process_has_lock)
    set_global_process_manager(process_manager)
    ```

3. The process should check whether the attempt to grab the lock was successful, and if not, it should terminate:

    ```python
    if not current_process_has_lock:
        msg = 'Another Core process is already running'
        logger.warning(msg)
        process_manager.sys_exit(1, msg)
    ```
    
So, the API is relatively simple.

### DIFF
--- a/src/tribler/core/utilities/process_manager/manager.py
+++ b/src/tribler/core/utilities/process_manager/manager.py
@@ -34,14 +34,40 @@ def get_global_process_manager() -> Optional[ProcessManager]:
         return global_process_manager
 
 
+def setup_process_manager(root_state_dir: Path, process_kind: ProcessKind, current_process_owns_lock: bool,
+                          creator_pid: Optional[int] = None) -> ProcessManager:
+    process_manager = ProcessManager(root_state_dir)
+    process_manager.setup_current_process(kind=process_kind, owns_lock=current_process_owns_lock,
+                                          creator_pid=creator_pid)
+    set_global_process_manager(process_manager)
+    return process_manager
+
+
 class ProcessManager:
-    def __init__(self, root_dir: Path, current_process: TriblerProcess, db_filename: str = DB_FILENAME):
+    def __init__(self, root_dir: Path, db_filename: str = DB_FILENAME):
         self.logger = logger  # Used by the `with_retry` decorator
         self.root_dir = root_dir
         self.db_filepath = root_dir / db_filename
         self.connection: Optional[sqlite3.Connection] = None
-        self.current_process = current_process
-        current_process.manager = self
+        self._current_process: Optional[TriblerProcess] = None
+
+    @with_retry
+    def setup_current_process(self, kind: ProcessKind, owns_lock: bool, creator_pid: Optional[int] = None):
+        current_process = TriblerProcess.current_process(manager=self, kind=kind, owns_lock=owns_lock,
+                                                         creator_pid=creator_pid)
+        self._current_process = current_process
+        with self.connect():
+            if current_process.primary:
+                if primary_process := self.get_primary_process(current_process.kind):
+                    raise RuntimeError(f'Previous primary process still active: {primary_process}. '
+                                       f'Current process: {current_process}')
+            current_process.save()
+
+    @property
+    def current_process(self):
+        if self._current_process is None:
+            raise RuntimeError('Current process is not set')
+        return self._current_process
 
     @contextmanager
     def connect(self) -> ContextManager[sqlite3.Connection]:
@@ -105,27 +131,38 @@ class ProcessManager:
 
         return 'unknown reason'
 
-    def primary_process_rowid(self, kind: ProcessKind) -> Optional[int]:
+    def get_primary_process(self, kind: ProcessKind) -> Optional[TriblerProcess]:
         """
         A helper method to load the current primary process of the specified kind from the database.
 
-        Returns rowid of the existing process or None.
+        Returns existing primary process or None.
         """
         with self.connect() as connection:
             cursor = connection.execute(f"""
                 SELECT {sql_scripts.SELECT_COLUMNS}
-                FROM processes WHERE kind = ? and "primary" = 1 ORDER BY rowid DESC LIMIT 1
+                FROM processes WHERE kind = ? and "primary" = 1 ORDER BY rowid
             """, [kind.value])
-            row = cursor.fetchone()
-            if row is not None:
+
+            primary_processes = []  # In normal situation there should be at most one primary process
+            rows = cursor.fetchall()
+            for row in rows:
                 process = TriblerProcess.from_row(self, row)
                 if process.is_running():
-                    return process.rowid
+                    primary_processes.append(process)
+                else:
+                    # Process is not running anymore; mark it as not primary
+                    process.primary = False
+                    process.save()
 
-                # Process is not running anymore; mark it as not primary
-                process.primary = False
-                process.save()
-            return None
+            if not primary_processes:
+                return None
+
+            if len(primary_processes) > 1:
+                for process in primary_processes:
+                    process.error_msg = "Multiple primary processes found in the database"
+                    process.save()
+
+            return primary_processes[0]
 
     def sys_exit(self, exit_code: Optional[int] = None, error: Optional[str | Exception] = None, replace: bool = False):
         """
--- a/src/tribler/core/utilities/process_manager/tests/test_manager.py
+++ b/src/tribler/core/utilities/process_manager/tests/test_manager.py
@@ -1,6 +1,6 @@
 import sqlite3
 import time
-from unittest.mock import Mock, patch
+from unittest.mock import MagicMock, Mock, patch
 
 import pytest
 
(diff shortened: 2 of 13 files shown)


src/tribler/core/utilities/process_manager/manager.py
```diff
@@ -35,13 +35,31 @@ def get_global_process_manager() -> Optional[ProcessManager]:
...
+    def setup_current_process(self, kind: ProcessKind, has_lock: bool, creator_pid: Optional[int] = None):
+        current_process = TriblerProcess.current_process(manager=self, kind=kind, has_lock=has_lock,
+                                                         creator_pid=creator_pid)
+        self._current_process = current_process
+        with self.connect():
+            if current_process.primary:
+                primary_process = self.get_primary_process(current_process.kind)
+                if primary_process is not None:
+                    raise RuntimeError(f'Previous primary process still active: {primary_process}. '
+                                       f'Current process: {current_process}')
```
drew2a, line 55: NIT: can be simplified

```suggestion
                if primary_process := self.get_primary_process(current_process.kind)
                    raise RuntimeError(f'Previous primary process still active: {primary_process}. '
                                       f'Current process: {current_process}')
```

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
 # pylint: disable=protected-access
 
 
-def test_become_primary(process_manager: ProcessManager):
-    # Initially process manager fixture creates a primary current process that is a single process in DB
-    p1 = process_manager.current_process
-    assert p1.primary
+def test_current_process_not_set(tmp_path):
```
drew2a, line 14: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
-    p2 = TriblerProcess.current_process(ProcessKind.Core, manager=process_manager)
-    p2.pid += 1
-    # The new process should not be able to become a primary process, as we already have the primary process in the DB
-    assert not p2.become_primary()
-    assert not p2.primary
 
-    with process_manager.connect() as connection:
-        # Here we are emulating the situation that the current process abnormally terminated without updating the row
-        # in the database. To emulate it, we update the `started_at` time of the primary process in the DB.
+def test_save(process_manager: ProcessManager):
```
drew2a, line 20: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
+    assert pm.get_primary_process(ProcessKind.GUI) is None
+    is_running.assert_not_called()  # No primary process was selected from the db => no is_running checks were called
 
-    with process_manager.connect() as connection:
-        rows = connection.execute('select rowid from processes where "primary" = 1').fetchall()
-        # At the end, the DB should contain only one primary process, namely p3
-        assert len(rows) == 1 and rows[0][0] == p3.rowid
 
+@patch('tribler.core.utilities.process_manager.process.TriblerProcess.is_running')
+def test_get_primary_process_2(is_running: MagicMock, tmp_path):
```
drew2a, line 38: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
+    p2 = TriblerProcess(manager=pm, kind=ProcessKind.Core, pid=200, primary=False, app_version='1', started_at=now-1)
+    p1.save()
+    p2.save()
+    assert pm.get_primary_process(ProcessKind.Core) is None
+    assert pm.get_primary_process(ProcessKind.GUI) is None
+    is_running.assert_not_called()  # No primary process was selected from the DB => no is_running checks were called
+
+
+@patch('tribler.core.utilities.process_manager.process.TriblerProcess.is_running', side_effect=[True])
+def test_get_primary_process_3(is_running: MagicMock, tmp_path):
```
drew2a, line 53: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
+    assert pm.get_primary_process(ProcessKind.GUI) is None  # No primary GUI processes in the DB
 
-def test_save(process_manager: ProcessManager):
-    p = TriblerProcess.current_process(ProcessKind.Core, manager=process_manager)
-    p.pid = p.pid + 100
-    p.save()
-    assert p.rowid is not None
+
+@patch('tribler.core.utilities.process_manager.process.TriblerProcess.is_running', side_effect=[False])
+def test_get_primary_process_4(is_running: MagicMock, tmp_path):
```
drew2a, line 72: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
+    assert pm.get_primary_process(ProcessKind.GUI) is None  # No primary GUI processes in the DB
+
+    last_processes = pm.get_last_processes()
+    assert len(last_processes) == 3
+    # The previous primary process status should be updated in the DB to non-primary
+    assert last_processes[-1].pid == 300 and not last_processes[-1].primary
+
+
+@patch('tribler.core.utilities.process_manager.process.TriblerProcess.is_running', side_effect=[True, True])
+def test_get_primary_process_5(is_running: MagicMock, tmp_path):
```
drew2a, line 95: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -11,48 +11,121 @@
...
+    assert pm.get_primary_process(ProcessKind.GUI) is None  # No primary GUI processes in the DB
+
+    last_processes = pm.get_last_processes()
+    assert len(last_processes) == 3
+    msg = "Multiple primary processes found in the database"
+    assert [p.error_msg for p in last_processes] == [None, msg, msg]
+
+
+@patch('tribler.core.utilities.process_manager.process.TriblerProcess.is_running', return_value=True)
+def test_setup_current_process(is_running: MagicMock, tmp_path):  # pylint: disable=unused-argument
```
drew2a, line 119: NIT: Adding a description of the test would be beneficial.

src/tribler/core/utilities/tiny_tribler_service.py
```diff
@@ -72,13 +75,16 @@ async def _start_session(self):
...
-
         root_state_dir = get_root_state_directory(create=True)
-        current_process = TriblerProcess.current_process(ProcessKind.Core)
-        self.process_manager = ProcessManager(root_state_dir, current_process)
+
+        self.process_lock = try_acquire_file_lock(root_state_dir / CORE_LOCK_FILENAME)
+        current_process_has_lock = bool(self.process_lock)
+
+        self.process_manager = ProcessManager(root_state_dir)
+        self.process_manager.setup_current_process(kind=ProcessKind.Core, has_lock=current_process_has_lock)
```
drew2a, line 84: I've encountered these lines for the third time now. Is there any chance they could be extracted or generalized?

src/tribler/gui/start_gui.py
```diff
@@ -38,12 +38,14 @@ def run_gui(api_port: Optional[int], api_key: Optional[str], root_state_dir, par
...
+    process_lock = try_acquire_file_lock(root_state_dir / GUI_LOCK_FILENAME)
+    current_process_has_lock = bool(process_lock)
+
+    process_manager = ProcessManager(root_state_dir)
+    process_manager.setup_current_process(kind=ProcessKind.GUI, has_lock=current_process_has_lock)
     set_global_process_manager(process_manager)  # to be able to add information about exception to the process info
-    current_process_is_primary = process_manager.current_process.become_primary()
 
-    load_logger_config('tribler-gui', root_state_dir, current_process_is_primary)
+    load_logger_config('tribler-gui', root_state_dir, current_process_has_lock)
```
drew2a, line 48: The name `current_process_has_lock` seems ambiguous (the same name also used on the core side). 
It's unclear whether the current process has detected an existing lock (indicating that it is a secondary) or if it is acquiring the lock (indicating it is primary).

Possible alternatives:
* `current_process_is_primary`
* `lock_has_been_acquired`
* `current_process_owns_lock`

drew2a (changes requested):
The PR looks good to me.

I have a minor question about the naming and a request for optionally adding descriptions for the tests (otherwise, it can be hard to guess what specific tests are testing).

As a side note, the test coverage for this PR could be increased.

src/tribler/gui/start_gui.py
kozlovsky: Agreed, renamed it to `current_process_owns_lock`

src/tribler/core/utilities/tiny_tribler_service.py
kozlovsky: Agreed. I added the `setup_process_manager` utility function that manages this initialization logic.

drew2a (approved):

Issue #7641, opened 2023-10-23
[7.13.0] TypeError: 'NoneType' object is not subscriptable in StartDownloadDialog.__init__()


kozlovsky:
```python
tribler.gui.utilities.CreationTraceback:
  File "run_tribler.py", line 115, in <module>
  File "tribler\gui\start_gui.py", line 58, in run_gui
  File "tribler\gui\tribler_app.py", line 32, in __init__
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "tribler\gui\utilities.py", line 438, in trackback_wrapper
  File "tribler\gui\utilities.py", line 435, in trackback_wrapper
  File "tribler\gui\tribler_app.py", line 36, in on_app_message
  File "tribler\gui\tribler_app.py", line 40, in handle_uri
  File "tribler\gui\tribler_window.py", line 1256, in handle_uri
  File "tribler\gui\tribler_window.py", line 639, in process_uri_request
  File "tribler\gui\tribler_window.py", line 866, in start_download_from_uri
  File "tribler\gui\dialogs\startdownloaddialog.py", line 100, in __init__
TypeError: 'NoneType' object is not subscriptable
```

[Here](https://github.com/Tribler/tribler/blob/cf0669e40a672ef9e45734c9baa7a17b1ed7393c/src/tribler/gui/dialogs/startdownloaddialog.py#L100):
```python
        self.dialog_widget.anon_download_checkbox.setChecked(
            self.window().tribler_settings['download_defaults']['anonymity_enabled']
```

It looks like sometimes an option can be missed from the config.

drew2a:
I can not reproduce the bug but I have a hypnotise of how it could happened.

Consider the following sequence:
1. Tribler starts the second instance of GUI and receives a URL from the first instance via the local socket (you can see it in logs)
2. GUI handles the URL:
https://github.com/Tribler/tribler/blob/5662d712dae1a6ec7c7f793336336bdea39ddc80/src/tribler/gui/tribler_window.py#L1253-L1256

But there is a chance that at this moment `self.tribler_settings` is still None as they have not been received yet. Therefore we see the `TypeError: 'NoneType' object is not subscriptable` as `self.tribler_settings` equals None.

PR #7662, opened 2023-10-31
Fix TypeError: 'NoneType' object is not subscriptable in StartDownloadDialog.__init__() 

drew2a:
This PR addresses issue #7641 by introducing a new flag `ui_started`, which is utilized to ascertain the appropriate time to proceed with URI handling.

### DIFF
--- a/src/tribler/gui/tests/test_gui.py
+++ b/src/tribler/gui/tests/test_gui.py
@@ -65,7 +65,7 @@ def fixture_window(tmp_path_factory):
     wait_for_signal(
         window.core_manager.events_manager.core_connected,
         timeout=20,
-        condition=lambda: window.tribler_started or (
+        condition=lambda: window.core_connected or (
                 window.core_manager.core_started and not window.core_manager.core_running)
     )
     if not window.core_manager.core_running:
--- a/src/tribler/gui/tribler_window.py
+++ b/src/tribler/gui/tribler_window.py
@@ -190,7 +190,8 @@ class TriblerWindow(QMainWindow):
         request_manager.set_api_key(api_key)
         request_manager.set_api_port(api_port)
 
-        self.tribler_started = False
+        self.core_connected = False
+        self.ui_started = False
         self.tribler_settings = None
         self.tribler_version = version_id
         self.debug_window = None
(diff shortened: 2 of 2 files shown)


kozlovsky (changes requested):
I think the suggested fix can solve this specific issue. However, the implementation does not look entirely correct to me.

Currently, the `tribler_started` flag means that:
1. The Tribler Core process has already started, and
2. The `EventRequestManager` of the Tribler GUI process has already established a connection to the Core process events stream.

The name of the flag does not make it intuitive that these (and only these) two conditions are satisfied at the moment when the flag is currently set.

When the current PR moves the setting of the `tribler_started` flag to the beginning of the `start_ui` method, the meaning of the `tribler_started` flag is changed to the:

3. ... and settings were already received from the Trible Core

This PR change makes the meaning of the `tribler_started` flag even less clear. Now, as the flag is moved to the `tribler_ui` function, it suggests that the purpose of the flag is somehow related to the UI, and the code reader can interpret this flag as "the Tribler UI is started". But then it should be placed at the end of the `start_ui` method, as at the beginning of the method, the UI is not started yet.

Furthermore, this change could affect the previous code. We have a condition:
```python
def on_core_connected(self, version):
    if self.tribler_started:
        self._logger.warning("Received duplicate Tribler Core connected event")
        return

    self._logger.info("Core connected")
    self.tribler_started = True
    self.tribler_version = version

    request_manager.get("settings", self.on_receive_settings, capture_errors=False)
```

But after the PR's change, it becomes possible for this code to be executed twice because the line `self.tribler_started = True` is now executed at a much later stage. If we rename the flag to the `tribler_core_connected` to properly reflect its current meaning, it will become clear that the flag should be set in its current place.

I suggest to make the name of the flag less ambiguous. It is better to have several flags with precise semantics for each flag:
* `tribler_core_started` - the Core process is started
* `tribler_core_connected` - this is what the `tribler_started` flag means currently before the PR changes.
* `settings_received` - this new flag we should check to avoid the bug
* `tribler_ui_starting` - this is how a possible flag at the beginning of the `start_ui` method could be named
* `tribler_ui_started` - this is how a possible flag at the end of the `start_ui` method could be named

We don't need all these flags. However, it may be good to have most of them to reduce ambiguity. The minimal subset that we currently need is:
* `tribler_core_connected` - instead of the current `tribler_started`, renamed to avoid ambiguity.
* `settings_received` - a new flag that we can check in the `handle_uri` method to avoid the bug we are trying to fix in the current PR. But I can't preclude that actually, in the `handle_uri` method, we should actually check another new `tribler_ui_started` flag that could be set at the end of the `start_ui` method.

kozlovsky (approved):
Looks good! After this PR the logic is much clearer now.

Issue #7519, opened 2023-06-29
Seeding option only follow HH:MM data format & not the others


absolutep:
**Describe the bug**

Seeding option only follows the HH:MM data & all the other options are neglected.
Meaning, if you selected unlimited seeding or seeding according to ratio, eg, 2:1. Tribler will not follow this rule but instead will follow the rule written in HH:MM

**To Reproduce**

Go to tribler settings >> write any amount of time in HH:MM option >> save >> quit >> check Tribler config file >> HH:MM format seeding time is denoted in overall seconds.

Now,

Go to Tribler settings >> seeding options >> select unlimited seeding >> save >> quit >> check Tribler config file >> HH:MM format seeding time is denoted in overall seconds and not the unlimited seeding or any ratio seeding.

**Expected behavior**

When Unlimited seeding is selected - Tribler should follow this rule.

When Specific ratio seeding is selected, for eg, 2:1, Tribler should follow this rule.

Instead Tribler only follows the HH:MM format rule, irrespective of your seeding choice.

**Screenshots**

Unlimited seeding is chosen & data entered in HH:MM format.

Tribler Config shows - Tribler choses HH:MM format data everytime irrespective of user's chosen option.

![1](https://github.com/Tribler/tribler/assets/97702149/931985c5-58e1-482b-9bad-71f0c34972b5)

![2](https://github.com/Tribler/tribler/assets/97702149/b261cbf6-1a32-4a88-b7cb-81630c99fc81)

![3](https://github.com/Tribler/tribler/assets/97702149/f13d61e6-17df-4ec0-ad3a-7440993cf12d)
![4](https://github.com/Tribler/tribler/assets/97702149/da343598-ce47-40ec-b17a-6649d9b05c83)

**Desktop (please complete the following information):**

- OS: [e.g. Windows 10]
- Tribler's version [e.g. 7.13.RC2]

drew2a:
@absolutep thank you for the report. I was able to reproduce the bug and hopefully can provide a fix for it soon.

drew2a:
After some investigation, I can confirm that this isn't a bug but rather a characteristic of how values are stored in the Config.

Tribler Config does not persist default values, meaning that if a value is set to its default, it won't be written to disk (and vice versa).

Here are the default values for `DownloadDefaultsSettings`:

```python
class DownloadDefaultsSettings(TriblerConfigSection):
    anonymity_enabled: bool = True
    number_hops: int = 1
    safeseeding_enabled: bool = True
[...]

PR #7663, opened 2023-11-01
Extract `parse_settings` to the `TriblerConfig`

drew2a:
I have some code left over from investigating #7519, which includes a refactoring of `parse_settings` and additional tests.

It is at the discretion of the reviewer to determine whether this refactoring is necessary or not.

### DIFF
--- a/src/tribler/core/components/restapi/rest/settings_endpoint.py
+++ b/src/tribler/core/components/restapi/rest/settings_endpoint.py
@@ -19,7 +19,7 @@ class SettingsEndpoint(RESTEndpoint):
 
     def __init__(self, tribler_config: TriblerConfig, download_manager: DownloadManager = None):
         super().__init__()
-        self.tribler_config = tribler_config
+        self.config = tribler_config
         self.download_manager = download_manager
 
     def setup_routes(self):
@@ -40,7 +40,7 @@ class SettingsEndpoint(RESTEndpoint):
     async def get_settings(self, request):
         self._logger.info(f'Get settings. Request: {request}')
         return RESTResponse({
-            "settings": self.tribler_config.dict(),
+            "settings": self.config.dict(),
             "ports": list(default_network_utils.ports_in_use)
         })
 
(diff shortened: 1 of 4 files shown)


src/tribler/core/components/restapi/rest/settings_endpoint.py
```diff
@@ -55,31 +55,12 @@ async def get_settings(self, request):
...
+        settings = await request.json()
+        self._logger.info(f'Received settings: {settings}')
+        self.config.update_from_dict(settings)
+        self.config.write()
 
-        # Perform some actions when specific keys are set
-        if section == "libtorrent" and (option == "max_download_rate" or option == "max_upload_rate"):
-            if self.download_manager:
-                self.download_manager.update_max_rates_from_config()
+        if self.download_manager:
```
drew2a, line 63: The condition `if section == "libtorrent" and (option == "max_download_rate" or option == "max_upload_rate"):` was always evaluating to True, making the check unnecessary.

src/tribler/core/config/tribler_config.py
```diff
@@ -152,6 +152,20 @@ def write(self, file: Path = None):
...
         conf.write()
 
+    def update_from_dict(self, config: Dict):
+        """ Update (patch) current config from dictionary"""
+
+        def update_recursively(settings: BaseSettings, attribute_name: str, attribute_value: Union[Any, Dict]):
+            """ Update setting recursively from dictionary"""
+            if isinstance(attribute_value, dict):
+                for k, v in attribute_value.items():
+                    update_recursively(settings.__getattribute__(attribute_name), k, v)
```
kozlovsky, line 162: **nitpick**: `getattr` should work fine here:
```suggestion
                    update_recursively(getattr(settings, attribute_name), k, v)
```

kozlovsky (approved):
A nice refactoring!

src/tribler/core/config/tribler_config.py
drew2a: Agree that it works fine here. What are the benefits of using `getattr` over `__getattribute__`?

kozlovsky: It is more common to use `getattr(a, "b")` instead of `a.__getattr__("b")`; it feels more high-level, and it is shorter and a bit clearer without all these underscores. Just as we usually write `len(collection)` instead of `collection.__len__()`, `iter(sequence)` instead of `sequence.__iter__()`, and `next(iter)` instead of `iter.__next__()`. The longer form with underscores feels a bit like working with private/protected attributes or implementation details, while all of `getattr`/`len`/`iter`/`next` feel more like a public API.

drew2a: Replaced.

drew2a:
@kozlovsky please note that I've added the ` validate_config` call to the end of the `update_recursively` function.

kozlovsky (approved):
Looks good

Issue #5399, opened 2020-06-22
[7.5.0] large torrent stuck on building circuits.


Dmole:
"Filesize 269.2 GB in 0 files"
100% CPU for 15+ minutes on this 22MB torrent.
Can't remove the torrent, had to kill -9 Tribler and restore a backup of the ~/.Tribler folder

ichorid:
Unfortunately, there is almost nothing we can do without more information.
@Dmole , could you provide us the torrent file that caused the issue? Or a glitchy `.Tribler` folder?

Dmole:
magnet:?xt=urn:btih:cccaad8d8b856c9609137890d5802189d841ad2c

ichorid:
Confirmed, trying to download this one results in "Stopped on error"

devos50:
@ichorid do you want to debug this one?

synctext:
"Filesize 269.2 GB in 0 files"
This already provides 90% certainty what is going on. Probably a .torrent file with numerous MBytes of filenames and hashes. We should test Tribler with a 10+ MByte legal torrent file. Thnx @qstokkink for finding this! Indeed one of the reasons people will dump Tribler forever.

qstokkink:
Since I complained about this behavior, I guess I'll assign myself to solve it. 😃

qstokkink:
I found a single torrent that managed to lock up my Tribler, on an AMD Ryzen Threadripper 3970X 😃. After some investigation, the culprit - for me at least - seems to be the REST API's `Request: GET http://localhost:[xxx]/downloads?get_pieces=1` locking up the main thread for almost 5 seconds. Roughly 4/5 seconds were spent in this loop:

https://github.com/Tribler/tribler/blob/633f8bdda3bdca26a2ea96208ee489c9312f157a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py#L146-L155

As a POC, I applied some quick-and-dirty optimizations to get the 5 seconds down to ~1 millisecond and the locking-up behavior stopped.

qstokkink:
Annoying, tables with tens of thousands of rows slow down Qt _a lot_. So, even though files are streamed in through the REST API with my quick-and-dirty fix, the user interface becomes slow after some time anyway if you click on the "Files" tab.

The only sane solution is to implement pagination for a torrent's files but this will require a substantial refactoring effort (that I was initially trying to avoid).

qstokkink:
[...]

PR #7664, opened 2023-11-02
Only load peers, pieces, and files for the selected download

qstokkink:
Related to #5399 (PR 1/4)

In order to filter for the selected download in the downloads list, a new REST API parameter `"infohash"` is added to `get_downloads()` (serving `/downloads`). If the parameter is omitted, `get_downloads()` functions as before. Otherwise, the `get_peers`, `get_pieces`, and `get_files` are only applied to the download with the given infohash.

Because of the change `set_pieces()` may encounter downloads without set pieces. This is now handled.

Tangentially related to the other changes, I added a `refresh_downloads()` whenever the detail tab/window is shown.

---

Test reruns for this PR:

| **name** | **reason** |
| --- | --- |
| guitest_nix / run (macos-latest) | teardown of test_trust_page (Signal <bound PYQT_SIGNAL finished of QProcess object at 0x11fe49ee0> not raised within 10 seconds) |
| guitest_nix / run (macos-latest) | Segmentation fault |

### DIFF
--- a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
@@ -178,6 +178,13 @@ class DownloadsEndpoint(RESTEndpoint):
                 'description': 'Flag indicating whether or not to include files',
                 'type': 'boolean',
                 'required': False
+            },
+            {
+                'in': 'query',
+                'name': 'infohash',
+                'description': 'Limit fetching of files, peers, and pieces to a specific infohash',
+                'type': 'boolean',
+                'required': False
             }],
         responses={
             200: {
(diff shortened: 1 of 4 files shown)


src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
```diff
@@ -315,30 +323,31 @@ async def get_downloads(self, request):
...
+
+                    download_json["peers"] = peer_list
+
+                # Add piece information if requested
+                if get_pieces:
+                    download_json["pieces"] = download.get_pieces_base64().decode('utf-8')
+
+                # Add files if requested
+                if get_files:
+                    download_json["files"] = self.get_files_info_json(download)
```
qstokkink, line 350: Info: The only change to these lines is an additional indent.

drew2a (approved):
👍

PR #7680, opened 2023-11-08
Upload binaries by default 

drew2a:
This PR enables uploading binaries for each PR by default.

### DIFF
--- a/.github/workflows/!PR.yml
+++ b/.github/workflows/!PR.yml
@@ -15,6 +15,8 @@ jobs:
     outputs:
       python-version: '3.8'
       source-ref: refs/pull/${{github.event.pull_request.number}}/merge
+      labels: ${{ toJson(github.event.pull_request.labels.*.name) }}
+
     steps:
       - run: echo done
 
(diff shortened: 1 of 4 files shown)


qstokkink (approved):

xoriole (commented):
- Should we actually be uploading the binaries by default?
- Does this have any potential to prolong the time taken for pull requests?

drew2a:
> Should we actually be uploading the binaries by default?

I don't know. It is up to us to decide. I personally never use binaries from the PR builds.

> Does this have any potential to prolong the time taken for pull requests?

Yes. For example, for the Windows build it adds 2m 43s:
https://github.com/Tribler/tribler/actions/runs/6811621558/job/18522334331

image

drew2a:
As an alternative approach, I could implement a condition where, if a PR has a specific label (such as 'upload binaries'), then the binaries would be uploaded.

drew2a:
I'm closing this PR. The new PR with the add binaries for PR with the label #7689.

PR #7685, opened 2023-11-09
Handle an absent `db_version` for Upgrader.

### LINKED ISSUE #7601: UpgradeError: "TypeError: 'NoneType' object is not subscriptable" in get_db_version
```
    File "tribler\gui\upgrade_manager.py", line 48, in run
    File "tribler\gui\upgrade_manager.py", line 91, in upgrade_state_dir
    File "tribler\core\upgrade\upgrade.py", line 97, in run
    File "tribler\core\upgrade\upgrade.py", line 380, in upgrade_pony_db_8to10
    File "tribler\core\upgrade\db8_to_db10.py", line 248, in get_db_version
TypeError: 'NoneType' object is not subscriptable
The above exception was the direct cause of the following exception:
    File "tribler\gui\upgrade_manager.py", line 251, in on_worker_finished
tribler.gui.exceptions.UpgradeError: TypeError: 'NoneType' object is not subscriptable
```

[Here](https://github.com/Tribler/tribler/blob/8d03b993a0c4ec4bff88b533987df2b482d7c4a9/src/tribler/core/upgrade/db8_to_db10.py#L248):
```python
def get_db_version(db_path):
    with contextlib.closing(sqlite3.connect(db_path)) as connection, connection:
        cursor = connection.cursor()
        cursor.execute('SELECT value FROM MiscData WHERE name == "db_version"')
        version = int(cursor.fetchone()[0])  # <-- here
    return version
```

Currently, two users were affected by this bug (one on Linux and another on Windows-7 32bit).

To avoid the error, the code should take into account a case when the database does not contain any `db_version` value at all (in that case, `cursor.fetchone()` returns `None`) and should handle it properly.

But normally, the `db_version` value is added immediately after the database is created, and it is not clear why it was missing on the user's machine; it is possible it was deleted manually from the database.

I consider this bug low-priority; we can wait until a bigger number of users is affected by it.

drew2a:
This PR fixes #7601

### DIFF
--- a/src/tribler/core/upgrade/db8_to_db10.py
+++ b/src/tribler/core/upgrade/db8_to_db10.py
@@ -245,5 +245,8 @@ def get_db_version(db_path):
     with contextlib.closing(sqlite3.connect(db_path)) as connection, connection:
         cursor = connection.cursor()
         cursor.execute('SELECT value FROM MiscData WHERE name == "db_version"')
-        version = int(cursor.fetchone()[0])
+        query_results = cursor.fetchone()
+        if not query_results:
+            return 0
+        version = int(query_results[0])
     return version
--- a/src/tribler/core/upgrade/tests/test_db8_to_db10.py
+++ b/src/tribler/core/upgrade/tests/test_db8_to_db10.py
@@ -1,0 +1,11 @@
+from unittest.mock import MagicMock, Mock, patch
+
+from tribler.core.upgrade.db8_to_db10 import get_db_version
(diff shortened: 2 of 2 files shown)


kozlovsky (approved):
One possible explanation of the missed `db_version` value is that the initial migration from the ancient database format was aborted before the `db_version` value was stored in the database at the end of the initial database migration. In that case, the fix should work - the content of the database can be incomplete, as the database was not fully migrated from the old format, but at the current moment, it is the best we can do with these old data.

PR #7692, opened 2023-11-15
Clean up licenses

### LINKED ISSUE #6590: There are three licenses in the Tribler repo instead of just single one
There are three licenses in the Tribler repo:

image

Which is ambiguous for users and hard to maintain for developers.

Probably we should merge them and use just a single one.

### LINKED ISSUE #7585: The GPL is not an EULA
**Describe the bug**
The Tribler installer includes the GPL text in an End User License Agreement (EULA) click-through page.  The GPL is not a license for end users of the software and it doesn't belong in the installer.  It serves no purpose legally or otherwise.

**To Reproduce**
1. Start the Tribler installer and see the License Agreement click-through

**Expected behavior**
The GPL text should not be in a click-through license during the install

**Screenshots**
image

- OS: Windows 10 64-bit
- Tribler's version: 7.13.0

**Additional context**
References:  
https://flameeyes.blog/2020/03/23/the-gpl-is-not-an-eula/
https://drj11.wordpress.com/2008/02/25/the-gnu-gpl-is-not-an-eula/
https://twitter.com/Foone/status/1528010830448451585
https://digipres.club/@foone/110719488736965037

drew2a:
This PR eliminates the EULA screen from the Windows and macOS installers and resolves [#7585](https://github.com/Tribler/tribler/issues/7585). Additionally, it removes the LGPL license from the root of the repository, addressing [#6590](https://github.com/Tribler/tribler/issues/6590). Previously, there were two licenses in the Tribler repo: GPL and LGPL. We should retain only one, as they are not the same.

If the decision is to keep LGPL, I will restore it and remove GPL instead.

Refs:
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository
* https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project
* https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License
* https://www.gnu.org/licenses/why-not-lgpl.en.html
* https://www.qt.io/licensing/open-source-lgpl-obligations


drew2a:
I've checked the macOS dmg from https://github.com/Tribler/tribler/actions/runs/6890483412?pr=7692 it is work.

synctext (approved):
Good cleanup! Let's do the most pure license: GPL.

xoriole (approved):

Issue #7647, opened 2023-10-25
Improve Core Process Error Handling: Restart on Crash and Prompt for Error Reporting


xoriole:
## Overview
Currently, when the core process of Tribler encounters a serious error, the application is made to crash. This behavior can be optimized to enhance user experience and provide a more robust error-reporting mechanism. This proposal is to ensure that whenever the core process crashes from an unhandled error, the GUI detects it and initiates a restart of the core process. Furthermore, if the new core process identifies that there was an unreported error from the previous run, it will notify the GUI to prompt the user for error reporting.

## Proposal Details
**1. Detect Core Process Crash in GUI :** Implement a check on the state change of the Core process in CoreManager that checks the status of the core process. If the core process is not running without an intentional shutdown, it should be interpreted as a crash. Additionally, the errors that are propagated via events endpoint that are not subclasses of NoCrashException, also crash the application. This should be handled as well.

**2. Restart the Core Process :** Upon detecting the crash, the GUI should initiate the restart of the core process.

**3. Save Sentry Event on Crash :**  Whenever there's a crash in the core process, save the sentry event locally. This will enable us to access crash details in the subsequent runs. If the event is sent to the error reporter, then the saved event can be removed.

**4. Detect Unreported Errors on Core Start :** When the core process starts, it should first check for any saved sentry events that were not reported in the previous runs.

**5. Communicate Unreported Errors to GUI :** If an unreported error is detected, the core process should notify the GUI about it.

**6. Prompt User for Error Reporting :** Upon receiving a notification about an unreported error, the GUI should display a popup dialog to the user, asking if they'd like to send the error report. This allows users to have a choice whether or not to send the error report.

## Benefits
- **Enhanced User Experience :** By automatically restarting the core process upon a crash, the disruption on using Tribler will likely reduce for the user.
- **Efficient Error Reporting :** Saving the errors that were not sent to the GUI during the crash will get collected in the next run. This increases the chances of finding out the actual cause of the error if the user reports the error. This will give us more insights.

This proposal seeks to improve the resilience of Tribler in the face of unexpected errors and optimize the error-reporting process. Feedback and further suggestions are welcome.

drew2a:
Sounds promising. However, one concern is the potential for an "endless core restart" loop if the same error repeatedly occurs.

1. Launch GUI
2. Start Core
3. Encounter error in Core
4. Restart Core
5. Encounter the same error in Core
6. And so on...

xoriole:
[...]

PR #7699, opened 2023-11-20
Improve Core Process Error Handling

xoriole:
This PR addresses
- https://github.com/Tribler/tribler/issues/7647

Here is an example of Sentry report sent:
- https://sentry.tribler.org/organizations/tribler/issues/2500/?project=10&query=&referrer=issue-stream&stream_index=0

The following additional entry is added to the report:
```
core_restart_logs: [
CoreRestartLog(pid=1232211, started_at='2023-11-27 14:36:14', uptime='0:00:11',  time_to_shutdown='0:00:04',  exit_code=1, exit_status='0'), 
CoreRestartLog(pid=1232278, started_at='2023-11-27 14:36:25', uptime='still running',  time_to_shutdown='unknown',  exit_code=None, exit_status='None')
],
```

### DIFF
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -187,7 +187,7 @@ class CoreExceptionHandler:
         self_copy.should_stop = False
         serialized_error = json.dumps(dataclasses.asdict(self_copy), indent=True)
 
-        with open(filepath, 'w') as exc_file:
+        with open(filepath, 'w', encoding='utf-8') as exc_file:
             exc_file.write(serialized_error)
 
     def delete_saved_file(self, reported_error: ReportedError):
@@ -203,7 +203,7 @@ class CoreExceptionHandler:
                 continue
 
             error_file_path = self.crash_dir / error_filename
-            with open(error_file_path, 'r') as file_handle:
+            with open(error_file_path, 'r', encoding='utf-8') as file_handle:
                 try:
                     saved_errors.append(ReportedError(**json.loads(file_handle.read())))
                 except JSONDecodeError:
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -1,8 +1,12 @@
+import dataclasses
 import errno
+import json
 import logging
+import os
 import re
 import sys
 from io import StringIO
+from json import JSONDecodeError
 from socket import gaierror
 from traceback import print_exception
 from typing import Callable, Dict, Optional, Set, Tuple, Type
@@ -52,6 +56,7 @@ class CoreExceptionHandler:
         self.report_callback: Optional[Callable[[ReportedError], None]] = None
         self.unreported_error: Optional[ReportedError] = None
         self.sentry_reporter = SentryReporter()
+        self.crash_dir = None
 
     @staticmethod
     def _get_long_text_from(exception: Exception):
@@ -135,7 +140,7 @@ class CoreExceptionHandler:
                 process_manager.current_process.set_error(exception)
 
             if should_stop:
-                reported_error.save_to_file()
+                self.save_to_file(reported_error)
 
             if self.report_callback:
                 self.logger.error('Call report callback')
@@ -154,6 +159,56 @@ class CoreExceptionHandler:
             self.sentry_reporter.capture_exception(ex)
             self.logger.exception(f'Error occurred during the error handling: {ex}')
             raise ex
+
+    def get_or_create_log_dir(self):
+        if self.crash_dir and not self.crash_dir.exists():
+            self.crash_dir.mkdir(exist_ok=True)
+        return self.crash_dir
+
+    def get_file_path(self, reported_error: ReportedError):
+        if not self.crash_dir:
+            return None
+
+        if not self.crash_dir.exists():
+            self.crash_dir.mkdir(exist_ok=True)
+
+        filepath = self.crash_dir / f"{reported_error.type}-{reported_error.created_at}.json"
+        return filepath
+
+    def save_to_file(self, reported_error: ReportedError):
+        filepath = self.get_file_path(reported_error)
+        if not filepath:
+            return
+
+        # While saving to file, set should_stop=False.
+        # This is because this file will be read on restart of the core, and
+        # we don't want to crash the core for the error from the last run.
+        self_copy = dataclasses.replace(reported_error)
+        self_copy.should_stop = False
+        serialized_error = json.dumps(dataclasses.asdict(self_copy), indent=True)
+
+        with open(filepath, 'w') as exc_file:
+            exc_file.write(serialized_error)
+
+    def delete_saved_file(self, reported_error: ReportedError):
+        self.get_file_path(reported_error).unlink(missing_ok=True)
+
+    def get_saved_errors(self):
+        if not self.crash_dir.exists():
+            return []
+
+        saved_errors = []
+        for error_filename in os.listdir(self.crash_dir):
+            if not error_filename.endswith('.json'):
+                continue
+
+            error_file_path = self.crash_dir / error_filename
+            with open(error_file_path, 'r') as file_handle:
+                try:
+                    saved_errors.append(ReportedError(**json.loads(file_handle.read())))
+                except JSONDecodeError:
+                    error_file_path.unlink()
+        return saved_errors
 
 
 default_core_exception_handler = CoreExceptionHandler()
--- a/src/tribler/core/components/reporter/exception_handler.py
+++ b/src/tribler/core/components/reporter/exception_handler.py
@@ -53,7 +53,16 @@ class CoreExceptionHandler:
         self.report_callback: Optional[Callable[[ReportedError], None]] = None
         self.unreported_error: Optional[ReportedError] = None
         self.sentry_reporter = SentryReporter()
-        self.crash_dir: Optional[Path] = None  # this path is set in ReporterComponent
+        self.crash_dir: Optional[Path] = None
+
+    def set_crash_dir(self, crash_dir: Path):
+        """
+        Sets the path to the directory where the errors will be saved to the file.
+        This is set in ReporterComponent.
+        """
+        self.crash_dir = crash_dir
+        if self.crash_dir and not self.crash_dir.exists():
+            self.crash_dir.mkdir()
 
     @staticmethod
     def _get_long_text_from(exception: Exception):
@@ -160,15 +169,6 @@ class CoreExceptionHandler:
             self.logger.exception(f'Error occurred during the error handling: {ex}')
             raise ex
 
-    def get_or_create_log_dir(self) -> Path:
-        """
-        Returns the path to the directory where the crash log files will be saved.
-        If the directory doesn't exist, it will be created.
-        """
-        if self.crash_dir and not self.crash_dir.exists():
-            self.crash_dir.mkdir(exist_ok=True)
-        return self.crash_dir
-
     def get_file_path(self, reported_error: ReportedError) -> Optional[Path]:
         """
         Returns the path to the file where the error will be saved.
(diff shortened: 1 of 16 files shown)


.github/workflows/pytest.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     env:
       PYTEST_SENTRY_DSN: ${{secrets.PYTEST_SENTRY_DSN}}
 
-    timeout-minutes: 10
+    timeout-minutes: 15
```
xoriole, line 47: The timeout-minutes are updated on main but it was necessary to update here because the tests were consistently crossing 10 mins timeout on this branch.

drew2a:
The field `core_restart_logs` was added to the tags: 

image


It feels that the more appropriate place for this field is `additional_information` or `Reporter`:
image

drew2a:
@xoriole could you please rebase it to the main branch?

src/tribler/core/components/reporter/exception_handler.py
```diff
@@ -134,6 +145,12 @@ def unhandled_error_observer(self, _, context):
             if process_manager:
                 process_manager.current_process.set_error(exception)
 
+            # If should_stop is True, the error is critical, so we first save it to the file to ensure
+            # that it will be reported to the GUI via events endpoint either immediately or later
+            # after core restart. This saved file will be deleted when the error is reported to the GUI.
+            if should_stop:
```
drew2a, line 151: From my understanding, if error reports are handled by the exception handler, then they will be sent to the GUI without any problem.

Could you please provide a way to reproduce the initial problem that this PR is aimed to solve?

xoriole: To reproduce a part of the solution, `CTRL` + `ALT` +  `SHIFT` + `C` will trigger a core error which is sent to the GUI as an error report and in turn restarts the core. The core_restart_logs should be visible on sentry then.

The other part where the core exception is logged and sent to sentry on the next restart is difficult to reproduce. To test this, I manually placed a saved exception on the crash_logs directory inside the state directory.

xoriole:
> The field `core_restart_logs` was added to the tags:
> 
> image
> It feels that the more appropriate place for this field is `additional_information` or `Reporter`: image

Thank you for pointing that out. It should be fixed now.
An example: [here](https://sentry.tribler.org/organizations/tribler/issues/2500/?project=10&query=&referrer=issue-stream&stream_index=0)

src/tribler/core/components/reporter/exception_handler.py
drew2a: > To reproduce a part of the solution, CTRL + ALT + SHIFT + C will trigger a core error which is sent to the GUI as an error report and in turn restarts the core. The core_restart_logs should be visible on sentry then.

Should I press it within the Tribler GUI? Is it `Ctrl+Opt+Shift+C` on macOS?

xoriole: It is a keyboard shortcut to trigger core exception.
https://github.com/Tribler/tribler/blob/8bb85b50d5d0c4a0580630820079974b98e5f2f9/src/tribler/gui/tribler_window.py#L237-L241

drew2a: Basically, this combination just calls `core_exception_handler.unhandled_error_observer`, which is not a way to reproduce the real use case of the improvements proposed in this PR.

```python
    async def core_test_exception(self, _):
        self._logger.info('Test Tribler Core exception was triggered')
        if self.core_exception_handler:
            try:
                raise TriblerCoreTestException('Tribler Core Test Exception')
            except TriblerCoreTestException as e:
                context = dict(should_stop=True, message='Test message', exception=e)
                self.core_exception_handler.unhandled_error_observer(None, context)
        else:
            self._logger.info('Exception handler is not set in DebugEndpoint')

        return RESTResponse({"success": True})
```

So for me, it is still not clear which case the improvement covers. I understand the theoretical point from the original issue but do not understand their relation with the implementation.

In practice, which type of errors could trigger the new logic and become visible for us but not visible with the old logic? Where should an error be raised in the core to prove the new logic adds value to error handling?

xoriole: > In practice, which type of errors could trigger the new logic and become visible for us but not visible with the old logic?

Here is the original issue this PR is developed to tackle:
- https://github.com/Tribler/tribler/issues/7603 

And, here is the linked sentry issue for it:
- https://sentry.tribler.org/organizations/tribler/issues/2003/events/f1a9c56138aa4dbea5598d00c2a72311/?project=7&query=is%3Aunresolved+version%3A7.13.0&referrer=previous-event&stream_index=2

In the case where the Core connection is lost, we do not know what caused the issue. From the uptime, we see that Tribler was running for 7205 seconds. Then suddenly the core connection was lost and we have no idea what caused it. The original error in the core that broke the GUI connection is unknown. 

With this PR, such a core error should be logged as a file (assuming the exception handler on the core catches such an exception) and sent to the GUI when the core connection is re-established after the core restart. Then the error report sent by the user should give us more information about the original cause of the previous exception.

> Where should an error be raised in the core to prove the new logic adds value to error handling?

Regarding the reproduction of such a case practically under a normal Tribler run, I'm not able to do so. Which is why the above suggestion was mentioned.

drew2a: I understand your intention and appreciate it. I'm just questioning the implementation, particularly whether the suggested change could reveal the real case of https://github.com/Tribler/tribler/issues/7603.

I'm uncertain about the change because we actually cannot imagine the potential case for the error; therefore, we cannot verify the solution, even theoretically. It is such a significant change you propose (15 new commits, +700 LOC) and it changes the interaction between Core and GUI pretty significantly, which could lead to unforeseen circumstances. I'm not saying the PR is good or bad; I'm just highlighting the risks and questioning the worthiness of taking them.

From my POV, if the error is successfully handled by the core's error handler, then it should be sent to the GUI without any problem, so the fix you suggested might be insufficient.

One of the explanations for https://github.com/Tribler/tribler/issues/7603 could be the possibility that someone (maybe even a user) just killed the process following the observation that the process consumes a lot of CPU and/or memory resources.

Another explanation could be the probability of some weird bug in `aiohttp` that led to an HTTP server crash, which in turn led to the `CoreConnectionError`. With this explanation, the suggested solution could work, but we need to verify it somehow.

drew2a: I came up with two more probable scenarios that could lead to a `CoreConnectionError` but are not related to `aiohttp`. I'll explain them tomorrow.

drew2a: I've added scenarios here: https://github.com/Tribler/tribler/issues/7647#issuecomment-1838527794

drew2a:
@xoriole, nice summary!

Regarding the PR, my suggestion is to treat it as general core process error handling rather than particular handling of the original `CoreConnectionError`.

If it works with the deadlock case and with the frozen asyncio case, then let's merge it. If it doesn't work, let's improve it and then merge.

It looks like a nice improvement, but we have to have confirmation that it improves error handling (and we can prove it by executing scenarios described in https://github.com/Tribler/tribler/issues/7647#issuecomment-1838527794). We can also extend this list, but I'm out of good ideas.

kozlovsky (changes requested):
In my opinion, this PR adds valuable functionality.

As I see it, the PR implements two independent features:

1. The first feature is restarting Core in case of a serious error without restarting GUI. In my opinion, it is beneficial for user experience. Not all Core problems can be fixed by restart, but some can.

2. The second feature is storing serious Core errors on disk before reporting them to GUI and re-sending errors to GUI in case sending them successfully the first time was not possible.

Both parts of the current PR rely on the established connection between GUI and Core. For that reason, in my opinion, the solution implemented in this PR is partial:

1. The PR logic only restarts Core if an error was successfully sent to GUI. If the Core error happens before the connection to GUI is established, Core will not be restarted.

2. If sending the error to the GUI immediately is impossible, the PR logic sends it again after the connection to the GUI is established after the restart. But if Core constantly crashes before the connection to GUI is established (the most interesting error case, actually!), GUI will never receive the errors stored on disk.

However, extending the logic in a subsequent PR is possible by changing how the serialized errors are handled. Instead of waiting until the restarted Core process reads errors from the disk and sends them to the GUI process via the newly established connection, the GUI process can directly read serialized errors from the disk even before the new Core process is started. It should not be hard to make this change in a subsequent PR to extend the range of the handled startup errors significantly.

Also, GUI can restart the crashed Core process even before the connection to GUI is established. GUI already does it in case of database corruption, and we can extend the range of errors handled this way.

In conclusion, the implemented logic is useful and can be extended further in a subsequent PR to cover a broader range of possible errors. I have some minor comments about unused attributes and logical flags handling.

src/tribler/core/components/reporter/reported_error.py
```diff
@@ -8,8 +16,81 @@ class ReportedError:
...
+        """
+        return parent_dir / self.get_filename()
+
+    def copy(self, **changes) -> 'ReportedError':
+        """
+        Returns a copy of the error report with the given changes.
+        """
+        return dataclasses.replace(self, **changes)
+
+    def serialized_copy(self) -> 'ReportedError':
```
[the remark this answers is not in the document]
xoriole: It is now used in EventsEndpoint.

src/tribler/gui/core_manager.py
```diff
@@ -235,29 +283,49 @@ def stop(self, quit_app_on_core_finished=True):
...
+        """
+        if self.is_restarting:
+            return
+
+        self._logger.info("Restarting Core Process ...")
+        self.is_restarting = True
+
+        self.old_core_process_pid = self.core_process.pid()
+        self.should_quit_app_on_core_finished = False
+        self.shutting_down = False
```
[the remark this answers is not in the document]
xoriole: I have removed this part of the code. I agree with shutting down to be the final call, however, I've opted not to do the refactoring part of this PR. That could be a separate PR dedicated to simplifying the states in the CoreManager.

src/tribler/gui/core_manager.py
```diff
@@ -158,7 +189,7 @@ def check_core_api_port(self, *args):
             return
 
         core_process = self.process_manager.current_process.get_core_process()
-        if core_process is not None and core_process.api_port:
+        if core_process is not None and core_process.api_port and core_process.pid != self.old_core_process_pid:
```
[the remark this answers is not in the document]
xoriole: I have removed this check. This was mostly for Flatpak but now removed.

src/tribler/gui/error_handler.py
```diff
@@ -139,3 +150,10 @@ def _stop_tribler(self, text):
 
         if self.tribler_window.debug_window:
             self.tribler_window.debug_window.setHidden(True)
+
+    def _restart_tribler(self, text):
```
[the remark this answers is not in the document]
xoriole: Renamed to `_restart_tribler_core`

src/tribler/gui/core_manager.py
```diff
@@ -303,6 +373,9 @@ def on_core_finished(self, exit_code, exit_status):
         if self.shutting_down:
             if self.should_quit_app_on_core_finished:
                 self.app_manager.quit_application()
+        elif self.is_restarting:
+            self.is_restarting = False
```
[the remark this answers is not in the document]
xoriole: This part of code was to check if this finished was part of restart. To make it more explicit, I have defined another variable `wait_for_finished_to_restart_core`.

src/tribler/core/components/restapi/rest/events_endpoint.py
```diff
@@ -162,12 +178,14 @@ def on_tribler_exception(self, reported_error: ReportedError):
...
-        message = self.error_message(reported_error)
         if self.has_connection_to_gui():
-            self.send_event(message)
-        elif not self.undelivered_error:
-            # If there are several undelivered errors, we store the first error as more important and skip other
-            self.undelivered_error = message
+            self.send_exception(reported_error)
+        else:
+            if reported_error.should_stop:
+                self.exception_handler.save_to_file(reported_error)
```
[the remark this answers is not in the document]
xoriole: Good point! Used serialized_copy()

xoriole:
@kozlovsky I have addressed your review comments and would appreciate your re-review.

kozlovsky (approved):

PR #7714, opened 2023-11-22
Fix GH actions download and install libsodium windows dependencies

xoriole:
`Setup windows dependencies for Tribler` GitHub action has been failing for me on my personal repo test for PR. The failure was caused by the following line not working:
```
C:\msys64\usr\bin\wget.exe -q https://download.libsodium.org/libsodium/releases/libsodium-${{inputs.libsodium-version}}-msvc.zip
```
No zip file is downloaded.

This issue is not visible on the Tribler repo because the libsodium is used from the cache and not downloaded, so the code with the issue is not normally executed.

Here, I have replaced wget with PowerShell `Invoke-WebRequest` along with defining the TLS protocol support type. The rest of the changes are converting batch script to PowerShell script.

### DIFF
--- a/.github/actions/windows_dependencies/action.yml
+++ b/.github/actions/windows_dependencies/action.yml
@@ -29,17 +29,23 @@ runs:
 
     - name: Download Libsodium
       if: steps.check_libsodium.outputs.files_exists == 'false'
-      env:
-        RELEASE_FOLDER: .\libsodium\x64\Release
-      shell: cmd
+      shell: pwsh
       run: |
-        C:\msys64\usr\bin\wget.exe -q https://download.libsodium.org/libsodium/releases/libsodium-${{inputs.libsodium-version}}-msvc.zip
+        echo Downloading libsodium-${{inputs.libsodium-version}}-msvc.zip
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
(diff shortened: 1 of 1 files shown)


drew2a:
@xoriole, could you please elaborate on the underlying cause of the issue and how replacing wget with PowerShell's Invoke-WebRequest resolves it?

Previously, `shell: cmd` was employed universally, even in actions pertaining to Windows, as it provided a straightforward method of action description. This approach required only knowledge of cmd for understanding GitHub Actions. Introducing PowerShell necessitates familiarity with both cmd and PowerShell. Therefore, I'm curious to know if it is indeed essential to switch to PowerShell, or if the existing cmd-based approach could suffice.

xoriole:
@drew2a The underlying cause is likely permission issue on Windows runner machine:
```
C:\msys64\usr\bin\wget.exe -q https://download.libsodium.org/libsodium/releases/libsodium-${{inputs.libsodium-version}}-msvc.zip
```
The above command did not download libsodium zip file. Here is a Test PR reproducing the issue:
- https://github.com/Tribler/tribler/pull/7715

Here is a link to action run logs:
- https://github.com/Tribler/tribler/actions/runs/6961909598/job/18944520771?pr=7715

kozlovsky (approved):

drew2a:
@xoriole thank you for the explanation. This bug is fixed in #7605 (merged to `main`).

xoriole:
@drew2a thank you for the reminder. I'll close this one.

Issue #7669, opened 2023-11-06
[Draft] Channels Retirement


drew2a:
#### Core Side:
- [x] Flatten `metadata.db` structure (completed by @egbertbouman) 
- [ ] Streamline RemoteQueryCommunity functionality. Eliminate all remote query API, preserving only the metadata retrieval mechanism.
- [ ] Transfer the search engine to the PopularityCommunity
- [ ] Shift endpoints for the PopularPage to PopularityCommunity
- [ ] Transition `metadata.db` to the DatabaseComponent 
- [ ] Eliminate the GigachannelManagerComponent, GigachannelCommunityComponent, and MetadataStoreComponent
- [ ] Rename PopularityCommunity to ContentDiscoveryCommunity

#### Data Side (migration):
- [ ] Remove channel entities from the DB (?)

#### GUI Side:
- [x] Revise the main UI (completed by @drew2a)
- [x] Adapt existing widgets to Core modifications (completed by @egbertbouman)
- [ ] Overhaul widgets, dialogs, and models

Post-implementation, users will encounter an updated GUI:
<img width="888" alt="image" src="https://github.com/Tribler/tribler/assets/13798583/30143b37-72f5-441f-8bca-d44164e66e27">

On the core side, the PopularityCommunity will inherit the search feature from the ChannelCommunity.

egbertbouman:
Just in case anyone is curious, I have a branch [here](https://github.com/egbertbouman/tribler/tree/giga_cleanup). Please bear in mind that this branch was meant to see if we *could* easily remove the channels, and shouldn't be considered stable. It took about 3-4 days to make this, while losing 7500 LoC.

qstokkink:
@egbertbouman In order to get a PR out of your branch, it _seems_ like these are the remaining TODOs:

1. Rebase to `Tribler/main`.
2. Remove commented-out code.
3. Restore tests for `torrent_metadata.py`.

Is that correct?

By the way, I'm not trying to solve ALL the [bullet points listed in O.P.](https://github.com/Tribler/tribler/issues/7669#issue-1979375773) in a single PR. I'll leave the rest for (a) follow up PR(s).

qstokkink:
@egbertbouman Cool, thanks 👍 That sounds like it should be a relatively quick job then ("famous last words"). I'll get started on this now then.

qstokkink:
The initial cut has now been merged in #7726. I'll move on to the (hopefully) smaller points now.

[...]

PR #7726, opened 2023-11-27
GigaChannel removal

### LINKED ISSUE #1368: Backing up a channel
es{Desearía ver que en la siguiente version de "tribler", se pudiera hacer una copia de respaldo a un canal propio o con el número de identificador se pudiera pasar de un "tribler" a otro "tribler" la administración de un canal.

Discúlpeme por mi ingles.}

en{Would like to see in the next version of "Tribler" could be done a backup to own channel or ID number you could pass a "Tribler" another "Tribler" the administration of a canal.

Excuse me for my English.}

### LINKED ISSUE #2284: Refactor the add torrent to channel endpoint
I would like to propose two changes to the add torrent to channel endpoint:
- We can easily pass the torrent file data using a HTTP multipart request instead of passing the base64 data of the torrent file. I've already done this here: https://github.com/devos50/tribler/commit/2d0efe15e3a12d2c231eb021b600447b269d77b2. After that, we can easily add torrents to your channel using something like:

```
curl -X PUT -F "file=@/path/to/my/file.torrent" -F "additional_data=bla"
```
- Currently, when adding a torrent to your channel using a url/magnet, we are making a PUT request to `mychannel/torrents/{theurl}`. This is problematic since I currently assume that torrents are identified by their infohash and not by their url. I think it might be better to add an extra field to the request when adding a torrent file to your channel, specifying the source (file/url/magnet).

@brussee what are your thoughts on this?

### LINKED ISSUE #3615: Redesign of the Search/Channels feature
We're about to ditch the last part of Dispersy - the Search/Channels feature. A new Community based on IPv8 will provide this functionality instead.
As we're not going to make this Community backward compatible with Dispersy, there is an opportunity to reimagine the Search/Channels concept to fix it's current shortcomings, and make it suitable for the future developments.

### Use cases.
We strive to provide the same kind of service that is provided by:
 * Highly organized repositories of torrents (trackers), like rutracker.org
 * Scientific articles repositories, like SciHub.
 * Music playing services, like Spotify.
 * Video playing services, like Youtube.
 * Scanned books archives, like Google Books.
 * Internet libraries, like lib.rus.ec

### On fuzzy searches.
We do not try to become the distributed Google. Google is good at finding knowledge, that is relevant to some words. It is used when the user _does not know_ what he or she is looking for. Our user _always_ knows what he or she wants. Therefore, we should disable metadata (or torrent contents) search by default, and *only search by the torrent name*. On the other hand, the user should be able to use metadata to _refine_ the search results. Like, "sort by date of creation" or "sort by tag/channel", etc. This model already works perfectly for all use cases indicated above (except Youtube).

### Data structurization.
We must be able to replicate the basic information organization structures that are used by our example use cases:
 * **Tracker forum** (rutracker.org): tree-like structure, traversable both up and down. When the user finds a torrent/channel, he or she should be able to see its immediate parent/children nodes, and the "root"(owner) of the hierarchy this channel is part of. Thus, **nested channel**.
* **Scientific articles database**: a single record in the database should be able to hold an arbitrary amount of metadata fields, and **each field could provide its own record hierarchy**. For example, top-level owner/channel "SciHub" could provide "pseudochannels" for "authors" "journals" "years" "tags", etc. These are essentially selecting the metadata field that would be used to **build the structure "on the fly"**. Advanced users would be able to select several criteria at once, to find intersection of metadata sets. Thus we replicate relational database with nested channels architecture.
 * **Music search**: should be fast and simple. It could be only achieved by implementing a **distributed cloud-based search**, with some data cached at user's device. We can't get new user do download XX Gigabytes search database, especially for mobile usage (that is basically everything for music playing). One search query should take 5-30 seconds, and it's results should include related info (like, the search for a single song should get that song _and_ reveal the link to the whole album). To solve this problem, at some point we could introduce **search mining**. Fast hosts producing highly relevant search results get credit.
* Libraries and book archives fall somewhere between scientific articles and tracker forum.

### Constraints on design choices

The design should consist of several _completely independent_ parts:
* Networking subsystem
    - Data transfer protocol
    - Trustworthy gossip protocol
 * Storage subsystem
    - Database/store
    - Cache management / mining algorithm
 * Search subsystem
    - Distributed search protocol
    - Distributed search algorithm
    - Data distribution algorithm
 * Metadata format

It is *very* important that we do not make any of these parts depend on specific **implementation** of any other one. We must be able to design the system in such way, that it would be trivial to exchange one implementation of some part for another one.

### LINKED ISSUE #4637: GigaChannel Manager restarts personal channel download each time Tribler starts
On startup, GigaChannel Manager checks for the existence of personal channel download. If it does not exist, it will restart it. Unfortunately, Libtorrent Wrapper starts downloads asynchronously, so there is a race condition between these two processes. If GigaChannel Manager wins the race, it will re-download the personal channel download.

### LINKED ISSUE #4645: Show precise number of torrents in the channel

### LINKED ISSUE #4650: Efficient order of checks processing payloads in GigaChannel
At the heart of GigaChannel is the [`process_payload`](https://github.com/Tribler/tribler/blob/01c88f20c13552035f78418206f41fcb037f7508/Tribler/Core/Modules/MetadataStore/store.py#L360) method, which acts as a firewall for all incoming GigaChannel data.

This method features a lot of checks that evoke costly SQL queries for each processed payload, such as checks for duplicates, old versions, etc. The order in which these checks run does not affect the fate of the packet (they all just drop it). However, some orders can be much more efficient in terms of SQL requests. The higher the probability of the check, the higher it should be moved in the checking order, so we do not waste i/o on rarely used checks.

One way to solve this problem is to gather some statistics on what checks trigger most often, and then reorder the checks accordingly. Unfortunately:
 1. we don't do telemetry on our users.
 2. in some modes, like uploading data from a channel torrent, patterns of triggered checks are very different from the average.

This problem reminds the problem of scheduling operations in the CPU conveyor and can be solved accordingly, by introducing a *dynamic* ordering of the checks, based on some recent stats. The ordering can be determined by, e.g. VSIDs heuristics.

### LINKED ISSUE #4672: Channels 3.0 - small files support
Channels 3.0 provides the required technological foundations for adding numerous small files to existing Bittorrent videos.

A lot of our cultural heritage requires multilangual support in the form of subtitles, thumbnails for visual navigation and screen art to provide stunning backgrounds. Popular channels of libraries or vlogger collectives might require a million support files. This issue provides a single swarm to serve them all. Each 3.0 channel has a single unique hash for all support files. This hash is updated regularly and all subscribed clients incrementally seed their content.

### LINKED ISSUE #4677: GigaChannel: BitTorrent is not enough
**Long story short**: metadata requires crowdsourcing, which requires low-latency delivery system, which BitTorrent can't provide. So,  *in addition* to BitTorrent, we need something like a DHT-based pub-sub.

### Motivation
Our primary objective is to provide *relevant* information to humans. Humans should be able to process it efficiently. Human's information processing capabilities are very limited, and they can't process more than a few dozens of lines of text at once. No one ever looks beyond the 3rd page of Google. Therefore, we must limit the information that we show to the user. There are three primary ways to do this:

1. Filter the information on some criteria (text tokens/tags/type, etc.).
2. Sort it based on some criteria (popularity/date/size, etc.).
3. Organize it in balanced trees (ontologies), so each level is small enough not to overwhelm the human.

All three ways work together nicely, complementing each other. For example, one can **search** for a word `foo` in a local database, **sort** it based on recency, click on a most promising entry and then **browse** the collection holding it to look for similar entries.

### Knowledge creation
People add information to the system. The information comes in the form of independent entries, possibly organized into collections. Every person has their domain of information facilitated by the public key infrastructure, so no one has (direct) power over other's creations.

A person can either add some original information (e.g., a personal podcast) or copy it from others. Humans can't meaningfully produce more than a dozen original entries per day. Therefore, the influx of truly original content per person is minimal. However, when a person copies stuff in their collection and shares it, they effectively produce new information. **The act of selection is an information-producing event.** We will call this information *the grouping information*.

### The problem of duplicates
When one can't browse personal channels, there is no grouping information. Therefore, if entry *E* comes both from peers *A* and *B*, it makes no sense to store it twice, and the second one can be dropped on receiving it into a local DB. As soon as we start to account for *any* kind of grouping information, cutting entries means losing/distorting this information. Of course, we can store database relationships instead of duplicate entries themselves. This storage scheme will help with the indexing, but will still result in the same *O(n)* linear storage requirements. Thus, **we must drop grouping information based on some criteria, lest it overwhelms the system.**

### Ontology tree balancing
When users can only produce single-level channels, it is impossible to build an ontology that includes more than about a 100^2 entries, even if all users cooperate on this. A human user can't look through more than a hundred channels, and can't look through more than a couple hundred entries in a channel. **A perfect ontology is a balanced tree** (or even better, a perfect encoding). Thus, users must be able to create multi-level channels, which means **more grouping information**. 

### Group selection
When users figure out their grouping information is dropped when it clashes with others, they will start organizing in communities to **coordinate**  their efforts. This organization will require some instruments for collective authoring and communication.

### Scaling
At some point, there will be channels with *thousands* of collective authors and sub-channels (collections). These collections will be updated a few times per day, which will result in **several updates per second** for the root channel. This is essentially the same problem that is faced by the Bitcoin ledger. Therefore, there should be only loose connections between the root channel and the sub-channels.

### The BitTorrent problem
The power of torrents comes from it exploiting "the network effect": one can start sharing the data as soon as one got at least one piece of the torrent. This is made possible by *hashing the data as a whole*, so anyone can check any piece for the correctness and immediately share it. Even with "mutable torrents" support, sharing a single collectively-authored channel means  either:
a. seeding thousands of tiny torrents for each sub-channel;
b. downloading the updated version several times per second, while simultaneously serving thousands of previous versions.

As a system, BitTorrent was never intended for this kind of usage: low latency updates, low volume transfers.

### The solution
BitTorrents should only be used for **bulk** transfers of big channels. The channel infohashes should be automatically updated, say, daily. For the online updates, we should employ something like a DHT-bases pub-sub system (e.g. [PolderCast](https://hal.inria.fr/hal-01555561/document)). Then, two systems will together cover all the bases:
* Pub-sub: low-latency, low throughput;
* BitTorrent: high-latency, high throughput;

### LINKED ISSUE #5015: "New arrivals" feed

### LINKED ISSUE #5067: Discuss how the number of items should be displayed besides the breadcrumb in the GUI
As discussed in #5056:

```
Well, channels/folders being marked as torrents is not a bug. It is correctly showing the total number of objects in the current view. We need to discuss this thing from the UI standpoint.
```

### LINKED ISSUE #5154: Identicons for users and channels' authors

### LINKED ISSUE #5271: Content not added to channel if metadata lookup fails
It seems that if metadata lookup fails, the torrent will not be added to your channel. Instead, it shows an error that the meta info lookup has failed. This makes it impractical to add content to your channel that is currently not being seeded/healthy, unless you have generated the torrent yourself.

I tried it with the following magnet link:

```
magnet:?xt=urn:btih:55b9043a942dc075e97807b54741a0fc7740b465&dn=RandomStuff
```

Furthermore, after adding the torrent to your channel, there is no visual feedback that the metadata of the torrent is currently being fetched from the network. This is confusing since it might seem like the operation has failed. However, this issue is less

### LINKED ISSUE #5522: Various serialization numbers (REGULAR_TORRENT, CHANNEL_TORRENT, ...) directly used by the GUI
In the GUI, we are directly importing stuff from the Tribler core (in `table contentdelegate.py`):

```
from tribler_core.modules.metadata_store.serialization import CHANNEL_TORRENT, COLLECTION_NODE, REGULAR_TORRENT
```

These variables should move to the `tribler_common` module since they are both used by the core and GUI.

### LINKED ISSUE #5587: The Plan, updated
(Continued from #5348)

## Tribler Mission
Our mission is to build a completely decentralized, anonymized, impossible-to-shutdown **information market**. The market will provide anyone willing to participate in it an opportunity to serve the community in the following ways:
 * seeding rare torrents
 * building ontologies
 * maintaining indexes
 * fighting spam
 * writing reviews
 * publishing original content
 * providing anonymity services (e.g. running a relay or an exit node)

### Incentivizing social good
Serving community is incentivized by assigning a "social score" to participants. The better your score, the better service you get. If your score is zero, you become "bottom feeder" who can only use surplus resources not claimed by people with higher score. Effectively, this creates "information abundance economy" where spare resources are given to the community for free, as a kind of "information welfare".

### Author payments
The system will provide fairness to artists. Anyone can send tokens to any address in the Tribler network, thus creating a donation-based system where authors can directly receive payments from their fanbase.

## Minimum Viable Product

The minimum architecture to get the network going includes the following components:
 * :heavy_check_mark: decentralized networking library (IPv8, :+1: )
 * :heavy_check_mark: torrenting library (`libtorrent`, :+1: )
 * :heavy_check_mark: tunnels and hidden services systems (too slow :turtle: )
 * traffic accounting mechanism (Trustchain :arrow_right: Bami)
 * social score calculation and prioritising (TBD)
 * content catalog, preview, discovery (Channels+PopularityCommunity, WIP, too slow :snail: )
 * content keyword search (Channels, stub, too narrow)
 * content collective authoring (Channels, TBD)

### 7.6
(Shipped December 2020)

| Component           | Tasks                                                                                                                                                                 | People                                           |
|---------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------|
| Channels            | Asynchronous queue<br>Fast previews via Bloom filters neighborhood<br>Improve channel processing w/o changing format<br>OR<br>implement `ATTACH`-based channel format | @kozlovsky<br>@ichorid                           |
| Torrents popularity | Improve torrent popularity gathering,<br>add separate "Now trending" view to the GUI                                                                                  | @kozlovsky<br>@drew2a<br>@xoriole                |
| Channels GUI        | Change channels GUI to be more intuitive, resembling popular services                                                                                                 | @ichorid<br>@drew2a<br>@devos50                  |
| GUI                 | Add tree-based torrent contents view<br>Solve "create torrent" screen problems<br>Solve minor GUI defects                                                             | @xoriole<br>@ichorid<br>@devos50<br>@drew2a      |
| Infrastructure      | Solve Pytest migration consequences<br>Make tests run in parallel<br>Use real Core for GUI tests<br>Stabilize flaky tests<br>Integrate nice view for failed tests     | @devos50<br>@ichorid<br>@egbertbouman<br>@drew2a |
| Dashboard           | "Trending torrents" view in the Dashboard<br>Exit nodes, bootstrap nodes, infrastructure health stats                                                                 | @drew2a<br>@xoriole<br>@kozlovsky                |

The main goal for 7.6 is to produce an imperfect, but usable application, by addressing the most pressing issues, while also opening up a road to future experiments.

### 7.7
(Shipped April 7.7)

| Component    | Tasks                                                                          | People                              |
|--------------|--------------------------------------------------------------------------------|-------------------------------------|
| Channels     | Collective authoring<br>Individual files support (e.g. music albums)           | @ichorid<br>@grimadas<br>@kozlovsky |
| Channels GUI | Collective authoring GUI<br>Individual files support GUI<br>Generative avatars | @drew2a<br>@ichorid<br>@xoriole     |
| Tunnels      | Zero-copy endpoint in low-level language                                       | @egbertbouman<br>@kozlovsky         |
| GUI-Core     | GUI running independently of Core<br>Tribler as Windows Service                | @xoriole                            |
| Content      | Content bootstrapping                                                          | @drew2a                             |
| Economy      | Social score mining via "rare torrents support"                                | @grimadas<br>@xoriole               |

The goal for 7.7 is twofold: enable the Collective Authoring system for Channels and radically improve Tunnels performance by moving the hot :hotsprings:  parts of tunnels data processing to an endpoint written in a lower-level language (like Rust or Go).
It is also possible to split these two goals into two smaller releases.

### 7.8

The focus for 7.8 should be remote keyword search, semantic Channels overlay and "torrent market" with different prices per MB for different torrents. Another focus is i18. Also, donation system should work by that moment.

### 8.0?

By the second half of 2021 we should plan to hire outsourcers to completely redesign the GUI to modern standards. The finished GUI will become a major compelling point to release **Tribler 8.0**. This should be a very good moment to "whitewash" our reputation and start a media campaign.

Everyone is welcome to discuss and suggest changes!

### LINKED ISSUE #6126: GUI indicator for preview / subscribed unclear
See the following screenshot:

Schermafbeelding 2021-05-24 om 09 33 40

Since the word 'preview' is directly to the left of the slider, it makes me think that I should toggle the slider to enable preview. Hovering over the slider, however, shows that the slider is for subscription.

One potential fix would be to rename the word 'preview' to 'subscribe'. We might also remove the word 'preview' and use icons to the left of the slider. Then I suggest to also add an icon to the left of the popularity indicator.

### LINKED ISSUE #6208: [Wireframes] CR to a channel
Naming:
* Collaborator: a user who wants to create a change to a channel
* Owner: the channel's owner
* Change Request (CR): a changeset for the channel created by the collaborator

Restrictions:
1. A collaborator can have only one CR for one channel at a time
2. Only subscribers can become collaborators

## CR Feature 

CR has to be implemented as a feature with two states:

* enabled
* disabled

A channel owner can either enable or disable the feature for his channel.

Every CR can be in one of the following states:

* created
* pending
* canceled
* added

#### CR Feature enabled: 

* a user see a button "Add Torrent"
* an owner see a list of CR
* the owner can approve or decline every CR
* the owner can disable the CR feature for a channel

#### CR Feature disabled: 

* a user does not see a button "Add Torrent"
* an owner does not see a list of pending changes
* the owner can enable the CR feature for a channel

 ## Wireframes

Left column: A subscriber's interface.
Right column: A channel's owner interface.

PR

### LINKED ISSUE #6217: Sybil-resistant channels crowdsourcing system (VSIDS)
The channels crowdsourcing system is based on a concept of a change request (CR) to a channel. Each CR **must** be confirmed by the channel's owner. This is really easy to achieve in a centralised system with fixed identities, high availability, and no risk of a Sybil attack.

### Problem statement ###
Design a system for suggesting atomic changes (CRs) to a Tribler Channel, that will be:
* resistant to spam and Sybil attacks (i.e. whitewashing, collusion, etc.)
* resistant to node churn (low node downtime)

### The Solution ###
Reuse the existing VSIDS-based Channels rating system. Rank CRs inside a channel the same way we rank channels, ask users if they "_recommend_" :+1:  a CR analogous to how they subscribe :heavy_check_mark:  to a channel.

#### Long description ####
Each CR is an atomic change to a particular Channel entity, such as adding or renaming a torrent or a channel folder. There are three kinds of participants in the system:
* Channel owner (CO)
* Suggestion author (SA)
* Channel subscriber (CS)

The system works as follows:
1. SA subscribes to a channel (C) created by CO. SA then creates a CR and stores it into the local Channels database.
2. CS's GigaChannelCommunity(GC) walks randomly to another peer. On introduction, GC asks the introduced peer: "do you have any _recommended_ PRs into channel C". Then, CS **pulls** the CR into CS's local database.
3. CS opens Tribler and sees that there is a new CR. CS "likes" :+1:  or "dislikes" :-1: the CR.
4. If CS :-1:  the CR, CR is immediately deleted from the database and its hash is added to a blacklist.
5. If CS :+1: the CR, CR becomes "_recommended_" in the local database.
6. If CR is already known to CS, and is not on a blacklist, it is bumped :arrow_up: according to VSIDS heuristic **local to the channel**.

This way, CO gets to know about CR eventually. If CO keeps their Tribler instance running long enough, they will get a good picture of what CRs are supported in the community of the channel, and which are not. If CO likes the CR, they will accept it and as soon as the channel is updated for all subscribers, CR entries will get deleted/merged in CS databases. If CR is rejected, it will die out eventually. 
In this system, CS-es will work as both a crowdsourced spam filter, change voters and intermediary CR storage. As a result, it is in everyone's best interest to keep their Tribler instances walking as long as possible: COs get more stats, CS-es get updates, influence and engagement, while SA's keep pushing their CRs to the community.

Note the system is very gentle on everyone's attention span: no one _have_ to work through tons of queued up :hankey: to get to the relevant stuff. Instead, there is a constant, community-curated **feed** of suggested changes sorted by relevance.

### Implementation ###
The system will reuse the existing components: MetadataStore, VSIDS, RQC. On the Core side, it will require minimal changes, such as support for per-channel VSIDS ratings. On the GUI side, it will require adding a sorted feed of proposed changes and the CR interface such as described in #6208

### LINKED ISSUE #6395: Content appearing and disappearing when refreshing a channel

### LINKED ISSUE #6481: Vadim's testament
Related to https://github.com/Tribler/tribler/issues/143

## Preface

Over the last few years, we successfully solved [every major technical problem](https://github.com/Tribler/tribler/issues/4566) inherited from previous Tribler dev generations. We radically updated Tribler codebase, established a [robust code architecture](https://github.com/Tribler/tribler/pull/6206) and [brought code support up to industry standards](https://github.com/Tribler/tribler/pull/5735). Also, we've cut down on [every](https://github.com/Tribler/tribler/pull/5350) [non-essential](https://github.com/Tribler/tribler/pull/5390) [feature](https://github.com/Tribler/tribler/pull/5549), focusing development on two things only: metadata delivery and anonymous downloads.

During this journey, we identified a number of technical and scientific problems that block the Tribler project from reaching its goals. We tended not to tackle those in fear of bogging down our understaffed team, focusing on "low-hanging fruits" instead. 

Gentlemen! :smoking:
I inform you that all the low-hanging fruits are gone, only the hard ones remain. Here is the list.

## Network-level problems
Solutions for these are obvious, but we never put enough effort into these, because we never had enough qualified manpower.

<details>
<summary>:no_entry_sign: DHT calls blocked over tunnels :no_entry_sign:</summary>

BitTorrent uses Mainline DHT to find nodes that seed an infohash. Mainline DHT is susceptible to various types of attacks, including DDOS. To solve this problem, BitTorrent libraries use spam control methods, blocking peers that send too many requests. Different clients employ different criteria for detecting DDOS attempts. The problem is Tribler DHT requests are all sent through exit nodes, which may look like a single node to DHT peers, [triggering spam control](https://github.com/Tribler/tribler/issues/3065). The result is ~half-dead~ impaired torrent info fetching and unreliable health info.
One solution to this problem could be caching or performing DHT requests on exit nodes on behalf of tunnel users. However, this could result in non-technical problems :copyright: :policeman: 

</details>

<details>
<summary>:snail: Slow tunnels performance :snail:</summary>

Our anonymous tunnels code is implemented in Python (though the crypto library is low level). [Performance is very, very bad](https://github.com/Tribler/tribler/issues/2548) compared to VPNs: we do 5 MBytes/s _at best_ and 0,5 MBytes average for an overseeded torrent on a fast modern PC, while a typical VPN (such as Wireguard) will use the whole bandwidth available to the host (about 20-80 MBytes/s). ~~The reason is Python[ copies strings on slicing](https://github.com/Tribler/tribler/issues/4459), resulting in useless waste of memory and CPU cycles~~([solved](https://github.com/Tribler/tribler/issues/6481#issuecomment-949361098)). The reason is slow data exchange between Python lower-level libraries. The solution is to implement a shim IPv8 tunnels - SOCKS endpoint in a [lower-level language](https://github.com/Tribler/tribler/issues/4567).

The problem is exacerbated by Libtorrent's [bad performance](https://github.com/Tribler/tribler/issues/2620) when [using uTP](https://github.com/arvidn/libtorrent/issues/3542).
</details>

<details>
<summary> :two: Support for BitTorrent 2.0 :two: </summary>

LibTorrent 2.0 is actively pushing BitTorrent 2.0 standard, which moves to a more secure, 32-byte SHA-256 hash. Supporting it will touch *every* part of Tribler codebase, as 20 bytes hashes size is hardcoded and expected everywhere.

</details>

## [Token economy](https://github.com/Tribler/tribler/issues/3326) 
Currently, our token economy does not do anything useful, but instead just perplexes the users and annoys the developers. Here are the reasons:

<details>

<summary>:hole: The exitnode blackhole problem:hole:</summary>

When Tribler downloads something through an exit node, the user pays it the corresponding amount of bandwidth tokens. The problem is, 99.99% of seeds are non-Tribler, meaning that exit nodes pay no one. Essentially, exit nodes act the role of super-seeds for the network, but they _never spend their tokens_. The result is, exit nodes become "supermassive black holes" of Tribler economy, constantly dragging regular users to a negative balance. And negative numbers piss off people, incentivizing them to either stop using Tribler, or just regularly delete their identities to [whitewash their balance](https://github.com/Tribler/tribler/issues/4015).
There is another problem adding more complexity to the issue: when someone from outside the Tribler network exchanges traffic with a Tribler hidden seeder,[ or just an anonymous downloader](https://github.com/Tribler/tribler/issues/5310), the traffic leaving the Tribler network will never be paid back, ultimately making the economy deflatory.
And no, the simple solution of prioritizing Tribler peers will destroy performance: instead of a hundred fast peers, we would use a single slow one.
One possible solution is to stop showing balances altogether and instead, [show the user's relative ranking.](https://github.com/Tribler/tribler/issues/3495)

</details>
<details>
<summary>:-1:  Hidden seeding is ~broken~ useless :-1:</summary>

~[Receiving UDP over SOCKS is broken in Libtorrent](https://github.com/arvidn/libtorrent/issues/6512)~
When a Tribler starts seeding a torrent in "hidden seeding mode", the torrent will only be available through exit nodes. Unexpectedly, the seeding ratio of the hidden torrent will always be near-zero. The reason is: BitTorrent protocol prefers the :racing_car:fastest:racing_car: seeds, but hidden seeding is always :turtle:slower:turtle: than direct seeding.
The result is, hidden seeding does not help with recuperating the bandwidth tokens user spent for anonymous downloads, further breaking Tribler token economy. This is a vicious circle: Tribler users don't seed because no one pays them for seeding, because there is no incentive for Tribler users to download from other Tribler users.
</details>

<details>
<summary>:arrow_up: Prioritization of users is semi-functional :arrow_down:</summary>

Currently, there is just a single mechanism of prioritizing users on exit nodes: if the user's balance goes too low, the user will have a lower probability of getting service. (@devos50 , correct me if I'm wrong) The mechanism is very primitive, barely working, and is trivial to circumvent by whitewashing.
</details>

<details>
<summary>:champagne:  The exitnode bottleneck and inflection to self-sufficiency :champagne:</summary>

A simple back-of-envelop calculation shows that for the Tribler network to become self-sufficient, there should be about ** one million** Tribler users online at every given moment. The inflexion point is 500 000 users: after that, most of the traffic will be served from inside the "Tribler bubble":
inf

However, even then **all** the traffic will have to go through the exit nodes. The reason is, in the current architecture of Tribler anonymization network there is no such thing as "hidden-seeding-only" exit nodes. I.e. if there is a million users and no exit nodes, hidden seeders will not be able to connect.
The trivial solution is to add a special class of "pseudo-exit-nodes" that only allow connections to hidden seeders. (@egbertbouman correct me on this if I am wrong.)

</details>

<details>
<summary> :moneybag: Credit mining fiasco :moneybag: </summary>

In an attempt to bootstrap Tribler into a self-sufficient ecosystem, we tried to implement a ["Credit mining" system](https://github.com/Tribler/tribler/issues/4363) that should have allowed users to "invest" some disc space and traffic into seeding Tribler torrents to get token rewards. Unfortunately, the only thing it provided to the users is [a constant stream of lost tokens and disappointment](https://github.com/Tribler/tribler/issues/3778). Eventually, we removed this feature.
The reasons why it failed are multiple (e.g. the hidden seeding problem described above), but the ultimate one is: **BitTorrent is a non-zero-sum game**. If everyone is using the same algorithm and downloading the same popular torrent in hopes to profit from it, every megabyte of the torrent they provide to others play against them. In fact, the simplest analysis shows the series of "wins" for a torrent starting from a single peer resembles a harmonic series, grows incredibly slowly because of [diminishing returns](https://en.wikipedia.org/wiki/Harmonic_series_(mathematics)).
The solution to this problem is two-fold:
1. devise different token prices for different torrents
2. stop trying to replicate the money economy and instead come up with a social rating system

</details>

<details>
<summary> :money_mouth_face:  Deanonymization by payouts :money_mouth_face: </summary>
In the current architecture, we do payouts immediately. In combination with an open ledger, this could allow deanonymizing people by their traffic patterns.
Solution: implement [deferred (and possibly fuzzy) payouts](https://github.com/Tribler/tribler/issues/4255).

</details>

<details>
<summary> :handshake:  Unite the economies of Metadata, Anonymity, and Seeding :handshake: </summary>

If Tribler is ever going to reach its goal of creating **the** [attack resilient economy for media](https://github.com/Tribler/tribler/issues/1), users must be able to provide value and trade different kinds of services in it. There are three primary ways how a user can benefit the media-economy:
1. provide anonymization services (being an exit node or an intermediary peer)
2. provide seeding services (storing data for rare torrents)
3. provide metadata enrichment services (create and maintain channels, categorize metatada, add tags, etc.)

**All three kinds of activities must reside in the same, single economic space**. This means either using a single token to reward all three kinds of activities, or creating three different systems of tokens that can be traded on a free market.

</details>

## [Content layer](https://github.com/Tribler/tribler/issues/3615)
Due to numerous social engineering mistakes, design errors and architectural miscalculations, the current Tribler content layer ["Channels 2.0"](https://github.com/Tribler/tribler/issues/3615) failed to reach its goal of becoming a decentralized alternative to web-based BitTorrent trackers.

<details>
<summary> :dizzy:  BitTorrent is unsuitable as the Channels backend :dizzy: </summary>

Big torrents collections cannot be created or maintained by a single user. Therefore, the bigger the collection, the more people a required to maintain it and edit it simultaneously. If the data is stored as a torrent, each change results in creating a new infohash for the whole collection, eventually leading to **swarm fragmentation**.  Thus, collaborating on, or just regularly copying data from a large external source becomes nearly impossible. Clearly, [BitTorrent is unsuitable ](https://github.com/Tribler/tribler/issues/4677)as collaboration platform backend.
Also, using torrents as a backend involves pretty complex logic for packing the data into append-only files and dealing with asynchronous events from an external entity (Libtorrent). 
The solution to this (and other) architectural problems would be abandoning BitTorrent as the Channels backend altogether and instead [fetch data on-demand from other peers](https://github.com/Tribler/tribler/discussions/5721). [Migrating to BitTorrent 2.0 will not help](https://github.com/Tribler/tribler/issues/4672), because the problem is not storing many small files _per se_, but the rate of change being proportional to the number of contributors of a channel (e.g. **quadratic** if every user is a contributor).

</details>

<details>
<summary> :pencil2: No crowdsourcing instruments :pencil2: </summary>

The ability for users to [author](https://github.com/Tribler/tribler/issues/31) [metadata](https://github.com/Tribler/tribler/issues/2455) [cooperatively](https://github.com/Tribler/tribler/issues/4642) is a critical [requirement](https://github.com/Tribler/tribler/issues/134) for Tribler Project to reach its [goals](https://github.com/Tribler/tribler/issues/1).
Channels 2.0 system was initially designed as permissioned, except for the top-level federated channels list (the "Discovered" tab). The plan was to begin from permissioned and then add [permissionless crowdsourcing elements](https://github.com/Tribler/tribler/issues/6217), such as the ability for users to create "[pull requests](https://github.com/Tribler/tribler/issues/6208)" into other's channels. Unfortunately, we became distracted by [gigantomania](https://github.com/Tribler/tribler/issues/21) and [non-essential features](https://github.com/Tribler/tribler/issues/5977).

The [upcoming tags system](https://github.com/Tribler/tribler/issues/6214) is a step in the right direction, although its decision to not use the Channels 2.0 backend pose the question about how those two systems are going to be integrated in the future.

Also, users must be able to [communicate with each other ](https://github.com/Tribler/tribler/issues/6043)to discuss information organization. No crowdsourcing system could exist without discussion between participants.

</details>

<details>
<summary> :mag: Channels search is lame :mag_right: </summary>

Our Channels search algorithm is very simple and inefficient: ask five random neighbours for some results on a keyword. That's all. No dynamic deepening of the search, no walks, no deduplication, no popularity concerns, no indexing. Just :five: random hosts :shrug:  This is a clear obstacle for the Tribler content layer to become usable. One thing that probably saves us at the moment is the proliferation of [Free-For-All (FFA) entries](https://github.com/Tribler/tribler/issues/3615#issuecomment-651091413) due to local caching when users search for popular keywords.

Some related issues: https://github.com/Tribler/tribler/issues/2250 https://github.com/Tribler/tribler/issues/2547

</details>

<details>
<summary>:zombie: Brain-dead data transfer: 99.99% of Channels data unused :zombie:</summary>

Channels 2.0 design is based on transferring the log of signed channel changes and then replaying it on the user machine to put those entries in the user's local DB. The processing is very slow and inefficient for bigger channels and unreliable for smaller channels. An analysis of possible solutions to the transfer problem [shows](https://github.com/Tribler/tribler/issues/3615#issuecomment-651091413) that there is no good design at all, **if we are downloading full channels**.
The real problem is **data integration**: there are no databases that allow transferring and merging indexes in sub-linear time. Also, a human being is only interested in a **handful** of torrents **from a million-torrents** channel, wasting 99.99%. Moving everything around is a brain-dead :zombie:  waste of bandwidth, CPU cycles and users' time. Google does not ask the user to download the whole index before usage, really :wink:
The solution is to [spread the data around in the network and fetch it dynamically, using the network itself as an index](https://github.com/Tribler/tribler/discussions/5721).

изображение

(Be warned that that transferring complete SQLite DB's is no option because of security issues.)

</details>

<details>
<summary>:popcorn: "Popular" suggestions are 95% garbage :popcorn: </summary>

The "Popular" tab is served contents by Popular Community. There are multiple problems with how that works: first of all, the info is not propagated transitively, for the fear of spam. Also, Popular Community uses the push-based gossip model, which still keeps the overlay susceptible to flood-spam attacks, but does not allow for the "initial boost" feature of pull-based gossip (that makes Channels discovery usable). The push-based model also prevents us from doing aggressive walks for research purposes, as we did with TrustChain and Channels. Also, there is no bias for newer torrents in the "Popular" tab. The result is, "Popular" tab only shows 2-3 torrents that are really popular at the moment: the rest are typically  2-15(!) years old torrents that showed some great number of seeds at that moment, and for some reason (probably due to DHT bugs) still sometimes show high number of seeds. The problem is exacerbated by the fact that BEP33 checks for seeds are unreliable, and there is basically no way to tell the real number of seeds for a torrent without connecting to the corresponding swarm. In Tribler, these connections always go through exit nodes, which often results in DHT spam filter triggers and unnecessary exit nodes load.
The problem of popular torrents is a complex one. Basically, we must design a distributed algorithm for collectively checking a set of entries (infohashes) and sorting those entries dynamically based on a dynamic property (number of seeds). And don't forget the exit nodes bottleneck and the constant danger of spam!

Some solutions to this problem could be:
* cache health data on exit nodes (dangerous due to non-technical problems)
* establish a separate class of "health-checker supernodes".
* stop propagating health data and instead go for some relative-popularity, time-based heuristic like VSIDS
* switch to pull-based gossip and add dynamic boost based on the state of the local database
* split the health checking work between neighbouring peers in a semi-structured way

</details>

<details>
<summary>:lips::straight_ruler:  Family filter is Victorean-nun-meets-Nazi overzealous :lips::straight_ruler: </summary>

Originally, our [Family Filter](https://github.com/Tribler/tribler/issues/1052) was a quick hack made of bag-of-words-grep from some Dutch porn site. That's early-80s state-of-the-art! We **definitely** need something smarter, e.g. [Bert](https://en.wikipedia.org/wiki/BERT_(language_model)).

</details>

<details>
<summary>:cat2::cat2: "Just show 100 most popular torrents" will never work :cat2::cat2: </summary>

Content popularity for file-sharing networks and the Web is fundamentally different: file-sharing is much more **flat** at the top.
Essentially, there are no "Google" or "Facebook" among torrent files. The reasons are many:
* movies are available in many languages
* movies are available in many resolutions
* movies are fast-lived, popular only for a short amount of time
Basically, file-sharing is about _streaming_ media: games, shows, movies, music - and the media really _streams_. Torrent popularity is short-lived. Thus, the strategy of "let's just show 1000 most popular things" will never work with torrents: **people's taste for movies is much more diverse than their taste for websites.**
Also, torrent collections are products of specific **communities**. Copying content will not copy the associated community: the collection will remain dead :skull:. Best case, if some algorithm would be copying contents continuously, Tribler will forever remain just a mirror of those web-based trackers, like the Internet Wayback Machine. Do people use that often? How many people even now of the Internet Archive?
Moreover, copying content from another platform disincentivizes Tribler users from creating functional channels and communities around those. Users just "choke" on those big piles of content, unable to change those or use in their own projects.
изображение

The way to solve this is to acknowledge that user communities [co-evolve](https://en.wikipedia.org/wiki/Enactivism) with the content they produce and their environment. Instead of focusing on copycatting :cat2: :cat2: contents from others, we should focus on developing efficient crowdsourcing tools for the Tribler community. One can spend infinite amounts of gas trying to start a fire - it will never become :fire:self-sufficient:fire:  if the logs are :droplet:wet:droplet:.

</details>

## User interface

Aside from the usual [discussion](https://github.com/Tribler/tribler/discussions/6525) about using Web stack instead of QT, our GUI has a **looooooong** way to go in regard to style and usability...

<details>
<summary>:goberserk: UI looks :goberserk: </summary>
In general, Tribler UI looks lame and outdated, like a thing made by a schoolkid in early 2000s (which it essentially is :man_facepalming: ).
First of all, none of the other torrent clients uses a dark scheme. The thing just does not associate well with what torrents clients do - sending files around. It would be very nice if we could provide both light and dark themes with Tribler: unfortunately, this will require refactoring the QT CSS mess - at the moment, the stylesheets are scattered all around `.ui` files and `.py` files. Some colours are even set up in the code manually. QT bugs do not help with this task either.
In general, the solution should be: "move all the stylesheets into a separate file, leave .UI files unstyled". Also, it will require creating position-specific subclasses for many widgets in the GUI.

Second, our UI is just... no eye-candy? Inconsistent? Here is an example of a sleek modern PyQT GUI ([PyOneDark](https://github.com/Wanderson-Magalhaes/PyOneDark_Qt_Widgets_Modern_GUI)):
GUI

Third, lots of little usability details are missing: the keyboard focus is not there where it is expected, we don't use keyboard shortcuts, we raise dialogs on every occasion, dialogs look boring without any icons, etc. In short, when one opens Tribler for the first time, their first reaction is: "OMG, that's **ugly** :japanese_ogre: ". Then they try to use it and it does not disappoint their expectations - the UX is as :hankey: as the UI.

With such looks, it will be **extremely** hard to reach 1 million users in the world of 15-seconds attention spans, where each app only gets a single chance to prove itself.

The solution is to either hire a professional GUI design team, or move to Web tech and reuse templates that are already there.

</details>

<details>
<summary>:neckbeard: Identifying torrent authors (identicons) :neckbeard:</summary> 

At the moment, when the user searches for content, they get a flat list related to channels, folders and torrents. The problem is, _the user can't see the author or the source channel of those_. This makes it impossible for users to identify good sources of information (e.g. channels to subscribe). 

The solution is two-fold:
1. show full paths for entries in the search results list, either in form of a pop-up hint, or inline
2. add [identicons](https://github.com/Tribler/tribler/issues/5154) for public keys/channels

</details>

<details>
<summary> :deciduous_tree: Add recursive channel search :deciduous_tree: </summary>

Currently, the "filter" input box in channels searches for contents only in the current folder/channel, not diving into the folder/channel child folders. It would be very useful to instead make it work recursively on channel's folders.
To do this with acceptable performance, we'll have to add some accelerator structure to Channels DB, such as [material paths, transitive closures or matrix encodings](https://vadimtropashko.files.wordpress.com/2011/07/ch5.pdf).

</details>

<details>
<summary>:scroll: Downloads table is too wide :scroll:</summary> 
See https://github.com/Tribler/tribler/issues/6452

изображение

Also, the table is not too responsive, especially when Tribler starts. A better way to represent the downloads list is to use a hiding list to the right half of the window. Also, that will enable a natural drag'n'drop way of moving torrents to/from Channels.
Home_Layout

</details>

<details>
<summary> :no_entry_sign: : Can't share files with Tribler, torrent creation broken :no_entry_sign: : </summary>
Tribler's primary goal is to enable users to easily share files. Our users asked for a feature to [share a folder of files](https://github.com/Tribler/tribler/issues/4729). At the moment, the torrent creation dialog is [completely broken](https://github.com/Tribler/tribler/issues/4674) (and it has been broken for a couple of years already).

This means two things:
* torrent creation is too complex
* users need a simpler way to share files

</details>

<details>
<summary> :house: Bring back the Home screen :house: </summary>

Home screen was removed because [it did not bring any value](https://github.com/Tribler/tribler/issues/5071) (it was just a placeholder). Nonetheless, users expect there to be a home screen, something like a personalized dashboard or a[ news feed](https://github.com/Tribler/tribler/issues/5015). This time, it should provide value to the user.

Some stuff that should be on the home screen:
* updates to subscribed channels
* changes to popular torrents list
* list of active torrents
* list of users' channel
* user's [identicon](https://github.com/Tribler/tribler/issues/5154)
* pull requests on the user's channels
* updates on users' pull requests
* new personal messages, etc.
* Tribler development news and new version notifications

</details>

<details>
<summary> :page_facing_up: Switch to paginated Channels interface :page_facing_up: </summary>

Implementing Channels interface with `QTableView` was a big mistake. Yes, it provided the fancy endless scrolling feature good for showing off to journalists, but endless scrolling is useless if all entries in the table are the same height. Google uses endless scrolling just for pictures - its search remains paginated for a reason. Moving to **paginated** `QListView` will result in the following upsides:
* better navigation - pages help with that
* rich, robust and easy to change entries representation (e.g. with thumbnails)
* **much** more simple code (the whole "index to delegate" thing with `QTreeView` is pure horror)
* enable us to create a **united model for downloads and torrent entries in Channels** - this will remove all the [synchronization issues](https://github.com/Tribler/tribler/issues/6303) between the Downloads list and Channels contents.

<details>

### LINKED ISSUE #6680: Channel regeneration leaves old entries intact on update
Channels system supports regenerating one's personal channel in case of e.g. Tribler folder loss or channel torrent corruption. To do so, every ChannelMetadata entry includes the `start_timestamp` which sets the lower bound of timestamps during the processing of `.mdblob`s in the channel torrent dir.
However, there is no procedure for _removing_ entries with timestamps lower than `start_timestamp`. As the result, there could be a situation when, for example:
1. the author adds torrent A into their channel. `start_timestamp` is 100, `timestamp` is 102.
2. the user downloads the channel and adds torrents A into their DB.
2. the author loses their channel folder and re-creates their channel by adding torrents B into it. This time, `start_timestamp` is 200, `timestamp` is `202`
4. the user downloads the new version of the channel and adds B to their DB. Thus, the user will have _both_ A _and_ B in their local DB, which is incorrect, because the author's new version contains only B.

This problem is the most probable cause of https://github.com/Tribler/tribler/issues/6395

To fix it, we should add a procedure to remove all the obsolete content from the channel on update. Unfortunately, the procedure will be very data-intensive for big channels, since we'll have to limit the removal by the top-level channel id, resolving an entry's parents recursively. Otherwise, we could accidentally remove entries from another channel by the same user.

### LINKED ISSUE #6711: Lz4 does not  compress well enough.
**Describe the bug**

**To Reproduce**
1. Populate channels folder with data (subscribe to channels)
2. navigate to channels folder via cmd
3. Run compact /c /s /a /i /exe:lzx
4. See that it was able to compress files  1,3 to 1 that have already were compressed with lz4

**Expected behavior**
Files are not able to be compressed further
**Screenshots**

 - OS: Windows 11
 - Tribler's version: 7.11.0

**Additional context**

### LINKED ISSUE #7489: Unable to close Tribler after subscribing to a channel.

### LINKED ISSUE #7491: channel is still updated after unsubscribe
unsubscribing from a channel does not cancel the update job.

**To Reproduce**

1. enable 2 channels
2. wait for the first channel to start updating
3. disable the second channel
4. wait for the first channel to finish updating

**Expected behavior**
dont update the unsubscribed channel.

**Environment**

- OS: linux
- Tribler version: 889eb2c85024cfdd9d80b7346fb8d7443df1215c

qstokkink:
Related to #7669

This PR voids many of the previous issues related to channels.

Fixes #1368
Fixes #2284
Fixes #3615
Fixes #4637
Fixes #4645 
Fixes #4650
Fixes #4672
Fixes #4677
Fixes #5015
Fixes #5067
Fixes #5154
Fixes #5271
Fixes #5522
Fixes #5587
Fixes #6208
Fixes #6126
Fixes #6217
Fixes #6395
Fixes #6481
Fixes #6680
Fixes #6711
Fixes #7489
Fixes #7491

The diff is so big that Codacy is broken. However, just before it died, it seemed like all issues were resolved.

The script test is broken because I removed the channel creation script test. I modified the workflow file but this will only take effect in PRs after this PR.

Almost all of the hard work is to be attributed to @egbertbouman. I came in afterward to rebase onto the latest main and tidy up some loose ends.

---

@egbertbouman is our "lucky" winner on the 25% reviewer assignment chance 🎉 but I would still like an additional review on https://github.com/Tribler/tribler/pull/7726#issuecomment-1831769159 because I feel these three source files could be better.


qstokkink:
@kozlovsky Could you take a look at the changes in:

- `src/tribler/core/upgrade/tags_to_knowledge/previous_dbs/tags_db.py`
- `src/tribler/core/upgrade/tests/test_upgrader.py`
- `src/tribler/core/upgrade/upgrade.py`

This works with the new database format but I'm not entirely confident in my implementation. Is there a better way to do this?

hbiyik:
image

PR #7756, opened 2023-12-06
Reisolate content discovery community

qstokkink:
I removed the isolation of the `ContentDiscoveryCommunity` during refactoring and I forgot to add it back. This PR resolves this.


qstokkink:
I'm guessing the `application_tester_draft` is not stable yet and can be ignored.

PR #7758, opened 2023-12-06
Made `TorrentDefNoMetainfo` inherit from `TorrentDef`

### LINKED ISSUE #7732: [main branch] AttributeError: 'TorrentDefNoMetainfo' object has no attribute 'torrent_info_loaded' (in get_files_info_json_paged)
I got this bug on my machine:

```python
An error occurred during the request "GET http://localhost:59116/downloads/8bfbb8a87948e26bd83be120c4b7eac0a126b661/files?view_start_path=.&view_size=20":

  File "C:\dev\tribler\src\tribler\core\components\restapi\rest\rest_manager.py", line 53, in error_middleware
    response = await handler(request)
  File "C:\dev\tribler\venv39\lib\site-packages\aiohttp\web_middlewares.py", line 117, in impl
    return await handler(request)
  File "C:\dev\tribler\src\tribler\core\components\libtorrent\restapi\downloads_endpoint.py", line 658, in get_files
    "files": self.get_files_info_json_paged(download, Path(view_start_path), view_size)
  File "C:\dev\tribler\src\tribler\core\components\libtorrent\restapi\downloads_endpoint.py", line 173, in get_files_info_json_paged
    if not download.tdef.torrent_info_loaded():
AttributeError: 'TorrentDefNoMetainfo' object has no attribute 'torrent_info_loaded'
```

[Here](https://github.com/Tribler/tribler/blob/596e36cdede452bcf875148a9ab535c86397ce90/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py#L173):
```python
    @staticmethod
    def get_files_info_json_paged(download: Download, view_start: Path, view_size: int):
        """
        ...
        """
        if not download.tdef.torrent_info_loaded():  # <-- here
            download.tdef.load_torrent_info()
            return [{
                "index": IllegalFileIndex.unloaded.value,
                "name": "loading...",
                "size": 0,
                "included": 0,
                "progress": 0.0
            }]
        ...
```

As I understand it, there are three ways to fix the problem:

1) **The most complete solution:** It looks like the `TorrentDefNoMetainfo` instance represents a temporary state of the `TorrentDef` object, so probably, at some moment, we can get rid of the `TorrentDefNoMetainfo` class completely and directly use `TorrentDef` class instance instead after we extend the logic of the `TorrentDef` class so it can handle the state when no meta info data is loaded yet. 

    This approach can simplify the logic in eight places where we currently have checks like:
    
    ```python
        if not isinstance(self.tdef, TorrentDefNoMetainfo):
            ...
    ```

    These checks will be unnecessary if the base torrent def object can handle all appropriate cases.

2) **The simplest solution:** We can just add yet another check `if isinstance(self.tdef, TorrentDefNoMetainfo)` to the `get_files_info_json_paged` endpoint

3) **The balanced solution:** We can add a simple method implementation to the `TorrentDefNoMetainfo` class:

    ```python
        def torrent_info_loaded(self):
            return False
    ```

qstokkink:
Fixes #7732

We could still remove the entirety of `TorrentDefNoMetainfo` but its existence still seems convenient.

### DIFF
--- a/src/tribler/core/components/libtorrent/tests/test_torrent_def.py
+++ b/src/tribler/core/components/libtorrent/tests/test_torrent_def.py
@@ -225,6 +225,9 @@ def test_torrent_no_metainfo():
     assert not tdef.is_private()
     assert tdef.get_name_utf8() == "video.avi"
     assert tdef.get_nr_pieces() == 0
+    assert tdef.torrent_info is None
+    tdef.load_torrent_info()
+    assert tdef.torrent_info is None
 
     torrent2 = TorrentDefNoMetainfo(b"12345678901234567890", VIDEO_FILE_NAME, "magnet:")
     assert len(torrent2.get_trackers()) == 0
--- a/src/tribler/core/components/libtorrent/torrentdef.py
+++ b/src/tribler/core/components/libtorrent/torrentdef.py
@@ -1,6 +1,8 @@
 """
 Author(s): Arno Bakker
 """
+from __future__ import annotations
+
 import itertools
 import logging
 from asyncio import get_running_loop
@@ -520,7 +522,7 @@ class TorrentDef:
             raise ValueError("File not found in single-file torrent")
 
 
-class TorrentDefNoMetainfo:
+class TorrentDefNoMetainfo(TorrentDef):
     """
     Instances of this class are used when working with a torrent def that contains no metainfo (yet), for instance,
     when starting a download with only an infohash. Other methods that are using this class do not distinguish between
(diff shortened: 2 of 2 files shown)


src/tribler/core/components/libtorrent/tests/test_torrent_def.py
```diff
@@ -225,6 +225,9 @@ def test_torrent_no_metainfo():
     assert not tdef.is_private()
     assert tdef.get_name_utf8() == "video.avi"
     assert tdef.get_nr_pieces() == 0
+    assert tdef.torrent_info is None
+    tdef.load_torrent_info()
+    assert tdef.torrent_info is None
```
xoriole, line 230: Since `TorrentDefNoMetainfo` inherits from `TorrentDef`, calling `tdef.save()` on line 231 or 234 raises `IndexError: list index out of range`.

xoriole (approved):

PR #7760, opened 2023-12-07
Modify `check_free_space function` to verify available disk space

### LINKED ISSUE #7745: [7.12.1, 7.13.0] FileNotFoundError in check_free_space: [WinError 3] System cannot find the specified path: '.'
A pretty frequent error:

```
  File "run_tribler.py", line 100, in <module>
  File "tribler\gui\start_gui.py", line 48, in run_gui
  File "tribler\core\check_os.py", line 51, in check_free_space
  File "__init__.py", line 1995, in disk_usage
  File "psutil\_pswindows.py", line 265, in disk_usage
FileNotFoundError: [WinError 3] System cannot find the specified path: '.'  
```

[Here](https://github.com/Tribler/tribler/blob/258fe001d9b1023f74a55dc2690b61ab6aebb5b8/src/tribler/core/check_os.py#L51):
```python
def check_free_space():
    logger.info('Check free space')
    try:
        free_space = psutil.disk_usage(".").free / (1024 * 1024.0)
```

Should we specify some absolute path here?

Or at least we can catch and ignore the error, and continue without checking

drew2a:
This PR modifies the function `check_free_space` to ignore `OSErrors` so they cannot lead to Tribler crashing.
Additionally, in the `start_gui.py` module, the `check_free_space` function now takes a parameter specifying the root state directory to be checked for free space.

Fixes #7745

### DIFF
--- a/src/tribler/core/check_os.py
+++ b/src/tribler/core/check_os.py
@@ -7,6 +7,7 @@ import traceback
 
 import psutil
 
+from tribler.core.utilities.path_util import Path
 from tribler.core.utilities.utilities import show_system_popup
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/tribler/core/check_os.py
```diff
@@ -45,17 +46,22 @@ def check_environment():
     check_read_write()
 
 
-def check_free_space():
-    logger.info('Check free space')
+def check_free_space(state_dir: Path):
+    """ Check if there is enough free space on the disk."""
+    logger.info(f'Checking free space for the folder: {state_dir}')
```
qstokkink, line 51: **NIT:** Using f-strings in logging calls is generally discouraged (see for example https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html).

qstokkink (approved):

src/tribler/core/check_os.py
drew2a: We had this discussion before: #5611 :)

drew2a: Thank you for the suggestion. For this PR, I left the "f-string" version, but we can discuss it at the next dev meeting. In case of an altered agreement, I'll change the formatting for all upcoming pull requests.

PR #7764, opened 2023-12-08
Fix "ValueError: relative path can't be expressed as a file URI"

### LINKED ISSUE #6345: ValueError: relative path can't be expressed as a file URI
Sentry Issue: [TRIBLER-8K](https://sentry.tribler.org/organizations/tribler/issues/882/?referrer=github_integration)

```
CreationTraceback: 
  File "src\run_tribler.py", line 209, in <module>
  File "src\tribler-gui\tribler_gui\tribler_window.py", line 285, in __init__
  File "src\tribler-gui\tribler_gui\tribler_window.py", line 739, in create_add_torrent_menu

ValueError: relative path can't be expressed as a file URI
  File "src\tribler-gui\tribler_gui\utilities.py", line 398, in trackback_wrapper
  File "src\tribler-gui\tribler_gui\utilities.py", line 395, in trackback_wrapper
  File "src\tribler-gui\tribler_gui\tribler_window.py", line 772, in on_add_torrent_browse_file
  File "pathlib.py", line 744, in as_uri
```

drew2a:
This PR adds a new test case to the `test_tribler_window.py` file. The test checks that the `on_add_torrent_browse_file` method in the `TriblerWindow` class works correctly. It mocks certain methods and uses patching to simulate user interaction with a file dialog. The test verifies that the appropriate methods are called and asserts their expected behavior.

The code changes in `tribler_window.py` modify the implementation of the `on_add_torrent_browse_file` method. It now handles cases where no filenames are selected from the file dialog, preventing unnecessary processing. Additionally, it resolves each filename to its absolute URI path before appending it to the list of pending URI requests.

These changes improve the functionality and reliability of adding torrent files through browsing in TriblerWindow.

Fixes #6345

### DIFF
--- a/src/tribler/gui/tests/test_tribler_window.py
+++ b/src/tribler/gui/tests/test_tribler_window.py
@@ -1,0 +1,29 @@
+from unittest.mock import Mock, patch
+
+import pytest
+from PyQt5.QtWidgets import QFileDialog
+
+from tribler.gui.tribler_window import TriblerWindow
+
+
+# pylint: disable=redefined-outer-name
+@pytest.fixture
+def tribler_window():
+    """ Create mocked TriblerWindow instance"""
+    with patch('tribler.gui.tribler_window.TriblerWindow.__init__', Mock(return_value=None)):
+        window = TriblerWindow(Mock(), Mock(), Mock(), Mock())
+        window.pending_uri_requests = []
+        return window
+
+
+def test_on_add_torrent_browse_file(tribler_window: TriblerWindow):
+    """ Test that the on_add_torrent_browse_file method works correctly"""
+    tribler_window.raise_window = Mock()
+    tribler_window.process_uri_request = Mock()
+
+    with patch.object(QFileDialog, 'getOpenFileNames', Mock(return_value=['.'])) as patched_getOpenFileNames:
+        tribler_window.on_add_torrent_browse_file()
+
+    assert tribler_window.raise_window.called
+    assert patched_getOpenFileNames.called
+    assert tribler_window.process_uri_request.called
--- a/src/tribler/gui/tribler_window.py
+++ b/src/tribler/gui/tribler_window.py
@@ -83,7 +83,6 @@ from tribler.gui.upgrade_manager import UpgradeManager
 from tribler.gui.utilities import (
     connect,
     create_api_key,
-    disconnect,
     format_api_key,
     get_font_path,
     get_gui_setting,
@@ -91,7 +90,6 @@ from tribler.gui.utilities import (
     get_ui_file_path,
     is_dir_writable,
     set_api_key,
-    show_message_box,
     tr,
 )
 from tribler.gui.widgets.instanttooltipstyle import InstantTooltipStyle
@@ -681,13 +679,16 @@ class TriblerWindow(QMainWindow):
 
     def on_add_torrent_browse_file(self, *_):
         self.raise_window()  # For the case when the action is triggered by tray icon
-        filenames = QFileDialog.getOpenFileNames(
+        filenames, *_ = QFileDialog.getOpenFileNames(
             self, tr("Please select the .torrent file"), QDir.homePath(), tr("Torrent files%s") % " (*.torrent)"
         )
-        if len(filenames[0]) > 0:
-            for filename in filenames[0]:
-                self.pending_uri_requests.append(Path(filename).as_uri())
-            self.process_uri_request()
+        if not filenames:
+            return
+
+        for filename in filenames:
+            uri = Path(filename).resolve().as_uri()
+            self.pending_uri_requests.append(uri)
+        self.process_uri_request()
 
     def start_download_from_uri(self, uri):
         uri = uri.decode('utf-8') if isinstance(uri, bytes) else uri


src/tribler/gui/tests/test_tribler_window.py
```diff
@@ -0,0 +1,29 @@
+from unittest.mock import Mock, patch
+
+import pytest
+from PyQt5.QtWidgets import QFileDialog
+
+from tribler.gui.tribler_window import TriblerWindow
+
+
+# pylint: disable=redefined-outer-name
+@pytest.fixture
```
qstokkink, line 10: Suggestion to adhere to our Pylint rules:

```suggestion
@pytest.fixture(name="tribler_window")
def fixture_tribler_window():
```

That aside: please make sure the tests pass and re-request my review.

drew2a: Thank you for the suggestion. From my point of view, for this particular case (for `test_*.py` files), it is preferable to use the following construction:

**Example 1**
```python
# pylint: disable=redefined-outer-name
@pytest.fixture
def fixture_name():
    ...
``` 

Instead of the suggestion by pylint:

**Example 2**
```python
@pytest.fixture(name="fixture_name")
def fixture_fixture_name():
    ...
```

Pylint is aimed at helping developers but not to be blindly obeyed. In many cases, pylint does a great job. However, in this particular case (from my point of view), the suggestion doesn't protect us from any error and does not make the code better. Instead, it makes the code more polluted as it requires unnecessary syntax complication.

I tried both approaches with Tribler, and based on my experience, **Example 1** is as correct as **Example 2**, but it looks better and requires less time to write and correct, which saves developers' time.

qstokkink: Sure. Consider these two examples. First, by not specifying a name, everything seems fine:

case1

Then, option two, we expose a problem with our fixtures by using the `name=` syntax:

case2

Because the reference no longer exists, we see that there is something wrong with our fixture test.

qstokkink: > However, the error with the "something" fixture in your first example will pop up with the first test run.

You misunderstand the core issue. The problem is that **it will not show up** in the test run. Here is the output of the first snippet, nothing pops up:

```
============================= test session starts ==============================
collecting ... collected 2 items

test_something.py::test_2 
test_something.py::test_1 

============================== 2 passed in 0.51s ===============================
PASSED                                         [ 50%]PASSED                                         [100%]
```

Here is the snippet again in more copy-pasteable form, because I can see you didn't run it before making your claim.

```python
from pytest import fixture


@fixture()
def something():
    return True


def test_1(something):
    assert something


def test_2():
    assert something
```

qstokkink: Your tests are equivalent to the following:

```python
def test_1(something, port):
    raise TypeError
    assert something.property == 'there is an object normally returns from fixtures'


def test_2():
    raise TypeError
    assert something.property == 'there is an object normally returns from fixtures'
```

You're not even reaching the naming shadowing code.

In fact, your example code is doubly-wrong. Even if I strip out the forced `TypeError`:

```python
def test_1(something, port):
    assert something.property == 'there is an object normally returns from fixtures'


def test_2():
    assert something.property == 'there is an object normally returns from fixtures'
```

You're ignoring the issue of name shadowing by explicitly testing for a property that is supposed to be there.

A more appropriate example is this:

```python
from pytest import fixture
from unittest.mock import Mock


@fixture
def something():
    mock = Mock(return_value='there is an object normally returns from fixtures')
    return mock()


@fixture
def port():
    return 1234  # or sometimes there is indeed the value like a port


def test_1(something, port):
    assert something != 'this is something else'
    assert port != 0


def test_2():
    assert something != 'this is something else'
    assert port != 0
```

I use `!=` instead of a bare `assert` in this case, just to show that **this is an entire class of problems** and not just some fringe case with bare asserts. You can really make this check as complex as you want.

qstokkink: Let me add this: I'll stop here. I don't think I can be convinced of your implementation.

That said, if you still insist that we should ignore the pytest docs, Pylint (+ Codacy) warnings, @kozlovsky's concerns and my own and that the examples I gave are not relevant in practice: we can assign a different reviewer to this PR.

drew2a: I appreciate your effort to analyze my examples. Although I don't agree with the conclusions, I'm not going to try to change your mind because this particular part is not that important to the general topic at hand.

The funny thing is that we both actually started the discussion about shadowing without having an example of real shadowing. We started from here:

```python
from pytest import fixture


@fixture()
def something():
    return True


def test_1(something):
    assert something


def test_2():
    assert something
```

This example does not have anything related to the shadowing. I haven't found any definition for shadowing in general but there's a definition for variable shadowing https://en.wikipedia.org/wiki/Variable_shadowing

In the case of fixtures there is a "function name" shadowing:

> If a fixture is used in the same module in which it is defined, the function name of the fixture will be shadowed by the function arg that requests the fixture

I understand it as 

```python
from pytest import fixture


@fixture()
def something():
    return True


def test_1(something):  # <- shadowing happens here and this fact could lead only to harmless warnings.
    assert something
```

I can also provide an example of real shadowing, although it does not relate to the original "function name" shadowing from the pytest documentation:

```python
from pytest import fixture


@fixture()
def something():
    return True


def something():
    return False


def test_1(something):
    assert something

```

Summarising my point: you're right that there is shadowing correctly reported by pylint, but it can be safely ignored in this particular code (that I wrote for this PR). 

From my point of view (which could be wrong), that warning can be ignored pretty much everywhere in Tribler. Anyway, the question of whether we should ignore the warning for the entire test suite should be raised in a dev meeting, as it applies to everyone.

For a "controversial" PR like this, it's good to learn how we can move forward and reach a consensus, even if not all parties are fully satisfied with the final decision.

The best-written guide for the review process, to my knowledge, is the "Google Engineering Practices Documentation." So, we can use it as a starting point to make a decision on whether a PR is good enough to be accepted or not:
https://github.com/google/eng-practices/blob/master/review/reviewer/index.md

drew2a: I guess I can provide the "function name shadowing" problem example:

```python
from pytest import fixture


@fixture()
def something():
    return True


def test_1(something):
    assert something

    # here by mistake a developer could call the "something" function but here is the fixture, eg. the function result
    result = something()
```

qstokkink: From your comment I deduce that you do not see the point I'm trying to make. The issue with name shadowing is not `test_1` but it is `test_2`. The problem with `test_2` is that it may be introduced by someone else in the future. So, even if `test_1` works right now, someone can get into trouble in the future due to your (avoidable) construction. If you want to relate it to the Google practices, this practice degrades the code quality.

The reason I suggested leaving this to another reviewer is to not make this PR controversial. Perhaps another reviewer can explain this issue in a different way that makes more sense to you. I would like to urge you to allow a second reviewer opinion on this.

drew2a: @egbertbouman @xoriole please, share your opinion 🙏

drew2a: @qstokkink, no-no, I think I understand you. But in test `test_2`, there is no shadowing.

And yes, someone in the future could introduce mistakes (arithmetic, logical, or even related to shadowing) to `test_2`, which has nothing in common with the shadowing mentioned in the original pylint warning.

qstokkink: @drew2a Judging from your comment, I still do not think that you do understand. There is **one** specific type of situation that can happen solely due to the construction that causes name shadowing (in `test_1`). Not multiple as you just said. I gave an instance of this type of situation in the example above (https://github.com/Tribler/tribler/pull/7764#discussion_r1422721124).

The shadowing warning you get for `test_1` causes trouble for `test_2`, which will not get a warning. It (`test_2`) will pass, even though it can and should fail.

drew2a: @qstokkink, with all due respect, it's not shadowing that causes the potential problem; it's just a badly written test that caused the problem.

Here's a slightly changed example that "has problems due to shadowing." It still has them, even without any shadowing:

```python
def something():
    return True


def test_2():
    assert something
```

qstokkink: @drew2a Yes, thank you, that is indeed the core issue: if `something` is now made into a `fixture`, which uses a `name` argument, exactly this type of error cannot happen.

src/tribler/gui/tribler_window.py
```diff
@@ -681,13 +679,16 @@ def on_create_torrent_updates(self, update_dict):
...
         )
-        if len(filenames[0]) > 0:
-            for filename in filenames[0]:
-                self.pending_uri_requests.append(Path(filename).as_uri())
-            self.process_uri_request()
+        if not filenames:
+            return
+
+        for filename in filenames:
+            uri = Path(filename).resolve().as_uri()
```
xoriole, line 689: I don't see any harm in using resolve() here but I wonder if this will fix the issue since the expected file is .torrent files which should exist as it is selected using the file dialog. The reproduction using `['.']` might not be representative enough for this issue.

drew2a: Changing the test from the original to the following produces the same reproducibility.

```python
def test_on_add_torrent_browse_file(tribler_window: TriblerWindow):
    """ Test that the on_add_torrent_browse_file method works correctly"""
    tribler_window.raise_window = Mock()
    tribler_window.process_uri_request = Mock()

    with patch.object(QFileDialog, 'getOpenFileNames', Mock(return_value=['./any.torrent'])) as patched_getOpenFileNames:
        tribler_window.on_add_torrent_browse_file()

    assert tribler_window.raise_window.called
    assert patched_getOpenFileNames.called
    assert tribler_window.process_uri_request.called
```

qstokkink (dismissed):
This PR has gotten stuck because we started debating the style rules enforced by our PR checks and we have no resolution yet. I want to get the PR unstuck by deferring the debate.

Please stick to our currently enforced rules in this PR (i.e., do not suppress `redefined-outer-name`) because they are the currently enforced rules.

Assuming you still believe we should change these rules, please open an issue so we can have a focused discussion on changing the style rules. _If_ the team votes to adopt the new style rules, you can come in and change the code back and I will approve that PR without question.

drew2a:
@qstokkink sure.

> Please stick to our currently enforced rules in this PR (i.e., do not suppress redefined-outer-name) because they are the currently enforced rules.

We don't have an agreement that a developer should not suppress any linter warnings. The linter is just a tool that helps developers write good code and prevent bugs. The final decision should be with the developer (and the reviewer).

This PR increases the general system health as it fixes a bug. Even if it were written imperfectly (from a reviewer's perspective), it should be approved (at least, I would approve it even if I don't like the style, as Tribler is in better shape with this change than without it).

However, it is your right as a reviewer to request changes, and I appreciate it.

synctext:
This is lengthy discussion on tomato versus tomato. Suppression of lint warning for code readability (another entity for the fixture).

xoriole (approved):
Since both approaches are used in the codebase. I'll approve the PR as a second reviewer.

qstokkink:
> Even if it were written imperfectly [...]

> Suppression of lint warning for code readability [...]

There is a misconception here: this issue here is the suppression of a Warning-level linting error. This is just below the Fatal/Error category that would instantly crash your code. It is in the category that causes long-term degradation of your code base. If it were just readabillity, it would be in the Convenction/Refactor/Information classes.

I cannot support adding techincal debt and piling on more technical debt because we already have it, is something I cannot support either.

That said, because another reviewer is now staking their reputation on approving this, I'll dismiss my own review and let it pass.

drew2a:
@qstokkink thank you for your understanding.

I don't want to start again about that "there is nothing wrong with suppressing a warning-level linting error for pylint fixtures". All our arguments are in the discussion above.

But I want to clarify that this PR does not have any [technical debt](https://en.wikipedia.org/wiki/Technical_debt), as it does not consist of any parts that should be rewritten in the future:
> In [software development](https://en.wikipedia.org/wiki/Software_development), or any other IT field (e.g., Infrastructure, Networking, etc.) technical debt (also known as design debt[[1]](https://en.wikipedia.org/wiki/Technical_debt#cite_note-Girish_2014-1) or code debt) is the implied cost of future reworking required when choosing an easy but limited solution instead of a better approach that could take more time.

I wanted to give an example of a Tribler PR that has technical debt, but instead, I'll finish with the funny note that 
- #7799 

almost justifies having technical debt as so-called "prototypes" almost repeat the definition of technical debt from above.

qstokkink:
Your Wikipedia link lists a few causes of technical debt that I think are applicable here:

> - Lack of process or understanding, where businesses are blind to the concept of technical debt, and make decisions without considering the implications.
> - Lack of alignment to standards, where industry standard features, frameworks, and technologies are ignored. Eventually integration with standards will come and doing so sooner will cost less (similar to "delayed refactoring").[8]: 7 
> - Lack of knowledge, when the developer doesn't know how to write elegant code.[9]

I could also link you to quite a few blogposts that will try to convince you that ignoring linter warnings is a form of technical debt. However, that only counts as anecdotal evidence of random people from the Internet.

That said, I agree with you that if nobody ever changes the code, it does not count as technical debt. However, I do predict that this code will change. Of course, this is only mostly based on experience. Nobody can predict the future.

Also, completely agreed that prototypes are - by definition - technical debt. No discussion there. I do consider them a necessary part of code base evolution though.

Lastly, also agreed that we are beating a dead horse. I did not want to start another discussion. I just wanted to be transparent on why I will not approve this PR even after a second opinion, for posterity: just so I could look back later and say "I told you so".

Issue #1722, opened 2015-11-12
We need to go through all the TODO/FIXME entries and clean them up/do something about them


whirm:

qstokkink:
Why was this closed? We still have a boatload of TODO's and FIXME's in the code.

ichorid:
Because this is useless. We need a separate issue per FIXME/TODO

qstokkink:
> We need a separate issue per FIXME/TODO

I agree, but we don't have this breakdown (yet). Until we do, this issue is not resolved.

synctext:
ToDo about our ToDo list..

qstokkink:
I think we now have sufficient team support to finally resolve this issue and stop the practice of communication through TODOs and switch fully to GitHub issues for communication between developers. When this is resolved, we should enable the pylint checks to enforce this.

I created an overview of all FIXME-type issues and what I believe to be the correct response to each of them.

---

```
************* Module tribler.gui.utilities
src/tribler/gui/utilities.py:393: [W0511(fixme), ] FIXME: this is a temp fix to cap the normalized value to 1.
```

**Recommended action:** Remove this comment without creating an issue.
**Rationale:** This is simply defensive programming and doesn't need to be fixed. The intent of the code is obvious.

---

```
************* Module tribler.gui.widgets.tablecontentdelegate
src/tribler/gui/widgets/tablecontentdelegate.py:143: [W0511(fixme), ] TODO: restore this behavior, so there is really some tolerance zone!
src/tribler/gui/widgets/tablecontentdelegate.py:539: [W0511(fixme), ] TODO: refactor this not to rely on inheritance order, but instead use interface method pattern
```

**Recommended action (143):** Remove this comment without creating an issue.
**Rationale (143):** This is a "nice to have" feature request that is likely to be superseded by https://github.com/Tribler/tribler/issues/7736
**Recommended action (143):** Remove this comment without creating an issue.
**Rationale (143):** This is a good pointer for refactoring but also likely to be superseded by https://github.com/Tribler/tribler/issues/7736

---

```
************* Module tribler.gui.widgets.settingspage
src/tribler/gui/widgets/settingspage.py:494: [W0511(fixme), ] TODO: do it in RESTful style, on the REST return JSON instead
```

**Recommended action:** Remove this comment without creating an issue.
**Rationale:** This comment seems to be out of date.

---

```
************* Module tribler.core.components.libtorrent.download_manager.download_manager
[...]

PR #7773, opened 2023-12-12
Moved metadata_store into database and content_discovery

qstokkink:
Related to #7669, following the plan of https://github.com/Tribler/tribler/issues/7669#issuecomment-1849549637

The changes of this PR are simply moves and dead code removal.

Unit tests have been added to cover missing lines (in our existing code). The lines in `database_component.py` that are part of `gui_test_mode` or `chant_testnet` are not covered.


src/tribler/core/components/database/database_component.py
```diff
@@ -1,23 +1,67 @@
...
-        db_path = self.session.config.state_dir / STATEDIR_DB_DIR / "tribler.db"
-        if self.session.config.gui_test_mode:
+        config = self.session.config
+        db_path = config.state_dir / STATEDIR_DB_DIR / "tribler.db"
+        if config.gui_test_mode:
             db_path = ":memory:"
 
         self.db = TriblerDatabase(str(db_path))
 
+        # TODO: merge the code below into the TriblerDatabase
```
drew2a, line 26: Is this a TODO for the next PR, or is it a forgotten TODO for the current PR?

drew2a (approved):
👍

src/tribler/core/components/database/database_component.py
qstokkink: This is for the next PR.

drew2a: Some of the developers and I in this group consider it bad, or let's say "not good," practice to merge changes with TODOs into the main branch.

I'm not stopping you from doing this by not giving acceptance to the PR. I'm just highlighting that there are many points of view on this practice, some of which consider it to be a bad practice that leads to code quality degradation.

https://www.reddit.com/r/learnprogramming/comments/r25bj7/are_is_committing_code_to_github_with_todos_bad/?rdt=46532

qstokkink: I agree with them. However, our current linters are configured to allow this. Notably, I did not have to disable any warnings to pass our PR checks.

Once we fix https://github.com/Tribler/tribler/issues/1722 we can adopt a more strict stance on this.

drew2a: Are you saying that if you understand there is a bad practice, but there are no configured linter issues, then it is okay for you to commit and merge the code despite its quality?

qstokkink: Even though I disagree with the practice personally, this is the approach that the team has chosen and I adapt my workflow to it.

If I were the only developer, I would make make own standards and conform to what I would like to see. Sadly, I am not the only developer and I have to conform to the existing rules and linter settings.

PR #7781, opened 2023-12-13
WIP Update build dependencies and remove unnecessary package versions

drew2a:
This PR cleans up `requirements-build.txt`

- Update build dependencies in `makedist_macos.sh` script
- Remove specific package versions from `requirements-build.txt`
- Update PyInstaller version to `5.13.2`

It is better to have a list of requirements in the `requirements.txt` instead of having it in a bash script, as if it were a file, Dependabot could review it as well.


drew2a:
Modern versions of pyinstaller (I've tried 5.13.2 and 6.3.0) still doesn't work with our build script (see #7484). So I'm closing the PR.

Issue #7632, opened 2023-10-16
Revise the `PopularityCommunity` metadata retrieval protocol.


drew2a:
Despite the protocol's apparent simplicity, `PopularityCommunity` is quite complex as it derives logic from `RemoteQueryCommunity`:
https://github.com/Tribler/tribler/blob/20fb22453c8c5349c6de4b8e679a5b590792b7df/src/tribler/core/components/popularity/community/popularity_community.py#L22-L23
 
This inheritance was implemented in #5736

The current algorithm for metadata retrieval is as follows:
https://github.com/Tribler/tribler/issues/7398#issuecomment-1721263221
> If a peer receives torrent health info for a torrent whose metadata is missing, the Popularity Community subsequently requests the missing metadata.

https://github.com/Tribler/tribler/blob/20fb22453c8c5349c6de4b8e679a5b590792b7df/src/tribler/core/components/popularity/community/popularity_community.py#L81-L92

As `RemoteQueryCommunity` is going to be removed in `8.0.0` we have to replace the algorithm for metadata retrieval.

drew2a:
As a starting point for discussion, I propose the following algorithm:

Popularity Community operates in this manner:

1. Upon introduction requests, send information about popular torrents.
2. Every 5 seconds, choose a random peer and send a torrent health request.
3. The chosen peer responds with a list of health information.
   
(Note: Steps 1, 2, and 3 remain unchanged from the current algorithm)

4. The requester identifies torrents for which knowledge is missing and sends a series of messages requesting this knowledge:

```python
@dataclass
class RequestKnowledgeMessage:
    infohash: str
```

5. The chosen peer responds with a series of messages containing the required knowledge.

https://github.com/Tribler/tribler/blob/20fb22453c8c5349c6de4b8e679a5b590792b7df/src/tribler/core/components/knowledge/community/knowledge_payload.py#L42-L45

synctext:
My idea is to first focus on stability, removing Gigachannels, keeping tags, and then radically alter the architecture. Lets not try to fix things which are not broken currently :thinking: 
<IMG src=https://github.com/Tribler/tribler/assets/325224/672389f8-c4da-4425-bb7d-455fc1ffb1e4 width=300>

drew2a:
> Removing 2 out of 3 Content Discovery Methods

[...]

Issue #7784, opened 2023-12-18
ERROR Slow coroutine step execution (duration=1.673 seconds): <Task pending name='TorrentChecker:check local torrents'


slrslr:
**Describe the bug**

**To Reproduce**

1. start Tribler with some torrents
2. check tribler-core-error.log
4. See error:

```
[tribler-core PID:39] 2023-12-18 06:51:58,923 - ERROR <watching_thread:91> tribler.core.utilities.slow_coro_detection._report_freeze(): Slow coroutine is occupying the loop for 1.206 seconds already: <Task pending name='TorrentChecker:check local torrents' coro=<interval_runner() running at ipv8/taskmanager.py:18> cb=[TaskManager.register_task.<locals>.done_cb() at ipv8/taskmanager.py:128]>
Stack for <Task pending name='TorrentChecker:check local torrents' coro=<interval_runner() running at ipv8/taskmanager.py:18> cb=[TaskManager.register_task.<locals>.done_cb() at ipv8/taskmanager.py:128]> (most recent call last):
  File "run_tribler.py", line 109, in <module>
  File "tribler/core/start_core.py", line 204, in run_core
  File "tribler/core/start_core.py", line 174, in run_tribler_core_session
  File "asyncio/base_events.py", line 603, in run_until_complete
  File "asyncio/base_events.py", line 570, in run_forever
  File "asyncio/base_events.py", line 1859, in _run_once
  File "tribler/core/utilities/slow_coro_detection/patch.py", line 37, in patched_handle_run
  File "asyncio/events.py", line 81, in _run
  File "ipv8/taskmanager.py", line 18, in interval_runner

[tribler-core PID:39] 2023-12-18 06:51:59,390 - ERROR <patch:52> tribler.core.utilities.slow_coro_detection._report_long_duration(): Slow coroutine step execution (duration=1.673 seconds): <Task pending name='TorrentChecker:check local torrents' coro=<interval_runner() running at ipv8/taskmanager.py:18> wait_for=<_GatheringFuture pending cb=[<TaskWakeupMethWrapper object at 0x7f3d63a8e160>()]> cb=[TaskManager.register_task.<locals>.done_cb() at ipv8/taskmanager.py:128]>
```

**Desktop (please complete the following information):**

- OS: Linux Debian 11 oldstable
- Tribler's version 7.13.0 (flatpak org.tribler.Tribler)

qstokkink:
[...]

PR #7786, opened 2023-12-21
WIP: Store preferable infohashes for queries

qstokkink:
Related to #7632

This PR adds integrations with local and remote search results and stores the preferable torrent (according to what the user clicked and downloaded).

Currently, this integration is only local and the propagation of preferable torrents depends on the emergent effect of them being checked more often in the torrent checker, causing them to be propagated in the content discovery community more. In a future PR, we can implement more direct interaction with other peers to share recommendations.

The uncovered lines of this PR are in a `TYPE_CHECKING` block and, therefore, impossible to cover.

### DIFF
--- a/src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
+++ b/src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
@@ -1,0 +1,201 @@
+from typing import Generator
+
+from pony.orm import db_session
+from pytest import fixture
+
+from tribler.core.components.database.db.layers.user_activity_layer import UserActivityLayer
+from tribler.core.components.user_activity.types import InfoHash
+from tribler.core.utilities.pony_utils import TrackedDatabase
+
+
+@fixture(name="layer")
+def fixture_activity_layer() -> Generator[UserActivityLayer, None, None]:
+    database = TrackedDatabase()
+    database.bind(provider="sqlite", filename=":memory:")
+    ual = UserActivityLayer(database)
+    database.generate_mapping(create_tables=True)
+    yield ual
+    database.disconnect()
+
+
+def float_equals(a: float, b: float) -> bool:
+    return round(a, 5) == round(b, 5)
+
+
+def test_store_no_losers(layer: UserActivityLayer) -> None:
+    """
+    Test that queries can be stored and retrieved.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), set())
+
+    with db_session():
+        queries = layer.Query.select()[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert len(queries[0].infohashes) == 1
+        assert list(queries[0].infohashes)[0].infohash == b"\x00" * 20
+        assert float_equals(list(queries[0].infohashes)[0].preference, 1.0)
+
+
+def test_store_with_loser(layer: UserActivityLayer) -> None:
+    """
+    Test that queries with a loser can be stored and retrieved.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        winner, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        loser, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(winner.preference, 1.0)
+        assert float_equals(loser.preference, 0.0)
+
+
+def test_store_with_losers(layer: UserActivityLayer) -> None:
+    """
+    Test that queries with multiple losers can be stored and retrieved.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        winner, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        loser_1, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+        loser_2, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x02" * 20)[:]
+        loser_3, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(winner.preference, 1.0)
+        assert float_equals(loser_1.preference, 0.0)
+        assert float_equals(loser_2.preference, 0.0)
+        assert float_equals(loser_3.preference, 0.0)
+
+
+def test_store_weighted_decay(layer: UserActivityLayer) -> None:
+    """
+    Test result decay after updating.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
+    layer.store("test query", InfoHash(b"\x01" * 20), {InfoHash(b"\x00" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        entry_1, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        entry_2, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+        entry_3, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x02" * 20)[:]
+        entry_4, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.8)
+        assert float_equals(entry_3.preference, 0.0)
+        assert float_equals(entry_4.preference, 0.0)
+
+
+def test_store_delete_old(layer: UserActivityLayer) -> None:
+    """
+    Test result decay after updating.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
+    layer.store("test query", InfoHash(b"\x04" * 20), {InfoHash(b"\x00" * 20),
+                                                       InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20)})
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        entry_1, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        entry_2, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+        entry_3, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x02" * 20)[:]
+        should_be_dropped = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+        entry_4, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x04" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.0)
+        assert float_equals(entry_3.preference, 0.0)
+        assert should_be_dropped == []
+        assert float_equals(entry_4.preference, 0.8)
+
+
+def test_store_delete_old_over_e(layer: UserActivityLayer) -> None:
+    """
+    Test if entries are not deleted if their preference is still over the threshold e.
+    """
+    layer.e = 0.0
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
+    layer.store("test query", InfoHash(b"\x04" * 20), {InfoHash(b"\x00" * 20),
+                                                       InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20)})
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        entry_1, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        entry_2, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+        entry_3, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x02" * 20)[:]
+        entry_4, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+        entry_5, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x04" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.0)
+        assert float_equals(entry_3.preference, 0.0)
+        assert float_equals(entry_4.preference, 0.0)
+        assert float_equals(entry_5.preference, 0.8)
+
+
+def test_get_preferable(layer: UserActivityLayer) -> None:
+    """
+    Test if a preferable infohash is correctly retrieved.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    assert layer.get_preferable(b"\x00" * 20) == b"\x00" * 20
+
+
+def test_get_preferable_already_best(layer: UserActivityLayer) -> None:
+    """
+    Test if a infohash returns itself when it is preferable.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    assert layer.get_preferable(b"\x01" * 20) == b"\x00" * 20
+
+
+def test_get_preferable_unknown(layer: UserActivityLayer) -> None:
+    """
+    Test if a infohash returns itself when it has no known preferable infohashes.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    assert layer.get_preferable(b"\x02" * 20) == b"\x02" * 20
+
+
+def test_get_random(layer: UserActivityLayer) -> None:
+    """
+    Test if the preferred infohash always gets returned from a random checked selection.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20), InfoHash(b"\x02" * 20)})
+    layer.store("test query", InfoHash(b"\x01" * 20), {InfoHash(b"\x00" * 20), InfoHash(b"\x02" * 20)})
+
+    random_selection = layer.get_preferable_to_random(limit=1)
+
+    assert len(random_selection) == 1
+    assert list(random_selection)[0] == b"\x01" * 20
--- a/src/tribler/core/components/database/db/layers/user_activity_layer.py
+++ b/src/tribler/core/components/database/db/layers/user_activity_layer.py
@@ -1,0 +1,128 @@
+from __future__ import annotations
+
+import random
+import typing
+from dataclasses import dataclass
+
+from pony import orm
+from pony.orm import db_session
+
+from tribler.core.components.user_activity.types import InfoHash
+from tribler.core.utilities.pony_utils import TrackedDatabase
+
+if typing.TYPE_CHECKING:
+    @dataclass
+    class InfohashPreference:
+        infohash: bytes
+        preference: float
+        parent_query: Query
+
+    @dataclass
+    class Query:
+        query: str
+        infohashes: typing.Set[InfohashPreference]
+
+
+class UserActivityLayer:
+
+    def __init__(self, database: TrackedDatabase, update_weight: float = 0.8, e: float = 0.01) -> None:
+        """
+        Create a new User Activity scheme for a particular database.
+
+        :param database: The database to bind to.
+        :param update_weight: The weight of new updates.
+        :param e: A small value to decide near-zero preference.
+        """
+        self.database = database
+
+        self.e = e
+        self.update_weight_new = update_weight
+        self.update_weight_old = 1 - self.update_weight_new
+
+        class Query(database.Entity):
+            query = orm.PrimaryKey(str)
+            infohashes = orm.Set("InfohashPreference")
+
+        class InfohashPreference(database.Entity):
+            infohash = orm.Required(bytes)
+            preference = orm.Required(float)
+            parent_query = orm.Required(Query)
+            orm.PrimaryKey(infohash, parent_query)
+
+        self.Query = Query
+        self.InfohashPreference = InfohashPreference
+
+    def store(self, query: str, infohash: InfoHash, losing_infohashes: typing.Set[InfoHash]) -> None:
+        """
+        Store a query, its selected infohash, and the infohashes that were not downloaded.
+
+        :param query: The text that the user searched for.
+        :param infohash: The infohash that the user downloaded.
+        :param losing_infohashes: The infohashes that the user saw but ignored.
+        """
+        # Convert "win" or "loss" to "1.0" or "0.0".
+        weights = {ih: 0.0 for ih in losing_infohashes}
+        weights[infohash] = 1.0
+
+        # Update or create a new database entry
+        with db_session:
+            existing = self.Query.get(query=query)
+            if existing is not None:
+                for old_infohash_preference in existing.infohashes:
+                    if old_infohash_preference.infohash in weights:
+                        new_weight = (old_infohash_preference.preference * self.update_weight_old
+                                      + weights.pop(old_infohash_preference.infohash, 0.0) * self.update_weight_new)
+                        old_infohash_preference.preference = new_weight
+                    else:
+                        # This infohash did not pop up, candidate for deletion
+                        new_weight = old_infohash_preference.preference * self.update_weight_old
+                        if new_weight < self.e:
+                            old_infohash_preference.delete()
+                        else:
+                            old_infohash_preference.preference = new_weight
+                if infohash in weights:
+                    weights[infohash] = self.update_weight_new
+            else:
+                existing = self.Query(query=query, infohashes=set())
+
+            for new_infohash, weight in weights.items():
+                existing.infohashes.add(self.InfohashPreference(infohash=new_infohash, preference=weight,
+                                                                parent_query=existing))
+
+    @db_session
+    def _select_superior(self, infohash_preference: InfohashPreference) -> InfoHash:
+        """
+        For a given InfohashPreference, get the preferable infohash from the parent query.
+        """
+        all_hashes_for_query = list(infohash_preference.parent_query.infohashes)
+        all_hashes_for_query.sort(key=lambda x: x.preference, reverse=True)
+        return typing.cast(InfoHash, all_hashes_for_query[0].infohash)
+
+    def get_preferable(self, infohash: InfoHash) -> InfoHash:
+        """
+        Given an infohash, see if we know of more preferable infohashes.
+
+        :param infohash: The infohash to find better alternatives for.
+        """
+        with db_session:
+            existing = self.InfohashPreference.select(infohash=infohash)[:]
+
+            if not existing:
+                return infohash
+
+            return self._select_superior(random.SystemRandom().choice(existing))
+
+    def get_preferable_to_random(self, limit: int = 1) -> set[InfoHash]:
+        """
+        Retrieve (a) random infohash(es) and then return the preferred infohash for each infohash.
+
+        This method selects up to the limit of random infohashes and then outputs the set of preferable infohashes.
+        This means that you may start with ``limit`` number of infohashes and worst-case, if they all share the same,
+        preferable infohash, end up with only one infohash as the output.
+
+        :param limit: The number of infohashes to randomly get the preferred infohash for (the output set may be less).
+        :returns: A set of infohashes of size 0 up to ``limit``.
+        """
+        with db_session:
+            random_selection = self.InfohashPreference.select_random(limit=limit)
+            return {self._select_superior(ih) for ih in random_selection}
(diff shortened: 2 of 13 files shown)


src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+    layer.store("test query", InfoHash(b"\x00" * 20), set())
+
+    with db_session():
+        queries = layer.Query.select()[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert len(queries[0].infohashes) == 1
+        assert list(queries[0].infohashes)[0].infohash == b"\x00" * 20
+        assert float_equals(list(queries[0].infohashes)[0].preference, 1.0)
```
drew2a, line 38: NIT Same logic, but a bit more correct and easier to read.

The suggested version uses `Query.get()` instead of `Query.select()` because there should only be a single entity in the database, which you assert later. It also confines the `with` block to only include lines that actually need the `db_session`. The suggested version avoids multiple, yet identical, conversions to a list and retrieving the first element of the list.

```suggestion
    with db_session():
        test_query = layer.Query.get()
        infohashes = list(test_query.infohashes)

    assert test_query.query == "test query"
    assert len(infohashes) == 1
    
    infohash = infohashes.pop()
    assert infohash.infohash == b"\x00" * 20
    assert float_equals(infohash.preference, 1.0)
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    with db_session():
+        queries = layer.Query.select()[:]
+        winner, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x00" * 20)[:]
+        loser, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(winner.preference, 1.0)
+        assert float_equals(loser.preference, 0.0)
```
drew2a, line 55: NIT: the version avoids unnecessary conversions to a list and list copying:

```suggestion
        test_query = layer.Query.get()
        winner = layer.InfohashPreference.get(lambda x: x.infohash == b"\x00" * 20)
        loser = layer.InfohashPreference.get(lambda x: x.infohash == b"\x01" * 20)

    assert float_equals(winner.preference, 1.0)
    assert float_equals(loser.preference, 0.0)

    assert test_query.query == "test query"
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+        loser_1, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x01" * 20)[:]
+        loser_2, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x02" * 20)[:]
+        loser_3, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(winner.preference, 1.0)
+        assert float_equals(loser_1.preference, 0.0)
+        assert float_equals(loser_2.preference, 0.0)
+        assert float_equals(loser_3.preference, 0.0)
```
drew2a, line 78: More compact yet correct version of the same logic. It uses `select` where it is intended to query a list and `get` when it is supposed to be just a single entity.

```suggestion
        test_query = layer.Query.get()
        winner = layer.InfohashPreference.get(lambda x: x.infohash == b"\x00" * 20)
        losers = list(layer.InfohashPreference.select(lambda x: x.infohash != b"\x00" * 20))

    assert test_query.query == "test query"
    assert float_equals(winner.preference, 1.0)
    assert len(losers) == 3
    assert all(float_equals(ls.preference, 0.0) for ls in losers)
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+        assert float_equals(loser.preference, 0.0)
+
+
+def test_store_with_losers(layer: UserActivityLayer) -> None:
+    """
+    Test that queries with multiple losers can be stored and retrieved.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20),
+                                                       InfoHash(b"\x02" * 20),
+                                                       InfoHash(b"\x03" * 20)})
```
drew2a, line 64: It is easier to read the test function if you specify the argument names:

```suggestion
    layer.store("test query", infohash=InfoHash(b"\x00" * 20), losing_infohashes={InfoHash(b"\x01" * 20),
                                                                                  InfoHash(b"\x02" * 20),
                                                                                  InfoHash(b"\x03" * 20)})
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.8)
+        assert float_equals(entry_3.preference, 0.0)
+        assert float_equals(entry_4.preference, 0.0)
+
+
+def test_store_delete_old(layer: UserActivityLayer) -> None:
+    """
+    Test result decay after updating.
```
drew2a, line 109: The description of the test likely suffers from missing specifications about the difference between it and the previous test.

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+        should_be_dropped = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+        entry_4, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x04" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.0)
+        assert float_equals(entry_3.preference, 0.0)
+        assert should_be_dropped == []
+        assert float_equals(entry_4.preference, 0.8)
```
drew2a, line 132: NIT: the version avoids unnecessary select querying and list copying:

```suggestion
        test_query = layer.Query.get()
        entry_1 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x00" * 20)
        entry_2 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x01" * 20)
        entry_3 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x02" * 20)
        entry_4 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x04" * 20)
        should_be_dropped = layer.InfohashPreference.get(lambda x: x.infohash == b"\x03" * 20)

    assert test_query.query == "test query"
    assert float_equals(entry_1.preference, 0.2)
    assert float_equals(entry_2.preference, 0.0)
    assert float_equals(entry_3.preference, 0.0)
    assert float_equals(entry_4.preference, 0.8)
    assert not should_be_dropped
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+        entry_4, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x03" * 20)[:]
+        entry_5, = layer.InfohashPreference.select(lambda x: x.infohash == b"\x04" * 20)[:]
+
+        assert len(queries) == 1
+        assert queries[0].query == "test query"
+        assert float_equals(entry_1.preference, 0.2)
+        assert float_equals(entry_2.preference, 0.0)
+        assert float_equals(entry_3.preference, 0.0)
+        assert float_equals(entry_4.preference, 0.0)
+        assert float_equals(entry_5.preference, 0.8)
```
drew2a, line 161: NIT: the version avoids unnecessary select querying and list copying:

```suggestion
        test_query = layer.Query.get()
        entry_1 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x00" * 20)
        entry_2 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x01" * 20)
        entry_3 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x02" * 20)
        entry_4 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x03" * 20)
        entry_5 = layer.InfohashPreference.get(lambda x: x.infohash == b"\x04" * 20)

    assert test_query.query == "test query"
    assert float_equals(entry_1.preference, 0.2)
    assert float_equals(entry_2.preference, 0.0)
    assert float_equals(entry_3.preference, 0.0)
    assert float_equals(entry_4.preference, 0.0)
    assert float_equals(entry_5.preference, 0.8)
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+    assert layer.get_preferable(b"\x00" * 20) == b"\x00" * 20
+
+
+def test_get_preferable_already_best(layer: UserActivityLayer) -> None:
+    """
+    Test if a infohash returns itself when it is preferable.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    assert layer.get_preferable(b"\x01" * 20) == b"\x00" * 20
```
drew2a, line 179: There is a type mismatch: "Expected type 'InfoHash', got 'bytes' instead"

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def test_get_preferable_unknown(layer: UserActivityLayer) -> None:
+    """
+    Test if a infohash returns itself when it has no known preferable infohashes.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20)})
+
+    assert layer.get_preferable(b"\x02" * 20) == b"\x02" * 20
```
drew2a, line 188: There is a type mismatch: "Expected type 'InfoHash', got 'bytes' instead"

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Test if the preferred infohash always gets returned from a random checked selection.
+    """
+    layer.store("test query", InfoHash(b"\x00" * 20), {InfoHash(b"\x01" * 20), InfoHash(b"\x02" * 20)})
+    layer.store("test query", InfoHash(b"\x01" * 20), {InfoHash(b"\x00" * 20), InfoHash(b"\x02" * 20)})
+
+    random_selection = layer.get_random(limit=1)
+
+    assert len(random_selection) == 1
+    assert list(random_selection)[0] == b"\x01" * 20
```
drew2a, line 201: This is an unnecessary conversion to a list and retrieval of the first item:

```suggestion
    assert random_selection.pop() == b"\x01" * 20
```

src/tribler/core/components/database/db/layers/tests/test_user_activity_layer.py
```diff
@@ -0,0 +1,201 @@
...
+    database.bind(provider="sqlite", filename=":memory:")
+    ual = UserActivityLayer(database)
+    database.generate_mapping(create_tables=True)
+    yield ual
+    database.disconnect()
+
+
+def float_equals(a: float, b: float) -> bool:
+    return round(a, 5) == round(b, 5)
+
```
drew2a, line 23: In this file, you are using the same construction over and over again:

```python
def test_example(layer: UserActivityLayer) -> None:
    layer.store("test query", InfoHash(b"\x00" * 20), set())

    with db_session():
        queries = layer.Query.select()[:]

        assert queries[0].query == "test query"
        assert len(queries[0].infohashes) == 1
        ...
```

This is basically a single test case that repeated in `test_store_no_losers`, `test_store_with_loser`, `test_store_weighted_decay`, `test_store_delete_old`, `test_store_delete_old_over_e`.

I would suggest that you extract it to a separate test. Then there will be no need to repeat this in other tests, and they will more accurately describe the specific test case they are testing, without any excess.

Like:

```python
def test_store_query(layer: UserActivityLayer) -> None:
    layer.store("test query", InfoHash(b''), set())

    with db_session():
        test_query = layer.Query.get()

    assert test_query.query == "test query"
```

src/tribler/core/components/database/db/layers/user_activity_layer.py
```diff
@@ -0,0 +1,128 @@
...
+
+        class Query(database.Entity):
+            query = orm.PrimaryKey(str)
+            infohashes = orm.Set("InfohashPreference")
+
+        class InfohashPreference(database.Entity):
+            infohash = orm.Required(bytes)
+            preference = orm.Required(float)
+            parent_query = orm.Required(Query)
+            orm.PrimaryKey(infohash, parent_query)
```
drew2a, line 50: The database does not appear to be normalized. As @kozlovsky is coming back from vacation at the same time as you, I summon him to review the database structure.

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+        query = data.get("query")
+        if query is None:
+            return
+
+        results = {tmd.infohash for tmd in data["results"]}
+        for infohash in results:
+            self.infohash_to_queries[infohash].append(query)
+        self.queries[query] = results | self.queries.get(query, set())
+
+        if len(self.queries) > self.max_query_history:
```
drew2a, line 71: Perhaps for this purpose (to store a limited amount of data), you could use a dedicated data structure. This would make it possible to better cover it with tests and to use it in other parts of Tribler's code. As a beneficial side effect, it would be easier to understand the logic of the dedicated data structure and of the current method.

We can extend this draft:

```python
class LimitedSizeDict(OrderedDict):
    def __init__(self, *args, size_limit=None, **kwargs):
        self.size_limit = size_limit

    def __setitem__(self, key, value):
        super().__setitem__(key, value)
        self._check_size_limit()

    def _check_size_limit(self):
        if self.size_limit is None:
            return
        while len(self) > self.size_limit:
            self.popitem(last=False)
```

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+        When a torrent finishes, check if we were tracking the infohash. If so, store the query and its result.
+        """
+        b_infohash = InfoHash(unhexlify(infohash))
+        queries = self.infohash_to_queries[b_infohash]
+        for query in queries:
+            losing_infohashes = self.queries[query] - {b_infohash}
+            fut = get_running_loop().run_in_executor(None,
+                                                     self.database_manager.store,
+                                                     query, b_infohash, losing_infohashes)
+            self.task_manager.register_task("Store query", functools.partial(UserActivityComponent._fut_to_task, fut))
```
drew2a, line 94: Two points:
1. This method seems overly complicated for its simple purpose — to call the `self.database_manager.store` function.
2. I'm uncertain if the logic will work correctly in cases where the user downloads two or more torrents from the query results.

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        Check a preferable torrent.
+
+        This causes a chain of events that leads to the torrent being gossiped more often in the ``ContentDiscovery``
+        community.
+        """
+        random_infohashes = self.database_manager.get_random(limit=1)  # Note: this set can be empty!
+        for infohash in random_infohashes:
+            self.task_manager.register_anonymous_task("Check preferable torrent",
+                                                      self.torrent_checker.check_torrent_health, infohash)
```
drew2a, line 106: Here's a simplified version: This version avoids a for loop for a set of just a single item and avoids an unnecessary intermediate call for `register_anonymous_task`.

If you really want queuing for `check_torrent_health`, it would be better to implement it on the `check_torrent_health` side.

```suggestion
        if not random_infohashes:
            return

        infohash = random_infohashes.pop()
        await self.torrent_checker.check_torrent_health(infohash)
```

As a side note, if you never use `get_random` with a limit different from 1 (excluding tests), then it is probably a sign that the function should not return a set but a single item. See: [YAGNI on Wikipedia](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

drew2a (changes requested):
The PR appears to be a nice feature that looks promising for Tribler.

It is an interesting concept that appears quite similar to ClickLog. If that's the case, then it would be beneficial to add links to the ClickLog documentation as a reference in the newly added classes and components.

I've suggested a few code improvements and raised some points for discussion.

Also, I'm adding @kozlovsky as a reviewer since there is a new database structure implementation involved.

kozlovsky (changes requested):
I like the idea of this PR! It is crucial for Tribler to understand which torrent users prefer for a specific query because it is tough to properly rank torrents for a particular query without anonymized user feedback. 

The system that analyzes user preferences and shares the anonymized aggregated results looks to me like a cornerstone of a future Tribler's search system.

Initially, I thought we needed to gather that information on the UI side, but this PR shows it is possible to do it entirely on the Core side, which brings simplicity.

With this, it may be hard to develop a good algorithm on the first try, and I believe the specific details of how calculations are performed can be changed in the future. Therefore, storing integer values instead of floats and calculating the resulting rank on the fly may be better, as it allows changing the formula in the future.

I mean, it is possible to keep the number of times it was searched for each Query and for each info hash that appeared in the query results to keep the number of times it was chosen or ignored. This information should be enough to calculate the rank and allow for changing the ranking formula in the future without the data being lost.

---

So what this PR literally does? Right now, it just adds repeated health checks for torrents that the user downloaded. This does not look very useful and can be implemented much more simply; we can have a table with a list of locally downloaded info hashes and randomly check the health of its items. With the current PopularityCommunity protocol, it will not help much, and if this is the end goal of the current PR, then a significant part of it is unnecessary. But I believe that in the future, we should spread not only info hash health info but also should gossip anonymized user preferences for query results (in the form "for query Q, the info hash H with title T was chosen by someone"), and this PR lays down a foundation for this.

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+            query, results = self.queries.popitem(False)
+            for infohash in results:
+                self.infohash_to_queries[infohash].remove(query)
+                if not self.infohash_to_queries[infohash]:
+                    self.infohash_to_queries.pop(infohash)
+
+
+    @staticmethod
+    async def _fut_to_task(future: Future):
+        await future
```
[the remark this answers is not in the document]
qstokkink: Yes, we can and originally I did this but Codacy does not like this. If you still want this, I can suppress the style checker.

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+
+
+class UserActivityComponent(Component):
+
+    def __init__(self, reporter: typing.Optional[SentryReporter] = None) -> None:
+        super().__init__(reporter)
+
+        self.infohash_to_queries: dict[InfoHash, list[str]] = defaultdict(list)
+        self.queries: OrderedDict[str, typing.Set[InfoHash]] = OrderedDict()
```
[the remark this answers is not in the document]
qstokkink: Correct. By design, I only consider downloads in the current session.

I'm also not sure if it makes sense to change this. For example, something a user searched for in a previous Tribler session could be 3 months ago. There would have to be some timeout/validity system etc. etc. We can add that at some point but I don't think this should be in this first prototype.

src/tribler/core/components/database/db/layers/user_activity_layer.py
kozlovsky: After some additional thought, I agree with @drew2a that it may be better to link the `InfohashPreference` entity with the `Resource` entity in the same way as the `Tracker` entity of the `HealthDataAccessLayer` is linked with the `Resource` entity via the `Tracker.torrents` / `Resource.trackers` relationships.

It has the following benefits:
1. The primary key of `InfohashPreference` becomes shorter, as it now will use resource id instead of infohash bytes.
2. With the current approach of `TriblerDatabase`, the torrent metadata is just a kind of `Resource`, and having `Resource` directly linked with `InfohashPreference` can simplify some future queries.

The drawback is when we want to search `InfohashPreference` knowing the specific info hash, we first need to find the `Resource` and then use it instead of the info hash value.

With this change, the code will be like:
```python
class UserActivityLayer:

    def __init__(self, knowledge_layer: KnowledgeDataAccessLayer) -> None:
        self.database = knowledge_layer.instance
        self.Resource = knowledge_layer.Resource

        class Query(self.database.Entity):
            query = orm.PrimaryKey(str)
            searched_counter = orm.Required(int, default=1)
            infohashes = orm.Set("InfohashPreference")

        class InfohashPreference(self.database.Entity):
            torrent = orm.Required(self.Resource)
            query = orm.Required(Query)
            chosen_counter = orm.Required(int, default=0)
            ignored_counter = orm.Required(int, default=0)
            orm.PrimaryKey(torrent, query)

        self.Query = Query
        self.UserPreference = InfohashPreference
```

(In this example, I renamed `parent_query` to `query`, as `parent_` prefix looks unnecessary)

And, in the `Resource` entity of the `KnowledgeDataAccessLayer` we will have:
```python
infohash_preferences = orm.Set(lambda: db.InfohashPreference, reverse="torrent")
```

qstokkink: I don't see how you could implement decay of previous search and results for the same (infohash, query) with this database format. Is it still possible? Because that is a requirement.

src/tribler/core/components/user_activity/user_activity_component.py
qstokkink: This is solving a real issue: it is related to #4320 and #7784.

IIRC, the observation that we should not block the main `asyncio` thread with db operations should be accredited to @egbertbouman but I cannot find his original comment for the life of me. Once I find it, I'll properly link it.

src/tribler/core/components/database/db/layers/user_activity_layer.py
kozlovsky: Indeed, it is not enough to implement the proper decay; I missed that requirement. But it leads me to some additional thoughts.

The current scheme implemented in this PR is single-user. I don't think decay is important when the statistics are accumulated only for a single user. But if we aggregate anonymized query-result-preference statistics about thousands of users, the decay indeed makes sense.

But then we have a new question on spreading and accumulating these statistics. It probably should be signed by a second key when gossiping to prevent spam. However, we cannot sign the dynamically decaying `preference` value of the `float` type. We can sign some discrete information that at the moment T, an anonymous user U performed the query Q and clicked on the infohash H.

So, if the goal is to aggregate decaying anonymous user clicks-at-query-results statistics across multiple users, the discrete signed piece of information should probably be (T, U, Q, H). Then, the decay can be implemented by taking the time stamps into account - the weight of the user's "vote" can be inversely proportional to the vote's age.

In that case, the entity attributes might be something like:

```python
class InfohashPreference(self.database.Entity):
    user = orm.Required(User)  # some new entity
    query = orm.Required(Query)
    torrent = orm.Required(self.Resource)
    last_clicked_at = orm.Required(datetime)
    signature = orm.Required(bytes)
    orm.PrimaryKey(user, query, torrent)
```

What do you think?

src/tribler/core/components/user_activity/user_activity_component.py
```diff
@@ -0,0 +1,110 @@
...
+from tribler.core.components.database.db.layers.user_activity_layer import UserActivityLayer
+from tribler.core.components.libtorrent.libtorrent_component import LibtorrentComponent
+from tribler.core.components.torrent_checker.torrent_checker.torrent_checker import TorrentChecker
+from tribler.core.components.torrent_checker.torrent_checker_component import TorrentCheckerComponent
+from tribler.core.components.user_activity.settings import UserActivitySettings
+from tribler.core.components.user_activity.types import InfoHash
+from tribler.core.sentry_reporter.sentry_reporter import SentryReporter
+
+
+class UserActivityComponent(Component):
```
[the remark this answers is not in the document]
qstokkink: Agreed. I had this in my original design as well (https://github.com/Tribler/tribler/issues/7632#issuecomment-1855636285) However, the impelementation was so small and trivial that I moved the code here. Note that this code is not any less testable due to this (100% coverage).

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
kozlovsky: As I just wrote [here](https://github.com/Tribler/tribler/pull/7786#discussion_r1439564727), I think that we should add the timestamp field to the `InfohashPreference`. Then, we can update this field in the database if the user clicks again at the same infohash.

In that case, the pure-db implementation looks relatively straightforward. When the user clicks on the search result, we add the `InfohashPreference` object with the current timestamp to the database and start the download. When the download is finished, we mark the corresponding `InfohashPreference` as successful. We are gossiping only successfully downloaded `InfohashPreference` objects. Still, the timestamp is recorded at the time when the user clicks on the torrent and not at the time when the download is complete. We update the timestamp if the user clicks again at the same torrent in the new query with the same query string.

src/tribler/core/components/database/db/layers/user_activity_layer.py
qstokkink: That is close to what I had in mind for the long term, in a different PR. I would prefer we discuss the grand design in the linked issue, not on this PR.

Just to touch on it, in short, the plan for now is to use emergent behavior, as follows:
1. Torrents that are "preferred" have their health checked more frequently locally (this PR).
2. Torrents that have their health checked recently (locally) are more frequently gossiped in the popularity community (already exists).
3. Torrents that are gossiped more by others have a higher chance of appearing in search results remotely (already exists).
4. _Effect:_ search results that are gossiped to more users are more likely to be downloaded. For actually popular content, that is downloaded, this forms a feedback loop: back to step 2.

In summary, this PR creates a soft bias and, therefore, an emergent effect that boosts the popularity of content that is searched for and downloaded.

Establishing shadow identities and more aggressively gossiping - while preventing spam - is something I'll leave for a follow-up PR. Ideally, we don't need to gossip preference directly and we can somehow merge gossiped ML models. However, this should only be implemented after careful experimentation. For now, this PR gives each user a local ranking to start the ML experimentation.

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
qstokkink: Sure, I do not doubt the feasibility of adding this. My main concern is this:

> I don't think this should be in this first prototype.

We can add features - like you suggested - later. This is not the goal of this initial prototype.

src/tribler/core/components/database/db/layers/user_activity_layer.py
kozlovsky: Thanks for the explanation; now I understand your approach better. Initially, I was misguided by the picture in #7632 with a "store signed" label, as my understanding was that it is only possible to sign discrete facts, not float values. If gossiping about individual provable facts is not the intention, then storing calculated preferences is probably fine.

Still, you can consider using `torrent = orm.Required(self.Resource)` instead of `infohash = orm.Required(bytes)` in the `InfohashPreference` entity to reduce the data size.

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
kozlovsky: Ok, I agree that we can switch to the pure-db implementation later if necessary. Let's hope Tribler restarts will not significantly skew the preference statistics about big torrents.

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
kozlovsky: I prefer the style checker suppression instead of the separate method, as I hope we can find a way to write elegantly without the Codacy warning.

src/tribler/core/components/database/db/layers/user_activity_layer.py
qstokkink: Sure, thank you for the suggestion. Once the initial prototype has been merged, we can look at refactoring and optimizations and I will definitely keep your suggestion in mind. That said, once this has been merged, this PR is also no longer my sole responsibility and others can also contribute their excellent suggestions to the communal code. We can grow the code over time.

I do realize, now, that I left `enabled = True` as the default setting. I'll make sure to keep this disabled by default so we still have the freedom to change things like the database format in future PRs.

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
qstokkink: I had to dive back into the docs and I think the following works:

```python
self.task_manager.register_task("Store query", asyncio.wait([fut]))
```

I do have to double-check that this is actually properly scheduling everything.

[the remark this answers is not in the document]
qstokkink: I'm open to suggestions, if anyone knows of a magic one-liner.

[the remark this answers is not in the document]
egbertbouman: You could just let the `TaskManager` create the future. Something like:

```python
self.task_manager.register_anonymous_task("Store query", get_running_loop().run_in_executor,
                                          None, self.database_manager.store,
                                          query, b_infohash, losing_infohashes)
```

src/tribler/core/components/user_activity/user_activity_component.py
qstokkink: I did manage to shrink this a little bit (see [conversation with @kozlovsky](https://github.com/Tribler/tribler/pull/7786#discussion_r1439313557)).

src/tribler/core/components/user_activity/user_activity_component.py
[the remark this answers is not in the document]
qstokkink: @egbertbouman [your message did not load in before I posted the above] thanks, that's even shorter! I adopted your approach.

synctext:
This Pr is a great example of how we want to work. I'm split between:
- solid code. solid code. solid code. Polish code till it's perfect. ```solid code == solid review```
- just get stuff done. Don't re-do the original PR effort. Only talk about actual faults.

[PR_discussion_21-pages.pdf](https://github.com/Tribler/tribler/files/13821120/PR_discussion_21-pages.pdf). These 663 of extra lines of code produced 21 pages of discussion on Github. Frankly I'm getting a bit desperate that Tribler is not going anywhere. Trying to make it too perfect is not good, paradoxically. After 18 years, 9 months we still don't have anything that users actually want (caveat: without doing almost any PR). Most code does not last more then a few years anyways. So let's not make it too perfect and polished. Just to clarify, obviously it is critical that the code does not have nasty bugs. How much time should we spend on reviews? Say 20% of the original PR engineering effort?

So let's focus the effort of code review on identification of actual faults.

qstokkink:
It does not make sense to work on this PR until #7799 is resolved. If it is unclear where protype code should go, it should not go anywhere.

I'll close this PR for now because it falsely suggests that it is a work in progress. If the team does decide that prototype code should live in Tribler, we can reopen this PR.

synctext:
Approach to resolve this. First get operational Gumby experiment of this feature.
Do another PR and release only on forum.tribler for experimental release of Tribler.
In general we have always had difficulty of the development process with zero bugs and the phd research side which requires novel never-done-before things.

PR #7795, opened 2024-01-03
Fix OSError: [WinError 123] for the `_check_watch_folder`

### LINKED ISSUE #7794: [7.13.1] OSError: [WinError 123] A sintaxe do nome do ficheiro, do nome do diretório ou da etiqueta do volume é inc...
Sentry Issue: [TRIBLER-1GS](https://sentry.tribler.org/organizations/tribler/issues/2543/?referrer=github_integration)

```python
OSError: [WinError 123] A sintaxe do nome do ficheiro, do nome do diretório ou da etiqueta do volume é incorreta: 'C:\\Users\\<attention>\\AppData\\Roaming\\.Tribler\\7.13\\file:\\C:\\Users\\<attention>\\Desktop\\<hash>.dlc'
  File "tribler\core\components\watch_folder\watch_folder.py", line 43, in _check_watch_folder_handle_exceptions
  File "tribler\core\components\watch_folder\watch_folder.py", line 56, in _check_watch_folder
  File "pathlib.py", line 1403, in is_dir
  File "pathlib.py", line 1194, in stat
```

drew2a:
This PR fixes #7794 by adding the following block to check the validity of the path:

```python
        if not directory.is_valid():
            self._logger.warning(f'Cancelled. Directory is not valid: {directory}.')
            return False
```

Where `is_valid` is the following:

```python
    def is_valid(self):
        """ Check if the path is valid.
        Returns: True if the path is valid, False otherwise.
        """
        try:
            return self.is_file() or self.is_dir()
        except OSError:
            return False
```

The bulk of this PR consists of tests for previously written code, which might not be part of the fix. However, I think it's acceptable to include these to increase coverage and satisfy the coverage tool.

### DIFF
--- a/src/tribler/core/components/watch_folder/tests/test_watch_folder.py
+++ b/src/tribler/core/components/watch_folder/tests/test_watch_folder.py
@@ -1,6 +1,6 @@
 import asyncio
 import shutil
-from unittest.mock import MagicMock, Mock, patch, AsyncMock
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 
@@ -36,7 +36,7 @@ async def watch_folder(tmp_path):
 
 async def test_watch_folder_no_files(watch_folder):
     # Test that in the case of an empty folder, downloads are not started
-    await watch_folder._check_watch_folder()
+    assert await watch_folder._check_watch_folder()
 
     assert not watch_folder.download_manager.start_download.called
 
@@ -46,7 +46,7 @@ async def test_watch_folder_no_torrent_file(watch_folder: WatchFolder):
     directory = watch_folder.settings.get_path_as_absolute('directory', watch_folder.state_dir)
     shutil.copyfile(TORRENT_UBUNTU_FILE, directory / "test.txt")
 
-    await watch_folder._check_watch_folder()
+    assert await watch_folder._check_watch_folder()
 
     assert not watch_folder.download_manager.start_download.called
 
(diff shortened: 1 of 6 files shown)


src/tribler/core/components/watch_folder/tests/test_watch_folder.py
```diff
@@ -120,3 +120,34 @@ async def test_watch_folder_no_crash_exception(watch_folder: WatchFolder):
     # Test that errors raised in `_check_watch_folder` reraise as `NoCrashException`
     with pytest.raises(NoCrashException):
         await watch_folder._check_watch_folder_handle_exceptions()
+
+
+async def test_watch_folder_invalid_dir(watch_folder: WatchFolder, tmp_path):
+    """ Test that the watch folder is disabled when the directory is invalid"""
+    watch_folder.settings.put_path_as_relative(
+        property_name='directory',
+        value=Path('incorrect path'),
```
kozlovsky, line 129: **nitpick:** This `'incorrect path'` is not actually incorrect; it just does not exist. Correspondingly, this path does not trigger `OSError` inside the `is_valid` method. But it is probably OK as handling OSError is covered by `test_is_invalid_by_os_exception`. As I understand it, making a path that is universally syntactically incorrect on all OSes is a bit tricky.

kozlovsky (approved):

src/tribler/core/components/watch_folder/tests/test_watch_folder.py
drew2a: You are right :) What if we use this construction:

```python
async def test_watch_folder_invalid_dir(watch_folder: WatchFolder, tmp_path):
    """ Test that the watch folder is disabled when the directory is invalid"""
    watch_folder.settings.put_path_as_relative(
        property_name='directory',
        value=Path('# incorrect path'),
        state_dir=Path(tmp_path)
    )

    assert not await watch_folder._check_watch_folder()

```

PR #7798, opened 2024-01-04
Remove `BandwidthCommunity`

### LINKED ISSUE #3731: Give the user insight in the history of received/spent bandwidth tokens
We should give the user more insight in the user interface regarding the history of bandwidth tokens. A basic list that shows the history of incoming/outgoing bandwidth tokens should already address this.

### LINKED ISSUE #4255: Improve the bandwidth token payout mechanism
As we have seen in issue #4234, the current payout mechanism is functional (as most of the time, people are earning bandwidth tokens for running Tribler idle), but it can and should be improved. Note that #4248 and #4250 addressed issues during payout.

- Get rid of the 5-second window during which a payout should be made - after receiving a `destroy` message, a circuit continues to live for another 5 seconds. During this time, a payout should be made. If the payout takes longer, the payout will most likely not reach the exit node.
- Remove requirement for honest behaviour of intermediaries - the current mechanism heavily relies on intermediary (relay) nodes in a circuit acting honest and transferring tokens to the next hop in a circuit. This can easily be abused. Possible solutions: incremental payouts? Detect whether a node did not pay and not join their circuits?
- Improve the visualization of the trust score. Instead of calling it MBs given/taken (which is the technical perspective), we should name it investment/yield for instance.
- Long term: Implement deferred payouts - this prevents a correlation attack based on TrustChain records, since one can infer which nodes did payout each other.

As more issues are discovered, they will be documented here.

### LINKED ISSUE #4939: Trust statistics 2.0
Every now and then, our users raise the question about their token balance. Even with good intentions of running Tribler to help the community, they are labelled as selfish users (in Trust graph) because of their negative token balance where they have no idea how they got their balance in the negative.

We should provide answers to them with more details breaking down the balance represented in a clear GUI. 
Something like:
- **Downloaded (- MB)**
  - Torrents
  - Channels
  - Bootstrap
- **Anonymous seeding**
  - Tribler Users (+ MB)
  - Non-Tribler Users (- MB)
- **Relayed (+ MB)**

### LINKED ISSUE #5212: Hidden seeding needs in-GUI explanation regarding tokens
Anonymous downloading of torrents costs bandwidth tokens. 
But it's unclear from the user interface and the FAQ whether the option "anonymous seeding using proxies" also costs tokens. Does it?
I think this should be clarified in either the user interface or the FAQ.

### LINKED ISSUE #7336: Refactoring: `BandwidthCommunity` may be removed.

drew2a:
This PR removes the `BandwidthCommunity` and closes #7336. It also contains a migration for removing `bandwidth.db` files.

I've decided to leave the `BandwidthCrawler` as an example of a simple crawler for Tribler, for the next generation of developers or PhD students.

`PR / scripttest_draft / run (pull_request_target)` is failing because, due to security reasons, the content of YML files is always taken from the target branch. Therefore, the checks for the current PR use the old (unchanged) `scripttest.yml`. After merging, this check should be correct.

Closes #5212
Closes #4939
Closes #4255
Closes #3731


drew2a:
@egbertbouman, I'm requesting your review as it turns out that the `BandwidthCommunity` was intertwined with the `TunnelCommunity`. I've tried to be as careful as possible, but it's hard to grasp all the logic from `TunnelCommunity` and verify on my own that my changes are correct. However, all tests have passed, and anonymous downloading is working (I've tried it both on new downloads and existing ones).

egbertbouman (commented):
Looks good to me. A couple of things to note:
* `BalanceRequestCache` can be removed, as well as the cache check in `should_join_circuit`.
* Since messages/cells have been removed, but old peers are still sending these messages, I'm guessing there is a lot of log spam now.
* While the network updates to the new code, there will be a bias toward people that are running the old code (e.g. new peers only compete for random slots, whereas old peers compete for all slots). Not really a problem, but  anyway.

drew2a:
I've removed `BalanceRequestCache` and the cache check in should_join_circuit: https://github.com/Tribler/tribler/pull/7798/commits/d6edc77bb40ad00a6bf7dabb8ad761cf9c41b658 (I'll swap the commits before merge).

> Since messages/cells have been removed, but old peers are still sending these messages, I'm guessing there is a lot of log spam now.

Yes. But as we've seen that the network updates to the new version quite quickly (it takes a couple of weeks), then it is probably worth it to wait without any temporary warning suppression.

> While the network updates to the new code, there will be a bias toward people that are running the old code (e.g. new peers only compete for random slots, whereas old peers compete for all slots). Not really a problem anyway.

Agreed. Also, I'll note here just in case, that I've increased the random slot count from 5 to 20 as it seems fair to keep the same amount of slots as before (random_slots (5) + competing_slots (15)).

egbertbouman (approved):

PR #7815, opened 2024-01-12
Refactor magnet link parsing logic and add new test cases

drew2a:
This PR refactors the `parse_magnetlink` function to use `libtorrent.parse_magnet_uri` for parsing magnet links, as it seems safer and more convenient to rely on the native libtorrent parsing mechanism instead of our own.

It is includes:
- Added a new test case for parsing a magnet link with a wrong hash.
- Added a new test case for parsing a magnet link with bytes input.
- Removed two existing test cases that were testing invalid infohashes, as they are no longer needed.

This refactoring was carried out during the work on:
- #7801

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/download_manager.py
@@ -11,7 +11,7 @@ from asyncio import CancelledError, gather, iscoroutine, shield, sleep, wait_for
 from binascii import unhexlify
 from copy import deepcopy
 from shutil import rmtree
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Union
 
 from ipv8.taskmanager import TaskManager
 
@@ -489,7 +489,7 @@ class DownloadManager(TaskManager):
         lt_session.set_ip_filter(ip_filter)
 
     async def get_metainfo(self, infohash: bytes, timeout: float = 30, hops: Optional[int] = None,
-                           url: Optional[str] = None, raise_errors: bool = False) -> Optional[Dict]:
+                           url: Optional[Union[str, bytes]] = None, raise_errors: bool = False) -> Optional[Dict]:
         """
         Lookup metainfo for a given infohash. The mechanism works by joining the swarm for the infohash connecting
         to a few peers, and downloading the metadata for the torrent.
@@ -583,14 +583,12 @@ class DownloadManager(TaskManager):
             self._logger.info('Magnet scheme detected')
             name, infohash, _ = parse_magnetlink(uri)
             self._logger.info(f'Name: {name}. Infohash: {infohash}')
-            if infohash is None:
-                raise RuntimeError("Missing infohash")
             if infohash in self.metainfo_cache:
                 self._logger.info('Metainfo found in cache')
                 tdef = TorrentDef.load_from_dict(self.metainfo_cache[infohash]['meta_info'])
             else:
                 self._logger.info('Metainfo not found in cache')
-                tdef = TorrentDefNoMetainfo(infohash, "Unknown name" if name is None else name, url=uri)
+                tdef = TorrentDefNoMetainfo(infohash, "Unknown name" if not name else name, url=uri)
             return await self.start_download(tdef=tdef, config=config)
         if scheme == FILE_SCHEME:
             self._logger.info('File scheme detected')
--- a/src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
@@ -10,10 +10,10 @@ from ipv8.REST.schema import schema
 from marshmallow.fields import String
 
 from tribler.core import notifications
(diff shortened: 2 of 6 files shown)


src/tribler/core/components/libtorrent/download_manager/download_manager.py
```diff
@@ -463,7 +463,7 @@ def update_ip_filter(self, lt_session, ip_addresses):
         lt_session.set_ip_filter(ip_filter)
 
     async def get_metainfo(self, infohash: bytes, timeout: float = 30, hops: Optional[int] = None,
-                           url: Optional[str] = None, raise_errors: bool = False) -> Optional[Dict]:
+                           url: Optional[Union[str, bytes]] = None, raise_errors: bool = False) -> Optional[Dict]:
```
egbertbouman, line 466: I realize this PR doesn't introduce this, so I'm not requesting you to change this, but it doesn't make a lot of sense to me for an url to be bytes.

src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
```diff
@@ -106,15 +108,17 @@ async def get_torrent_info(self, request):
                 return RESTResponse({"error": str(e)}, status=HTTP_INTERNAL_SERVER_ERROR)
 
             if response.startswith(b'magnet'):
-                _, infohash, _ = parse_magnetlink(response)
-                if infohash:
-                    metainfo = await self.download_manager.get_metainfo(infohash, timeout=60, hops=hops, url=response)
+                infohash = self.get_infohash_from_magnet(response)
```
egbertbouman, line 111: You're not downloading the metainfo anymore.
I'm not sure this is your intention, but if it is, why even parse the magnet link? Since the metainfo will never be set, it's just going to return an error later on.

src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
```diff
@@ -148,3 +152,19 @@ async def get_torrent_info(self, request):
...
+        Returns:
+            bytes: The infohash of the magnet link. If the magnet link is invalid, None is returned.
+        """
+        infohash = None
+        try:
+            _, infohash, _ = parse_magnetlink(magnet)
+        except RuntimeError as e:
+            self._logger.warning(f'Error while getting an ingo hash from magnet: {e.__class__.__name__}: {e}',
+                                 exc_info=e)
+            self.last_error = e
```
egbertbouman, line 169: Since it's possible to have multiple simultaneous requests, I don't think it's a good idea to store exceptions like this. Sure, it will work right now, but if someone ever adds an await directly after calling this function, it could break.

src/tribler/core/components/libtorrent/download_manager/download_manager.py
drew2a: I also realized this after adding typings in this PR. I've created a related issue: https://github.com/Tribler/tribler/issues/7823

src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
drew2a: Thank you! No, indeed, that wasn't my intention; it was a misprint during the refactoring. I've fixed it.

However, a side note: if I was able to make this change without any failed tests, it indicates suboptimal test coverage for the `TorrentInfoEndpoint`.

src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py
drew2a: I can rewrite the whole endpoint in DownloadsEndpoint style:

```python
    async def add_download(self, request):
        ...
        try:
            download = await self.download_manager.start_download_from_uri(uri, config=download_config)
        except Exception as e:
            return RESTResponse({"error": str(e)}, status=HTTP_INTERNAL_SERVER_ERROR)

        return RESTResponse({"started": True, "infohash": hexlify(download.get_def().get_infohash())})

    async def start_download_from_uri(self, uri, config=None):
        """all the logic located here"""
```

egbertbouman: Rewriting the whole thing seems a bit much. You could just change `get_torrent_info` a bit so that so only need to parse the magnetlink once. Then drop `get_infohash_from_magnet`  and catch the exception DownloadsEndpoint-style

drew2a: Or, I could simply surround the `parse_magnetlink` call with a try-except block and leave the refactoring for later.

egbertbouman: Sure, that's fine too.

drew2a: I've implemented the second option, surrounding the `parse_magnetlink` call with a try-except block.

egbertbouman (approved):

PR #7816, opened 2024-01-12
Add typings for the TorrentDef class

drew2a:
This refactoring was carried out during the work on:
- https://github.com/Tribler/tribler/issues/7801

### DIFF
--- a/src/tribler/core/components/libtorrent/torrentdef.py
+++ b/src/tribler/core/components/libtorrent/torrentdef.py
@@ -9,20 +9,21 @@ from asyncio import get_running_loop
 from contextlib import suppress
 from functools import cached_property
 from hashlib import sha1
+from os import PathLike
+from typing import Dict, Iterator, List, Optional, Set, Tuple, Union
 
 import aiohttp
 
 from tribler.core.components.libtorrent.torrent_file_tree import TorrentFileTree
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/libtorrent/torrentdef.py
```diff
@@ -237,19 +240,19 @@ def set_tracker(self, url):
...
 
-    def get_tracker(self):
+    def get_tracker(self) -> Optional[str]:
         """
         Returns the torrent announce URL.
         """
         return self.torrent_parameters.get(b'announce', None)
 
-    def get_tracker_hierarchy(self):
+    def get_tracker_hierarchy(self) -> List:
```
kozlovsky, line 249: It is possible to define the resulted type more strictly as `List[List[str]]`
```suggestion
    def get_tracker_hierarchy(self) -> List[List[str]]:
```
It is a list of "tiers", where each tier is a list of torrent URLs

See: https://github.com/rakshasa/libtorrent/blob/master/doc/multitracker-spec.txt

kozlovsky (changes requested):
Looks correct except the `exts` parameter typing in `get_files`

kozlovsky (approved):

PR #7824, opened 2024-01-15
[WIP] Add dependencies for application tester to run

xoriole:
While fixing the application tester jobs on Jenkins, I found that some dependencies need to be installed to run the application tester from the Tribler repo application tester code. This is a subset of Tribler dependencies.

### DIFF
--- a/.github/workflows/application_tester.yml
+++ b/.github/workflows/application_tester.yml
@@ -70,6 +70,10 @@ jobs:
         if: runner.os == 'Linux'
         uses: pyvista/setup-headless-display-action@v1
 
+      - name: Install application tester dependencies
+        run: |
+          pip install -r ./scripts/application_tester/requirements.txt
+
       - name: Application Tester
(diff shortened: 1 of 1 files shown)


drew2a:
When I added the Application Tester to the GitHub actions, I assumed that the installed requirements from the main Tribler were sufficient for the Application Tester to run. Therefore, I removed the dedicated `requirements.txt` as I thought it was better to have fewer `requirements.txt` files :)

Don't forget to fix the https://github.com/Tribler/tribler/blob/main/.github/workflows/application_tester.yml if you think is necessary to have the dedicated `requirements.txt`.

P.S. Some dependencies look suspicious to me, but as it's currently in draft, I'll leave the review part to the reviewer.

drew2a:
As @qstokkink pointed out, the suggested file is a duplicate of an existing one: https://github.com/Tribler/tribler/blob/main/doc/contributing.rst

PR #7831, opened 2024-01-17
Add dependencies for application tester to run

xoriole:
While fixing the application tester jobs on Jenkins, I found that some dependencies need to be installed to run the application tester from the Tribler repo application tester code. This is a subset of Tribler dependencies.

Several dependencies are required because of the application tester code dependence on `TriblerConfig` class.

Earlier PR https://github.com/Tribler/tribler/pull/7824 was used to test if application tester-specific dependencies are installed during tests. It was done by adding an extra dependency `requests==2.31.0` and using that dependency in the application tester code. The test is successful and the action run for that is available [here](https://github.com/Tribler/tribler/actions/runs/7555493318/job/20570601421#step:10:14).

### DIFF
--- a/.github/workflows/application_tester.yml
+++ b/.github/workflows/application_tester.yml
@@ -70,6 +70,10 @@ jobs:
         if: runner.os == 'Linux'
         uses: pyvista/setup-headless-display-action@v1
 
+      - name: Install application tester dependencies
+        run: |
+          pip install -r ./scripts/application_tester/requirements.txt
+
       - name: Application Tester
         timeout-minutes: 5
         env:
--- a/scripts/application_tester/README.md
+++ b/scripts/application_tester/README.md
@@ -3,8 +3,13 @@
 > **_NOTE:_** code moved from: https://github.com/Tribler/application-tester/
 
 This repository hosts code for the Tribler application tester, which aims to automatically test Tribler using code injection.
-To run the Tribler application tester, it requires a path to the Tribler executable, for example:
 
+To run the Tribler application tester, dependencies specified in `scripts/application_tester/requirements.txt` must be installed. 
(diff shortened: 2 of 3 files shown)


drew2a (changes requested):
Looks good to me, except the `file-read-backwards` dependency seems suspicious as it is used only in Tribler and in Exit Node code (NIT).

Please update the `readme.md` of the Application Tester and add a section about the necessity of installing prerequisites.

xoriole:
> Looks good to me, except the `file-read-backwards` dependency seems suspicious as it is used only in Tribler and in Exit Node code (NIT).
> 
> Please update the `readme.md` of the Application Tester and add a section about the necessity of installing prerequisites.

Here is the stacktrace for dependency `file_read_backwards`
```
/home/user/tribler/venv/bin/python /home/user/tribler/scripts/application_tester/main.py -p python src/run_tribler.py -d 120 
```
```python
  File "/home/user/tribler/scripts/application_tester/main.py", line 8, in <module>
    from tribler.core.utilities.utilities import make_async_loop_fragile
  File "/home/user/tribler/src/tribler/core/utilities/utilities.py", line 24, in <module>
    from tribler.core.utilities.rest_utils import HTTPS_SCHEME, HTTP_SCHEME, scheme_from_url
  File "/home/user/tribler/src/tribler/core/utilities/rest_utils.py", line 6, in <module>
    from tribler.core.utilities.path_util import Path
  File "/home/user/tribler/src/tribler/core/utilities/path_util.py", line 11, in <module>
    from file_read_backwards import FileReadBackwards
ModuleNotFoundError: No module named 'file_read_backwards'
```

drew2a (approved):

PR #7834, opened 2024-01-17
KnowledgeCommunity : Update outdated descriptions

drew2a:
This PR updates outdated descriptions in `KnowledgeCommunity`. The fact that they are outdated became apparent while working on:

- #6214


kozlovsky (approved):
Looks good. I noticed many docstrings in this file (including the two docstrings you are fixing now) start with a leading space. It is not an error but a stylistic issue. There is even a warning in flake8: `H401: docstring should not start with a space`. It may be a good time to add a separate commit in this PR to remove all leading whitespace from this and other docstrings of these two files. But if you want to leave it out of the current PR, we can do it later.

drew2a:
@kozlovsky yes it is a bit out of scope of the current PR. You can open the issue with the suggestion. As we don't use `flake8` we should discuss it first.

Issue #7836, opened 2024-01-18
TypeError: Object of type bytes is not JSON serializable - `DownloadsEndpoint`


xoriole:
Encountered when running application tester on `main` branch from source.

```python3
Traceback (most recent call last):
  File "/home/user/tribler/src/tribler/core/components/restapi/rest/rest_manager.py", line 53, in error_middleware
    response = await handler(request)
  File "/home/user/tribler/venv/<discount>/python3.10/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/home/user/tribler/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py", line 379, in get_downloads
    return RESTResponse({"downloads": downloads_json, "checkpoints": checkpoints})
  File "/home/user/tribler/src/tribler/core/components/restapi/rest/rest_endpoint.py", line 66, in __init__
    body = json.dumps(body)
  File "/usr/<discount>/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/<discount>/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/<discount>/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/<discount>/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f\'Object of type {o.__class__.__name__} \'
TypeError: Object of type bytes is not JSON serializable
```

Additional tags:
```
{
  'source': 'core',
  'tribler_stopped': False,
  'version': '7.14.0-GIT',
  'machine': 'x86_64',
  'os': 'Linux-6.5.0-14-generic-x86_64-with-glibc2.35',
  'platform': 'linux'
}
```

PR #7842, opened 2024-01-22
Fix TorrentDefNoMetainfo.get_name_utf8() to return string

xoriole:
Fixes https://github.com/Tribler/tribler/issues/7836

The downloads response was not JSON serializable because the name was still being returned as bytes in case of serialization of name from torrent def of instance `TorrentDefNoMetainfo`. This PR fixes the issue.

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/download_manager.py
@@ -590,7 +590,7 @@ class DownloadManager(TaskManager):
                 tdef = TorrentDef.load_from_dict(self.metainfo_cache[infohash]['meta_info'])
             else:
                 self._logger.info('Metainfo not found in cache')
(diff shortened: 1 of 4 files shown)


drew2a (approved):
Looks good to me, despite the fact that in `TorrentDefNoMetainfo` there is a violation of the Liskov Substitution Principle, at least in the constructor. However, this is not your fault, but rather a design issue from the author of the class.

qstokkink:
> there is a violation of the Liskov Substitution Principle, at least in the constructor.

@drew2a could you provide more details on this violation? AFAIK the substitution principle does not apply to constructors.

drew2a:
@qstokkink, you are right; the Liskov Substitution Principle is not applicable to constructors. It was a false alarm.

drew2a:
I apologize to the unknown author of this class 🙏

Issue #7823, opened 2024-01-15
`DownloadManager. get_metainfo()` receives an URL in bytes


drew2a:
Here the call passes the URL in bytes:

https://github.com/Tribler/tribler/blob/5f13897dd7d2dc45e610b1a392401ff1e7b609f4/src/tribler/core/components/libtorrent/restapi/torrentinfo_endpoint.py#L108-L111

it looks like a bug as the URL should be a string.

https://github.com/Tribler/tribler/blob/5f13897dd7d2dc45e610b1a392401ff1e7b609f4/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L491-L492

PR #7845, opened 2024-01-23
Add tests for `recursive_unicode` function

drew2a:
While working on https://github.com/Tribler/tribler/issues/7823 I became suspicious that `recursive_unicode` was not functioning correctly. To confirm this, I wrote tests and found that the function works as expected. This PR includes these tests.


egbertbouman (approved):
Looks good to me, although it would be nice to also test if `UnicodeDecodeError` is being handled as expected.

drew2a:
@egbertbouman tested 👍

Issue #7847, opened 2024-01-23
Application tester re-run fails with OS error [Errno 98] Address already in use


xoriole:
Running application tester immediately after the previous run was terminated results in an `OS error [Errno 98] Address already in use`.

```python
[tribler-gui PID:922794] 2024-01-23 16:59:58,364 - ERROR <start_gui:86> tribler.gui.start_gui.run_gui(): Tribler configuration conflicts with the current OS state: REST API port 20100 already in use
Traceback (most recent call last):
  File "/home/user/tribler/src/tribler/gui/start_gui.py", line 77, in run_gui
    window = TriblerWindow(process_manager, app_manager, settings, root_state_dir,
  File "/home/user/tribler/src/tribler/gui/tribler_window.py", line 167, in __init__
    raise RuntimeError(
RuntimeError: Tribler configuration conflicts with the current OS state: REST API port 20100 already in use
********************************************************************************
Tribler Exception
********************************************************************************
RuntimeError: Tribler configuration conflicts with the current OS state: REST API port 20100 already in use
********************************************************************************
ERROR - Executor(125): Tribler process terminated suddenly. Code: 1
ERROR - Executor(103): Tribler process finished unexpectedly with the return code 1
```

Operating System: Ubuntu 22.04
Branch: main

PR #7850, opened 2024-01-24
Set socket reuseaddr to true while checking free port

xoriole:
Fixes https://github.com/Tribler/tribler/issues/7847

### DIFF
--- a/src/tribler/core/utilities/network_utils.py
+++ b/src/tribler/core/utilities/network_utils.py
@@ -74,6 +74,7 @@ class NetworkUtils:
         try:
             for socket_class in self.socket_class_set:
                 with socket_class() as s:
+                    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
                     s.bind(('', port))
             return True
         except OSError:
--- a/src/tribler/core/utilities/tests/test_network_utils.py
+++ b/src/tribler/core/utilities/tests/test_network_utils.py
@@ -1,3 +1,5 @@
+import socket
+
 import pytest
 
 from tribler.core.utilities.network_utils import FreePortNotFoundError, NetworkUtils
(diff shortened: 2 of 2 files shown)


drew2a:
Nice improvement. Could you educate me a bit and share how this change fixes the issue?

xoriole:
> If the local application dies while a connection is active, the port will be tied up in TIME_WAIT. This is also true if the application dies while a connection is pending.
> TIME_WAIT is the state that typically ties up the port for several minutes after the process has completed.

@drew2a Here is a nice write-up about `Bind: Address Already in Use` issue
https://hea-www.harvard.edu/~fine/Tech/addrinuse.html

image

drew2a:
Thank you for the article!

I now understand that you modified `NetworkUtils.is_port_free()` to allow it to return TRUE for ports that are in the `TIME_WAIT` status.

drew2a (changes requested):
Allowing `NetworkUtils.is_port_free()` to return TRUE for ports that are in the TIME_WAIT status leads to two consequences:
1. It allows the GUI to reuse a port that is in `TIME_WAIT` status (which solves the issue with the Application Tester).

https://github.com/Tribler/tribler/blob/becc7277c966e344e7be37d5c0598064890587d5/src/tribler/gui/tribler_window.py#L166-L170

2. It allows `NetworkUtils.get_random_free_port()` and `NetworkUtils.get_first_free_port()` to use ports in `TIME_WAIT` status.

Both cases are risky as there could be a third-party application (not Tribler) that is leaving the port in `TIME_WAIT` to restore it later.

Although this fix resolves the issue with the Application Tester, it could lead to more severe problems in production.

I suggest solving the problem by using a random port for each consecutive Application Tester run (currently, it uses `20100` all the time).

xoriole:
I have implemented another PR https://github.com/Tribler/tribler/pull/7859 as an alternative solution to the current PR. 
Closing this PR in favor of https://github.com/Tribler/tribler/pull/7859.

PR #7853, opened 2024-01-25
Fix `DownloadState.get_ratio()`

### LINKED ISSUE #7817: per torrent ratio and up/down do not match
**Describe the bug**
per torrent ratio and up/down do not match

**To Reproduce**

1. download a torrent
2. restart part way through

**Current behavior**
total ratio is shown with session up / down

**Expected behavior**
total ratio should be shown with total up / down

**Screenshots**

wut(1)

drew2a:
This PR fixes #7817 by changing the function to the following:
```python
    def get_all_time_ratio(self) -> float:
        if not self.lt_status or not self.all_time_download:
            return 0
        return self.all_time_upload / self.all_time_download
```

### DIFF
--- a/src/tribler/core/components/libtorrent/conftest.py
+++ b/src/tribler/core/components/libtorrent/conftest.py
@@ -69,7 +69,10 @@ def mock_lt_status():
     lt_status.download_rate = 43
     lt_status.total_upload = 100
     lt_status.total_download = 200
-    lt_status.all_time_upload = 100
+    lt_status.total_payload_upload = 30
+    lt_status.total_payload_download = 100
+    lt_status.all_time_upload = 200
+    lt_status.all_time_download = 1000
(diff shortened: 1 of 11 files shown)


kozlovsky (changes requested):
I added the comment to the original issue #7817. It looks like the previous formula is correct (except when the `total_done` is zero), and instead of changing the formula, we should display the `all_time_upload` / `all_time_download` counters.

drew2a:
@kozlovsky thank you. I've changed the function to the following:

```python
    def get_all_time_ratio(self) -> float:
        if not self.lt_status or not self.all_time_download:
            return 0
        return self.all_time_upload / self.all_time_download
```

Also, I've changed the UI, please check the screenshot in the description for this issue.

kozlovsky (approved):

PR #7868, opened 2024-01-30
Add `float('inf')` to `all_time_ratio` calculation

### LINKED ISSUE #7865: For seeders, the torrent ratio is shown as zero
Extracted from #7817.

When a user seeds a new torrent, it only uploads it without downloading. Torrent ratio is calculated as `uploaded / downloaded`, and, as `downloaded` is zero, the ratio should be infinite. However, Tribler displays a zero ratio in that case.

Consider the case when a user seeds torrent A and downloads torrent B (without uploading for torrent B). These two situations are opposite and should be displayed differently, but Tribler currently shows zero ratio for both torrents, and it can be confusing for users who are seeders. In my opinion, it is a minor UI bug. It is minor because `ratio` has a purely informative purpose, and not many users are pure seeders and are affected by this issue.

To fix the issue, we can show the ratio ♾ for cases when `downloaded == 0 and uploaded > 0`.

drew2a:
This PR fixes #7865

image


src/tribler/core/components/libtorrent/download_manager/download_state.py
```diff
@@ -156,8 +156,12 @@ def total_payload_download(self) -> int:
     def get_all_time_ratio(self) -> float:
         """ Returns the accumulated seeding ratio of the download across multiple sessions.
         """
-        if not self.lt_status or not self.all_time_download:
+        if not self.lt_status:
             return 0
+
+        if not self.all_time_download:
+            return 0 if not self.all_time_upload else float("inf")
```
kozlovsky, line 163: **nitpick:** I slightly prefer `math.inf` to `float("inf")`, but both are possible.

kozlovsky (approved):
Interestingly, the JSON standard does not allow Infinity or NaN values, but Python extends the format by including them: https://docs.python.org/3/library/json.html#infinite-and-nan-number-values.

As Tribler GUI and Core are both in Python, it should be fine.

src/tribler/core/components/libtorrent/download_manager/download_state.py
drew2a: Agree. Changed.

PR #7872, opened 2024-01-31
Add `MemoryError` to the list of ignored exceptions for `TriblerProce…

### LINKED ISSUE #7839: [7.13.0] MemoryError on Windows
Sentry Issue: [TRIBLER-1J0](https://sentry.tribler.org/organizations/tribler/issues/2585/?referrer=github_integration)

```
MemoryError: 
(4 additional frame(s) were not displayed)
...
  File "tribler\core\utilities\process_manager\manager.py", line 95, in primary_process_rowid
  File "tribler\core\utilities\process_manager\process.py", line 167, in is_running
  File "__init__.py", line 682, in status
  File "psutil\_pswindows.py", line 679, in wrapper
  File "psutil\_pswindows.py", line 1054, in status
```

drew2a:
This PR fixes #7839 by extending the list of ignored exceptions.


egbertbouman (commented):
Could you explain why it's OK to ignore the `MemoryError`?

drew2a:
@egbertbouman, sure.

I followed the original logic of the `is_running(self)` function, as it already ignores `psutil.Error` errors. So, I reasoned that adding `MemoryError` to the list of ignored errors wouldn't significantly alter the system's behavior (1). 

I think in this case, ignoring is okay because it's preferable to swallow an error in non-critical parts of the program than to cause a crash for the user and stop the application when we can continue it (2).

Regarding point (2) – we've had numerous discussions about whether it's better for Tribler to be more fragile or less fragile, but I can't find the specific issue.

egbertbouman (approved):

PR #7875, opened 2024-02-01
Split *nix and Windows error formatting

### LINKED ISSUE #7863: Error code formatting for windows returns `Error message: Unknown error`
We use `os.strerror(exit_code)` to transform an error code into its string representation, as documented in https://docs.python.org/3/library/os.html#os.strerror

https://github.com/Tribler/tribler/blob/1c6baadb719dca4869fc4fd4dd36d1e158f48b96/src/tribler/gui/core_manager.py#L278-L289

However, it seems not to work on Windows OS. As seen in #7855, the error code `-1073741819` was transformed to `Error message: Unknown error` instead of something like `Application Error 0xc0000005 (Access Violation)`.

drew2a:
This PR addresses #7863 by splitting error formatting for Windows and Linux.
The change does not completely resolve #7863, but it does improve error handling.

It borrows the idea from the original issue, specifically the idea (to use `ctypes.FormatError`) demonstrated in this snippet:

```python
import ctypes
import os


def error_code_to_hex(error_code: int) -> str:
    """Convert a signed integer error code to a hexadecimal string."""
    v = error_code & 0xffffffff
    return hex(v)


def get_error_message(error_code: int):
    h = error_code_to_hex(error_code)
    print(f'\nError: {h}')
    print(f"\t{ctypes.FormatError(error_code)}")
    print(f"\t{ctypes.WinError(error_code)}")
    print(f'\t{os.strerror(error_code)}')


get_error_message(-1073741819)
get_error_message(-1073740940)
get_error_message(1)
```

```
Error: 0xc0000005
        <no description>
        [WinError -1073741819] <no description>
        Unknown error

Error: 0xc0000374
        <no description>
        [WinError -1073740940] <no description>
        Unknown error

Error: 0x1
        Incorrect function.
        [WinError 1] Incorrect function.
        Operation not permitted
```

### DIFF
--- a/src/tribler/gui/core_manager.py
+++ b/src/tribler/gui/core_manager.py
@@ -277,23 +277,36 @@ class CoreManager(QObject):
     @staticmethod
     def error_code_to_hex(error_code: int) -> str:
         """Convert a signed integer error code to a hexadecimal string."""
-        v = error_code & (2 ** 32 - 1)
+        v = error_code & 0xffffffff
         return hex(v)
+
+    @staticmethod
+    def format_str_error(error_code: int):
+        """ Return the error message corresponding to the given error code."""
+        try:
+            return os.strerror(error_code)
+        except ValueError:
+            # On platforms where strerror() returns NULL when given an unknown error number, ValueError is raised.
+            return '<no description>'
+
+    @staticmethod
+    def format_win_error(error_code: int):
+        """ Return the Windows error message corresponding to the given error code."""
+        import ctypes  # pylint: disable=import-outside-toplevel
+        return ctypes.FormatError(error_code)
 
     @staticmethod
     def format_error_message(exit_code: int, exit_status: int) -> str:
         hex_error_code = CoreManager.error_code_to_hex(exit_code)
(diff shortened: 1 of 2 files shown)


src/tribler/gui/core_manager.py
```diff
@@ -277,23 +277,36 @@ def get_last_core_output(self, quoted=True):
...
+            return os.strerror(error_code)
+        except ValueError:
+            # On platforms where strerror() returns NULL when given an unknown error number, ValueError is raised.
+            return '<no description>'
+
+    @staticmethod
+    def format_win_error(error_code: int):
+        """ Return the Windows error message corresponding to the given error code."""
+        import ctypes  # pylint: disable=import-outside-toplevel
+        return ctypes.FormatError(error_code)
```
kozlovsky, line 296: Unfortunately, from what I see, `ctypes.FormatError` does not return correct error descriptions for the actual exit codes we have. Based on your PR, I implemented a different idea: I found the complete list of Windows errors, parsed it and added a new utility function to retrieve the correct error name. Please look at #7885; it may be an approach that we should choose.

PR #7878, opened 2024-02-02
Optimize requests for downloads

### LINKED ISSUE #7360: 10x speedup of all downloads in Tribler when switching to Search tab and performing a search

drew2a:
This PR fixes #7360

The list of optimizations includes:
1. Backend: Removal of the `get_files` parameter (as it is no longer used).
2. Backend: Optimization of querying for a specific infohash.
3. Frontend: Splitting the requests for downloads into two categories: one for selected downloads and another for non-selected downloads. Extended information like pieces and files is requested only for selected downloads.
4. Frontend: The logic for receiving data was conceptually rewritten to allow for partial updates of downloads. Now, it is possible to update just a single download, whereas previously, updating all downloads was mandatory. This opens up possibilities for further improvements.

### DIFF
--- a/src/tribler/core/components/libtorrent/tests/test_torrent_def.py
+++ b/src/tribler/core/components/libtorrent/tests/test_torrent_def.py
@@ -1,5 +1,5 @@
 import shutil
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 import pytest
 from aiohttp import ClientResponseError
@@ -7,6 +7,7 @@ from aiohttp import ClientResponseError
 from tribler.core.components.libtorrent.torrentdef import TorrentDef, TorrentDefNoMetainfo
 from tribler.core.tests.tools.common import TESTS_DATA_DIR, TORRENT_UBUNTU_FILE
 from tribler.core.utilities.path_util import Path
+from tribler.core.utilities.unicode import hexlify
 from tribler.core.utilities.utilities import bdecode_compat
 
 TRACKER = 'http://www.tribler.org/announce'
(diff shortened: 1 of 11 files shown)


src/tribler/core/components/libtorrent/tests/test_torrent_def.py
```diff
@@ -234,6 +236,28 @@ def test_torrent_no_metainfo():
...
+    assert not tdef.get_infohash_hex()
+
+
+def test_get_infohash_hex(tdef: TorrentDef):
+    # Test that get_infohash_hex returns the infohash in hex format
+    tdef.infohash = b'0' * 20
+    assert tdef.get_infohash_hex() == hexlify(tdef.infohash)
+
+
+def test_get_infohash_hex_cashed(tdef: TorrentDef):
```
xoriole, line 250: Typo: `cashed` --> `cached`

xoriole (changes requested):

src/tribler/core/components/libtorrent/tests/test_torrent_def.py
drew2a: Thank you, fixed!

PR #7879, opened 2024-02-02
Update tribler.sh

Dmole:
headless and custom env for Debian 12

### DIFF
--- a/src/tribler.sh
+++ b/src/tribler.sh
@@ -1,6 +1,11 @@
 #!/bin/sh
 # Run Tribler from source tree
 
+# Options
+HEADLESS=false
+# Use a custom env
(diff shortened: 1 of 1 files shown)


src/tribler.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 # Run Tribler from source tree
 
+# Options
+HEADLESS=false
+# Use a custom env
+#PYTHONPATH=/home/user/Downloads/tribler_env/lib/python3.11/site-packages
```
drew2a, line 7: These lines appear to be specific to a particular user and it doesn't seem appropriate to commit them to the main repository.
```suggestion

```

drew2a (changes requested):
Thank you for the contribution. The PR looks good to me, except for one small detail that could be easily fixed.

drew2a (approved):

PR #7896, opened 2024-02-09
Fix remote search

### LINKED ISSUE #7893: Remote search does not yield any results
Running the current [main](https://github.com/Tribler/tribler/tree/471b34833c220fac0b22f2bff503d2b328400047) from source with an empty state directory does not show any search results.

In the beginning, there are no peers so it is expected that there are no results.
However, even after the peers are discovered, no results from the remote peers are available. 

The recent refactoring of PopularityCommunity and RemoteQueryCommunity to ContentDiscoveryCommunity kept the same community ID, so the peers should be discoverable and the results should be available. But, it is not the case which seems like a bug.

To reproduce, the local search needs to be disabled manually from the code (one way is to return empty [] results from `/search/local` present in database_endpoints.py).

The issue seems platform-independent but tested on Ubuntu.

xoriole:
There are two issues addressed in this PR:
1. The request parameters included a deprecated parameter `exclude_deleted`. This parameter is popped out when doing the actual search on the metadata store now.
2. The `metadata_type` in the request parameter is received as string `'300'` which on conversion to json was turning it to list `[3, 0, 0]`. Since channels are removed, there is only one metadata type which is a single string, a simple integer conversion should be sufficient if the value is of string type.

Out of scope of this PR:
1. This PR does not address removing the deprecated parameter `exclude_deleted` from the remote search endpoint call from the GUI side.
2. Still the search query processing logic checks if the metadata_type is either an integer or list of integers. Since there is only one metadata type available, this logic should be changed. This change is out of scope for this PR.

Fixes #7893

### DIFF
--- a/src/tribler/core/components/content_discovery/community/tests/test_content_discovery_community.py
+++ b/src/tribler/core/components/content_discovery/community/tests/test_content_discovery_community.py
@@ -293,22 +293,26 @@ class TestContentDiscoveryCommunity(TestBase[ContentDiscoveryCommunity]):
             self.torrent_metadata(1)(title="ubuntu torrent", infohash=random_infohash())
             self.torrent_metadata(1)(title="debian torrent", infohash=random_infohash())
 
-        notifier = Mock()
-        self.overlay(0).composition.notifier = {notifications.remote_query_results: notifier}
-        self.overlay(0).send_search_request(**{"txt_filter": "ubuntu*"})
+        async def assert_response_received_on_search(search_query: dict):
+            notifier = Mock()
+            self.overlay(0).composition.notifier = {notifications.remote_query_results: notifier}
+            self.overlay(0).send_search_request(**search_query)
 
-        await self.deliver_messages()
+            await self.deliver_messages()
 
-        notifier.assert_called()
-        notifier.reset_mock()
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/content_discovery/community/tests/test_content_discovery_community.py
```diff
@@ -293,13 +293,26 @@ async def test_popularity_search(self):
...
-        await self.deliver_messages()
+            await self.deliver_messages()
+
+            notifier.assert_called()
+
+        # Basic search query
+        query1 = {"txt_filter": "ubuntu*"}
+        await assert_response_received_on_search(query1)
+
+        # Query with unexpected parameters like 'exclude_deleted'
```
drew2a, line 309: NIT: It tests a query with "expected" unexpected parameters. Perhaps it would be better to slightly revise the comment.

drew2a (approved):

src/tribler/core/components/content_discovery/community/tests/test_content_discovery_community.py
xoriole: Substituted `unexpected parameters` with `deprecated parameters`

Issue #7684, opened 2023-11-09
Anonymous download performance enhancements


egbertbouman:
I'm currently doing some performance enhancements to our core, in particular to the IPv8 socket/tunnel community. By moving some code to rust (https://github.com/Tribler/py-ipv8/issues/1119), we're getting much better download speeds:

![rust_25mb](https://github.com/Tribler/tribler/assets/3785124/ba5151b1-7688-4700-adf9-9247669cad4f)

synctext:
25.5 MByte/sec download speed :boom: 
Real results, after starting abandoned experiments in 2019: https://github.com/Tribler/tribler/issues/4567
:rocket:

ichorid:
I wonder how much of increase that will be, if you move _both_ sides of the tunnel (e.g. tunnel helpers too) to Rust... :star_struck:

egbertbouman:
I don't really expect to see speeds beyond what I'm currently seeing. If I look at the speed with just a single circuit, it won't go beyond 3MB/s. So, while testing, I worked around the issue by creating 20 circuits.
At 25MB/s the most CPU intensive thread will use a full core at the downloader side. Of course, it all depends on which machine you use.

synctext:
As discussed in Dev Meeting...this month, Jan 2024 all Rust matters on both IPv4 and IPv6 should come together. Stable performant code is the ambition.

egbertbouman:
IPv8 has now been refactored to support a separate Rust endpoint (https://github.com/Tribler/py-ipv8/pull/1250). There is also a [release](https://pypi.org/project/ipv8-rust-tunnels/) on PyPI.

qstokkink:
For the occasion, I gave the Rust implementation a run on my bare metal machine (AMD Ryzen Threadripper 3970X 32-Core, 64 GB 2666 MHz RAM, from an M2 drive with a gigabit Internet connection).

As a benchmark, using the same torrent, I gathered three results that are more-or-less repeatable:

 - A 0 hop download using the Libtorrent 2.0.7 "simple client" C implementation: the maximum we can achieve with Tribler.
 - A 0 hop download using "Tribler Experimental" and Libtorrent 1.2.19 with everything switched off except the download manager, IPv8, and the tunnel community.
 - A 1 hop download using "Tribler Experimental" and Libtorrent 1.2.19 with everything switched off except the download manager, IPv8, and the tunnel community.
 
 Here are the screenshots:
 
 
[...]

Issue #4567, opened 2019-06-14
Faster Tunnel crypto by re-implementing IPv8 crypto in "Rust" programming language


synctext:
This is a long-term issue for re-implementing our crypto tunnel core for raw speed, zero-copy protocol stack, and usage of work stealing thread pool. Everything in Rust. This means leaving our ideally-suitable-for-rapid-prototyping Python stack, making it less easy to modify and freezing the wire format and behavior.

This issue partly addresses #1 issue ***Tribler anonymous downloads are fast and secure***.

This is an honor students project at TUDelft, complimentary to ongoing tunnel re-factoring and tweaking like: #4459.

Current code repo: https://github.com/ip-v8/rust-ipv8

jonay2000:
@synctext could i be assigned too? thanks!

jonay2000:
also, could the main repo link be edited to this? https://github.com/ip-v8/rust-ipv8

jonay2000:
the other assignees were correct though :)

devos50:
> also, could the main repo link be edited to this? https://github.com/ip-v8/rust-ipv8

I edited the link in the first post.

devos50:
Also, not entirely related to this ticket, but could you guys let me know when the message serialization method is working? I have an experiment where I test the market community under a high system load and it seems that one of the bottlenecks is IPv8 message serialization. If there is something faster, I can integrate it into this experiment and 1) check the speed improvement and 2) check the implementation for correctness. 👍

jonay2000:
Yes of course. We use the [Serde](https://github.com/serde-rs/serde) library for it which provides zero copy serialization, extendable to custom objects. In many cases we could basically use the default serialization strategy as python does this too with the struct library. We provide some "atoms" which correspond to python types which we guarantee to be serialized in a certain way. An example of this is the Varlen16 struct which guarantees to serialize to 2 bytes of length + data of max 65536 bytes. There is also a Varlen8, Varlen32 and Varlen64 (which arent strictly necessary as pyipv8 doesnt use them, but adding them was really easy and might prove useful). Similarly we have a Bits struct which serializes to an u8. etc. For these we implemented a custom serde serializer process. As long as you use these types in your structs instead of the builtin rust types you are guaranteed that python is able to interpret them. Serde obviously does all the hard work.

[...]

Issue #4459, opened 2019-04-12
Tunnel improvements


egbertbouman:
There are still some improvements that I would like to make to the TunnelCommunity in the coming weeks:
* Let peers within the TunnelCommunity announce how they would like to participate in the network. For instance, a node may want to:
  * Relay data
  * Act as an IPv8 and/or uTP exit
  * Be an introduction/rendezvous point for hidden services
* Set up an experiment for donating to the network and obtaining tokens for end-to-end encrypted relaying:
  * Observe operational performance on these machines
  * Tweak and boost the performance
* Since CPU still seems to be the bottleneck, I'd like to make the code a bit more efficient.
Some examples of what can be improved:
  * Message (de)serialization
  * Before Twisted writes to a socket, it first spends a lot of time in `isIPAddress` /  `isIPv6Address`
  * Remove `isinstance` calls from byte counting functions
  * Packet demultiplexing (at Community level)
  * Crypto (also investigate https://github.com/Tribler/tribler/issues/3099)

Function duration breakdown of on_cell:
![image](https://user-images.githubusercontent.com/3785124/56023541-5cd68a00-5d0e-11e9-8eec-86fca34cdb86.png)

devos50:
Should this be assigned to the Tribler 7.3 milestone?

egbertbouman:
@devos50 Since 7.3 is already in beta, I think this is something for 7.4.

synctext:
We can hopefully have the doubling of tunnel speed for final Beta of 7.3... Plus trust animation. Then into code freeze and stability only.

synctext:
ToDo: discuss raw speed, faster download starts, robustness versus security properties such as equal packet size. @egbertbouman

egbertbouman:
Just a few things that are on my tunnel wishlist:
- [ ] Create a live circuit speed test (bytes/s up and down, including hidden services). I'd like to add some speed test specific messages that allow us to do a speed test without involving a real-world BitTorrent swarm. Using the `peer_flags` we can extend the tunnel protocol without breaking compatibility.
- [ ] Build more endpoints/tools that allow us to gain insight into the inner workings of the tunnel community, especially hidden services. For instance, the status of the PexCommunities, the sizes of hidden swarms, and tunnel packet loss.
[...]

PR #7908, opened 2024-02-19
Enable rust endpoint for anonymous downloads

egbertbouman:
This PR enables the rust endpoint for anonymous downloads.

Closes https://github.com/Tribler/tribler/issues/7684.  Closes https://github.com/Tribler/tribler/issues/4567. Closes https://github.com/Tribler/tribler/issues/4459.

Note that this PR also changes the tunnel community ID, as IPv8 breaks the tunnel protocol by adding (partial) IPv6 support.


xoriole (approved):
Nice!
ipv8-rust

PR #7932, opened 2024-03-18
Process manager should not attempt to unlink the processes database file in case of OperationalError

### LINKED ISSUE #7931: [7.13.2] PermissionError when unlinking processes.sqlite "The process cannot access the file because it is being used by another process"
```
  File "tribler\core\utilities\process_manager\manager.py", line 92, in connect
sqlite3.OperationalError: database is locked
  During handling of the above exception, another exception occurred:
  File "run_tribler.py", line 109, in <module>
  File "tribler\core\start_core.py", line 191, in run_core
  File "tribler\core\utilities\process_manager\manager.py", line 40, in setup_process_manager
  File "tribler\core\utilities\process_manager\utils.py", line 29, in new_method
  File "tribler\core\utilities\process_manager\manager.py", line 59, in setup_current_process
  File "contextlib.py", line 113, in __enter__
  File "tribler\core\utilities\process_manager\manager.py", line 108, in connect
  File "pathlib.py", line 1321, in unlink
```

Here:
```python
    @contextmanager
    def connect(self) -> ContextManager[sqlite3.Connection]:
        ...
        connection = None
        try:
            ...
        except Exception as e:
            logger.exception(f'{e.__class__.__name__}: {e}')

            if connection:
                connection.close()

            if isinstance(e, sqlite3.DatabaseError):
                self.db_filepath.unlink(missing_ok=True)  # <-- here

            if isinstance(e, sqlite3.OperationalError) and str(e) == 'unable to open database file':
                ...

            raise
```

The reason for the error is that `OperationalError` is a subclass of `DatabaseError`. Therefore, the condition `if isinstance(e, sqlite3.DatabaseError)` mistakenly handles `OperationalError` as well just as a generic `DatabaseError`.

The possible solution is to swap handling of `OperationalError` and `DatabaseError` to handle all cases of `OperationalError` first.
Also it may be good to wrap the `unlink` call in try/except to suppress the possible `PermissionError`.

kozlovsky:
This PR fixes #7931 by handling all `OperationalError` exceptions before `DatabaseError` exceptions, and by suppressing possible `PermissionError` exceptions when trying to remove the database filw after the `DatabaseError`

### DIFF
--- a/src/tribler/core/utilities/process_manager/tests/test_manager.py
+++ b/src/tribler/core/utilities/process_manager/tests/test_manager.py
@@ -282,3 +282,33 @@ def test_unable_to_open_db_file_reason_added(process_manager):
                        match=r'^unable to open database file: parent directory.*does not exist$'):
         with process_manager.connect():
             pass
+
+
+@patch('sqlite3.connect', MagicMock(side_effect=sqlite3.OperationalError('error text')))
+def test_connect_operational_error(process_manager):
+    # All OperationalError exceptions except "unable to open database file" should be re-raised
+    with pytest.raises(sqlite3.OperationalError, match=r'^error text$'):
+        with process_manager.connect():
+            pass  # pragma: no cover
+
+
+@patch('sqlite3.connect', MagicMock(side_effect=sqlite3.DatabaseError('error text')))
+@patch('pathlib.Path.unlink', MagicMock(side_effect=PermissionError))
+def test_connect_database_error_suppress_permission_error(process_manager):
+    # All OperationalError exceptions except "unable to open database file" should be re-raised
+    with pytest.raises(sqlite3.DatabaseError, match=r'^error text$'):
+        with process_manager.connect():
+            pass  # pragma: no cover
+
+
+@patch('sqlite3.connect', MagicMock(side_effect=sqlite3.DatabaseError('error text')))
+def test_connect_database_error_raise_non_permission_error(process_manager):
+
+    class TestError(Exception):
+        pass
+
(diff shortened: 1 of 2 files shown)


src/tribler/core/utilities/process_manager/tests/test_manager.py
```diff
@@ -282,3 +282,33 @@ def test_unable_to_open_db_file_reason_added(process_manager):
...
+    pass
+
+
+@patch('sqlite3.connect', MagicMock(side_effect=sqlite3.DatabaseError('error text')))
+@patch('pathlib.Path.unlink', MagicMock(side_effect=TestError))
+def test_connect_database_error_raise_non_permission_error(process_manager):
+    # All OperationalError exceptions except "unable to open database file" should be re-raised
+    with pytest.raises(TestError):
+        with process_manager.connect():
+            pass  # pragma: no cover
```
drew2a, line 314: NIT: The `TestError` class could be defined inside the `test_connect_database_error_raise_non_permission_error` function since it's only used there. This will increase the readability of the entire test file.

```suggestion

@patch('sqlite3.connect', MagicMock(side_effect=sqlite3.DatabaseError('error text')))
@patch('pathlib.Path.unlink', MagicMock(side_effect=TestError))
def test_connect_database_error_raise_non_permission_error(process_manager):
    class TestError(Exception):
        pass

    # All OperationalError exceptions except "unable to open database file" should be re-raised
    with pytest.raises(TestError):
        with process_manager.connect():
            pass  # pragma: no cover
```

drew2a (approved):

PR #7941, opened 2024-03-26
Download Manager: remove unused code.

drew2a:
This PR removes unused code from `Download Manager`.

Related: 
- https://github.com/Tribler/tribler/issues/7886


egbertbouman (approved):
It would be nice if you could remove the duplicate base64 import. Other than that, this seems good to me.

PR #7958, opened 2024-04-08
Update dependencies and use the latest PyInstaller version

kozlovsky:
This PR attempts to fix MacOS segfaulting by upgrading Qt-related dependencies and updating PyInstaller to the latest version.

After the installation on MacOS, it is necessary to open the terminal and perform the following command:
```
xattr -c /Applications/Tribler.app
```
Or, alternatively:
```
xattr -d com.apple.quarantine /Applications/Tribler.app
```

Without it, MacOS falsely reports that the application is corrupted, but it just means that it is not signed.

### DIFF
--- a/build/mac/makedist_macos.sh
+++ b/build/mac/makedist_macos.sh
@@ -21,8 +21,8 @@ export RESOURCES=build/mac/resources
 
 python3 -m venv build-env
 . ./build-env/bin/activate
-python3 -m pip install pip==23.0.1 # pin pip version to avoid "--no-use-pep517" issues with the latest version
-python3 -m pip install PyInstaller==4.2 --no-use-pep517
+python3 -m pip install --upgrade pip
+python3 -m pip install PyInstaller==6.5.0
 python3 -m pip install --upgrade -r requirements-build.txt
 
 # ----- Build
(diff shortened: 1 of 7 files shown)


xoriole:
As an important consideration, these dependencies should be tested to work on Ubuntu (18.04) with glibc 2.27 as well. That usually places some constraints on PyQt5 and Pillow.

kozlovsky:
> This fact can be a show-stopper for our users

The only way to properly solve this is to obtain a Developer ID certificate to sign binaries. Until we have this certificate, asking users to execute a command in the terminal looks acceptable to me.

It is not strategically viable to use obsolete versions of PyInstaller and dependencies forever to get a slightly less ugly warning when Tribler first starts.

drew2a:
We currently don't have any ugly warnings on macOS.

We utilize the well-known "right-click" -> "open" trick, which remains the same for all applications from an unidentified developer.

https://support.apple.com/guide/mac-help/open-a-mac-app-from-an-unidentified-developer-mh40616/mac

What you're suggesting introduces something "unknown" to Mac users. It's quite suspicious (speaking as a Mac user myself) and not convenient.

I'm not the Product Owner, so I can't stop you from making such a risky change, but @synctext should at least be informed about it.

qstokkink:
Thank you for your hard work. Due to recent changes, this PR no longer relevant and I will close this PR.

PR #7986, opened 2024-04-18
Remove the torrent grouping feature based on names

drew2a:
This PR disables the torrent grouping feature based on torrent names as was requested by @synctext.
The feature has been introduced in 
 - #7025
 
image


xoriole (approved):

kozlovsky (approved):
Also, now unused `group_by_name` field should be removed from `SearchResultsModel` and `RemoteTableModel`.

Issue #8007, opened 2024-04-27
After upgrade to 7.14.0 http_port keeps changing and api does not work


heldersepu:
**Describe the bug**
I expected my settings to be maintained but some how the `http_port` changed

I had:
```
[api]
http_enabled = True
http_port = 52194
```

after upgrade
```
[api]
    http_enabled = True
    http_port = 32883
```

- OS: [e.g. Ubuntu 22.04]
- Tribler's version [e.g. 7.14.0]

____

I just tested running from source:

![image](https://github.com/Tribler/tribler/assets/545474/84d9fa12-96e8-4f95-9857-d7f3ae4ab12f)

here is the output of the `triblerd.conf` after opening an closing a couple of times

```
heldersepu@bat:~$ cat ~/.Tribler/7.14/triblerd.conf
[api]
    http_enabled = True
    http_port = 42585
    key = *****

heldersepu@bat:~$ cat ~/.Tribler/7.14/triblerd.conf
[api]
    http_enabled = True
    http_port = 46817
    key = *****

```

heldersepu:
Actually it looks like the port is now changing every time that tribler closes

heldersepu:
![Screenshot from 2024-04-27 19-11-10](https://github.com/Tribler/tribler/assets/545474/4f4bedbc-2e34-4f0d-8562-7bd54c98e568)

synctext:
Wow, thank you for finding this weird bug!

Btw could you possibly test if the experimental version on our front-page has for you the same weird bug?

heldersepu:
@synctext sorry I do not see any link to the experimental version

heldersepu:
I just tested running from source:

![image](https://github.com/Tribler/tribler/assets/545474/84d9fa12-96e8-4f95-9857-d7f3ae4ab12f)

here is the output of the `triblerd.conf`

```
heldersepu@bat:~$ cat ~/.Tribler/7.14/triblerd.conf
[api]
    http_enabled = True
    http_port = 42585
    key = *****

heldersepu@bat:~$ cat ~/.Tribler/7.14/triblerd.conf
[api]
    http_enabled = True
    http_port = 46817
    key = *****

```

synctext:
Very helpful!
Btw Website Tribler.org has bleeding edge downloads on front-page.

heldersepu:
some more logs that might help:

```
[...]

PR #8008, opened 2024-04-28
Add on_paste_event to handle pasting of magnet links

heldersepu:
A simple change to add magnet to the downloads with paste (`QKeySequence("Ctrl+v")`)
this speeds up the many click steps, ideally more options will be added in the future..


drew2a:
Hi, @heldersepu!

Thank you for your improvement. I see a lot of errors in the checks for this PR. However, they are not related to the changes in this PR and are caused by something else.
I'll figure out the reasons for them, fix them, and then return to this PR.

drew2a:
@heldersepu, could you please rebase your branch on `main`? The issues seem to be fixed.

heldersepu:
@drew2a done

drew2a:
Thank you, @heldersepu!

Two comments:
1. You merged the `main` branch into yours instead of rebasing your branch onto `main`. It's fine for now, but perhaps when the PR is ready, it would be nice to rebase it before merging (though it's not necessary).
[When do you use git rebase instead of git merge?](https://stackoverflow.com/questions/804115/when-do-you-use-git-rebase-instead-of-git-merge)
2. The test `test_clipboard` causes the GUI tests to fail. You should probably mock `QApplication` or some parts of it.

drew2a (changes requested):
The test `test_clipboard` is causing the GUI tests to fail. It should be fixed.

There are two ways to address this, one proper but time-consuming, and the other quick but less desirable, which is to simply remove the test.

If you choose the proper method, here are a couple of options:
1. Keep the test as is but try to mock `QApplication` or some parts of it.
2. Move the test to `test_gui.py` and adapt it to the style used in that file.

To verify correctness, you can run the entire test suite for the GUI folder. Currently, it results in an error.

heldersepu:
... but Merging is so much easier... github gives us a button they call [Sync fork]
on my branch is just one click an done, why will anyone still use rebase 

Looking into the failing tests, I noticed the `copy_to_clipboard` did not have any tests now I see why, I'm just going to remove that test, making sure copy/paste works is counter productive

drew2a:
> why will anyone still use rebase

It gives you a cleaner commit history. If you spend a lot of time doing bug fixing and figuring out when and how an error was introduced, you start to appreciate the clarity of a well-maintained commit history :)

drew2a (approved):

drew2a:
@heldersepu, thank you for your effort and cooperation.

heldersepu:
It really is my pleasure @drew2a ...
I've been using Tribbler for a while and always found a few quirks that could be improved, but never put the time to read the code; I just ran into a bug with the http_port used for the api that I could not leave alone: https://github.com/Tribler/tribler/issues/8007 that one unfortunately I could not fix there seems to be much more discussion that we need to settle before a permanent code fix is applied

PR #8020, opened 2024-05-06
Add setting to remember download columns width

heldersepu:
Are you tired of having to resize the columns in the download page all the time? me too...

This PR adds a couple of setting to set column width, for now i prefer to just have it as a "hidden" config option and in future versions we can add it in the UI, and more fancy stuff

____

This is how it looks like with my favorite options, I like the name to be the widest and the progress a bit wider than the default

Screenshot from 2024-05-06 13-17-44

____

Here is the new setting:
 
Screenshot from 2024-05-07 19-55-30

____

My goal is to make small changes as I get myself familiar with the code base, then I would be able to take on some more challenging issues

### DIFF
--- a/src/tribler/gui/qt_resources/mainwindow.ui
+++ b/src/tribler/gui/qt_resources/mainwindow.ui
@@ -1380,6 +1380,30 @@ color: white;margin-top:10px;</string>
 																				</property>
 																			</widget>
 																		</item>
+																		<item row="17" column="0">
+																			<widget class="QLabel" name="label_7">
+																				<property name="font">
+																					<font>
+																						<weight>75</weight>
+																						<bold>true</bold>
+																					</font>
+																				</property>
+																				<property name="styleSheet">
+																					<string notr="true">font-weight: bold;
+color: white;margin-top:10px;</string>
+																				</property>
+																				<property name="text">
+																					<string>Downloads table</string>
+																				</property>
+																			</widget>
+																		</item>
+																		<item row="18" column="0">
+																			<widget class="QCheckBox" name="downloads_header_state_checkbox">
+																				<property name="text">
+																					<string>Remember header state (keeps columns width and order)</string>
+																				</property>
+																			</widget>
+																		</item>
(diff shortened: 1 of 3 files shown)


egbertbouman:
Thanks for your PR!
Saving the column settings seems like a nice feature to have. However, I don't think these settinge belongs in the core. We normally store GUI settings in `self.window().gui_settings`. Could you try storing it there?

heldersepu:
@egbertbouman correction applied... thanks for the tip, code looks simpler that way
I also rebased and squashed my commits

egbertbouman:
Thanks, looks good!  I'm fine with merging this PR as-is, but I just noticed that the header also has `saveState`/`restoreState` functionality.

If you're up for it, you could save/restore the entire state of the downloads list in `TriblerWindow.onClose`/`TriblerWindow.restore_window_geometry`. We're already doing this for the position and size of the window. For instance, saving would be:
```python
self.gui_settings.setValue("download_columns", self.downloads_list.header().saveState())
```

heldersepu:
Let me experiment with that  `self.downloads_list.header().saveState()` will let you know more tomorrow

heldersepu:
@egbertbouman I made the following changes:
- added a new section to the settings page, screenshot provided above
- using the header `saveState/restoreState` functionality as suggested, (nice bonus it also saves the sort)
- new function `on_header_change` to save the state if use has it set

only disagreement with your suggestion I'm not saving onclose or restore_window_geometry instead doing it on:
- `sectionResized`
- `sortIndicatorChanged`
just in case if something crash it saved the state already

egbertbouman (approved):

egbertbouman:
Nice, it's running smoothly and there's even a checkbox for it now 👍
Thanks again!

heldersepu:
it was fun! ... thanks @egbertbouman  for pushing me in the right direction 
... if you can, take a look at an issue I opened a few days ago: https://github.com/Tribler/tribler/issues/8007 would love more eyes on that

PR #8022, opened 2024-05-08
Refactor RESTManager and add config.write when port changes

### LINKED ISSUE #8007: After upgrade to 7.14.0 http_port keeps changing and api does not work

heldersepu:
With this change the `http_port` is updated as soon as it changes
fixes #8007

### DIFF
--- a/src/tribler/core/components/restapi/rest/rest_manager.py
+++ b/src/tribler/core/components/restapi/rest/rest_manager.py
@@ -18,7 +18,7 @@ from tribler.core.components.restapi.rest.rest_endpoint import (
     RESTResponse,
 )
 from tribler.core.components.restapi.rest.root_endpoint import RootEndpoint
-from tribler.core.components.restapi.rest.settings import APISettings
+from tribler.core.config.tribler_config import TriblerConfig
 from tribler.core.utilities.network_utils import default_network_utils
 from tribler.core.utilities.process_manager import get_global_process_manager
 from tribler.core.version import version_id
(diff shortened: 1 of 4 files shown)


xoriole (changes requested):
Thank you for your PR @heldersepu.

Instead of passing the config object to RESTManager and rewriting the `config` as `config.api`, I think the issue can be solved by simply calling `config.write()`  when the RESTManager has started (in `restapi_component.py`).

heldersepu:
@xoriole We debated that in the issue:
https://github.com/Tribler/tribler/issues/8007#issuecomment-2100938488

in a nutshell:
- That config was not a real config(TriblerConfig) but of type APISettings, bad name
- No need for two params: `config=config.api, state_dir=config.state_dir` that can be just one
- This way we only invoke the write only when necessary.

qstokkink:
Thank you for your hard work. Recent changes to the Tribler code base have implemented your suggested feature. Therefore, this PR no longer relevant and I will close it.

PR #8025, opened 2024-05-11
Add a way to hide columns in the download table

heldersepu:
This PR adds a simple way to hide columns in the download table

Kazam_screenshot_00000

right clicking in the header `header().customContextMenuRequested,` will show a small menu 
in that menu we have options:
- allow the users to hide the clicked column `header().hideSection(index)`
- restore all hidden columns

### DIFF
--- a/src/tribler/gui/widgets/downloadspage.py
+++ b/src/tribler/gui/widgets/downloadspage.py
@@ -90,6 +90,8 @@ class DownloadsPage(AddBreadcrumbOnShowMixin, QWidget):
         connect(self.window().remove_download_button.clicked, self.on_remove_download_clicked)
         connect(self.window().downloads_list.header().sectionResized, self.on_header_change)
         connect(self.window().downloads_list.header().sortIndicatorChanged, self.on_header_change)
+        self.window().downloads_list.header().setContextMenuPolicy(Qt.CustomContextMenu)
+        connect(self.window().downloads_list.header().customContextMenuRequested, self.on_header_right_click_item)
         connect(self.window().downloads_list.itemSelectionChanged, self.on_selection_change)
         connect(self.window().downloads_list.customContextMenuRequested, self.on_right_click_item)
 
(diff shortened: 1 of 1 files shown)


kozlovsky (changes requested):
Thank you for the PR. While hiding a column can be useful for some users, the way it is implemented has some usability issues:

* **Unexpected Behavior:** Double-clicking a column header typically relates to sorting or adjusting the width of the column in many applications. Hiding a column by double-clicking could be counterintuitive for users accustomed to standard UI patterns.

* **Accidental Clicks:** Users might accidentally double-click a header, expecting it to sort or adjust, and instead find the column has disappeared. Imagine the Name column disappears after a user accidentally double-clicks it. This could lead to confusion and frustration. It may be hard for a user to understand what has just happened, and it will look like a bug.

* **Lack of Reversibility:** The absence of an immediate way to bring back the hidden column without restarting the application is a significant usability issue. Users should have the ability to easily undo the action, ideally through the same interface mechanism used to hide it (e.g., right-click the context menu to show/hide columns).

A more intuitive approach might be to implement a right-click context menu on the column headers. This would not only prevent accidental hiding but also enhance functionality by giving users more control.

heldersepu:
@kozlovsky right-click context menu added

src/tribler/gui/widgets/downloadspage.py
```diff
@@ -549,6 +551,28 @@ def on_export_download_request_done(self, filename, data):
                 tr("Torrent file exported"), tr("Torrent file exported to %s") % str(dest_path)
             )
 
+    def show_all_headers(self):
```
kozlovsky, line 554: Currently, when I click on "Restore all", I get the following error:
```python
  File "C:\dev\tribler\src\run_tribler.py", line 126, in <module>
    main()
  File "C:\dev\tribler\src\run_tribler.py", line 122, in main
    run_gui(api_port, api_key, root_state_dir, parsed_args)
  File "C:\dev\tribler\src\tribler\gui\start_gui.py", line 89, in run_gui
    exit_code = app.exec_()
  File "C:\dev\tribler\src\tribler\gui\utilities.py", line 463, in trackback_wrapper
    callback(*args, **kwargs)
  File "C:\dev\tribler\src\tribler\gui\widgets\downloadspage.py", line 571, in on_header_right_click_item
    connect(restore_action.triggered, self.show_all_headers)


The above exception was the direct cause of the following exception:

  File "C:\dev\tribler\src\tribler\gui\utilities.py", line 466, in trackback_wrapper
    raise exc from CreationTraceback(traceback_str)
  File "C:\dev\tribler\src\tribler\gui\utilities.py", line 463, in trackback_wrapper
    callback(*args, **kwargs)
TypeError: show_all_headers() takes 1 positional argument but 2 were given
```

To fix it, you need to specify the second unused argument for the signal handler:
```suggestion
    def show_all_headers(self, _=None):
```

kozlovsky (changes requested):
It looks good, but one fix should be added.

After you apply the fix, can you please squash all commits in the PR branch to make the commit history easier to read?

heldersepu:
> After you apply the fix, can you please squash all commits in the PR branch to make the commit history easier to read?

Fix applied, and all commits have been squashed

kozlovsky (approved):
Great, thank you for the contribution!

PR #8033, opened 2024-05-22
Refactor proxy settings and add error handling

### LINKED ISSUE #8014: [7.14.0] ValueError: invalid literal for int() with base 10 (in DownloadManager.set_proxy_settings)
```
  File "tribler/core/components/component.py", line 42, in start, 
  File "tribler/core/components/libtorrent/libtorrent_component.py", line 31, in run, 
  File "tribler/core/components/libtorrent/download_manager/download_manager.py", line 150, in initialize, 
  File "tribler/core/components/libtorrent/download_manager/download_manager.py", line 340, in get_session, 
  File "tribler/core/components/libtorrent/download_manager/download_manager.py", line 296, in create_session, 
  File "tribler/core/components/libtorrent/download_manager/download_manager.py", line 354, in set_proxy_settings, 
ValueError: invalid literal for int() with base 10: '//rutorrent.frixa.appboxes.co'
```

[Here](https://github.com/Tribler/tribler/blob/8d03b993a0c4ec4bff88b533987df2b482d7c4a9/src/tribler/core/components/libtorrent/download_manager/download_manager.py#L354):
```python
    def set_proxy_settings(self, ltsession, ptype, server=None, auth=None):
        """
        Apply the proxy settings to a libtorrent session. This mechanism changed significantly in libtorrent 1.1.0.
        """
        settings = {}
        settings["proxy_type"] = ptype
        settings["proxy_hostnames"] = True
        settings["proxy_peer_connections"] = True
        if server and server[0] and server[1]:
            settings["proxy_hostname"] = server[0]
            settings["proxy_port"] = int(server[1])  # <-- here
```

It looks like the user specified an incorrect value for the `LibtorrentSettings.proxy_server` config option.

The error does not look serious, but instead of a generic `ValueError`, it is better to raise a specific error that describes the problem to the user so he can fix the configuration settings.

drew2a:
Reworked the 'set_proxy_settings' function in the download manager to handle invalid port values more gracefully. The function now attempts to convert the port value to an integer and logs any ValueError exceptions that occur during this process. This prevents crashes when non-integer values are provided for the port. Also, reorganized import statements according to PEP8 guidelines.

Additionally, a new test case 'test_set_proxy_settings_invalid_port' has been added to verify this behavior. It checks if proxy settings are not set when an invalid port number is provided.

Fixes #8014

### DIFF
--- a/src/tribler/core/components/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/components/libtorrent/download_manager/download_manager.py
@@ -8,11 +8,11 @@ import logging
 import os
 import time as timemod
 from asyncio import CancelledError, gather, iscoroutine, shield, sleep, wait_for
-from binascii import unhexlify
 from copy import deepcopy
 from shutil import rmtree
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Tuple
 
+from binascii import unhexlify
 from ipv8.taskmanager import TaskManager
 
 from tribler.core import notifications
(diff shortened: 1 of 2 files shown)


src/tribler/core/components/libtorrent/download_manager/download_manager.py
```diff
@@ -362,17 +362,24 @@ def get_session(self, hops=0):
...
 
-    def set_proxy_settings(self, ltsession, ptype, server=None, auth=None):
+    def set_proxy_settings(self, ltsession, ptype, server: Optional[Tuple] = None, auth: Optional[List] = None):
         """
         Apply the proxy settings to a libtorrent session. This mechanism changed significantly in libtorrent 1.1.0.
         """
+        proxy_port = None
+
+        try:
+            proxy_port = int(server[1])
```
xoriole, line 372: Since `[server](server: Optional[Tuple] = None)`, this can raise `TypeError: 'NoneType' object is not subscriptable`

drew2a: Thank you! You are right. 
It has been fixed.

src/tribler/core/components/libtorrent/download_manager/download_manager.py
```diff
@@ -362,17 +362,25 @@ def get_session(self, hops=0):
...
         settings["proxy_hostnames"] = True
         settings["proxy_peer_connections"] = True
-        if server and server[0] and server[1]:
-            settings["proxy_hostname"] = server[0]
-            settings["proxy_port"] = int(server[1])
+        if server:
+            proxy_port = None
+
+            try:
+                proxy_port = int(server[1])
```
xoriole, line 377: Under the possibility of `server[1]` being None which causes `int(server[1])` to raise `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`, I would suggest to keep the initial if statement:
```python3
if server and server[0] and server[1]:
```

drew2a: You are right about the technical possibility of server[1] being None. This could be due to wrongly formatted settings. If we start considering these corner cases, then we should also check that the value of server has exactly two items, as it is technically possible to form an incorrect parameter. For example:

```python
    fake_dlmgr.set_proxy_settings(session, 3, (None,))
```

To address these corner cases, I have to rewrite the function completely. Additional tests have also been added to highlight and test this possibility.

xoriole (approved):

PR #8034, opened 2024-05-22
Add breadcrumb ordering

drew2a:
We have noticed that breadcrumbs are not always sorted by timestamp:

image

This PR fixes it by Introducing a new feature to order breadcrumbs by timestamp in ascending order. This is achieved by adding a new function, `order_by_utc_time`, which sorts the breadcrumbs based on their timestamps.

Ref:
- https://github.com/getsentry/sentry/issues/8133
- https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/


kozlovsky (changes requested):
Unfortunately, it does not work: https://sentry.tribler.org/organizations/tribler/issues/2658/events/e962cfb599ab49ddbf9f874d1720642d/

You can see that breadcrumbs still come in two independently sorted groups.

(As a side note, I noticed that the user comment is currently omitted from the Sentry report, even when included. It is not related to the current PR, I will investigate it)

drew2a:
You are right. It is strange. I'll investigate further.

drew2a:
@kozlovsky I think I've fixed it by https://github.com/Tribler/tribler/pull/8034/commits/957b3296a5e803f842c7b58cdd116bc13c78df1d

https://sentry.tribler.org/organizations/tribler/issues/2658

kozlovsky:
Nope, it is still two separate lists: https://sentry.tribler.org/organizations/tribler/issues/2658/events/0a59f86f4eeb4c908fd860e2da8343c0/

kozlovsky (approved):
It seems it is fixed now!

Issue #7736, opened 2023-11-30
Rewrite the GUI to use a Qt webview (HTML+Javascript)


qstokkink:
Currently, our GUI uses a mix of Qt XML (QML), custom Qt Python widgets and CSS. Modifying or extending this XML (which can be up to [6000 lines](https://github.com/Tribler/tribler/blob/main/src/tribler/gui/qt_resources/mainwindow.ui)) is rather painful. Furthermore, implementing custom Python Qt widgets takes a lot of time and tends to be error-prone, just for the end-goal of reinventing the wheel to mimick common web interfaces.

We can downsize and simplify our code base by switching from XML+custom Python to HTML and Javascript. The latter combo has very mature (and bug free!) libraries. We can make this switch relatively easily and without sacrificing features by using Qt's WebViews, which allows us to keep tray icons etc.

As an aside, more people understand HTML and Javascript so this should also make it easier for developers to contribute to our code base.

drew2a:
From my very first day of working with Tribler, I believed that we should change our direction towards HTML and Javascript. The way QT is used in Tribler has led to the fact that none of the developers want to (and sometimes can) work with it. This leads to us not changing our UI and not fixing bugs in the GUI, preferring instead to work with the core.

Our GUI does not look quite attractive. To make it attractive, we should spend a lot of effort. Having a lot of effort to make on one hand and no wish from developers to apply this effort on the other hand means the Tribler GUI in QT will never be good. In contrast, if you choose HTML+Javascript, you have so many templates that will make the UI look and feel cool and modern with almost no effort from the developer's side.

So, @qstokkink you have my full support on this topic.

egbertbouman:
We have some old (Angular) prototypes:
https://github.com/egbertbouman/tribler-webui for the main UI
https://github.com/egbertbouman/tribler-debug-ui for the debug panel

synctext:
Big investment in het past of re-writing our wxPython in favor of PyQt. Adding Javascript is something I would like to avoid, but we're living in modern times. 
Just QT +Webviewer + Javascript seems "clean" on our 3 supported OSes.

https://stackoverflow.com/questions/34184704/qwebview-not-loading-external-javascript
```
    QString html = readFile(":/qt.html");
    view->setHtml(html, QUrl("qrc:/"));
```
application.qrc
```
<!DOCTYPE RCC><RCC version="1.0">
<qresource>
    <file alias="qt.png">resource/qt.png</file>
[...]

PR #8043, opened 2024-05-28
Performance improvements for getting/selecting files using the web API

egbertbouman:
While working on the webui (https://github.com/Tribler/tribler/issues/7736), I noticed that getting and selecting files was very slow, to the point that the core was locking up. This PR fixes this problem.

I tested this PR using a torrent with 28700 files in it, and got a 11x speedup for getting files and a 50x speedup for selecting files. (on my machine it now takes 0.4s to get 28.7K files and 0.13s to change the selected files).

With these speedups, I don't have to use to the tree logic when building the webui, and I can just use standard client-side pagination. This makes things much simpler.

### DIFF
--- a/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
+++ b/src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
@@ -107,7 +107,8 @@ class DownloadsEndpoint(RESTEndpoint):
         """
         files_json = []
         files_completion = {name: progress for name, progress in download.get_state().get_files_completion()}
-        selected_files = download.config.get_selected_files()
+        # We use a dict for performance reasons
+        selected_files = {fi: True for fi in (download.config.get_selected_files() or range(len(files_completion)))}
         file_index = 0
         for fn, size in download.get_def().get_files_with_length():
             files_json.append({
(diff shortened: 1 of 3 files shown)


src/tribler/core/components/libtorrent/restapi/downloads_endpoint.py
```diff
@@ -465,7 +466,7 @@ async def update_download(self, request):
             num_files = len(download.tdef.get_files())
             if not all([0 <= index < num_files for index in selected_files_list]):
                 return RESTResponse({"error": "index out of range"}, status=HTTP_BAD_REQUEST)
-            download.set_selected_files(selected_files_list)
+            download.set_selected_files(selected_files_list, no_tree=True)
```
kozlovsky, line 469: This way, we leave the `file_instance.selected` attribute for `tdef.torrent_file_tree` items not updated; it may be OK to do this, but I think at least we should note this in a comment, so further we should not rely on values of tree item `.selected` attribute.

kozlovsky (changes requested):
The performance speedup looks promising! I think I found one bug that mixes up `0` and `1`.

egbertbouman:
The web UI is now part of Tribler Experimental, and I'll focus on working the web UI from there. 

Given that these changes are mostly useful for the web UI anyway, I'm closing this PR.

PR #8063, opened 2024-06-17
Update signing script for macos build

xoriole:
This PR updates the MacOS (app and installer dmg) build signing.
The signing part is extracted to separate scripts from the build script. Further, app and dmg signing are separated.
The build is fully running on mac_mini on Jenkins which removes the dependency on external dependent machines previously used for signing.

Additionally, there is some refactoring done on the build script, extracting the environment variables into a separate file `env.sh`.

Here is a working build from this PR:
https://jenkins.tribler.org/job/Tribler/job/Build/job/Build-macOS/174/

### DIFF
--- a/build/mac/Readme.md
+++ b/build/mac/Readme.md
@@ -1,0 +1,38 @@
+# MacOS Build and Signing Procedure
+
+## Introduction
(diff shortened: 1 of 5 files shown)


drew2a (approved):
Nice improvement!

NIT: Perhaps it's worth adding a small README.md to the root folder of `build/mac` with roughly the same description you wrote in the PR.

drew2a (approved):
👍

Issue #8086, opened 2024-08-08
Double clicking a torrent in Windows (build) doesn't download it


qstokkink:
After registering Tribler in the installer as the app for `.torrent` files, double clicking a `.torrent` does not open Tribler or download it.

Issue #8085, opened 2024-08-08
Open Tribler when starting the executable


qstokkink:
It would be nice to open Tribler when the Tribler executable is started.

PR #8089, opened 2024-08-12
Fix command line arguments and detect duplicate sessions

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8085.
Fixes https://github.com/Tribler/tribler/issues/8086.


egbertbouman:
validate

src/run_tribler.py
```diff
@@ -60,38 +74,52 @@ async def main() -> None:
...
+    if config.get("api/key") is None:
+        config.set("api/key", os.urandom(16).hex())
+        config.write()
 
+    logger.info("Creating session. API port: %d. API key: %s.", config.get("api/http_port"), config.get("api/key"))
     session = Session(config)
+
+    torrent_uri = parsed_args.get('torrent')
+    if torrent_uri and os.path.exists(torrent_uri) and torrent_uri.endswith(".torrent"):
+        torrent_uri = Path(torrent_uri).as_uri()
```
qstokkink, line 97: When looking at this, I just recalled that the Windows installer also [registers itself for the magnet extension](https://github.com/Tribler/tribler/blob/42a8ac190f8f9f8bc4733c34c8092dd1375dd746/build/win/resources/tribler.nsi#L184-L190). Do you want to add support for that in this PR or should be open a separate issue?

qstokkink (approved):
Assuming we split off an issue for https://github.com/Tribler/tribler/pull/8089#discussion_r1713742371, I'll preemptively approve.

src/run_tribler.py
qstokkink: Another alternative, "none of the above", is to remove support for `.magnet` files.

egbertbouman: `DownloadsEndpoint.add_download` calls `DownloadManager.start_download_from_uri` which seems to have the ability to deal with magnet links. So, I'm hoping this will be fixed too. I haven't tested it, though.

qstokkink: I guess we/you can add `torrent_uri.endswith(".magnet")` to this filter then and if anything breaks we'll just open an issue.

egbertbouman:
validate

src/run_tribler.py
egbertbouman: Done, .magnet support should work now. I also fixed an issue with the default hop count.

PR #8092, opened 2024-08-19
Added versioning manager

### LINKED ISSUE #8079: Support versions and upgrading again
One of the temporarily-removed Tribler features for TriblerExperimental was [the version check](https://github.com/Tribler/tribler/blob/main/src/tribler/core/components/version_check/versioncheck_manager.py).

Essentially, we should query the following two URLs for new versions:

- `https://release.tribler.org/releases/latest?current={version_id}`
- `https://api.github.com/repos/tribler/tribler/releases/latest`

Since we are using Python 3.10 for TriblerExperimental, we can use modern version determination (instead of a `version.py`), see https://docs.python.org/3.10/library/importlib.metadata.html#overview:

```python
(example) $ python
>>> from importlib.metadata import version  
>>> version('wheel')  
'0.32.3'
```

If we get a `PackageNotFoundError` here, we are running from source and we should probably not be checking for new versions.

qstokkink:
Fixes #8079

This PR:

 - Adds versions and upgrading again.

Notes:
 - The `https_certfile` still exists in the main (non-versioned) `state_dir`, other files are again placed in the version's state dir.
 - The latest "disk" version (`TO`) _may not be equal_ to the latest release version `version("tribler")`. For example, a version `8.1` may operate from the `8.0` state directory. This is not necessarily pretty but it avoids file copies and renaming.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -9,6 +9,7 @@ lint.ignore = [
     "ARG002",
     "ARG005",
     "ASYNC109",
+    "ASYNC110",
     "BLE001",
     "COM812",
     "COM819",
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -12,11 +12,12 @@ import webbrowser
 from pathlib import Path
 
 import pystray
-import tribler
 from aiohttp import ClientSession
 from PIL import Image
+
+import tribler
 from tribler.core.session import Session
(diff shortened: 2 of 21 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8093, opened 2024-08-19
Updated IPv8 pointer

qstokkink:
This should get rid of the annoying `ip-v8.org` bootstrap resolution warnings.


qstokkink:
validate

egbertbouman (approved):

Issue #5648, opened 2020-10-16
Tribler exe is not signed


xoriole:
From the builder logs
```
E:\jenkins\workspace\Build-Tribler_release\Build-Win64\tribler\dist>signtool.exe sign /f c:\build\certs\certificate.pfx /p "[*******]" /d "Tribler" /t "http://timestamp.digicert.com" Tribler_*.exe 
'signtool.exe' is not recognized as an internal or external command,
operable program or batch file.
```
The builder is missing `signtool.exe`

xoriole:
`signtool.exe` is in path now but we'll need new certificates to resolve this issue.

qstokkink:
This is ultra-high priority. Any executable that is not signed using an EV SSL certificate will be flagged as malware by anti virus software (possibly causing issues #7929 and #5979)

qstokkink:
- Ideally, we would have an EV SSL certificate. However, this is rather expensive.

- The next-best thing would be a long-lived self-generated key.

- The next-best thing after that would be a one-time key per release.

- The very worst case would be to not sign at all (current).

I'd suggest slowly making our way up this list of options and start by signing releases in the build job with a fresh key every time.

qstokkink:
Assuming we will never splurge on a full-on EV certificate, and with the Windows builds signed again, I think this issue is now as resolved as it will ever be.

PR #8097, opened 2024-08-20
Added one-time signing keys to windows build

qstokkink:
Related to #5648

This PR:

 - Fixes the Windows build not being signed at all*.
 - Updates the build action to strip the "v" out of the tag.

I'm not 100% sure if the paths are correct for GitHub Actions but there is only one way to find out 🙂 

*Ideally, we would have a beter solution (see #5648) but a cheap signature is still better than none at all.


qstokkink:
validate

qstokkink:
Hold up, typo: "Delf".

qstokkink:
validate

egbertbouman (approved):

PR #8101, opened 2024-08-20
Store GUI settings in TriblerConfig

egbertbouman:
Currently, Tribler uses the browser's `localStorage` to store the GUI settings. However, since we're using a default port of 0, the GUI settings are likely to be reset after every restart.

This PR addresses this issue by avoiding the use of `localStorage` and storing the GUI settings alongside the core settings instead.

### DIFF
--- a/src/tribler/core/restapi/settings_endpoint.py
+++ b/src/tribler/core/restapi/settings_endpoint.py
@@ -69,9 +69,16 @@ class SettingsEndpoint(RESTEndpoint):
 
         return RESTResponse({"modified": True})
 
-    def _recursive_merge_settings(self, existing: dict, updates: dict) -> None:
+    def _recursive_merge_settings(self, existing: dict, updates: dict, top: bool = True) -> None:
         for key in existing:
+            # Ignore top-level ui entry
+            if top and key == "ui":
+                continue
             value = updates.get(key, existing[key])
             if isinstance(value, dict):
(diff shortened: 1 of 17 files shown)


egbertbouman:
validate

src/tribler/core/restapi/settings_endpoint.py
```diff
@@ -71,7 +71,13 @@ async def update_settings(self, request: web.Request) -> RESTResponse:
 
     def _recursive_merge_settings(self, existing: dict, updates: dict) -> None:
         for key in existing:
+            if key == "ui":
+                continue
```
qstokkink, line 75: Could you double-check this? Looking at later code, I feel like you only need to ignore one "ui" key from the top level and not all "ui" keys in all nested dictionaries. I may be misinterpreting this though.

qstokkink (dismissed):
I'll pre-approve, in case https://github.com/Tribler/tribler/pull/8101#discussion_r1723399193 turns out to be a non-issue.

src/tribler/core/restapi/settings_endpoint.py
egbertbouman: Point taken, better to be safe then sorry. I've cooked up a fix for this.

qstokkink (approved):

egbertbouman:
validate

PR #8120, opened 2024-08-28
Added an error reporter

### LINKED ISSUE #8117: Forward errors to GitHub issues
When we have an uncaught core error, the core responds with an HTTP 500 message to the GUI. However, unless we look at the raw responses coming in (on the GUI side), we won't know of these errors.

We should let a user know that an error has occurred and allow them to forward these hidden stack traces to some error reporting facility (e.g., straight to GitHub issues).

qstokkink:
Fixes #8117

This PR:

 - Adds an error reporting feature in the GUI.


qstokkink:
validate

xoriole (approved):
Here the dependency looks like the reporter must have a Github account but that is a design choice that is decided. Therefore, I approve it.

PR #8129, opened 2024-09-04
WIP: Added a new version notification on startup

### LINKED ISSUE #8128: Add a popup when Tribler starts when there is a new version
Currently, a user has to go to the version settings to discover that a new Tribler version exists. We can provide a small popup after starting, when a new version is available.

qstokkink:
Fixes #8128

This PR:

 - Adds a version check call in the `Header.tsx`, displaying a toast when a new version is available.
 - Updates the `Toast` to be defined in the `Header.tsx` instead of `SaveButton.tsx`.
 
See #8128 for reasons why this does not use the new version event (this was my original intention).


qstokkink:
validate

egbertbouman (approved):

PR #8132, opened 2024-09-04
Open the download dialog when clicking on the name of a torrent

egbertbouman:
This PR makes torrent names on the search/popular pages clickable, whereas before you needed to double-click the table entry itself. This essentially reverts the behaviour back to that of Tribler 7.14.


egbertbouman:
validate

qstokkink (approved):

PR #8136, opened 2024-09-05
Fixed PyInstaller copying file as folder

qstokkink:
Fixes the current build failure on Mac. According to [the docs](https://pyinstaller.org/en/stable/spec-files.html#adding-data-files):

```
        The first string specifies the file or files as they are in this system now.
        The second specifies the name of the folder to contain the files at run-time.
```


qstokkink:
validate

egbertbouman (approved):

qstokkink:
https://github.com/Tribler/tribler/actions/runs/10721519240 fixed 🚀

PR #8142, opened 2024-09-06
POC: Add GUI error handling per request

### LINKED ISSUE #8141: Don't use Axios interceptor for error handling
In #8120, I introduced an error reporter that uses an Axios `interceptor`. The intent was that this would only intercept unhandled errors.

What actually happens is that this interceptor is called BEFORE any handlers. This means that every single non-2xx HTTP code is reported as an error, regardless of whether it is handled later or not.

Regrettably, the implication here is that we have to attach the error handler manually to every single request in the Tribler + IPv8 services. For example

```typescript
async getDownloadFiles(infohash: string): Promise<BTFile[]> {
    return (await this.http.get(`/downloads/${infohash}/files`)).data.files;
}
```

.. turns into something like ..

```typescript
async getDownloadFiles(infohash: string): Promise<BTFile[]> {
    try {
        return (await this.http.get(`/downloads/${infohash}/files`)).data.files;
    } catch (error) {
        if (axios.isAxiosError(error) && error.response?.status == "404") {
            return new Array();
        } else {
            handleHTTPError(error);
        }
    }
}
```

This will roughly quadruple the file sizes of the services. It would be nice if there was some way to get this shorter.

qstokkink:
Fixes #8141

This PR:

 - Adds error handlers for all REST requests.

This PR is meant as a proof of concept. I'm not necessarily sure myself that this is the best approach.

**CURRENTLY BROKEN**: simply raising a `RuntimeError` in the `TorrentInfoEndpoint` will (still) break the GUI, disallowing the error reporter from being shown.

### DIFF
--- a/src/tribler/ui/src/services/ipv8.service.ts
+++ b/src/tribler/ui/src/services/ipv8.service.ts
@@ -1,7 +1,7 @@
 import { Circuit } from "@/models/circuit.model";
 import { OverlayStats } from "@/models/overlay.model";
 import axios, { AxiosInstance } from "axios";
-import { handleHTTPError } from "./reporting";
+import { handleHTTPError, handles } from "./reporting";
 
 
 export class IPv8Service {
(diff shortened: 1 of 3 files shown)


qstokkink:
validate

qstokkink:
validate

qstokkink:
I don't think this POC can be saved, see https://github.com/Tribler/tribler/issues/8141#issuecomment-2339970534
We need a different approach.

PR #8149, opened 2024-09-12
Fixed the select event loop not being set on Windows

### LINKED ISSUE #8146: Enabling statistics causes peers to drop to 0
Run 1:

screenshot 1

Run 2:

screenshot 2

My peer counts go up after starting Tribler and then quickly go down to near zero.

qstokkink:
Fixes #8146

This PR:

 - Fixes the select event loop not being set on Windows.
 - Fixes the tunnel component not respecting its "enable" config value.

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -134,5 +134,6 @@ async def main() -> None:
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8153, opened 2024-09-13
Query peers for popular torrents

### LINKED ISSUE #8083: Query peers for popular torrents to populate the popular torrents page
To augment the popular torrents page, peers can be queried for their popular torrents.

The previous implementation for this was:

https://github.com/qstokkink/TriblerExperimental/blob/136e8c891654b17b14413bf641551123a8707193/src/tribler/gui/tribler_window.py#L791-L792

Some changes have been made to the REST API but a similar approach can be followed.

qstokkink:
Fixes #8083

This PR:

 - Adds querying of peers for popular torrents.
 - Removes xxx filtering from endpoints (this no longer does anything).

### DIFF
--- a/src/tribler/ui/src/pages/Popular/index.tsx
+++ b/src/tribler/ui/src/pages/Popular/index.tsx
@@ -1,6 +1,6 @@
 import SimpleTable from "@/components/ui/simple-table";
 import SaveAs from "@/dialogs/SaveAs";
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { triblerService } from "@/services/tribler.service";
 import { Torrent } from "@/models/torrent.model";
 import { ColumnDef } from "@tanstack/react-table";
@@ -57,11 +57,31 @@ export default function Popular() {
     const [open, setOpen] = useState<boolean>(false)
     const [torrents, setTorrents] = useState<Torrent[]>([])
     const [torrentDoubleClicked, setTorrentDoubleClicked] = useState<Torrent | undefined>();
+    const [request, setRequest] = useState<string>("");
 
     useInterval(async () => {
         const popular = await triblerService.getPopularTorrents(true);
         setTorrents(filterDuplicates(popular, 'infohash'));
+        const remoteQuery = await triblerService.searchTorrentsRemote('', true);
+        setRequest(remoteQuery.request_uuid);
     }, 5000, true);
+
+    useEffect(() => {
+        (async () => { triblerService.addEventListener("remote_query_results", OnSearchEvent) })();
+        return () => {
+            (async () => { triblerService.removeEventListener("remote_query_results", OnSearchEvent) })();
+        }
+    }, [request]);
+
+    const OnSearchEvent = (event: MessageEvent) => {
+        const data = JSON.parse(event.data);
+        if (data.uuid !== request)
+            return;
+
+        for (const result of data.results) {
+            setTorrents((prevTorrents) => [...prevTorrents, result]);
+        }
+    }
 
     const handleDownload = useCallback((torrent: Torrent) => {
         setTorrentDoubleClicked(torrent);
--- a/src/tribler/ui/src/services/tribler.service.ts
+++ b/src/tribler/ui/src/services/tribler.service.ts
@@ -147,12 +147,12 @@ export class TriblerService {
         return (await this.http.get(`/metadata/search/completions?q=${txt_filter}`)).data.completions;
     }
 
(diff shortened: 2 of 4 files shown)


src/tribler/ui/src/pages/Popular/index.tsx
```diff
@@ -57,12 +57,32 @@ export default function Popular() {
...
+    const [request, setRequest] = useState<string>("");
 
     useInterval(async () => {
         const popular = await triblerService.getPopularTorrents(true);
         setTorrents(filterDuplicates(popular, 'infohash'));
+        const remoteQuery = await triblerService.searchTorrentsRemote('', true);
+        setRequest(remoteQuery.request_uuid);
     }, 5000, true);
 
+    useEffect(() => {
```
egbertbouman, line 69: Why reset the event listener every time the request changes?

src/tribler/ui/src/services/tribler.service.ts
```diff
@@ -147,12 +147,12 @@ export class TriblerService {
...
-    async searchTorrentsLocal(txt_filter: string, hide_xxx: boolean): Promise<Torrent[]> {
-        return (await this.http.get(`/metadata/search/local?first=1&last=200&metadata_type=300&exclude_deleted=1&fts_text=${txt_filter}&hide_xxx=${+hide_xxx}`)).data.results;
+    async searchTorrentsLocal(txt_filter: string): Promise<Torrent[]> {
+        return (await this.http.get(`/metadata/search/local?first=1&last=200&metadata_type=300&exclude_deleted=1&fts_text=${txt_filter}`)).data.results;
     }
 
-    async searchTorrentsRemote(txt_filter: string, hide_xxx: boolean): Promise<{ request_uuid: string, peers: string[] }> {
-        return (await this.http.put(`/search/remote?fts_text=${txt_filter}&hide_xxx=${+hide_xxx}&metadata_type=300&exclude_deleted=1`)).data;
+    async searchTorrentsRemote(txt_filter: string, popular: boolean): Promise<{ request_uuid: string, peers: string[] }> {
+        return (await this.http.put(`/search/remote?fts_text=${txt_filter}&popular=${+popular}&metadata_type=300&exclude_deleted=1`)).data;
```
egbertbouman, line 155: Could be that I missed something, but isn't the `popular` argument filtered out by calling `sanitize_parameters` (in `SearchEndpoint.remote_search`)?

src/tribler/ui/src/pages/Popular/index.tsx
qstokkink: This was copied over from the search endpoint. I don't know of a better way.

src/tribler/ui/src/services/tribler.service.ts
qstokkink: You are absolutely correct! This was just pulling in random torrents now. Fixing..

src/tribler/ui/src/pages/Popular/index.tsx
egbertbouman: Ah, I missed that. It should be OK to pass an empty list as a parameter (like we do on the downloads page). But since we're doing elsewhere too, I'm fine with it as-is. Maybe there's a reason for it.

qstokkink:
validate

src/tribler/ui/src/services/tribler.service.ts
qstokkink: Solved 👍

src/tribler/ui/src/pages/Popular/index.tsx
qstokkink: I can try it out and see what happens. I'll report back.

src/tribler/ui/src/services/tribler.service.ts
qstokkink: 🤔 hold on, now that this is "fixed" there are no more torrents coming in.

src/tribler/ui/src/pages/Popular/index.tsx
qstokkink: For some reason, replacing `[request]` with `[]` causes `data.uuid !== request` to always evaluate to `false`. So, you never see any results (don't ask me why).

src/tribler/ui/src/services/tribler.service.ts
qstokkink: I assume that remote peers also sanitize away the `popular` parameter. So, I got a bit creative and I replaced the `popular` with `sort_by=HEALTH`, which should do the same thing.

qstokkink:
validate

egbertbouman (approved):

Issue #6801, opened 2022-03-06
Ability to seed more than 5 torrents at a time


absolutep:
**Is your feature request related to a problem? Please describe.**

Seeding is limited to only 5 torrents at a time.

Which is detrimental to torrents because there are no settings to disregard - slow speed seeding torrents(ex - 2kBps torrent) which are inadvertently included in set of 5 maximum torrents being seeded.

This poses an issue where a torrent which can be seeded at 50kBps cannot enter the seeding pool because it is being crowded by, slow torrents(ex - 2kBps speed ones) and thus disregards the potential seeding order.
Very bad for seeding torrents.

**Describe the solution you'd like**

**There should be a settings option which will allow simultaneous seeding of more than 5 torrents** or there should be an option to disregard slow speed seeding torrents by a stipulated speed limit ( **as shown in the screenshot**)
![Untitled](https://user-images.githubusercontent.com/97702149/156915848-be417273-7296-4a24-8722-68ae604841a0.png)

devos50:
Thank you for your report/suggestion. By default, Tribler (and `libtorrent`, the torrent library used by Tribler) should not limit the number of torrents that are seeding simultaneously. If you go to 'settings', the setting 'unlimited seeding' should be turned on (in the 'seeding' tab). Could you verify if this is the case for your Tribler instance?

absolutep:
> If you go to 'settings', the setting 'unlimited seeding' should be turned on (in the 'seeding' tab).

unlimited seeding is turned ON

absolutep:
Here is the screenshot of the issue
![Untitled](https://user-images.githubusercontent.com/97702149/156925123-c128d2bb-fbe1-4ea6-900c-cf9d4466c019.png)


It does not matter if I use anonymous mode or non-anonymous mode. It never works for more than 5 torrents.

qstokkink:
There are settings in `libtorrent` that limit the number of seeded torrents. We/someone should double-check that the `libtorrent` configuration's `active_seeds` and `active_limit` are set to something sensible in the "normal Tribler" configuration, like they are set for the `seedbox/seeder.py` script:

https://github.com/Tribler/tribler/blob/52bbf04454fc893a372343079ee4b03fba4a641d/src/seedbox/seeder.py#L179-L182

On a side note, exposing `high_performance_seed()` to users through the Tribler config screen may also be interesting.

absolutep:
**This is right now**

[...]

PR #8154, opened 2024-09-13
READY: Set active downloads/seeds to unlimited

qstokkink:
Related to #6801

This PR:

 - Updates `active_downloads` and `active_seeds` to be unlimited by default (capped to 500 in total by libtorrent internally).

This seems like a better default than a max of 3 downloads and 5 seeds.

screenshot

A follow-up PR should make this configurable in the GUI.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -287,7 +287,10 @@ class DownloadManager(TaskManager):
                                             "enable_upnp": int(self.config.get("libtorrent/upnp")),
                                             "enable_dht": int(self.config.get("libtorrent/dht")),
                                             "enable_lsd": int(self.config.get("libtorrent/lsd")),
-                                            "enable_natpmp": int(self.config.get("libtorrent/natpmp"))}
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (commented):
I think these settings apply to auto-managed downloads. Are downloads auto-managed nowadays?

qstokkink:
I believe so, from [the docs](https://www.libtorrent.org/manual-ref.html):

> By default, torrents are auto-managed, which means libtorrent will pause, resume, scrape them and take them out of upload-mode automatically.

egbertbouman:
But isn't the flag cleared by:

https://github.com/Tribler/tribler/blob/79a150949e976a6d9f3f77686b8fb2ed0c19d65e/src/tribler/core/libtorrent/download_manager/download.py#L284-L288

qstokkink:
Indeed it is. 

The auto managed flag IS deprecated. So, just to be sure, I double-checked whether `libtorrent` somehow automatically enabled auto managing anyway. This does not seem to be the case. I can download more than 3 torrents at the same time:



I guess the original issue was then already resolved at some point. So, I will close that.

PR #8163, opened 2024-09-18
Fixed upgrader script causing incorrect database relationships

### LINKED ISSUE #8162: Upgrader causes database errors
Our upgrade script uses `iterdump` to get the scheme and content of an existing database.

https://github.com/Tribler/tribler/blob/b8b2563898194b601ddc9d05f92da41a009e8168/src/tribler/upgrade_script.py#L104

Then, if an existing database is found, the lines of the dump are sequentially executed to inject the old data into the new database:

https://github.com/Tribler/tribler/blob/b8b2563898194b601ddc9d05f92da41a009e8168/src/tribler/upgrade_script.py#L110-L114

This does exactly that. However, the content that is included in this dump also **includes ids**, for example:

```
INSERT INTO "Peer" VALUES(1,X'...','2024-01-05 13:09:23.894782');
INSERT INTO "Peer" VALUES(2,X'...','2024-01-05 13:11:05.888214');
INSERT INTO "Peer" VALUES(3,X'...','2024-01-05 13:11:05.901082');
INSERT INTO "Peer" VALUES(4,X'...','2024-01-05 13:11:10.887803');
```

This is where the approach becomes less fine: if an id already exists, this imported data gets dropped. However, it gets worse: if another entry that points to this id _does get imported_ in another table, it may point to a completely different object with the same id. I believe this causes errors like #8111.

---

In short, I think we need to change our approach and reconstruct objects from the existing data. Instead of copying the data, we would create actual new objects (in the orm) from the old data. The downside is that _this is much slower_ than our blind `iterdump` approach.

qstokkink:
Fixes #8162

This PR:

 - Updates the upgrader script to explicitly import old data instead of simply dumping the old data into the new database.

Notes:
 
 - It's definitely longer but at least it works.

### DIFF
--- a/src/tribler/upgrade_script.py
+++ b/src/tribler/upgrade_script.py
@@ -17,12 +17,15 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from configobj import ConfigObj
+from pony.orm import db_session
 
 if TYPE_CHECKING:
     from tribler.tribler_config import TriblerConfigManager
 
 FROM: str = "7.14"
 TO: str = "8.0"
+
+# ruff: noqa: N802,RUF015,W291
 
 
 def _copy_if_not_exist(src: str, dst: str) -> None:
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8174, opened 2024-09-24
Fixed entitlements line endings

qstokkink:
It took me a few hours to figure out why a file with the exact same content didn't work. As it turns out, the line endings were different.


qstokkink:
validate

egbertbouman (approved):

PR #8193, opened 2024-10-04
Show file trees in SaveAs / DownloadDetails components

egbertbouman:
This PR brings back file trees, similar to the way file selection worked before introducing the web UI.


egbertbouman:
validate

qstokkink (approved):

qstokkink (approved):
For some reason, my cancel button froze but that was a one-time thing it seems.

Issue #8195, opened 2024-10-06
Create an error mode session


qstokkink:
As shown in https://github.com/Tribler/tribler/issues/8194#issuecomment-2394007907, some errors can occur before we have a GUI. In these cases, we should still launch some minimal "backup" GUI and notify the user of the error.

In particular for https://github.com/Tribler/tribler/issues/8194#issuecomment-2394007907, a corrupted private key, we should ask the user what they want to do with the corrupt material. There is no obvious solution to this: casually removing a private key is a very bad idea. However, we can give "human" advice instead of the raw traceback.

For other cases, unknown failures, the error should be reported "as usual".

PR #8196, opened 2024-10-07
Display a messagebox when an error occurs during startup

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8195.

Error in case of a currupted private key:
image

In case of a import error:
image


egbertbouman:
validate

qstokkink (changes requested):
Please merge the `show_error` function into the `run_tribler.py` file, or rename the `utilities.py` file to something more descriptive if this is expected to be used in other files in the future.

*I had the displeasure of dealing with the aftermath of the previous `utilities` and it still triggers my PTSD: https://github.com/Tribler/tribler/tree/release/7.15/src/tribler/core/utilities*

egbertbouman:
I can't say I agree, but OK.  I've chucked it into `run_tribler.py`. :shrug:

qstokkink (approved):
Thanks! I have quite strong feelings about this matter.

egbertbouman:
validate

PR #8203, opened 2024-10-11
Added workaround for deprecation of macos-12 runner

### LINKED ISSUE #8200: Libtorrent is not available for GitHub Actions CI
We currently specify the following version for libtorrent:

https://github.com/Tribler/tribler/blob/feca8bc6a9691cab36d7ab9a7a51df17e3df9a24/requirements.txt#L6

This release does not run in MacOS > 12. However, starting the third of december `macos-12` will cease to exist on GitHub Actions:

https://github.com/actions/runner-images/issues/ 10721

In order to keep testing on GitHub, we need to either create our own `libtorrent` release that is is compatible with Actions (`macos-13`, `macos-14`, or `macos-15`) or test on a virtual machine running MacOS 12. Because a virtual machine is exceptionally slow, our unit tests would become unbearably slow. The most workable solution seems to be to create our `whl` for `libtorrent` and install that in our GitHub Actions Mac CI PR workflow.

qstokkink:
Fixes #8200

This PR:

 - Updates the unit tests to use `macos-14` instead of `macos-12`.
 - Updates `create_torrent_file` to use the `v1_only` flag.
 - Removes `macos-12` from the build job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-12, ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest]  # macos-12 is no longer on GitHub Actions, we use the Mac mini for this
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -37,7 +37,7 @@ jobs:
           cd src
           python run_unit_tests.py -a
   macos:
-    runs-on: macos-12
+    runs-on: macos-14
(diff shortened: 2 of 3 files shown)


qstokkink:
It seems like the newest libttorent version demands a minimum piece size of 16384 for v2-compatible torrents. From the libtorrent source, I gather that passing the `v1_only` flag may fix this.

https://github.com/Tribler/tribler/blob/04095a69300311018d0a8fb7f730b2090ee3ffb3/src/tribler/core/libtorrent/torrents.py#L152

That said, this flag is so new that it doesn't exist on our typing yet and we'll have to manually insert the integer value.

qstokkink:
validate

egbertbouman (approved):

PR #8221, opened 2024-10-16
Fix download sorting

egbertbouman:
No table supports sorting right now, except for the downloads table.
However, sorting isn't actually working. This PR fixes this.


egbertbouman:
validate

qstokkink (approved):

PR #8226, opened 2024-10-18
Fixed startup version check not checking for error

qstokkink:
This PR:

 - Fixes the startup version check not checking for error responses.

When browsing to the Tribler URL with an invalid key, you would see this (doubled because I run `npm run dev`):

screenshot


qstokkink:
validate

egbertbouman (approved):

Issue #8209, opened 2024-10-13
[8.0.1] Downloads list disappears


herrJones:
After navigating to other sections, the list with downloads gets cleared.
It still functions in the background though. Only a webinterface problem it seems.

This situation is only resolved by stopping and restarting Tribler.

X64 - Win11 - 8.0.1

(on the other hand: v8.0.x immediately resolved the connectivity errors I had since v7.12 or so)

egbertbouman:
Thanks for the report! To figure out whether it's indeed a problem on the webui side, could you please try refreshing the page? You can also try opening /api/downloads in your browser and see if that return an empty list (e.g. `http://127.0.0.1:8085/ui/#/downloads/all` would become `http://127.0.0.1:8085/api/downloads`).

herrJones:
Hello,
That does not seem to work


When I restart and try, I get:


All downloads remain in 'searching for circuits' forever, so maybe the error in the json causes something?
    Op maandag 14 oktober 2024 om 10:20:33 CEST schreef Egbert Bouman ***@***.***>:  
 
 


Thanks for the report! To figure out whether it's indeed a problem on the webui side, could you please try refreshing the page? You can also try opening /api/downloads in your browser and see if that return an empty list (e.g. http://127.0.0.1:8085/ui/#/downloads/all would become http://127.0.0.1:8085/api/downloads).

egbertbouman:
It sounds like the error is outside of the webui. Just to be sure, there is nothing in `/api/downloads`?

herrJones:
Hello,
Sorry for the confusion.
When the downloads-list in the GUI is empty, the api call returns nothing.When the downloads-list is not empty, I do get output, but an invalid character seems to be detected somewhere.
A larger (and redacted) screenshot below



    Op maandag 14 oktober 2024 om 13:02:22 CEST schreef Egbert Bouman ***@***.***>:  
 
 


It sounds like the error is outside of the webui. Just to be sure, there is nothing in /api/downloads?

egbertbouman:
I'm not seeing any screenshots. I guess that's because you're responding by email.

nitmeare:
Same here under Mint/Xubuntu.... after some random interactions sometimes the download list just disappears. Hard restart (kill tribler) is needed to make it visible again.

egbertbouman:
[...]

PR #8228, opened 2024-10-20
Fix empty download list when ratio is infinity

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8209.

Also fixes:
* An issue with erroneously not seeding at all when `seeding_mode = 'ratio'`.
When upload was 0 and download > 0, the ratio became ∞, and the seeding ratio was immediately reached. This stopped the download before seeding could even begin.
* Seeding settings in the web UI.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_state.py
+++ b/src/tribler/core/libtorrent/download_manager/download_state.py
@@ -6,7 +6,6 @@ Author(s): Arno Bakker
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

qstokkink (approved):
Very nice! This bug scared the pants off of me.

egbertbouman:
validate

qstokkink (approved):

Issue #5591, opened 2020-09-29
Export .torrent file on all selected torrents


balupton:
Currently Tribler only allows "Export .torrent file" on a single selected item.

It would be preferable if you could select multiple items, and have it enabled, such that you can export the torrent file for each selected item at once.

devos50:
Thanks for your suggestion!

balupton:
Is there any workaround in the meantime?

drew2a:
Hi, @balupton,

If I understand your request correctly, you want to export multiple torrents at once. As a workaround, you can use the watch folder functionality:
- #1038
- #2065

balupton:
@drew2a does the watch folder batch export all existing torrents to torrent files inside the watch folder? If so, that'll work. However, isn't it just for importing torrent files into Tribler, and not for getting torrent files out of Tribler?

drew2a:
@balupton, I think I misunderstood your initial request. Do you want to export multiple torrents from Tribler to `.torrent` files? If so, there is no workaround. 

However, if you want to import multiple torrent files from `.torrent` to Tribler, you can use the watch folder.

PR #8230, opened 2024-10-22
Add ability to export multiple .torrents at once

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/5591.

After merging, users will be able to export multiple .torrent files as a zip file. Compression is done in Javascript.

### DIFF
--- a/src/tribler/ui/package-lock.json
+++ b/src/tribler/ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tribler-webui",
-  "version": "0.1",
+  "version": "0.1.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


src/tribler/ui/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "tribler-webui",
-  "version": "0.1",
+  "version": "0.1.0",
```
qstokkink, line 3: Question: are all these edits intended?

egbertbouman: Yep, if I do `npm i jslint` the version becomes `0.1.0` for some reason.

qstokkink (approved):

PR #8231, opened 2024-10-22
Fixed the tray icon not appearing on Mac

### LINKED ISSUE #8229: Tray icon does not appear on Mac
The `pystray` tray icon does not appear on Mac. This happens both when running from source and when running from the build. I uncovered that the following line is to blame:

https://github.com/Tribler/tribler/blob/35cc11e744232cad218f5700af40c357bdabc36e/src/run_tribler.py#L197

Only on Mac, `icon.run()` cannot be run on anything other than the main thread.

BUT if you do run `icon.run()` on the main thread, it blocks the main thread until the application exits. The `run_detached` function exists but this doesn't work without extra effort. In particular, we need to run an event loop or, rather, something that consumes and runs events on the main Python thread.

I made a proof-of-concept of an asyncio-compatible consuming "event loop" (the meat is in `runEventLoop` and `main` is essentially our `run_tribler.py` with two changed lines):
 
```python
import asyncio
import sys
from asyncio import run, sleep

import pystray
from PIL import Image

from AppKit import NSApp, NSEventMaskAny
from Foundation import NSDate, NSDefaultRunLoopMode

import tribler

async def runEventLoop():
    while True:
        event = NSApp().nextEventMatchingMask_untilDate_inMode_dequeue_(NSEventMaskAny, NSDate.now(), NSDefaultRunLoopMode, True)
        NSApp().sendEvent_(event)
        await sleep(0)


async def main():
    image_path = tribler.get_webui_root() / "public" / "tribler.png"
    image = Image.open(image_path.absolute().resolve())
    menu = (pystray.MenuItem('Open', lambda: print("open")),
            pystray.MenuItem('Quit', lambda: sys.exit(0)))
    icon: pystray._darwin.Icon = pystray.Icon("Tribler", icon=image, title="Tribler", menu=menu)

    icon.run_detached(None)  # <---
    asyncio.ensure_future(runEventLoop())  # <---

    for i in range(10):
        print(10-i, "...")
        await asyncio.sleep(1)
    icon.stop()


if __name__ == "__main__":
    run(main())
```

We may want to beautify this a bit.

qstokkink:
Fixes #8229

This PR:

 - Fixes the tray icon not appearing on `Mac` because it was not spawned in the main thread.

Regarding the note in `mac_event_loop`: the asyncio main thread blocks while the tray icon context menu is open. I **assume** nobody will keep their menu open, but **should they do so** their Tribler core won't do anything anymore. Even so, it would be nice to fix that in the future. For now, I think this is better than nothing.

The choice for `asyncio.sleep(0.01)` instead of `asyncio.sleep(0)` was because the latter chokes up the core. With `0.01`, the core seems unaffected.

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -2,6 +2,10 @@ from __future__ import annotations
 
 import sys
 import traceback
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pystray import Icon
 
 
 def show_error(exc: Exception, shutdown: bool = True) -> None:
@@ -21,11 +25,11 @@ def show_error(exc: Exception, shutdown: bool = True) -> None:
             import win32api
 
             win32api.MessageBox(0, text, title)
-        elif sys.platform == 'Linux':
+        elif sys.platform == 'linux':
             import subprocess
 
             subprocess.Popen(['xmessage', '-center', text])  # noqa: S603, S607
-        elif sys.platform == 'Darwin':
+        elif sys.platform == 'darwin':
             import subprocess
 
             subprocess.Popen(['/usr/bin/osascript', '-e', text])  # noqa: S603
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

src/run_tribler.py
```diff
@@ -25,7 +29,7 @@ def show_error(exc: Exception, shutdown: bool = True) -> None:
             import subprocess
 
             subprocess.Popen(['xmessage', '-center', text])  # noqa: S603, S607
-        elif sys.platform == 'Darwin':
+        elif sys.platform == 'darwin':
```
egbertbouman, line 32: Maybe also change `Linux` to `linux` while you're at it? I guess the reason for the confusion is that the result of `platform.system()` starts with a capital letter, and `sys.platform` does not.

qstokkink:
validate

src/run_tribler.py
qstokkink: Sure, done 👌

egbertbouman (approved):

Issue #8242, opened 2024-11-02
gtk-4.0 required for Tribler 8.0 on Ubuntu LTS 20.04.06


synctext:
Install the rock-stable old Ubuntu LTS 20.04.06
Try to install Tribler V8.0.3

```Apollo Guidance Computer
sudo apt-get install ~/Downloads/tribler_8.0.3_aarch64.deb 
[sudo] password for tribleruser: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'tribler' instead of '/home/tribleruser/Downloads/tribler_8.0.3_aarch64.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 tribler : Depends: gir1.2-gtk-4.0 but it is not installable
E: Unable to correct problems, you have held broken packages.
```

qstokkink:
First, I assume you need the `x64.deb` and not `aarch64.deb`. Second, could you try the following?

```
sudo dpkg -i ~/Downloads/tribler_8.0.3_x64.deb
sudo apt install gir1.2-gtk-3.0
tribler
```

synctext:
GTK3 is not acceptable for Tribler8 :bug: 
```
sudo dpkg -i ~/Downloads/tribler_8.0.3_x64.deb 
[sudo] password for tribleruser: 
(Reading database ... 484243 files and directories currently installed.)
Preparing to unpack .../tribler_8.0.3_x64.deb ...
Unpacking tribler (8.0.3) over (7.14.0) ...
dpkg: dependency problems prevent configuration of tribler:
 tribler depends on gir1.2-gtk-4.0; however:
  Package gir1.2-gtk-4.0 is not installed.

dpkg: error processing package tribler (--install):
 dependency problems - leaving unconfigured
Processing triggers for mime-support (3.64ubuntu1) ...
Processing triggers for gnome-menus (3.36.0-1ubuntu1) ...
Processing triggers for desktop-file-utils (0.24-1ubuntu3) ...
Processing triggers for bamfdaemon (0.5.3+18.04.20180207.2-0ubuntu2) ...
Rebuilding /usr/share/applications/bamf-2.index...
Processing triggers for hicolor-icon-theme (0.17-2) ...
Errors were encountered while processing:
 tribler
[...]

PR #8252, opened 2024-11-11
Use Ubuntu 20.04 for building

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8242.

I removed `cpio=2.13+dfsg-7` since it isn't needed nor is version `2.13+dfsg-7` available on Ubuntu 20.04.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       matrix:
         include:  # macos-12 is no longer on GitHub Actions, we use the Mac mini for this
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             architecture: x64
           - os: ubuntu-latest
             architecture: aarch64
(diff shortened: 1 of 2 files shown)


egbertbouman:
validate

qstokkink (dismissed):
I verified on my Ubuntu 24.04 that this works. 

IIRC `cpio`, `alien` and `rpm` were necessary for the `cx_Freeze` build process before it used `makedist_debian.sh`, maybe the other two are also unnecessary now.

synctext:
Work on Ubuntu 20.04 :+1:

egbertbouman:
validate

egbertbouman:
Turns out the `alien` and `rpm` dependencies were needed after all (or al least one of them is..).
The only change is I made is an `apt-get update` which seemed to be needed.

qstokkink (approved):
Thanks for checking 👍

PR #8256, opened 2024-11-11
Fixed Debian metadata XML duplicate release tag

### LINKED ISSUE #8255: AppData file invalid due to duplicate release
I created a test build for Flatpak and I got the following error when using `8.0.3_x64.deb`:

```python
Processing application org.tribler.Tribler
org.tribler.Tribler:               AppData problem: tag-invalid : <release> version was duplicated
Error loading AppData file: AppData file /app/share/appdata/org.tribler.Tribler.appdata.xml was not valid
```

I extracted `org.tribler.Tribler.metainfo.xml` and found the following:

```xml
  <releases><release version="8.0.3" date="2024-11-04" /><release version="8.0.3" date="2024-11-04" /></releases>
```

This should be written here:

https://github.com/Tribler/tribler/blob/42a8ac190f8f9f8bc4733c34c8092dd1375dd746/build/debian/update_metainfo.py#L5-L12

However, I don't think the script is being called twice and I only see one `write()` being called.

UPDATE: I can reproduce this by calling `update_metainfo.py` once.
UPDATE2: The problem is that `SubElement` already appends itself and then is appended again.

qstokkink:
Fixes #8255

This PR:

 - Fixes a double `<release>` tag being written (erroring out Flatpak builds).

### DIFF
--- a/build/debian/update_metainfo.py
+++ b/build/debian/update_metainfo.py
@@ -7,6 +7,5 @@ if __name__ == "__main__":
     tree = parse(metainfo_xml)
     root = tree.getroot()
     releases_tag = root.find("releases")
-    releases_tag.append(SubElement(releases_tag, "release", {"version": getenv("GITHUB_TAG"),
-                                                             "date": strftime("%Y-%m-%d", localtime())}))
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

Issue #8251, opened 2024-11-08
NotFoundError: Failed to execute 'removeChild' on 'Node'


iyagicom:
Hi! I was using Tribler and THIS happened! :cry:
```
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at Yx (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:26251)
    at Er (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:25938)
    at Yx (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:26699)
    at Sn (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:27553)
    at Qx (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:28013)
    at Sn (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:27703)
    at Qx (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:27816)
    at Sn (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:27703)
    at Qx (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:28013)
    at Sn (http://127.0.0.1:44027/ui/assets/index.4e0ae77f.js:32:27703)
```

synctext:
Big oops. Sorry. Something really broke Tribler.

We need a bit more info to fix this. Please tell us if you are a Mac or Windows users. You used latest experimental 8.0.3 of Tribler?

If you have any memory of what caused Tribler to crash, that would be most helpful. Where you doing search, downloading, playing with anon mode, or background seeding?

qstokkink:
`4e0ae77f` is the fingerprint for https://jenkins.tribler.org/job/Tribler/job/Build/job/Build-macOS/229 (you can find it in the corresponding logs). 
This is Mac `Tribler-8.0.0.dmg`.

This might already be fixed in a newer version. That said, it would be nicer to publish a source map with every release to deobfuscate tracebacks.

qstokkink:
> This might already be fixed in a newer version.

It is probably not. I followed the code into line 32 of the traceback in the minified `index.4e0ae77f.js`.

[...]

PR #8258, opened 2024-11-12
Disable translation services

egbertbouman:
Hopefully fixes https://github.com/Tribler/tribler/issues/8251.

I've disabled translation services for the entire page. To me, this seems reasonable, since we have our own localization stuff anyway. We can always revisit this later, but I think for now stability is more important.


egbertbouman:
validate

qstokkink (approved):

Issue #8260, opened 2024-11-13
long swarm names mess up the GUI


synctext:
Latest Tribler 8 on Ubuntu 20.04
After a search request the list of search results is shown. The width of the results seems to be hard-linked to the largest search results. No truncation seems to occur. The result is that columns get pushed aside, shifting to the right. A scroll bar appears! That is a problem.

Checking out if a 8KByte or even an 100KByte swarm name is possible: https://www.libtorrent.org/reference-Torrent_Info.html#torrent_info 
Our Libtorrent does not have an easily detectable limiter, so we could try 4GByte+1Byte (32-bit pointer overflow) :scream: See what that does to IPv8 torrent discovery community :monocle_face:

qstokkink:
It looks like we can do this easily enough: https://github.com/TanStack/table/issues/ 578

egbertbouman:
This is supposed to be handled by the `line-clamp-1` CSS class that should cause overflow to be removed and replaced by `...`.
It looks like `inline-block` is preventing this from working correctly, so we only need to remove that.

qstokkink:
I tried to reproduce this by creating a torrent with a long name. However, I can't see the button anymore:

![](https://github.com/user-attachments/assets/8c2d65d9-430e-4858-9cbe-ee2b5976ff71)

PR #8261, opened 2024-11-13
Fix dealing with large swarm names

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8260.


qstokkink:
@egbertbouman Do you want to deal with https://github.com/Tribler/tribler/issues/8260#issuecomment-2473058221 in this PR, or keep that for a separate one?

egbertbouman:
I can't reproduce that issue. For me, the overflow is removed.

However, the fix should be the same, so I added it anyway.

qstokkink (approved):

egbertbouman:
validate

PR #8264, opened 2024-11-14
READY: Consolidate error responses

### LINKED ISSUE #8263: Consolidate REST error responses
We inconsistently use REST error responses.

2 instances:

```typescript
{
  error: {
    handled: boolean,
    message: string,
  }
}
```

3 instances:

```typescript
{
  error: {
    handled: boolean,
    code: string,
    message: string,
  }
}
```

52 instances:

```typescript
{
  error: string
}
```

This is bad because it causes error handling constructions like this:

https://github.com/Tribler/tribler/blob/6213050af39f9d074914e6007d8d5da8fac38c66/src/tribler/ui/src/dialogs/SaveAs.tsx#L156-L163

qstokkink:
Fixes #8263

This PR:

 - Updates all instances of error RESTResponses to conform to `{error: {handled: bool, message: string}}`.
 
 Notes:
 - In `rest_endpoint.py` I do prefix the exception class because it is lost after the `str()` conversion. In `rest_manager.py` I don't prefix the exception class because it is already part of the traceback.

### DIFF
--- a/src/tribler/core/content_discovery/restapi/search_endpoint.py
+++ b/src/tribler/core/content_discovery/restapi/search_endpoint.py
@@ -75,10 +75,16 @@ class SearchEndpoint(RESTEndpoint):
         try:
             sanitized = DatabaseEndpoint.sanitize_parameters(request.query)
         except (ValueError, KeyError) as e:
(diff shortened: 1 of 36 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8266, opened 2024-11-14
Un-version dlcheckpoints

### LINKED ISSUE #8265: Don't store download checkpoints in versioned directory
We currently store our download configurations in a versioned directory. This means that every data directory upgrade requires users to port their download checkpoints, or their downloads will not be visible. We've had multiple reports by users that have seen their downloads disappear.

One thing we could do, is make a more invasive "UPGRADE NOW" dialog when we detect that users can upgrade old downloads to a new version. However, it probably makes more sense to stop versioning the download checkpoints (pulling them up one level from `.Tribler/VERSION/dlcheckpoints` to `./Tribler/dlcheckpoints`).

qstokkink:
Fixes #8265

This PR:

 - Updates `dlcheckpoints` to no longer be stored in a versioned directory.

Notes:

 - The next pre-release will require a large banner that tells users to copy over their `dlcheckpoints` from a previous pre-release folder. Importing from the 7.14 release automatically creates the directory though.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -93,7 +93,7 @@ class DownloadManager(TaskManager):
         super().__init__()
         self.config = config
 
-        self.state_dir = Path(config.get_version_state_dir())
+        self.state_dir = Path(config.get("state_dir"))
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

Issue #1261, opened 2015-02-21
Remember my sort preferences


Lottiotta:
(Most of the issues on this are really complex and way over my head, so I completely understand if this is a low priority, but it bugs me so here it is.)

I always like my torrents sorted by progress, ascending. I would like it if Triblr could remember that. I'm on a Mac, if that's relevant.

Thanks!

PR #8267, opened 2024-11-14
Add sorting to all tables

egbertbouman:
This PR adds sorting to all tables in the UI. I've included the tables in the debug panel, but disabled translating the header name.

Also fixes https://github.com/Tribler/tribler/issues/1261. Note that I'm only storing the sorting state for `Download` / `Popular` / `Search` tables. The sorting state is stored in local storage.


egbertbouman:
validate

qstokkink (approved):

PR #8306, opened 2024-12-04
READY: Added watch folder again

### LINKED ISSUE #8280: WatchFolder settings missing in new version 8.0.5
When checking all of the settings pages in the Web UI (version 8.0.5), there is no setting to update or set the WatchFolder location.
I checked the  ~/.Tribler folder for the config file in case it was imported from the previous version (7.13), but no mention of the word "watch" or the folder name of my last watch folder.

qstokkink:
Fixes #8280

This PR:

 - Adds the watch folder back again.
 - Fixes a bug in the settings endpoint that would not allow new settings to be written.
 - Removes the unused `watch_folder_corrupt_file` notification.
 - Removes the unused family filter and hide tags settings.


qstokkink:
validate

egbertbouman (commented):
I'm not sure I understand the new logic. Why are we removing downloads now?

Also, why use `replace_task`? What's wrong with processing torrents one at a time?

qstokkink:
> Why are we removing downloads now?

🤔 In retrospect, maybe this was overly ambitious. I turned this into folder synchronization, while all we needed was a watch folder. I'll remove this.

> Also, why use `replace_task`? 

If the previous task is still pending and we re-enter the loop (i.e., for some reason this has not completed after 10 seconds), as opposed to catching a `RuntimeError` if the task already exists, I chose to cancel the previous task and retry with a new task. I opted for this because there is a (extremely small) chance that a new `.torrent` with newer metainfo was added for the same infohash, and `try .. except RuntimeError .. ` is less aesthetically pleasing to me.

qstokkink:
validate

qstokkink:
@egbertbouman fixed up.

If you don't accept my reasoning and/or feel strongly about avoiding `replace_task`, please offer an alternative implementation.

egbertbouman (approved):
The `interval_runner` of  `register_task` waits 10s seconds in between tasks, so if you simply await it (and don't create an additional task), there is no problem. Having said that, I don't have strong feelings about this. I was just curious.

qstokkink:
I don't really see how you could avoid creating additional tasks, with `start_download` and `TorrentDef.load` being async. That said, if you _do_ see a way to make this sexier/learner/meaner, that would be cool. For now, I'll merge this - but feel free to make a follow-up PR with sexification.

egbertbouman:
> I don't really see how you could avoid creating additional tasks, with start_download and TorrentDef.load being async.

You could just make `check` async, which shouldn't be a problem since it's being run by the `TaskManager`.

qstokkink:
Ah! That's the missing link. I never once thought about making `check` async.

Issue #8302, opened 2024-12-03
SELECT ALL Feature


absolutep:
At the moment, Tribler 8.0.5 lacks a "**Select All**" feature.

### Usecase = Change anonymity of all torrents, Delete all torrents, STOP/RESUME all torrents, Edit Trackers etc. in the Tribler in an instant.

Reference image of qBittorrent where you can do all the aforementioned.

![Untitled](https://github.com/user-attachments/assets/2190a4ca-6ad1-4167-9941-3ea6cb5c9d0f)

egbertbouman:
That would be useful, indeed. We'll try to address this in the next release.

PR #8307, opened 2024-12-04
Various UI enhancements/fixes

egbertbouman:
This PR contains the following enhancements/fixes to the web UI:
* Add a search button to the header.
* When clicking a menu, navigate directly to the last visited submenu. If no such submenu exists, use the first submenu.
* Hide search suggestions when the list is empty. This fixes a 1px line appearing when clicking on the search input.
* Show a spinner during shutdown making it easier to see when shutdown has completed.
* Allow selecting all downloads using ctrl+a (fixes https://github.com/Tribler/tribler/issues/8302).
* Use infohash as `RowId` in the downloads table. Since table selection is keyed by `RowId`, this fixes an issue with the selected download(s) magically changing after changing the hop count.

### DIFF
--- a/src/tribler/ui/package-lock.json
+++ b/src/tribler/ui/package-lock.json
@@ -41,6 +41,7 @@
         "react-router-dom": "^6.16.0",
         "tailwind-merge": "^1.14.0",
         "tailwindcss-animate": "^1.0.7",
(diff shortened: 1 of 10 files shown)


egbertbouman:
validate

qstokkink (approved):

PR #8316, opened 2024-12-06
Fixed mimetype being wrong in winreg

### LINKED ISSUE #8279: Tribler 8 won't load in any browser
Upgraded to Tribler 8, launches a completely blank browser window.  First, I don't want to run this in a browser, and secondly it doesn't see to be working properly.  I use other localhost/127 programs without issue, however tribler loads an empty window. 

Frustrating as the previous version isn't listed for download on the website.

qstokkink:
Fixes #8279

This PR:

 - Fixes a blank screen appearing if the Windows registry entry for `.js`, `Content Type` is set to something other than `application/javascript`.
 
 Notes:
 - For good measure, I also added `html` explicitly to avoid similar problems
 - I locally reproduced the error and confirmed this fix.


qstokkink:
validate

egbertbouman (approved):

PR #8329, opened 2024-12-10
Added flathub shield to README

### LINKED ISSUE #8318: Package the app for non-Debian based Linux
- The request is to add packaging formats for Linux distributions that aren't Debian based.
- Flatpaks and AppImages are distribution-agnostic, so they would do the trick.
- Alternatively, you could add other common formats like .rpm.

qstokkink:
Fixes #8318

This PR:

 - Adds a link to the flathub build in the README.
 - Updates the README to have a bit more space around the download link, to draw more attention.
 
 Preview:
 
 
preview

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,13 +1,28 @@
 |doi| |openhub|
 |python_3_10| |contributors| |pr_closed| |issues_closed|
 
+
+
+|
+|
+
+.. raw:: html
+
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8338, opened 2024-12-13
Don't notify metainfo if there is no metainfo

qstokkink:
This PR:

 - Fixes torrents with the name `metainfo request` being injected into the database.


qstokkink:
validate

egbertbouman (approved):

Issue #8322, opened 2024-12-08
Tribler found a bug!


juanmacartur:
Hi! I was using Tribler and THIS happened! :cry:
```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/asyncio/events.py", line 80, in _run
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/asyncio/tasks.py", line 674, in cancel
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/asyncio/tasks.py", line 674, in cancel
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/asyncio/tasks.py", line 674, in cancel
  [Previous line repeated 491 more times]
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/asyncio/tasks.py", line 670, in cancel
RecursionError: maximum recursion depth exceeded while calling a Python object

rE@http://127.0.0.1:35781/ui/assets/index.70dc79d9.js:80:336

```

qstokkink:
Thanks for reporting! This is a very obscure bug.

_[dev stuff]_ Apparently this happens when cancelling a `gather()` call, the `cancel()` points to this: https://github.com/python/cpython/blob/3.10/Lib/asyncio/tasks.py#L655-L681 That would mean that one of the `Future` s that we `gather` loops in on itself on cancel.

qstokkink:
This is a bit like finding a needle in a haystack for us. Could anyone tell me when this error occurred for them? What page were you on, or did you click something, or were you shutting down?

qstokkink:
Thanks for the input! I'll leave my Tribler running with two downloads on the download screen. Hopefully, I can get it to fail.

juanmacartur:
Apparently I had run out of space on the partition where the downloads folder is located.

egbertbouman:
Maybe we're gathering `shutdown_task_manager` within the `TaskManager` itself?

Something like:

```python
class SomeTaskManager(TaskManager):

    def __init__(self):
        super().__init__()
        self.register_task("break_stuff", self.break_stuff, delay=1)

    async def break_stuff(self):
        await asyncio.gather(self.shutdown_task_manager())
```

gtubolcev:
UPD: It turned out that I had no free space on device too, just like [juanmacartur] said.

qstokkink:
[...]

PR #8339, opened 2024-12-13
Fix infinite recursion while shutting down TrackerSession

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8322.


egbertbouman:
validate

qstokkink (approved):

Issue #8370, opened 2025-01-07
Enhancements to multi-selection


qstokkink:
### Discussed in https://github.com/Tribler/tribler/discussions/8359

<div type='discussions-op-text'>

<sup>Originally posted by **balupton** December 27, 2024</sup>
- [ ] ability to use cmd+click to additionally select the clicked item (current shift+click behaviour)
- [ ] ability to use shift+click to additionally select all between the last clicked item and the newly clicked item (missing functionality)
- [ ] do not select text when cmd/shift+clicking, the text selection makes it seem that it does not work, should be `cursor: pointer !important; user-select: none;`

in this video, I've applied the stylesheet change, and demonstrating how shift+click doesn't select in-between

https://github.com/user-attachments/assets/1cb47e45-f541-4bcd-990e-02ebd2cf8ca1

</div>

PR #8394, opened 2025-01-17
Allow use of shift/control keys while changing row selection

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8370.


egbertbouman:
validate

qstokkink (approved):

Issue #8397, opened 2025-01-18
Search results and out of order sort orders


teddyrogers:
Tribler 8.0.7, if I do a few searches in Tribler the results and sort order gets itself (ironically) out of order and messed up when clicking the columns. For example names are not alphabetised, sizes are not in order, created date, etc.

To replicate;

1. Click on "Popular"
2. Type in any random search term
3. Sort by size, click on the, "Size", column
4. Type in a new random search term
5. Sort by alphabetised, click on the, "Name", column
6. Type in a new random search term
7. Sort by size, click on the, "Size", column
8. ...the more you repeat the worse it gets.

Most of the time, after two searches, I have a references from the previous search stuck at the top of the list. Changing any of the sort orders columns does not make them move in the order list.

Issue #8378, opened 2025-01-10
Sorting issues when searching old torrents


absolutep:
**Sorting issues occur when searching old torrents. The torrent I searched is less than 4 years old. Sorting does not work on size, torrent health, creation date or name even after letting tribler run for more than 5 minutes in search. After 5 minutes total number of searched torrents does rise but even after that sorting issues do exist**

As shown in image 

![1](https://github.com/user-attachments/assets/b62e665b-1b69-4988-bbee-d1e52209ca50)

PR #8399, opened 2025-01-20
Fix for processing search events

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8378.
Fixes https://github.com/Tribler/tribler/issues/8397.


egbertbouman:
validate

qstokkink (approved):

qstokkink:
By the way, is this also related to #8397?

egbertbouman:
Sounds like the same problem. Duplicates will remain stuck, even between searches.

PR #8400, opened 2025-01-20
Removed pinned marshmallow version

qstokkink:
This PR:

 - Removes the `marshmallow` version pinning. This is no longer necessary, since `3.25.0`.


qstokkink:
validate

PR #8404, opened 2025-01-21
Ignore internal ConnectionResetErrors

### LINKED ISSUE #8355: ConnectionResetError when connected to unstable Wi-Fi
Tribler 8.0.7, I have noticed when there is a connection issue a red error appears at the top of the WebGUI 

image

```
Tribler encountered an error! What do you want to do?

  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\asyncio\events.py", line 80, in _run
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\asyncio\selector_events.py", line 115, in _read_from_self
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host

n_@http://127.0.0.1:61260/ui/assets/index.8a64a9d4.js:76:336
```

It seems to be consistent, is there a better way for Tribler to detect and handle dropped connections?

qstokkink:
Fixes #8355

This PR:

 - Updates `_except_hook` to ignore `ConnectionResetError`.

These exceptions are a result of our connections, but they escape our control and get raised internally (CPython issue 116091). So, I think this is the only way to handle the error.

### DIFF
--- a/src/tribler/core/session.py
+++ b/src/tribler/core/session.py
@@ -7,7 +7,7 @@ from asyncio import AbstractEventLoop, Event
 from contextlib import contextmanager
 from os.path import isfile
 from traceback import format_exception
-from typing import TYPE_CHECKING, Any, Generator, Type, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import aiohttp
 from ipv8.keyvault.crypto import default_eccrypto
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

Issue #8381, opened 2025-01-11
Importing a magnet and downloading a single file replicates it for subsequent magnets


teddyrogers:
Tribler 8.0.7, I think it easier to explain in steps to reproduce the bug...

1. Find a couple of magnet links that contain a few files, for example one magnet contains 20 files and another with 30 files;
2. Import the first magnet into Tribler and wait for the "Destination" window to open;
3. Once the metadata has been retrieved you will see the list of files in the torrent;
4. Under size column click the top "✓" to deselect downloading all files in the torrent;
5. Select a single file to download, in this example I will select the 18th file down in the file list;
6. Click, "Download" to accept the torrent;
7. Once the download has started import the second magnet link;
8. Instead of waiting for the "Destination" window to open simply click, "Download";
9. Wait for the torrent to go out of "Stopped" status and for Tribler to retrieve the metadata;
10. Once the metadata has been downloaded go to the file list;
11. You will now see only one file has been selected for download in the second torrent... the 18th file down!

It would seem Tribler retains the file selection from the previous imported/ added torrent and replicates this for subsequent torrent files.

qstokkink:
I tried programmatically reproducing this and now I'm confused 🙂

```patch
diff --git a/src/tribler/ui/src/dialogs/SaveAs.tsx b/src/tribler/ui/src/dialogs/SaveAs.tsx
index e3b29bd36..c3d9a1b30 100644
--- a/src/tribler/ui/src/dialogs/SaveAs.tsx
+++ b/src/tribler/ui/src/dialogs/SaveAs.tsx
@@ -114,7 +114,7 @@ export default function SaveAs(props: SaveAsProps & JSX.IntrinsicAttributes & Di
     const [params, setParams] = useState<Params>({
         destination: '',
         anon_hops: 0,
-        selected_files: [],
+        selected_files: [0, 3],
         safe_seeding: false,
     });
 
@@ -125,7 +125,8 @@ export default function SaveAs(props: SaveAsProps & JSX.IntrinsicAttributes & Di
             // Reset state
             setError(undefined);
             setExists(false);
-            setFiles([])
+            setFiles([]);
+            setParams({...params, selected_files: []});
             const newSettings = await triblerService.getSettings();
             if (newSettings === undefined) {
                 setError(`${t("ToastErrorGetSettings")} ${t("ToastErrorGenNetworkErr")}`);
[...]

PR #8403, opened 2025-01-21
WIP: Fixed exception on bad torrent URL

qstokkink:
Fixes #8360

This PR:

 - Fixes `get_torrent_info` not checking for decoding errors when processing an HTTP/HTTPS response.

PR #8413, opened 2025-01-24
Fix for using selected_files from previous downloads

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8381.

Also fixes:
* Unique key error in `SelectRemotePath`.
* ~~An issue with the `SaveAs` dialog only being shown after the metadata has been downloaded, removing any feedback that something is happening. Only happened when coming from `AddTorrent` (introduced in https://github.com/Tribler/tribler/pull/8403).~~
* Stop requesting metainfo as soon as the user starts typing. Instead, wait until "Add" is pressed.

### DIFF
--- a/src/tribler/ui/src/dialogs/SaveAs.tsx
+++ b/src/tribler/ui/src/dialogs/SaveAs.tsx
@@ -121,7 +121,7 @@ export default function SaveAs(props: SaveAsProps & JSX.IntrinsicAttributes & Di
             setMoveCompleted(false);
             setError(undefined);
             setExists(false);
-            setFiles([])
+            setFiles([]);
             const newSettings = await triblerService.getSettings();
             if (newSettings === undefined) {
                 setError(`${t("ToastErrorGetSettings")} ${t("ToastErrorGenNetworkErr")}`);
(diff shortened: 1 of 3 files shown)


egbertbouman:
validate

egbertbouman:
validate

src/tribler/ui/src/dialogs/SaveAs.tsx
```diff
@@ -155,6 +158,8 @@ export default function SaveAs(props: SaveAsProps & JSX.IntrinsicAttributes & Di
                 setError(`${t("ToastErrorGetMetainfo")} ${t("ToastErrorGenNetworkErr")}`);
             } else if (isErrorDict(response)) {
                 setError(`${t("ToastErrorGetMetainfo")} ${response.error.message}`);
+                // Disable for all errors, except metainfo errors (timeouts)
+                setDisabled(response.error.message !== "metainfo error");
```
qstokkink, line 162: This is not enough to revert #8403, it reintroduces #8403 for slow loading sites. For example, entering `https://www.wish.com/` and clicking `Download`, leads to:



I appreciate your cause of "removing any feedback that something is happening", but this is not sufficient. Please either fix your fix, undo your unfix, or add a spinner or loadbar? I'm O.K. with anything but breaking it again.

qstokkink (changes requested):

src/tribler/ui/src/dialogs/SaveAs.tsx
egbertbouman: OK, I've removed my fix.

qstokkink (approved):
Reflecting on my earlier request for changes:

 - We used to have a good magnet link experience and a bad HTTP(S) experience.
 - In #8403, that flipped around. As you pointed out, the magnet link experience is now bad.

I think the only way out of this cycle, is to make some edits to the core endpoint.

egbertbouman:
validate

egbertbouman:
> I think the only way out of this cycle, is to make some edits to the core endpoint.

Or just disable the "Download" button while fetching from HTTP(s). An easy fix.

qstokkink:
I'm all for an easy fix. The "add torrent" endpoint does **a lot** of stuff though, so I think we should be careful about covering our bases. I'm a bit wary of this code now, because my own "easy fix" ignored magnets while improving http.

For example, I know the endpoint follows HTTP 300 redirects as well. So, could we get a situation where a magnet link is allowed to be added without metainfo, but a magnet link from a HTTP 300 redirect does have to wait? 🤷‍♂️

egbertbouman:
I'll leave this one to you then.

PR #8418, opened 2025-01-27
Allow magnet links to resolve later when adding URIs

### LINKED ISSUE #8416: Adding magnet links waits for metainfo to show SaveAs
After #8403, adding magnet links requires fetching the metainfo. This is causes a long wait, where no user feedback shows. After 60 seconds, a little toast may pop up to show that fetching the metainfo failed. This is not what we want.

Essentially, we want the following after adding:
 - Files:
   1. Read the torrent file. Stop if the file does not contain a torrent.
   2. Open the `SaveAs` dialog with the file contents.
 - Magnet links:
   1. Stop if the data is not a valid magnet format.
   2. Open the `SaveAs` dialog for the magnet infohash.
 - Http(s):
   1. Resolve the redirects, if necessary.
   2. Download the resource.
    a. In case of a torrent, use the "files" logic.
    b. In case of a magnet link, use the "magnet links" logic.

Essentially, what we want is to make sure that something is "torrent-like" (either a torrent file or an infohash/magnetlink) before opening the `SaveAs` dialog. This is currently not offered by either the pre-#8403 or post-#8403 method.

I broke it, so, I'll fix it: assigned myself.

qstokkink:
Fixes #8416

This PR:

 - Adds a `skipmagnet` parameter to the `torrentinfo` endpoint, allowing magnet links to resolve later in the `SaveAs` dialog (or not at all).

This should both do all the fancy HTTP forwarding and forward magnet links quickly to the SaveAs dialog. However, the price is another parameter for `getMetainfo()`.

### DIFF
--- a/src/tribler/core/libtorrent/restapi/torrentinfo_endpoint.py
+++ b/src/tribler/core/libtorrent/restapi/torrentinfo_endpoint.py
@@ -107,11 +107,23 @@ class TorrentInfoEndpoint(RESTEndpoint):
         summary="Return metainfo from a torrent found at a provided URI.",
         parameters=[{
             "in": "query",
-            "name": "torrent",
+            "name": "uri",
             "description": "URI for which to return torrent information. This URI can either represent "
                            "a file location, a magnet link or a HTTP(S) url.",
             "type": "string",
             "required": True
+        }, {
+            "in": "query",
+            "name": "hops",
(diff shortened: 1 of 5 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8429, opened 2025-02-03
Added RSS watcher

### LINKED ISSUE #8317: Add an RSS watcher
Searching for content is annoying. I have to browse to my favorite websites and search for torrents. After I found a good candidate, I have to download the torrent file and add it to Tribler. I would like this to be streamlined:

 - Add a website's RSS feed to Tribler ONCE.
 - Whenever I use Tribler search, the content that appeared in the RSS also appears in Tribler and I never have to exit my browser tab.

qstokkink:
Fixes #8317

This PR:

 - Adds the ability to auto-pull new metadata from RSS streams.

Screenshot


qstokkink:
validate

egbertbouman (commented):
Maybe I missed it, but I don't think the RSS watchers are shut down properly. Looks good otherwise.

qstokkink:
@egbertbouman all the tasks in `RSSWatcher` are registered to the `community` (managed by IPv8) of the `RSSComponent`, so these tasks shut down when IPv8 is shut down. This is equivalent to calling `RSSWatcherManager.stop()` without clearing the data structures. It's not necessary (yet) to add an explicit shutdown procedure.

If you like, I could add a `register_shutdown_task` to also clear out the (`dict`) data structures. This would harden the code to possible future modifications that do require shutdown logic.

egbertbouman:
@qstokkink Ah, I missed that. As long as all tasks get cancelled properly, it's fine by me.

egbertbouman (approved):

PR #8438, opened 2025-02-13
Refactor streaming

egbertbouman:
This PR refactors the current broken streaming logic, and addresses various issues causing the videoplayer to stop/crash in the middle of the video.

I've also added a streaming option to the GUI, but enabled it only when in dev mode and using Chrome, since only Chrome supports MKV files (unofficially). As long as the video file is supported by Chrome, it's working fine.

The big remaining issue is that some commonly used audio codes (e.g. (E)AC3) aren't supported, so you'll end up with video without audio. This happens a lot, unfortunately. The next step would be to transcode video files in realtime (with ffmpeg) in order to improve compatibility.


egbertbouman:
validate

egbertbouman:
validate

qstokkink (approved):

Issue #6591, opened 2021-11-27
Feature request: automated publishing to Docker Hub


taeh98:
Hello, I hope this finds you well.

Please could you add an action to this repository that would automatically make it publish new releases to Docker Hub? The availability of an image on Docker Hub allows for greater portability and ease with which to use and extend Tribler (for example, to run Tribler in a container and interact with it via its API). I see you have [existing images on Docker Hub](https://hub.docker.com/u/tribler), but all of them are several years old and out of date. By making an action to publish an image to Docker Hub automatically, this problem would be solved. It would remove the overhead of manually publishing images with each release (or alternatively not doing so and the image becoming out of date). The following links give some more context as to how this could be achieved:

- https://docs.github.com/en/actions/publishing-packages/publishing-docker-images
- https://github.com/marketplace/actions/build-and-push-docker-images
- https://docs.docker.com/ci-cd/github-actions

Thank you for your time and consideration.

ichorid:
@xoriole , when was the last time we built Docker image? Why did we stop to do so? Was it too hard to maintain?

qstokkink:
Small 2024 update: PR #8108 introduces automatic publishing to GitHub Packages.

qstokkink:
The latest main version of Tribler is now available here: https://github.com/Tribler/tribler/pkgs/container/tribler

Actually pushing to Docker hub is still TODO but that should not be too much additional work.

jkaberg:
@qstokkink this doesn't seem to work properly? None of the [8 series are pushed](https://github.com/Tribler/tribler/releases) as containers. An optimal setup would build and push stable releases to the :latest tag but also versioned :tags would be cool. Quite common to have some kind of :dev tag for bleeding edge as well. 

Also it would be nice with some docs on how to use the containers, especially now that the 8 series are released with the headless setup and web ui.

qstokkink:
You are right: currently, the Docker GitHub package is "barely functional" and stuck to the bleeding edge version. Getting everything packed with proper builds and tags, is something I plan to get done for the 8.1.0 release. I also want to bring back `compose`.

qstokkink:
For 8.1.0 I'll at least get a "stable" release up and running. We handle the remainder of this issue once we gain control over our own account again then.

qstokkink:
Alright, we now have a "stable" version living in `docker pull ghcr.io/tribler/tribler:latest`, instead of the bleeding edge experimental code.

[...]

PR #8440, opened 2025-02-18
Updated docker build script to get the latest stable release

qstokkink:
Related to #6591

This PR:

 - Updates `build.Dockerfile` to provide the **latest-stable full-GUI Tribler** (instead of headless) experience for X.Org-supporting platforms, including the tray icon and webbrowser opening.
 
Docker containers aren't really supposed to be able to manipulate the host and Python `webbrowser` doesn't normally forward to external services. So, some funky stuff is happening:
1. We fool Python into thinking the `x-www-browser` is installed. This is actually bash script that sends a `org.freedesktop.portal.OpenURI.OpenURI` event to the D-Bus with the URL it was called with.
2. We inherit the X.Org files from disk (volume), X.Org environment variables, and the user id from the host. This allows the host to interpret D-Bus events as its own.
3. _(Side note)_ We also need `network_mode: host` to forward events, but we were already using this for networking anyway.
4. We circumvent AppArmor to not block any D-Bus event (I would have appreciated a less priviledged approach - but I guess this will do for now).

I'll have to test if GitHub Packages correctly bundles this when it is built "for real" instead of locally on my machine. But, assuming there are no issues, this will be all I do for #6591 until the Tribler 8.2.0. milestone rolls around.

### DIFF
--- a/build/docker/build.Dockerfile
+++ b/build/docker/build.Dockerfile
@@ -1,37 +1,27 @@
-# libtorrent-2.0.9 does not support python 3.11 yet
-FROM python:3.10-slim
+# Build & run instructions:
+# 1. docker build -f build/docker/build.Dockerfile .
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8442, opened 2025-02-18
Add multiple session detection, avoid multiple tabs

### LINKED ISSUE #8379: Adding torrents opens multiple Tribler browser tabs
Tribler 8.0.7, if Tribler is open in a browser tab and I click on a .torrent file a new browser tab is opened with another Tribler page.

Is it possible to prevent this behaviour, stop multiple tabs and Tribler pages from opening if one already exists?

qstokkink:
Fixes #8379

This PR:

 - Adds an `/info` route to the `EventsEndpoint`.
 - Fixes multiple tabs getting opened when double-clicking torrents, when an existing session is present. When not adding a torrent, a tab is still opened (otherwise, a remote session could lock you out from connecting to the core).
 - Updates `find_api_server` to query the events endpoint and return basic session/events info.

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -61,6 +61,7 @@ try:
     import argparse
     import asyncio
     import encodings.idna  # noqa: F401 (https://github.com/pyinstaller/pyinstaller/issues/1113)
+    import json
     import logging.config
     import os
     import threading
@@ -230,7 +231,7 @@ async def main() -> None:
         session = Session(config)
 
         torrent_uri = load_torrent_uri(parsed_args)
-        server_url = await session.find_api_server()
+        server_url, initial_message = await session.find_api_server()
 
         headless = parsed_args.get("server")
         if server_url:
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

egbertbouman (commented):
Couldn't you just make an (events) API request, after getting the url from `server_url` (like we do in `start_download`). Maybe this would have been less complicated.

qstokkink:
@egbertbouman do you mean to `/api/events`? Assuming so, this was [my initial approach as well](https://github.com/Tribler/tribler/issues/8379#issuecomment-2665216875), and the answer I found is: no, that is in fact more complicated.

All in all, the issue I found is that the events endpoint _streams_ in a response. This is not the same as `start_download`. It means that we need to piece together the bytes response until we find the double-newline delimiter somewhere in the continuous stream and extract the first element, then close off the connection. It's possible, but requires active draining of a connection into some buffer. To avoid all that stream buffering and token parsing, I just added a new single-shot message, so we can use the new call just like `start_download`.

egbertbouman (approved):
Fair enough. I had a go anyway:
```python
async with aiohttp.ClientSession() as session:
    response = session.get(server_url + "/api/events", headers={"X-Api-Key": config.get("api/key")})
    await response.content.readline()
    initial_message = (await response.content.readline()).decode()
    session_count = int(json.loads(initial_message[5:]).get("sessions", 0))
```

Of course this is without exception handling.

qstokkink:
🤔 interesting code. I took some time to evaluate it, but ultimately I still think the implementation of this PR has a lower chance of scary things happening.

Issue #8272, opened 2024-11-24
High amount of Disk write requests in 8.0.4


nitmeare:
I found that tribler makes very high amounts of write requests to the disk, sometimes around 400-600KB/s constantly even when all Downloads are stopped this is going on.

![image](https://github.com/user-attachments/assets/d0ae6b57-86ac-4cb8-b45c-bbf3708c9bc9)

Upon further investigation with Procmon i found that Tribler is doing thousands of WriteFile Requests per second inside.
C:\Users\clusterfuck\AppData\Roaming\.Tribler\8.0\sqlite

![image](https://github.com/user-attachments/assets/0beb7e77-df98-48c8-af9d-ca297302584c)

According to Perfmon around 24.000+ write req/s average:

![image](https://github.com/user-attachments/assets/1fafcabc-88a3-4d5a-99d3-d0ad37a03d29)

This may wear out some SSDs on the long term, cause higher CPU usage and maybe it is even slowing down the tribler process itself. 

So perhaps it would be usefull if someone has a closer look on what is going on there.

qstokkink:
Thanks for reporting! That is not intended.

Our database (ORM library) should take care of batching writes to your disk to avoid exactly this issue. Perhaps we misconfigured it? We'll check it out.

synctext:
Thnx for diving into this! Wow, 24.000+ write req/s average 😱😨😱

@qstokkink Please try to reproduce this in the Tribler lab. Hopefully this can be fixed 💪

qstokkink:
@egbertbouman mitigated the disk writes in #8274 However, my earlier claim..

> Our database (ORM library) should take care of batching writes to your disk to avoid exactly this issue.

..shockingly, did not turn out to be true. We will need to make some bigger changes in our own code base to properly solve this issue. Because this issue requires a big refactoring effort, we'll schedule this for the next stable release, `8.1.0`.

absolutep:
very very high disk usage still persists in version 8.0.7, hanging (freezing) the computer. especially, for torrents above 5 GB.

qstokkink:
Our first step toward solving this will be #8402. After that has been merged, we'll reevaluate the disk usage.

qstokkink:
After #8402, we still have quite a bit of non-torrent disk I/O. This seems to be due to the following two features:
 1. `torrent_checker`
 2. `rendezvous`

We should evaluate how we can lower the disk writes of these features.

qstokkink:
[...]

PR #8450, opened 2025-02-21
Avoid micro-writes in the Rendezvous component

qstokkink:
Related to #8272

This PR:

 - Updates the `Rendezvous` component to only write certificates every 60 seconds and on shutdown.


qstokkink:
validate

egbertbouman (approved):

PR #8454, opened 2025-02-24
WIP: Added periodic checkpointing

### LINKED ISSUE #8382: A forceful Tribler quit does not save torrent progress
Tribler 8.0.7, after adding a few torrents let the downloads run as normal. Using 50% as an example reset the machine, do not gracefully quit Tribler.

With the machine rebooted and Tribler running again note the torrents remain in the list. However, the status is back to downloading at 0%. Unless I force Tribler to complete a recheck of the torrent/s it will re-download all 50% of the torrent file again.

I assume progress data of a torrent is saved on completion or during a graceful quit of Tribler. Either that or cache was not able to save to disk in time?

qstokkink:
Fixes #8382

This PR:

 - Adds checkpointing of downloads, once every minute.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -987,7 +987,7 @@ class DownloadManager(TaskManager):
         """
         self.state_cb_count += 1
 
-        for ds in states_list:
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

PR #8470, opened 2025-03-05
Fix issue with opening magnet link from browser

### LINKED ISSUE #8460: Opening a magnet link from a browser does not work if Tribler is not already running.
I have this small issue 

Triber v8.0.7
Ubuntu 24.10
Brave browser.

When I open magnet links through Brave and make them open in Tribler, it will start the program, but it will NOT add the torrent. Only when I do it again when Tribler is already running, the torrent is added.

I think this is a Linux specific issue as I believe I did this in windows with 8.x at some point and it worked.

SharkStudiosSK:
Fixes #8460

Add support for opening magnet links from a browser when Tribler is not already running.

* Add a check for a magnet link argument in the `main` function of `src/run_tribler.py`.
* Start the download if a magnet link argument is provided in `src/run_tribler.py`.

### DIFF
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -147,7 +147,7 @@ def init_config(parsed_args: Arguments) -> TriblerConfigManager:
         config.write()
 
     if "CORE_API_KEY" in os.environ:
-        config.set("api/key", os.environ.get("CORE_API_KEY"))
+        config.set("api/key", os.environ.get("CORE_API_KEY")))
         config.write()
 
     if config.get("api/key") is None:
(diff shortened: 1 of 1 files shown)


qstokkink:
This PR:

- Introduced a syntax error at line 150.
- Fixes something that is not broken. In fact, with the proposed code, a magnet link is now added twice. This actively hurts our code base.

Because none of the code proposed by this PR contributes to Tribler, or seems to address any issue we have. I'll close this. However, we do appreciate that you want to contribute to make Tribler better!

nextgenthemes:
@qstokkink well it's broken for me, I have not tried this PR but ...

qstokkink:
@nextgenthemes I appreciate that it is broken for you. However, making random changes is not going to fix it.

nextgenthemes:
@qstokkink I never even hinted that I think this should be pulled. You just made a FALSE claim, stating as fact that nothing is broken. Now I do not care helping out anymore by investigating this more, you can close my issue.

What you call a "random change" to me seems to be an attept to fix by issue. Hardly random.

PR #8480, opened 2025-03-07
Fix for initializing column visibility

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8475.

PR #8482, opened 2025-03-07
Fix for storing the sorting state

egbertbouman:
Fix storing/loading the sorting state, which I broke in https://github.com/Tribler/tribler/pull/8480.


egbertbouman:
validate

egbertbouman:
validate

qstokkink (approved):
I didn't test this version, but it looks good.

PR #8484, opened 2025-03-07
Add message statistics using ipv8-rust-tunnels

egbertbouman:
This PR:
* Adds message statistics in Rust and updates the statistics UI. The statistics are always enabled and can be viewed from `Debug > IPv8 > Traffic`.
* Adds the ability to test circuits in Rust. The resulting download speed can be viewed live from the tunnel debug panel.

IPv8 message stats:
image

Tunnel speed tests:
image

### DIFF
--- a/src/tribler/test_unit/core/restapi/test_ipv8_endpoint.py
+++ b/src/tribler/test_unit/core/restapi/test_ipv8_endpoint.py
@@ -1,3 +1,5 @@
+from unittest.mock import patch
+
 from ipv8.test.base import TestBase
 
 from tribler.core.restapi.ipv8_endpoint import IPv8RootEndpoint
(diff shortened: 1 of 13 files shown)


egbertbouman:
validate

src/tribler/test_unit/core/restapi/test_ipv8_endpoint.py
```diff
@@ -12,8 +14,9 @@ def test_binding(self) -> None:
         """
         Test if all the IPv8 endpoint paths exist.
         """
-        endpoint = IPv8RootEndpoint()
-        endpoint.setup_routes()
+        with patch('asyncio.get_running_loop', lambda: None):
```
qstokkink, line 17: This shows an IPv8 regression. The `ipv8/REST/tunnel_endpoint.py` should be changed: there is no reason to store the event loop in the `__init__`.

PR #8495, opened 2025-03-10
Various `8.1.0` fixes

### LINKED ISSUE #8485: 8.1.0 prerelease used 59gb of memory
After upgrading to the 8.1.0 prerelease from 8.0.7 on a m1 Mac mini running the latest macOS resulted in the force quit out of memory macOS modal after about an hour of use, in which tribler was the culprit at 59gb of memory on a 16gb of ram machine. I had to force quit it from that modal as the menu bar icon was unresponsive.

I have I guess 200 torrents in which half of them are completed.

Haven't reopened it since the force quit.

Let me know what steps are needed to send the logs.

### LINKED ISSUE #8489: Undeletable Torrents
Tribler 8.1.0, I have a couple of torrents that will not delete/ remove the, "dlcheckpoints", file.

After deleting the torrent in Tribler, closing and restarting Tribler the torrents reappear in the torrent list.

The only difference I can see between these two torrents and all the other torrents is the, "saveas", points to a network location and not a local drive.

qstokkink:
Fixes #8489 (experimental fix)
Fixes #8485

This PR:

 - Updates `DownloadManager.remove_config()` to use `Path.unlink()` instead of `os.access()` + `os.remove()`. At best, this fixes #8489, and in the worst case this only modernizes the code.
 - Updates `DownloadManager.create_session()` to use the `always_pwrite` flag.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -294,7 +294,8 @@ class DownloadManager(TaskManager):
             "allow_i2p_mixed": 1,
             "announce_to_all_tiers": int(self.config.get("libtorrent/announce_to_all_tiers")),
             "announce_to_all_trackers": int(self.config.get("libtorrent/announce_to_all_trackers")),
-            "max_concurrent_http_announces": int(self.config.get("libtorrent/max_concurrent_http_announces"))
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8507, opened 2025-03-11
Don't allow tracker changes while loading

### LINKED ISSUE #8494: Tribler removes added trackers
Tribler 8.1.0, import a magnet link in to Tribler with no trackers...

Image

...and at, "MetaData 0%", manually add some trackers in the, "Trackers" tab.

Change the anonymity/ hops. After the change the added trackers are removed from the, "Trackers", list.

I think this is also related to issue https://github.com/Tribler/tribler/issues/8376

qstokkink:
Fixes #8494

This PR:

 - Updates the GUI to show a loading spinner while the download's trackers are not available yet.

This avoids users adding a tracker while the trackers cannot be changed, leading to errors.

### DIFF
--- a/src/tribler/ui/src/pages/Downloads/Trackers.tsx
+++ b/src/tribler/ui/src/pages/Downloads/Trackers.tsx
@@ -76,6 +76,12 @@ export default function Trackers({ download }: { download: Download }) {
         }
     ]
 
+    if (download.trackers.length === 0)
+        return <svg className="animate-spin h-3 w-3 text-black dark:text-white ml-4 mt-4" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24">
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

src/tribler/ui/src/pages/Downloads/Trackers.tsx
```diff
@@ -76,6 +76,12 @@ export default function Trackers({ download }: { download: Download }) {
         }
     ]
 
+    if (download.trackers.length === 0)
```
egbertbouman, line 79: Doesn't this mean that you'll never be able to add trackers to a download with no trackers? Maybe the safer option would be to look at `status_code`?

egbertbouman (approved):

PR #8526, opened 2025-03-21
Fix table scrollbar issue

egbertbouman:
Fixes an issue with multiple scrollbars becoming visible next to a table after resizing (see screenshot). By adding `[&>[data-radix-scroll-area-viewport]]:max-h-[calc(100vh-57px)]` to the `ScrollArea` parent of the `Table`, we set it's height dynamically based on the viewport size minus the header size (57px).
Since these issues occur only when changing the height of the browser, I've ignored issues with tables in the debug panel.

image

This PR also:
* Fixes`ResizableHandle` being shown when it isn't needed.
* Removes duplicate SVG code for showing spinners.

### DIFF
--- a/src/tribler/ui/src/components/icons.tsx
+++ b/src/tribler/ui/src/components/icons.tsx
@@ -1,3 +1,5 @@
+import { cn } from "@/lib/utils"
+
 type IconProps = React.HTMLAttributes<SVGElement>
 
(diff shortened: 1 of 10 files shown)


egbertbouman:
validate

qstokkink (approved):

PR #8530, opened 2025-03-31
Fixed memory leak in the logging endpoint

### LINKED ISSUE #8485: 8.1.0 prerelease used 59gb of memory

qstokkink:
Fixes #8485 (once and for all, for the third time 🙂)

This PR:

 - Fixes a memory leak in the logging endpoint due to the `MemoryHandler` flushing mechanism.

I set the capacity to 400 records. This equates to roughly one minute of logging output before records start rotating out.

### DIFF
--- a/src/tribler/core/restapi/logging_endpoint.py
+++ b/src/tribler/core/restapi/logging_endpoint.py
@@ -6,6 +6,24 @@ from aiohttp import web
 from tribler.core.restapi.rest_endpoint import RESTEndpoint, RESTResponse
 
 
+class RotatingMemoryHandler(MemoryHandler):
+    """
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

Issue #8523, opened 2025-03-19
Destination: None


teddyrogers:
Tribler 8.1.2, I restarted Tribler after logging on to find the Tribler process had committed 22GB on a 16GB machine. Quit Tribler gracefully and everything shutdown successfully. Restarted Tribler and came across most of the seeding torrents now showing, "Downloading 0%". I manually performed a hash check on these and nothing happened. I also noticed none of these torrents were downloading from seeds. Upon further investigation discovered all the torrents now had "Destination" under the "Details" tab set to "None".

![Image](https://github.com/user-attachments/assets/26d3ecbe-4ca9-4c7b-a347-02b4302fe07e)

teddyrogers:
@egbertbouman I note the "None" issue has been fixed for 8.2.0 release. I wanted to check this is also fixed when importing a number of torrents they are automatically assigned to move on completion to, "None" directory? See screenshot below...

![Image](https://github.com/user-attachments/assets/9c0159fe-f476-4e5b-8f6c-a5838c0eb26f)

egbertbouman:
@teddyrogers Thanks for letting us know! I just checked, and this doesn't happen on the main branch anymore.

teddyrogers:
@egbertbouman @egbertbouman would it be possible to release a Tribler 8.1.4 release with this included please?

It is very annoying having torrents seemingly "stall" only to find out they have value "none" as the destination folder. Having to manually sanity check every torrent after a restart of Tribler to make sure "none" is set with a torrent. Lost days of seeding time with some torrents, oblivious to the fact "none" is the torrent destination.

Can this become an issue in the swarm, if peers are requesting chunks of data and Tribler is not sending anything, can Tribler's IP be added to banned/ ignore lists? Is Tribler sending any data?

If Tribler is unable to locate the file or send data to peers I think it better to stop and show an error? Currently, completed torrents sit in Triblers torrents list seeding with seed/ peer totals as if nothing is wrong.

PR #8531, opened 2025-03-31
Fix default completed_dir

egbertbouman:
Possibly fixes https://github.com/Tribler/tribler/issues/8523.

When downloading normally, `completed_dir` is set to `""` by the UI. However, when creating a new torrent and setting it to seed automatically, `completed_dir` is set to `None` (this comes from `download_defaults`). This is then cast to str to become `"None"`. This PR fixes this issue by not setting `None` in `from_defaults`.

Note that `completed_dir = None` in the config file is interpreted as a directory called None.  The same goes for `saveas = None`.


egbertbouman:
validate

qstokkink (approved):

Issue #8517, opened 2025-03-17
SHIFT + Torrent Selection -> Out of Order Selections


teddyrogers:
Tribler 8.1.2, when I select the first torrent in the torrent list, hold SHIFT and select a torrent further down in the list the expected torrents are not all selected. If a scroll down, torrents at various locations are highlighted.

I assume there is a sort order that is being used and when using SHIFT + selection is used it is following that sort order and not the order that is presented in the torrents list?

PR #8532, opened 2025-03-31
Fix shift+up/down when sorting

egbertbouman:
Second part of fixing https://github.com/Tribler/tribler/issues/8517.

Previously, we fixed shift+click, but it turns out there's a problem with shift+up/down as well. As before, this issue only occurs when sorting is set manually.


egbertbouman:
validate

Issue #8492, opened 2025-03-09
Enable Pagination


teddyrogers:
Tribler 8.1.0, after disabling pagination there is no option to re-enable. Scrolling to the bottom of the page does not present any options. 

To re-enable pagination involves clicking to another torrent and then back to the wanted torrent with pagination.

Maybe a right click menu option could resolve this?

egbertbouman:
The pagination options are temporary. As soon as you move to another page, "disable pagination" will be disabled again.

However, if we really require these options to be changed in-place, I think we should just remove it. This would also fix https://github.com/Tribler/tribler/issues/8365.

teddyrogers:
> I think we should just remove it.

Do you mean to remove the, "disable pagination", feature?

If so, I would rather pagination be disabled altogether and have a listed view with a keyword search box.

Pagination is clunky when there are thousands of files and it gets even worse when there are folders, thrown into the mix, a few levels down. to wade through.

The issue I have come across in a listed view is the browser struggles the more files there are, refer https://github.com/Tribler/tribler/issues/8365.

teddyrogers:
Forgot to add a right click menu option in the, "Files", tab may be a simple way to enable and disable pagination.

Issue #8365, opened 2024-12-31
"Disable Pagination", causes severe slowdown/ unresponsive behaviour


teddyrogers:
Tribler 8.0.7, I have a MAME torrent with thousands of files that are contained in one folder. When I disable pagination, because it should be easier to navigate to view and/ or select individual torrents, Tribler page *really* struggles. Tribler page becomes very, very slow and unresponsive to the point where it seems to give up, including me and, where I now choose not to use Tribler for this torrent.

It does not help that the torrent archive also contains BitComet's endless number of, "____padding_file", that Tribler is unable to properly handle.

This may be an extreme case though other clients handle these perfectly okay, e.g. qBittorrent and BitComet.

PR #8533, opened 2025-04-02
Add virtual scrolling

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8365.
Fixes https://github.com/Tribler/tribler/issues/8492 by removing the need for pagination.

This PR adds virtual scrolling to all tables, including file trees. For lists of a couple of thousand items, the difference is quite dramatic. Before you could barely scroll, now there's no slowdown at all. I've tested with up to 100k items without any issues. I think/hope it's good enough to disable pagination.

A downside is that column sizes can change as you scroll through the list. That's just because the table layout is set to auto, and only a small set of rows is being rendered at a time. I didn't want to go into calculating column sizes manually, but we can always revisit this later.

Note that there is room for additional performance improvements by only requesting updates from the core for the items that are (close to being) visible. Maybe something to consider in the future.

### DIFF
--- a/src/tribler/ui/package-lock.json
+++ b/src/tribler/ui/package-lock.json
@@ -27,6 +27,7 @@
         "@radix-ui/react-tabs": "^1.0.4",
         "@radix-ui/react-tooltip": "^1.0.7",
         "@tanstack/react-table": "^8.10.7",
+        "@tanstack/react-virtual": "^3.13.5",
         "@types/video.js": "^7.3.58",
         "axios": "^1.6.8",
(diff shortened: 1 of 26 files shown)


egbertbouman:
validate

qstokkink (approved):
Nice 👍 This fully obsoletes the `TorrentFileTree` now then, I think.

egbertbouman:
Thanks! I wasn't brave enough to remove `TorrentFileTree`, as some people may prefer it.
It's also the normal way of displaying files (I think that e.g., QBittorrent uses trees too).

PR #8552, opened 2025-04-23
Removed TorrentFileTree

qstokkink:
This PR:

 - Removes the `TorrentFileTree` class.
 
 I did not remove several (now unused) path-based methods that were added alongside the `TorrentFileTree`. They might become useful in the near future, when we want to optimize the GUI.


qstokkink:
validate

egbertbouman (approved):

Issue #8325, opened 2024-12-09
Filter files... (Similar to "Filter by name")


teddyrogers:
It would be nice if Tribler had an option to filter files whilst adding a torrent, so that individual or groups of file types can be added...

![image](https://github.com/user-attachments/assets/2aff416a-07ad-451b-8a7f-fe8ac9348940)

Additionally it would be nice if a similar/ same option was added in the, "Files", list as it is possible to add torrents directly to Tribler - after associating torrent or magnet files to Tribler in the browser - bypassing the above, "Add torrent", method. It also makes it more beneficial here as Tribler has page numbering to scroll through in torrents with a large number of files. As an example, qBittorrent filter files method seems to work fine - and no page numbering!

![image](https://github.com/user-attachments/assets/a0cedde9-60bc-4388-a597-1af5fe406242)

qstokkink:
Thanks for the suggestion! I think we can look at this for the next feature release (8.1).

PR #8564, opened 2025-04-29
Add column setup

egbertbouman:
This PR fixes https://github.com/Tribler/tribler/issues/8325 by adding a column setup popover:
image

The popover appears when you right-click on the header. Right now it's just sorting and filtering (only for a small set of columns), but perhaps in the future we'll add other options as well. If anyone has better ideas I'm all ears.


qstokkink (approved):
Wow, cool 😎

PR #8565, opened 2025-04-29
WIP: Removed bitarray

qstokkink:
This PR:

 - Removes the `bitarray` dependency.

From my tests, the new method is as fast as (or 1~2 ms faster than) the old `bitarray` implementation. Not much of a difference, but this does mean that we ship one less compiled library in our installer.


qstokkink:
validate

egbertbouman (approved):

PR #8566, opened 2025-04-29
Updated handle decorators to pass valid handles

qstokkink:
This PR:

 - Updates `check_handle` and `require_handle` to pass valid handles, avoiding `cast()` everywhere.


qstokkink:
validate

PR #8586, opened 2025-05-09
Fixed checkpoint loading and completion hash checking

### LINKED ISSUE #8521: Hashchecking Randomly?
Tribler 8.1.2, with, "Check torrent after completion", enabled, after Tribler is re/started, hashchecking is performed on some already completed and previously hash checked files. I do not know what prompts "random" torrents to be chosen for another hash check.

I think a better way to prevent hash checks recurring unnecessarily would be a flag in the dlcheckpoints file, after a torrent has completed and hash checked. The only occasions where any further hash checks need to be performed is when the user requests it.

### LINKED ISSUE #8578: Hashchecking everytime Tribler is started
Tribler is rechecking/hash-checking all torrents after every restart.

This is a huge problem because some of my torrents exceed 100 GB and I have more than 80 torrents.

This happened after installing = https://github.com/Tribler/tribler/actions/runs/14734509352

windows 10

qstokkink:
Fixes #8521
Fixes #8578

This PR:

 - Adds a pretty print string to `TorrentDef`.
 - Fixes resume data not being loaded from checkpoints.
 - Fixes hashchecking on startup not properly detecting already-finished torrents.
 - Fixes the printed ATP for each download always being an empty dictionary.
 - Updates `DownloadConfig.get_engineresumedata()` to return `add_torrent_params` instead of a `dict`.
 - Removes `Download.get_atp()` (use `.tdef.atp` instead).

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -173,9 +173,12 @@ class Download(TaskManager):
         Wait for the torrent to finish downloading, then recheck.
 
         Note: a finished recheck causes a ``torrent_finished_alert``: hooking into that causes an infinite loop!
+        Note2: our own self.lt_status is too old.
+        Note3: the state flip-flops too much to be reliable so we use completed_time instead.
         """
         await self.future_added
-        if self.get_state().get_progress() == 1.0:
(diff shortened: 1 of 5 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8587, opened 2025-05-09
Update npm dependencies

egbertbouman:
This PR updates all npm dependencies to their newest version.

PR #8588, opened 2025-05-09
Fix for storing column states

egbertbouman:
This PR fixes storing the sorting state in `SimpleTable`, which broke after merging https://github.com/Tribler/tribler/pull/8587.


egbertbouman:
validate

qstokkink (approved):

PR #8591, opened 2025-05-15
Fix DHT readiness check

egbertbouman:
Currently, downloads can get stuck in the "Loading" state indefinitely when the DHT fails to get nodes. This could happen e.g. when you're behind a VPN.

There's a `timeout` that cancels the `DownloadManager.dht_ready_task`, which ensures that a single download starts. However, all remaining downloads remain stuck (awaiting a cancelled task throws a `CancelledError`, so `_async_add_torrent` never completes).

Note that this wasn't an issue until recently, when we started using `timeout`, which actually cancels the task itself. Before, the task kept running until shutdown (or enough nodes were discovered).

This PR:
* Fixes the above mentioned behaviour.
* Changes the check such that once the timeout passes, all subsequent downloads won't have to wait anymore.
* Ensures that we're checking the right DHT session. Currently, a 1-hop download waits until the 0-hop session DHT has enough nodes, which doesn't make sense.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -10,7 +10,7 @@ import dataclasses
 import logging
 import os
 import time
-from asyncio import CancelledError, Future, gather, iscoroutine, shield, sleep, timeout, wait_for
+from asyncio import CancelledError, Future, gather, iscoroutine, shield, sleep, wait_for
 from binascii import hexlify, unhexlify
 from collections import defaultdict
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


egbertbouman:
validate

qstokkink (approved):
_Trivia below_

I didn't understand how `asyncio.timeout` could possibly throw a `CancelledError`. According to the docs, and after checking the CPython source code, any `CancelledError` thrown within the context manager should be transformed into a `TimeoutError`: https://github.com/python/cpython/blob/3.13/Lib/asyncio/timeouts.py#L135-L142

As it turns out, Python 3.12 [doesn't have the same code](https://github.com/python/cpython/blob/3.12/Lib/asyncio/timeouts.py) as what the up-to-date Python 3 docs were written for (Python 3.13+) and **does allow** `CancelledError`s to escape.

I didn't notice this version inconsistency when porting the code. It's probably best to check if the same thing doesn't occur for the other `asyncio.timeout()` usages.

_End trivia_

Issue #8444, opened 2025-02-18
Download Upload speed indicator


absolutep:
I would like to request total download / total upload speed indicator at the bottom of Tribler GUI. If possible also give an indicator of total uploaded/downloaded data of the current session only.

As shown in image below

![Image](https://github.com/user-attachments/assets/4ce5895e-9e9d-494d-9583-4bd2c3a854e6)

egbertbouman:
We've been talking about some kind of network status, and upload/download stats will certainly be part of this. We'll try to address this in 8.2.

absolutep:
> If possible also give an indicator of total uploaded/downloaded data of the current session only.

### can we get this please!!?? @egbertbouman  @qstokkink 

Today, in my metered connection I exhausted my high speed data limit and saw tribler was ON.

I myself switched Tribler ON but I wanted to know how much overall data Tribler eats [current session] so that I can manage my FUP limit of my ISP.

**Please, enable this feature on war footing, this is an extremely important feature especially for people who are under metered connections.** 

Thanks.

egbertbouman:
@absolutep I'll start working on it. It will probably be merged somewhere next week, but you'll still have to wait until v8.2.

PR #8594, opened 2025-05-16
Add statusbar

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8444.

image


egbertbouman:
validate

src/tribler/core/components.py
```diff
@@ -158,8 +159,8 @@ def finalize(self, ipv8: IPv8, session: Session, community: Community) -> None:
         """
         When we are done launching, register our REST API.
         """
-        session.rest_manager.get_endpoint("/api/downloads").mds = session.mds
-        session.rest_manager.get_endpoint("/api/statistics").mds = session.mds
+        session.rest_manager.get_endpoint("/api/downloads").mds = session
```
qstokkink, line 162: This doesn't look right. Shouldn't this be `session.mds` still?

egbertbouman: You're right. I don't think `DownloadsEndpoint.mds` is actually used. anymore. That explains the lack of errors.

src/tribler/core/libtorrent/download_manager/download_manager.py
```diff
@@ -135,6 +134,14 @@ def __init__(self, config: TriblerConfigManager, notifier: Notifier,
         self.dht_readiness_timeout = config.get("libtorrent/dht_readiness_timeout")
         self._last_states_list: list[DownloadState] = []
 
+        self.register_task("Request libtorrent session stats", self._request_session_stats, interval=5)
```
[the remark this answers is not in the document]
egbertbouman: Sure thing. I didn't really think about it since there's other tasks getting registered as well (although those aren't really doing anything with an empty `ltsessions` dict).

qstokkink (approved):
Thanks 👍

PR #8601, opened 2025-05-19
Fixed safe_seeding reenabling through default

### LINKED ISSUE #8549: Disabled seeding proxies, on completion, defaults back to "Anonymity" setting
Tribler 8.1.3, add a torrent and disable anonymous download and seeding with proxies. Allow the torrent to complete. On completion the torrent changes from zero hops to the default number of hops set in "Anonymity" setting.

Should the seeding proxies remain disabled with zero hops?

qstokkink:
Fixes #8549

This PR:

 - Fixes `safe_seeding` reenabling itself if the download default of `safe_seeding: true` is used.


qstokkink:
validate

egbertbouman (approved):

PR #8605, opened 2025-05-22
Added url copy menu for system tray

### LINKED ISSUE #8604: Adding url copy menu for system tray
* Since a pull request template wants to reference some issue number—I am writing identical issue here.

Hello,
Thanks for the project. I've been using tribler in windows and I appreciate its existence.
I just have a minor annoyance that it doesn’t present me that magic url with api key to open tribler in the browser of my choice. I try to keep tribler relatively private so I want to open it on portable firefox instance, but it keeps opening on my default browser (edge) so I have to be real fast to copy url before its loading to open it on any other browser.
After few months of this experience I decided to see if I can do anything about it by modifying tribler. I am no python expert by any means but with enough internet and ai chat I was able to add MenuItem that provides an option to copy api url (that magic url) to clipboard.
I have only tested this in windows and so far it seems to work for me. I tried to make it work across mac and linux as well but please feel free to edit if those implementations are not right.
Please take a look at it and integrate it/edit it/etc.
Thanks.

thejjw:
Fixes #8604

This PR:

 - Adds url copy menu for system tray
 
Hello,
Thanks for the project. I've been using tribler in windows and I appreciate its existence.
I just have a minor annoyance that it doesn’t present me that magic url with api key to open tribler in the browser of my choice. I try to keep tribler portable so I want to open it on portable firefox instance, but it keeps opening on my default browser (edge) so I have to be real fast to copy url before its loading to open it on any other browser.
After few months of this experience I decided to see if I can do anything about it by modifying tribler. I am no python expert by any means but with enough internet and ai chat I was able to add MenuItem that provides an option to copy api url (that magic url) to clipboard.
I have only tested this in windows and so far it seems to work for me. I tried to make it work across mac and linux as well but please feel free to edit if those implementations are not right.
Please take a look at it and integrate it/edit it/etc.
Thanks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,6 @@ lz4
 pillow
 pony
 pystray
+pyperclip
+windows-toasts
 pywin32;sys_platform=="win32"
--- a/src/run_tribler.py
+++ b/src/run_tribler.py
@@ -72,6 +72,8 @@ try:
     from aiohttp import ClientSession
     from PIL import Image
 
+    import pyperclip
+
     import tribler
     from tribler.core.session import Session
     from tribler.tribler_config import VERSION_SUBDIR, TriblerConfigManager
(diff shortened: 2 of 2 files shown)


qstokkink:
First, thank you for your contribution. Part of the reason why we want issues to be linked, is to discuss potential changes before they are implemented - so we/you don't spend time creating code that is not merged. In this case, after some offline discussion, we ended up deciding not to accept your changes.

The two main reasons for not accepting this PR are:
 1. This PR introduces two new dependencies. Even without the toast dependency, this is still one added dependency for a likely little-used feature for average users. At the same time, this is a risk for build complications and it increases our installer size.
 2. The intended use for API keys is programmatic access using the REST API. To fix browser tab opening, modifying the `BROWSER` environment variable is preferred instead (as outlined in #8604). Ideally, we don't want typical users to even touch their API key.

Issue #8548, opened 2025-04-21
Move after Completion not consistent


kotlinbuddy:
I'm facing an issue where I've set Move after completion in settings, but it doesn't happen consistently. I have Win 11, Amd Ryzen 5600X, Tribler_8.1.3_x64.exe

In Tribler 8.1.2, I was facing this issue atleast 80% of the time. I had to manually move most of the torrents.

In Tirbler 8.1.3, this issue has gone down to 10% of the time. But I still have to do it sometimes.

I've noticed that first there is a move, then hashcheck. Could be related to [8521](https://github.com/Tribler/tribler/issues/8521)? I'm also facing that issue, I've added my observations for that issue on that thread. Maybe it could be related, if there is some race condition or logical issue which causes hashchecks to skip file movement. I'm not sure on this.

Is there a log file I can see? The debug tab scrolls too fast for me.

qstokkink:
I think the fix of #8531 has something to do with this. We'll have to check.

If this is indeed fixed by #8531, you should slowly stop seeing this issue as your old torrents get new configuration data.

qstokkink:
We will keep an eye on this, but I assume this is fixed in the next release (Tribler 8.2) after both #8531 and #8613.

PR #8613, opened 2025-05-26
Fix loading metainfo

egbertbouman:
Currently, downloads aren't loading their metainfo correctly, and as a result re-download their metainfo after every restart. This PR fixes this, and also addresses the issue that stopped torrents always show "Stopped 0%" after a restart, instead showing the actual progress. For some reason, downloads that are seeding don't have this 0% issue (which is probably why we haven't seen any bug reports yet).

Also fixes an issue where a stopped download can't be moved after a restart, since moving file storage doesn't appear to be working when libtorrent's `handle.get_metainfo` returns false. Possibly also related to https://github.com/Tribler/tribler/issues/8548.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -1028,9 +1028,6 @@ class DownloadManager(TaskManager):
             return False
 
         metainfo = config.get_metainfo()
-        if not metainfo:
-            self._logger.error("Could not resume checkpoint %s; metainfo not found", filename)
-            return False
         if not isinstance(metainfo, dict):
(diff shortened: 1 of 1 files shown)


egbertbouman:
validate

qstokkink:
Agreed/confirmed on `write_resume_data_buf` and loading the checkpoint: that is also how I believe it works and should work.

> attach it to the atp.

My previous comment was for the safety of this little bit of the puzzle. The question I have is: "can we do that attachment by just setting `atp.ti = ...`? Or, will that lose information in some non-`None` `atp.ti` we are replacing?" So far, I don't have an answer to that. So, I also have no alternative to offer.

I am genuinely interested in the answer. My earlier comment was not a "you're doing it wrong, but I'm not telling you how"-comment. I legitimately just don't know and I'm looking for help.

Issue #8474, opened 2025-03-06
Ability to set torrent speed limits for a single torrent


absolutep:
Ability to set torrent speed limits for a single torrent [on a per torrent basis].

Similar to as shown in image

![Image](https://github.com/user-attachments/assets/502f8702-2489-4cf3-b06b-ca0bb8d379af)

qstokkink:
Thanks for the suggestion. This has been on our wishlist for a few years now (though we never made an issue). I think this is a good fit for `8.2`.

PR #8616, opened 2025-06-02
Allow individual upload/download limits

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8474.

image


egbertbouman:
validate

PR #8619, opened 2025-06-03
Added warning popdown for max tracker setting

### LINKED ISSUE #8514: Max concurrent tracker connections -> Loading 0%
Tribler 8.1.2, if a value in, "Max concurrent tracker connections", is set too high (>200) on a restart Tribler hangs with torrents all stuck at, "Loading 0%". It is not possible to access settings to change the value back as the settings page/s are all blank. Tribler will not quit/ shutdown gracefully and the Tribler process needs to be terminated. The "Max concurrent tracker connections" value has to be manually changed back in, "configuration.json".

A sanity check and/ or max limit should be applied when using this settings value. Some info for the user should be included underneath the setting.

Other than a large value be applied, there may be another issue to investigate as to what is causing Tribler to hang.

qstokkink:
Fixes #8514

This PR:

 - Adds a small warning note when the maximum number of trackers is set to a high value (>150).

[warningtrackers.webm](https://github.com/user-attachments/assets/0778dbb0-8630-41f0-96fa-9a6bf50a760a)

### DIFF
--- a/src/tribler/ui/public/locales/en_US.json
+++ b/src/tribler/ui/public/locales/en_US.json
@@ -45,6 +45,7 @@
     "MinusOneIsUnlimited": "-1 = unlimited",
     "EnableAnnounceAll": "Announce to all torrent trackers",
     "MaxTrackerConnections": "Max concurrent tracker connections",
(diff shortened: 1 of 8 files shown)


qstokkink:
validate

egbertbouman (approved):

Issue #8356, opened 2024-12-24
Add (Multiple) Torrents does not display "Download torrent" menu options


teddyrogers:
Tribler 8.0.7, when adding/ importing a single torrent Tribler displays the "Download torrent" menu option to select the destination folder and drop down menu to select individual files within the torrent.

When adding two or more torrents at a time this menu does not appear. If importing lots of torrents and they need to go to a specific directory or, certain files in the torrents are required, those torrent all need to be imported individually.

Can this be improved to process a batch of imported torrents individually, to display the "Download torrent" each time?

Additionally there could be a separate menu option to import torrents silently, as it does now, when adding two or more?

PR #8620, opened 2025-06-04
Support for showing multiple SaveAs dialogs

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8356.

When "Always ask download settings" is enabled, and the user adds multiple torrents, Tribler will now show the `SaveAs` dialog for one torrent at a time.

I didn't add an additional menu option for adding torrents silently, as that defeats the purpose of having a GUI setting.


egbertbouman:
validate

PR #8621, opened 2025-06-05
Updated the main run script to not allow the w3m browser

### LINKED ISSUE #8602: "Starting tribler..."   ...and nothing?
Hello, I get this and nothing more on Ubuntu 24.10. If that is correct, should there be any user guide?

>tribler
Starting Tribler...
/usr/share/tribler ~
>

qstokkink:
Fixes #8602

This PR:

 - Updates `run_tribler.py` to remove `w3m` from the supported browsers (if it was found).

When testing, `w3m` appeared as the very last option in my `_tryorder` list. So, I don't really understand how it can be the first choice when (according to the information in #8602) `firefox` is available in the `_tryorder`. Even so, this is the heavy-handed approach to also solve that by completely removing `w3m` support.


qstokkink:
validate

egbertbouman (approved):

PR #8625, opened 2025-06-10
Allow setting selecting files to None

egbertbouman:
I got an error that suggested "files" in the download checkpoints were somehow getting set to `None` (which is then converted to a string after restart). This was while working on the download core, so I'm not sure this happens on the main branch too.
To avoid potential issues, this PR allows you to call `set_selected_files` with `None` as an argument.


egbertbouman:
validate

Issue #8596, opened 2025-05-17
What is Tribler downloading? Ghost?


absolutep:
What is Tribler downloading?

The torrent is not even loaded yet. 

And how did it download such a huge amount? 

I just started Tribler??

![Image](https://github.com/user-attachments/assets/92d8b8c7-91b8-4f42-84ca-b3d91439cdf0)

absolutep:
Completely uninstalled and reinstalled Tribler version = https://github.com/Tribler/tribler/actions/runs/15072687477

I mean I understand overhead, when I am downloading/uploading something, but I have not downloaded anything and still it has crossed 1 GB within a minute

![Image](https://github.com/user-attachments/assets/27766999-a42e-498f-bbb6-e3f4b71f8153)

teddyrogers:
@absolutep at a guess most this is most likely due to you being a proxy and forming a part of the Tribler network. Possibly this also includes passing on and aggregating all search queries and file history in the Tribler Community.

If the above is correct it would be good if those statistics could be separated out between what is being shared as a proxy, for the Tribler network, and what you are torrenting.

Check out: Debug -> Tunnels -> Circuits Tab

qstokkink:
@egbertbouman since you made both the anonymization and the little status bar, do you want to look into what is going on here?

qstokkink:
By the way, this shouldn't have anything to do with anonymization. This is the result of downloading a 34 kB torrent:

![Image](https://github.com/user-attachments/assets/c3c68ee4-cd6f-4df0-8420-a0c671c7c477)

Perhaps the `Metadata` state is causing this?

absolutep:
> By the way, this shouldn't have anything to do with anonymization. This is the result of downloading a 34 kB torrent:
> 
> ![Image](https://github.com/user-attachments/assets/c3c68ee4-cd6f-4df0-8420-a0c671c7c477)
> 
> Perhaps the `Metadata` state is causing this?

In the first image, I did try to download the torrent.

But in the second image that you have also quoted, I never tried to download anything, 

just switched tribler ON. after completely uninstalling it, remove roaming files, restarting my pc and then installing latest build 165.

so I do not think second image has anything to do with torrenting.

qstokkink:
Thanks for the info: that helps nail down the issue. It seems that the anonymization is not to blame here then.

qstokkink:
[...]

PR #8626, opened 2025-06-10
Fix ghost downloads

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8596 which can occur due to poorly managed metainfo downloads. To be safe, I'll also limit the maximum download rate of metainfo downloads, but I'll do this in a separate PR.

This PR:
* Fixes an issue where if a download can get metainfo after the timeout, it will fully download to the temp folder without ever being removed. It has been part of our code base for 2 years now (https://github.com/Tribler/tribler/commit/81c4bb839b0a8215040d844aa680e75fee8685dc). To fix this, I've added a `finally` clause to `get_metainfo`.
* Fixes an issue where if `get_metainfo` gets called without the DHT being ready yet, it will timeout and remove the download. However, it won't remove the handle from the session, since there's no handle available yet. If the handle becomes available later, it fully downloads to the temp folder. To fix this, the `DownloadManager` now removes unmanaged handles. In the future refactoring of `TorrentChecker`, we should make sure it won't start checking until the DHT is ready, but I think this fix is still needed due to possible `get_metainfo` requests from REST endpoints.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -505,6 +505,10 @@ class DownloadManager(TaskManager):
                 logger.debug("Got alert for download without handle %s: %s", infohash, alert)
         elif infohash:
             logger.debug("Got alert for unknown download %s: %s", infohash, alert)
+            if alert_type == "add_torrent_alert":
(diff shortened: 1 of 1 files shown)


egbertbouman:
validate

PR #8627, opened 2025-06-10
Added torrent backup folder

### LINKED ISSUE #5592: Preference to automatically create accompanying .torrent file
Would be nice if there was a preference to automatically create an accompanying .torrent file of each added torrent.

It can either be:

- stored alongside where the torrent downloads to, and should be moved alongside the torrent data on "Move file storage"

- stored in their own directory

qstokkink:
Fixes #5592

This PR:

 - Adds a setting to automatically store `.torrent` files for all manually-started downloads.

I chose not to add a unit test in this case, as it would be "99% mock".

settingscreen

result


egbertbouman (commented):
I'm sure about calling `write_backup_torrent_file` from the endpoint.
By doing this, doesn't this potentially keep the connection open indefinitely? Our plan to implement queueing would only worsen this. And what about after a restart?

qstokkink:
Correct, and when restarting before the handle is available, there will be no torrent. If you care about these things, I can address them (of course, the impact on the code base will be bigger).

egbertbouman (approved):
If it's not an oversight, it's not a bug. Just making sure this behaviour is intentional.

qstokkink:
The approach for this PR was "good enough for now". That said, I do agree with you that it could use more love.

I do think it would be nice to have operations that continue even after restart. This also goes for adding the default trackers. I'll make an issue for that.

Issue #3724, opened 2018-07-10
Limit max amount of active downloads


Thomasedv:
Make Tribler limit the amount of active downloads (and potentially uploads) to favor speed for certain torrents. Batch adding a large number of torrents may take many hours before the first torrent is fully downloaded, and if you could tell tribler to only take the 3 at a time, would greatly shorten the time before the first torrent is downloaded. As the user may want one torrent before the others, e.g. the first episode of a series. (If the torrents aren't all in one torrent)

Right now, the user needs to manually stop and start all the torrents themselves. And without batch selecting in the GUI, it quickly becomes a hassle, as well has having to remember to start the next ones afterwards.

Dmole:
Could be scripted thought the RPC API untill it's implemented.

heldersepu:
I raised this before: issue #608 in the comment the made it look as if it was an easy task:

Comment from @NielsZeilemaker
> We should be able to implement/expose this in the gui, as libtorrent allows us to do exactly this:
> http://www.rasterbar.com/products/libtorrent/manual.html#queuing

Comment from @egbertbouman
>GUI stuff should be simple. Besides that, we only need to turn on auto-management on for all torrents and it should work.

But the issue was closed as wontfix, I hope that

heldersepu:
> Could be scripted thought the RPC API untill it's implemented.

@Dmole Can you provide a sample script?

Dmole:
> > Could be scripted thought the RPC API until it's implemented.
> 
> @Dmole Can you provide a sample script?

https://tribler.readthedocs.io/en/next/restapi/downloads.html
So something like
```
T=http://localhost:8085
while read -r H ; do
    curl -X PUT "$T/torrents/$H/start"
done < <(curl -X GET "$T/downloads?get_peers=0&get_pieces=0" | grep infohash)
```
I left out string sanitation, and the limiting logic for brevity, but the main issue is;
the documentation is lacking so you would have to look at the source for the correct start/stop commands.

heldersepu:
@Dmole Thanks for the docs link, ye that documentation is not the best, but I think I can hack my way from there... 

Do you see this feature been added any time soon?

devos50:
I made some improvements to the documentation in #4521.

heldersepu:
@devos50 Improvements are good! 
take a look here:
https://tribler.readthedocs.io/en/next/restapi/downloads.html#Tribler.Core.Modules.restapi.downloads_endpoint.DownloadsEndpoint
[...]

Issue #3172, opened 2017-10-22
Just run one check at the time!


Nubro01:
- [ ] I have tried with the latest pre-release version and I still can reproduce the issue.
Oct 20th builtd
##### Tribler version/branch+revision: #####
Tribler 7
##### Operating system and version: #####
Windows 10 Insider 17017
##### Steps to reproduce the behavior: #####
Start multiple Torrent checks
##### Expected behavior: #####
If same location then que the checking.
##### Actual behavior: #####
all checks are started at the same time
##### Relevant log file output: #####
When 3 checks are running, where ALL files that are checked are on the same drive, then within Windows taskmanager the drive is at 100% and disk read is just 6MBs.
When I stop two, the drive goes to 54% and the read speed is 100 MBs, so much faster, which is logical, because the files are spread over the disk, so increased head movement, slows down the whole process.
I do not know if you guys have access to the destination/location, but if you have to check a Torrent and one is already running, check if this new check is in the same location, if it is, que it. if it is not, then start it.

qstokkink:
We should probably make this configurable. Something like a `Maximum parallel checks` setting.

ichorid:
I would say Tribler should only check a single torrent at a time by default, though the setting to disable this limitation should be available in the core config file. Checking lots of torrents in parallel makes no sense for all but fastest SSDs because the checking procedure is always i/o bound.

devos50:
Not sure whether parallel torrent checking is IO bound. I think it is network bound since getting a response from the DHT usually take up to a few second. Since we only use a single UDP socket for lookups, we also do not keep multiple sockets open. We should aim for some performance experiments with torrent lookups.

ichorid:
@devos50 , by "torrent checking" in this thread we mean "checking the files on the disk against the data in the torrent file", i.e. rehashing. It has nothing to do with the network at all :wink:

Nubro01:
The latest test version nr 7.13 shows stopped Torrents as white and 0% in the progress bar. When you start them they start checking, while that is not needed! If you start them all at ones, they all start checking, while active torrents are also seeding from the same disk. 3 tasks is already slowing things down on a hard disk and although an SSD might have 4 channels a harddisk has not, so things really slow down. Please create a que for torrents to be checked and do not start them all at once!

Nubro01:
[...]

PR #8631, opened 2025-06-11
Add queueing

egbertbouman:
This PR adds support for queuing.

Fixes https://github.com/Tribler/tribler/issues/3724.
Fixes https://github.com/Tribler/tribler/issues/3172.

A summary of changes made:
* Added Queued / Moving download states. Until now we never showed any feedback that libtorrent was moving files.
* Added auto_managed flag to download config. The default value can be changed in Settings->General. Custom values can be provided in the `SaveAs` dialog. Once running, the context menu can be used to view or toggle auto_managed.
* Added session settings for auto management. Values can be changed in Settings->General.
* Added a new "#" column showing the position in the queue for a specific session. For example, "1-3" means that a download is in the 1-hop session, in queue position 3. This column is enabled by default.
* Moving up/down + to the top/bottom of the queue can be done using the action buttons on the top of the download table, or using the context menu.
* Updated `ActionButtons`/`ActionMenu` so that the download list updates after every click. This fixes issues with the results of actions not being directly visible.
* Added an extra anonymized libtorrent session for getting metainfo. This avoids confusion with metainfo downloads sharing the same queue as normal downloads. I've set a hard-coded limit of 500KiB/s for this session (to limit issues like https://github.com/Tribler/tribler/issues/8596). In the future, when refactoring the swarm health stuff, we can revisit this.

image

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download_manager.py
+++ b/src/tribler/core/libtorrent/download_manager/download_manager.py
@@ -189,6 +189,8 @@ class DownloadManager(TaskManager):
         if self.config.get("libtorrent/upnp"):
             (await self.get_session()).start_upnp()
 
+        self.get_session(-1).add_done_callback(lambda s: self.set_session_limits(-1))
+
     def start(self) -> None:
         """
         Start loading the checkpoints from disk.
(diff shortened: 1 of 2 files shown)


egbertbouman:
validate

egbertbouman:
validate

PR #8634, opened 2025-06-13
Fix SOCKS5 hop count

egbertbouman:
Fix for allowing multiple SOCKS5 sessions to the `TunnelCommunity`.
Requires `ipv8-rust-tunnel` 0.1.30.


egbertbouman:
validate

PR #8635, opened 2025-06-16
Fix download config issues

egbertbouman:
This PR:
* Fixes an issue with files possibly being set to "None" after updating the hop count.
* Fixes an issue with the auto_managed flag in `add_download`.


egbertbouman:
validate

qstokkink (approved):

PR #8649, opened 2025-07-08
Add tracker scraping

egbertbouman:
This PR:
* Adds tracker scraping for active downloads. After merging, the trackers tab will show the seeds/leeches in the swarm according to the tracker. To make room for the extra data, I removed the buttons from the table and added a context menu. I've also updated the seeds/leeches in the download table to show what is hopefully the most accurate data.
* Fixes an issue with slow shutdown when no dowloads are running.


egbertbouman:
validate

qstokkink (approved):

PR #8651, opened 2025-07-09
Add trackers from TorrentMetadata to downloads

egbertbouman:
This PR fixes an issue with trackers in the database not being added to downloads.

After adding the trackers, I noticed Tribler freezing up because torrentinfo requests weren't getting cancelled. To fix this I've updated `SaveAs` to cancel pending requests.


egbertbouman:
validate

qstokkink (approved):

PR #8696, opened 2025-07-31
Updated move_storage to prefer existing file

### LINKED ISSUE #8689: Moving no progression bar!
As soon as my downloads are finished I move them to their family folder like D:\Tribler\Films because since 8.1 I am unable to pick a location when selecting a torrent from thePirateBay. The move within the same drive looks like a move of the filename.
I seed until a ratio of 9. When my harddisk got full I copied my D:\Tribler\Films to my NAS on S:\Films and that took a while. 
Asuming that a tribler move would check the destination folder, I started moving my Tribler library entries and to my surprise the files where physically moved again and not in a que, hence my D: drive got very busy, reading for seeding and reading for moving. I had to stop seeding to give the move some extra disk access time.
I also saw that the move DOES NOT have a progress indicator bar and I had to look at the network speed to get an indication whether that move had fineshed and I could start a new one! If moves where queued then I would not have to do it manually!

Please consider implementing a progression bar when moving to a **different** disk location.
Also check if files are already present!

Regards
Ron Nuberg

qstokkink:
Fixes #8689

This PR:

 - Updates `move_storage()` to use `dont_replace` mode.

This mode doesn't just flat-out leave files that exist. The behavior is a bit more nuanced. The actual behavior is that a file that already exists **does** get overwritten if its contents are wrong, otherwise the file is left.

Suppose we have `readme.txt`, last modified at `14:26`.

screenshot before

Now I start a download that moves-after-completion into this directory with the same file. The target file is not affected:

screenshot after

Now, _I change the contents of the txt file_, at `14:40`:

screenshot case 2 before

Again I start a download that moves-after-completion into this directory. Once finished, the file is detected to be different and replaced, at `14:42`:

screenshot case 2 after

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -659,7 +659,7 @@ class Download(TaskManager):
         Move the output files to a different location.
         """
         if self.tdef.torrent_info is not None:
-            handle.move_storage(str(new_dir))
+            handle.move_storage(str(new_dir), lt.move_flags_t.dont_replace)
         self.config.set_dest_dir(new_dir)
         return True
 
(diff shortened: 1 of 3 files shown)


qstokkink:
validate

egbertbouman (approved):

Issue #8685, opened 2025-07-28
Stuck on moving 100%


qstokkink:
> Based on the title of this, I assume this is related to my issue.
> 
> I have a completion directory that is different from the download. When restarting Tribler, with 8.2.0 and 8.2.1 all the completed downloads get "Moving 100%" status.
> 
> <img width="226" height="1074" alt="Image" src="https://github.com/user-attachments/assets/41140513-f10a-4644-af10-f05d017dcbf3" /> 

 _Originally posted by @balupton in [#8629](https://github.com/Tribler/tribler/issues/8629#issuecomment-3114175909)_

qstokkink:
I can't reproduce this. I've tried the following things:

 - Download file 1.txt to folder A, move to folder B with a 1.txt already present.
 - Download file 1.txt to folder A, move to folder B with a 1.txt already present, but different.
 - Download file 1.txt to folder A, move to folder B with a 1.txt already present, but read-only.
 - Download file 1.txt to folder A, move to folder B that doesn't exist.

There's also still the option that these moves are legitimately queued for some reason.

balupton:
happy for you to remote into my machine and debug

qstokkink:
Thanks, but I'm not comfortable with logging into other people's machines.

I forgot to ask two things:
 1. Do you have auto management enabled on these torrents?
 2. Is the location, which these files are being moved to, reachable/writable?

qstokkink:
As an aside, trying to move into an unwritable directory nets me this - also somewhat odd - `Seeding 0%` state:

<img width="1534" height="84" alt="Image" src="https://github.com/user-attachments/assets/9cc9ca38-adc5-4b4d-af6c-116b959fcbba" />

balupton:
<img width="2296" height="1576" alt="Image" src="https://github.com/user-attachments/assets/3fd41257-b550-46ca-bc19-dbbf11de54cb" />

<img width="2278" height="796" alt="Image" src="https://github.com/user-attachments/assets/6863ee14-479e-4779-b031-200c1d4d8c62" />

the "Save files to" is an external jbod raid of SSDs, the "create backup torrents in" and "move after completion" is an external jbod raid of HDDs.

the moving 100% torrents are old ones, that have already moved a long time ago, and are already 100% complete

I just rebooted my machine, and the issue seems to have resolved itself

[...]

Issue #8697, opened 2025-08-01
Updating to Tribler 8.2.2 created hashchecking frenzy


teddyrogers:
Tribler 8.2.2, updated Tribler from 8.2.1 to 8.2.2. On starting Tribler every torrent began, "hashchecking", including the downloaded and seeding torrents.

This makes me think back to the enhancement request of Tribler queuing hashchecking torrents to be completed consecutively rather than concurrently. It would save a lot of disk thrashing and likely speed up the process.

MARIUSVAMPIRE:
me too.

SpudmanWP:
Are you using the "Move after completion" option?
It's only happening to my already downloaded files and they are all moved after 100% so I cannot test the diff.

SpudmanWP:
OK, I have a fix:
I have two Folders, yours may vary, but you get the basic idea.

 /home/owner/Desktop/TriblerDownloads/
/home/owner/Desktop/TriblerDownloads/_Complete/

<img width="754" height="251" alt="Image" src="https://github.com/user-attachments/assets/c9242dad-7684-4182-8d4e-e83dd7a0fc66" />

All of the files that I have problems with are already complete and reside in the "Complete" folder.

1. Select the files you have problems with
2. Pause the job(s)
3. Right-click on those records and change the "Location" to the root Download location, make sure that "Move after Completion" is correct and selected, then click "Move Storage"

<img width="387" height="125" alt="Image" src="https://github.com/user-attachments/assets/c4a2cec1-70dc-4746-a14c-52ec0e36e16a" />

4. Restart the affected records.  It will hash once, move the files to the correct locations, then switch to seeding.

MARIUSVAMPIRE:
@SpudmanWP OK. It's solved.

teddyrogers:
@qstokkink hash checking has been an ongoing drama and irritation over many Tribler versions. If I recall correctly I have created about three or four separate similar hash checking issues in various Tribler versions.

If Tribler knows when a torrent has completed to perform the, "Check torrent after completion", then please add in the dlcheckpoints file a flag to state a hash check has been performed on torrent completion. Tribler can then check this flag, and if a hash check has already been completed further hash checks are ignored.

SpudmanWP:
Ugh... It's happening on new downloads that are moved after completion :(

SpudmanWP:
Update:
I'm disabling moving after completion because it just did it again to all files after a reboot.

teddyrogers:
[...]

PR #8699, opened 2025-08-02
Fix issue with downloads moving after restart

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8685. Possibly also https://github.com/Tribler/tribler/issues/869 and https://github.com/Tribler/tribler/issues/8697

Tribler was calling `move_storage` after every restart because `get_completed_dir` and `get_dest_dir` are returning different types of objects (one being a `str` and the other `Path`). Because of this the paths are never considered equal, and we kept calling `move_storage` after every restart.


egbertbouman:
validate

PR #8715, opened 2025-08-12
Added 'completed on' column

### LINKED ISSUE #6693: Add a "Completed on" column to the transfer window
I have quite a few downloaded elements in my Tribler now. However, there are days when I don't look at my Tribler at all, and I'll have to remember which files are new. Thus, I would like to have a "Completed on" column with the "Added on" column, so I can easily see the newest completions.

qstokkink:
Fixes #6693

This PR:

 - Adds a `Completed on` column to the GUI and downloads REST API.

Note that the ATP is persisted by Tribler, but this is not automatically updated by libtorrent.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -9,6 +9,7 @@ import asyncio
 import base64
 import itertools
 import logging
+import time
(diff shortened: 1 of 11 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8716, opened 2025-08-13
WIP: Updated code base to use ATPs instead of metainfo dicts

### LINKED ISSUE #8671: Finish `add_torrent_params` porting
In #8550, I converted the `TorrentDef` to manage its entire state in a `add_torrent_params` object. This was a nice first step, but there is more to convert.

Essentially, Tribler uses three concepts that are outdated:

 - [x] Metainfo, which has been replaced by the `add_torrent_params` object.
 - [X] Resume data, which has been replaced by the `add_torrent_params` object.
 - [x] Storing atp info (v1/v2) as a dictionary, which has been "replaced"[^1] by the `add_torrent_params` object.

All in all, using `add_torrent_params` as the main mode of state storage in Tribler will allow us to get rid of a lot of code.  However, just like #8440, this will involve touching _a lot_ of _critical_ moving parts. So, I'll schedule this for the next major `8.3.0` release.

[^1]: Strictly speaking, this is still supported. It's just harder to manage.

qstokkink:
Fixes #8671

This PR:

 - Fixes/updates Tribler's dependency on `TorrentDef.atp.info_hash`, as it is always written as a zero-hash in the resumedata of checkpoints (due to `libtorrent.write_resume_data_buf()`).
 - Updates "all" the usages of metainfo dicts, to use `libtorrent.add_torrent_params` instead.

This is super scary. I'll keep this in WIP for a bit, to continue testing. However, feedback is already welcome!

_Current issue: v2-only torrents._

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -25,7 +25,6 @@ from ipv8.util import succeed
 from tribler.core.libtorrent.download_manager.download_config import DownloadConfig
 from tribler.core.libtorrent.download_manager.download_state import DownloadState, DownloadStatus
 from tribler.core.libtorrent.download_manager.stream import Stream
-from tribler.core.libtorrent.torrentdef import TorrentDef
 from tribler.core.libtorrent.torrents import check_handle, get_info_from_handle, require_handle
 from tribler.core.notifier import Notification, Notifier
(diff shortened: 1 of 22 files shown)


qstokkink:
validate

qstokkink:
The v2-only torrents are a bigger problem than I initially thought. I'll close this PR (for now) to avoid continuously triggering the CI.

Issue #8723, opened 2025-08-23
Tribler Crash & Restoring Metadata


teddyrogers:
Tribler 8.2.3 Build 172...

@qstokkink I have been running it for a while now and Tribler process crashed (on Windows). The process simply terminated and the GUI popped up stating it lost connection.

1. When Tribler restarted a LOT of torrent metadata was lost. After Tribler downloaded the metadata (of the torrents 100% completed, and seeding, before the crash) Tribler wanted to (re)download these from 0%. No hashcheck (I had to manually force this). Ratio's look to have been retained. Important note here, for point two, all these torrents had hops with anonymity enabled.
2. This second issue, it may not be solely Build 172, I think I recall coming across it in a previous version. Of the torrents that were struggling to download metadata I disabled anonymity to speed things up. I believe some of the torrents downloaded the metadata, with no anonymity, and of these Tribler automatically started a hashcheck and marked these as seeding 100%. I believe if the torrents had been left with anonymity enabled the hashcheck would not have been performed and torrents would have been (re)download from 0%, as noted in the first point.
3. Of the torrents now with no anonymity, and showing seeding 100%, there is a weird issue where they are displaying size 0 bytes, refer screenshot below. If I click on the details tab they show 0 bytes downloaded yet seem to have kept upload data. Forcing a hashcheck on these does nothing. If I change these back to x hops with anonymity they go back to metadata 0%.

<img width="480" height="323" alt="Image" src="https://github.com/user-attachments/assets/3f003e3f-9201-48b8-b399-d84c18a90466" />

Additionally, are there two different methods Tribler uses to retrieve and manage torrents when there is no anonymity and when anonymity is enabled?

qstokkink:
The sudden crash likely has to do with some data being used by both C and Python and then deleted by C while Python still uses it. I'll go over the changes of #8722 again and explicitly copy the memory. That's slower, and really _shouldn't_ be necessary, but safer.

Your three points all have to do with the way Tribler now stores its checkpoints. Old downloads will seemingly have a size of 0 (they should still download/seed normally!). Redownloading the metadata once, for example by changing the hop count, should fix this. When I made #8722, I figured this wouldn't be too big of a deal, but I understand that this is annoying: I'll make a small, one-time, checkpoint upgrader.

> are there two different methods Tribler uses to retrieve and manage torrents when there is no anonymity and when anonymity is enabled?

[...]

PR #8725, opened 2025-08-26
Restore legacy metainfo support

qstokkink:
Fixes most of #8723, excluding the crash

This PR:

 - Updates `DownloadManagerload_checkpoint()` to deal with legacy metainfo again.

This should be removed again A.S.A.P. but, for now, it will ease the transition to the new 8.3.0 format.

I went back and tested with a checkpoint from 8.0.7 to verify behavior.


qstokkink:
validate

PR #8730, opened 2025-09-04
Add placeholders to show P2P settings interface format

### LINKED ISSUE #8702: Changing local listening interface prevents Tribler from starting
Tribler 8.2.2, if you use a device name for the local listening interface (Peer-to-Peer settings) Tribler refuses to start, and the process automatically quits with no user prompt or information. The only way to recover is to manually edit, "configuration.json". This could be because it is an incorrect device name or IP?

A question around this setting, does it need to be an IP address or can interface device names be used such as, "Ethernet4", similar to qBittorrent's implementation?

Under, "BitTorrent features", there is a similar local listening interface setting and it seems to accept device names, though I don't know if it is really working? The logs do not seems to have highlighted any issues.

Maybe some extra information, similar to what you see under the port setting under general settings, would be helpful on how it is meant to be used. 

After changing the interface setting, and saving, a user prompt/ message could be made if it is incorrect.

qstokkink:
Fixes #8702

This PR:

 - Adds placeholder values to show the P2P interface specification format, should they be removed (for example, accidentally).

### DIFF
--- a/src/tribler/ui/src/pages/Settings/Connection.tsx
+++ b/src/tribler/ui/src/pages/Settings/Connection.tsx
@@ -69,6 +69,7 @@ export default function Connection() {
                 </Label>
                 <Input
                     id="ipv8_ipv4"
+                    placeholder="0.0.0.0"
                     value={
                         (
                             settings?.ipv8?.interfaces || [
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8742, opened 2025-09-16
Updated IPv8 pointer

qstokkink:
This PR:

 - Updates `IPv8` to the newest version.
 - Updates `mypy` to no longer ignore `ipv8` imports.

### DIFF
--- a/src/tribler/core/tunnel/community.py
+++ b/src/tribler/core/tunnel/community.py
@@ -3,10 +3,10 @@ from __future__ import annotations
 import hashlib
 import math
 import time
-from asyncio import Future, open_connection, timeout
+from asyncio import CancelledError, open_connection, timeout
 from binascii import hexlify, unhexlify
 from collections import Counter
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 from ipv8.messaging.anonymization.community import unpack_cell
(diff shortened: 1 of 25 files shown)


qstokkink:
validate

EDIT: I guess it would be a good idea to extract the OpenSSL fixes into their own PR. Validation will still fail for that PR, but it at least the changes won't be hidden in a massive PR.

src/tribler/core/tunnel/community.py
```diff
@@ -190,14 +190,15 @@ def update_torrent(self, peers: set[tuple[str, int]], download: Download) -> Non
             if self.find_circuits():
                 self.readd_bittorrent_peers()
 
-    def remove_circuit(self, circuit_id: int, additional_info: str = "", remove_now: bool = False,
-                       destroy: bool = False) -> Awaitable[None]:
+    @task
```
qstokkink, line 193: Next to the big ball of typing nonsense, this is - in my opinion - the only controversial change in this PR to please mypy. This changes the behavior of the method:

Old:
    Execute the body of this `remove_circuit` and allow awaiting of the Coroutine of `super().remove_circuit`.

New:
    Allow awaiting of this `remove_circuit` and the Coroutine of `super().remove_circuit` together.

I think this is not an issue for anything but the one test that changed, but I may be wrong.

egbertbouman (approved):

PR #8745, opened 2025-09-17
Fixed missing OpenSSL libs for libtorrent

qstokkink:
Extracted from #8742

This PR:

 - Fixes `ImportError: DLL load failed while importing libtorrent: The specified module could not be found.` 
 
The correct way to fix the missing OpenSSL libraries for the Windows runner is to build OpenSSL ourselves. This avoids any supply chain attacks and it guarantees security. However, this takes an additional 5 minutes per PR. Instead, I'm downloading `dll` files from the Internet because it's MUCH faster. For testing, the security risk should be negligible. For building, we still use the "correct" secure way, of course.

### DIFF
--- a/.github/workflows/periodic_unittests.yml
+++ b/.github/workflows/periodic_unittests.yml
@@ -24,6 +24,9 @@ jobs:
       - run: python -m pip install -r requirements.txt
       - name: Run unit tests
         run: |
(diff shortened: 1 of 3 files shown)


qstokkink:
"validate"-ing this PR is guaranteed to fail until it is merged. So, I've skipped it.

egbertbouman (approved):

PR #8746, opened 2025-09-18
Add columns

### LINKED ISSUE #6113: Users' suggesstions for the GUI
Let's add **all** the user's suggestions over the years here!

[Forum post](https://forum.tribler.org/t/a-few-suggestions/6618)

- [ ] when I’m seeding many torrents, I often find myself wanting to keep track of the biggest breadwinners so I can delete ones that aren’t very popular, so it would be very helpful if there was a column for “total amount uploaded” for each torrent so they can be sorted by how much “money” they make. Currently the best option I know of is to sort by upload speed, but that only works if it’s being actively downloaded right now
- [ ] also if the columns are too cramped in the UI design, you might consider making it so you can right click and choose “select columns” to customize it like some softwares have
- [ ] I always wondered what the number in parentheses under seeds/peers meant, it would often show a high number but then (0) afterwards which made me think I had something setup wrong and it couldn’t connect… I actually was able to figure out what it means by looking at the code on github, but it would be nice if this was explained in the FAQ or docs somewhere
- [ ] along the same train of thought, it would be useful if tribler showed some kind of bright red warning in the corner in case something is unable to connect due to incorrect router/port settings instead of having to hope it’s working
- [ ] it seems unclear whether the “ratio” is showing the normal ratio or your personal ratio in the tribler P2P community… I think it would actually be useful to have both
- [ ] does a torrent’s “health” tell us its stats on tribler or on the open internet? The former seems like it would be more useful, but I don’t know which one we’re looking at
- [ ] in the dialog for adding a torrent, there is a checkbox for whether to download anonymously which is nice, but it would be even nicer if you could choose which anonymity level (1-3) you want :wink:
- [ ] is it possible to make the torrent items in the list less shorter? it feels like it wastes a lot of screen space vertically (but this is just a personal preference :stuck_out_tongue:)
- [ ] it seems that deleting a torrent from your list doesn’t remove it from your channel(s) if it’s in any of them, so it might be useful to include a “also delete from channels” checkbox when deleting a torrent

### LINKED ISSUE #8743: [Feature Request] Add "Latest Activity" column
Please add a "Latest Activity" to Tribler.  It should relate to the current Status.  

For example, if currently in "Downloading", then the last time a packet was received should be tracked but all upload sharing should be ignored.  Once the Status switches to "Seeding", then the last time it was shared can be displayed.

This will help us cancel incomplete downloads that have not received data in a long time.  Similarly, it will allow us to delete older files that nobody has asked for in a long time.

qstokkink:
Fixes #6113
Fixes #8743

This PR:

 - Adds a slider to select the number of hops when starting a download.
 - Adds a "Total up" column to the downloads page.
 - Adds a "Last up" column to the downloads page.


qstokkink:
validate

src/tribler/ui/src/dialogs/SaveAs.tsx
```diff
@@ -334,6 +335,21 @@ export default function SaveAs(props: SaveAsProps & JSX.IntrinsicAttributes & Di
...
                     </label>
+                    <div className="grow" />
+                    <div className="relative items-enter mb-3">
+                        <Slider
+                            className="w-64"
+                            value={[params.anon_hops === undefined ? 0 : params.anon_hops]}
+                            min={0}
+                            max={3}
+                            step={1}
+                            onValueChange={(value) => {console.log(value); setParams({...params, anon_hops: value[0]})}}
```
egbertbouman, line 346: Maybe remove the `console.log`?

qstokkink:
validate

PR #8782, opened 2025-10-10
Fix torrent discovery

egbertbouman:
Currently, our main source of torrent discovery is through swarm health information. When we receive an infohash that we haven't seen before, we immediately request the `TorrentMetadata`. These requests are expected to be mostly successful, because peers only send swarm information if they have the `TorrentMetadata` as well. It turns out, these requests always fail, because we're requesting an infohash that is *double* hex-encoded.

After merging this fix, torrent discovery should be fixed. This means that, combined with the earlier tracker fixes, we should be getting more tracker information. So, hopefully, this will also (partially) fix our swarm health problem.

### DIFF
--- a/src/tribler/core/content_discovery/community.py
+++ b/src/tribler/core/content_discovery/community.py
@@ -105,7 +105,7 @@ class ContentDiscoveryCommunity(Community):
         for field in self.composition.binary_fields:
             value = parameters.get(field)
             if value is not None:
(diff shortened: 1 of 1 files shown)


egbertbouman:
validate

teddyrogers:
@egbertbouman is this captured in build 178?

egbertbouman:
@teddyrogers No, not yet. I'll make a new build soon.

egbertbouman:
The new build: https://github.com/Tribler/tribler/actions/runs/18403318013

Issue #8731, opened 2025-09-04
Tracker status, "Not contacted", yet we have contact


teddyrogers:
Tribler 8.2.3 Build 173, is showing trackers as not contacted yet it is able to determine there are seeds and leeches. Refer to the screenshot below...

<img width="648" height="353" alt="Image" src="https://github.com/user-attachments/assets/87722796-788b-4d8a-b428-2b5ee51adda0" />

PR #8784, opened 2025-10-14
Fix tracker alerts

egbertbouman:
I was having problems downloading certain popular torrents, where a tracker was available, but the download wouldn't start. It turns out some trackers are (intermittently) blocking the exit nodes due to leeching.

It took a while to figure out what was happening, because Tribler provided no feedback. Apparently tracker alerts weren't getting processed. This PR fixes this. Also fixes https://github.com/Tribler/tribler/issues/8731.


egbertbouman:
validate

PR #8832, opened 2025-12-04
Fix state freeze after updating the hop count

egbertbouman:
I noticed that `apt.flags` can magically change to 8 (the `apply_ip_filter` flag). The same goes for the output of `handle.flags()`. Despite this, everything works as if the flags are still set correctly. However, since we're using this value when updating the hop count, this can freeze the download state (because the `update_subscribe` flag isn't getting set).

Since we always want state updates anyway, I've manually set the  flag in `_async_add_torrent`.


egbertbouman:
validate

Issue #8838, opened 2025-12-21
A rustError occurred with new version


danp5648:
**Traceback and/or steps to reproduce**
Startup 
see photo

![Image](https://github.com/user-attachments/assets/2e218ed3-402b-4945-9cbc-27bbde2b444f)

Dmole:
By "new version" do you mean "v1.2.3" ?

danp5648:
No. beta version. 1.2.3.4

PR #8844, opened 2026-01-12
Catch exceptions while starting SOCKS5 servers

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8838.


egbertbouman:
validate

PR #8845, opened 2026-01-13
Update TorrentChecker

egbertbouman:
This PR does the following:

* Updates `DownloadManager.get_metainfo` to spend more time in the swarm when doing a health check. It's currently set to 4 minutes. As health checks now potentially take a lot more time, I've increased the maximum simultaneous checks to 5,
* Splits health checks in 2 parts: checking trackers, and if that doesn't yield any results, join the swarm for 4 minutes.
* Removes `FakeDHTSession`/`FakeBep33DHTSession` as I think it makes more sense to call `get_metainfo` separately (as a fallback for torrents without a working tracker).
* Disables checking random trackers, as this only does a partial check. I didn't remove the underlying code, in case we decide to turn this back on in the future.
* Adds a debug panel showing completed health checks.

### DIFF
--- a/src/tribler/core/tunnel/community.py
+++ b/src/tribler/core/tunnel/community.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import hashlib
 import time
 from binascii import hexlify, unhexlify
-from collections import Counter
 from typing import TYPE_CHECKING, cast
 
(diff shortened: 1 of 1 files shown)


egbertbouman:
validate

Issue #8525, opened 2025-03-20
Setting do disallow people to disabled anonymity or at least make it harder.


nextgenthemes:
I really do not like the fact that I can accidentally disallow anon. Especially now with the right click menu it's at the very top and if you hover over it and click it ....

Putting at the bottom of the list would be a small improvement, even it would not make logical sense the way the list is ordered.

Extra confirmation, or a setting to just disallow it globally forever and make the option to completely disappear. If people want to download torrents without even 1 hop, they're probably using other clients torrent anyway, or am a wrong about that?

PR #8846, opened 2026-01-15
Ask for confirmation before disabling anonymity

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8525.


egbertbouman:
validate

Issue #4689, opened 2019-07-18
Show torrent description


ichorid:
Torrenn description stored in the torrent file should be shown when opening a torrent for downloading. Currently, we only show the title and the files list.

Related to #4674

PR #8848, opened 2026-01-16
Show torrent description in SaveAs dialog

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/4689.


egbertbouman:
validate

PR #8854, opened 2026-01-20
Fixed downloads list vis and sort not saving

### LINKED ISSUE #8851: Problem with saving column settings
Hello,

I want to enable the "Added on" column and sort downloads by it. After doing so and saving the settings, the corresponding lines in the configuration file are updated as below.

```
   "ui": {
        "sorting": "{\"download-sorting\":[{\"id\":\"time_added\",\"desc\":true}],\"popular-sorting\":[]}",
        "columns": "{\"download-columns\":{\"queue_position\":true,\"name\":true,\"size\":true,\"all_time_upload\":false,\"progress\":true,\"num_seeds\":true,\"num_peers\":true,\"speed_down\":true,\"speed_up\":true,\"hops\":true,\"eta\":false,\"last_upload\":false,\"time_added\":true,\"time_finished\":false}}"
    }
```

But when Tribler is opened again, it resets those settings and the corresponding lines in the file are changed. (the "download-sorting" value becomes null, and  the "time_added" value of "columns" becomes false)

```
    "ui": {
        "sorting": "{\"download-sorting\":[]}",
        "columns": "{\"download-columns\":{\"queue_position\":true,\"name\":true,\"size\":true,\"all_time_upload\":false,\"progress\":true,\"num_seeds\":true,\"num_peers\":true,\"speed_down\":true,\"speed_up\":true,\"hops\":true,\"eta\":false,\"last_upload\":false,\"time_added\":false,\"time_finished\":false}}"
    }
```

This doesn't happen with other settings like download location.

I use Windows 11 and Firefox.

qstokkink:
Fixes #8851

This PR:

 - Fixes the sorting state and column visibility of the downloads list being overwritten by nothing on startup.

The idea for both states is to add a new `undefined` state, only while starting, to avoid overwriting the settings with the default replacement value and sending this to the core process before receiving the actual settings.

There was already a `sortingState !== undefined` check in place (line 430 in SimpleTable). So, I assume this is how it worked at some point in the past already.

### DIFF
--- a/src/tribler/ui/src/components/ui/simple-table.tsx
+++ b/src/tribler/ui/src/components/ui/simple-table.tsx
@@ -1,6 +1,7 @@
 import {Dispatch, SetStateAction, useEffect, useRef, useState} from "react";
 import {Table, TableBody, TableCell, TableHead, TableHeader, TableRow} from "@/components/ui/table";
 import {
+    OnChangeFn,
     getCoreRowModel,
     useReactTable,
     flexRender,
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #8866, opened 2026-01-28
Updated resume data to manually populate from status

### LINKED ISSUE #8827: change anonymity hub count resest ratio to startup number
I seed anonimous until a ratio off 9, then seeding is stopped and I manually change anonymity hop count to 2. As a result every entry has its ratio reset to the number that it had at startup. SO it was 9 and after a hop count of to the ratio changes to 8.53 and I have to seed it again until 9 is reached for the second time.

qstokkink:
Fixes #8827

This PR:

 - Updates `on_save_resume_data_alert` to populate the `tdef.atp` from the `lt_status` and `handle`.

Notes:

 - For some reason the `name` in the resume data's `params` is being updated, even though the rest seems to be stale (from my testing). So, I do still use that.
 - I kept track of the things that could be set on the `atp`, but are currently not, in the "Skip:" comment. I skipped them because they are either static (e.g., `info_hash`), temporary (e.g., `peers`), or in a special category of "yeah, no way are we reconstructing that" (e.g., `unfinished_pieces`).
 - There is something to be said for an alternate design of updating the ATP as status/handle alerts come in. Essentially, you would be slowing down in those alert handlers all the time, instead of only when save_resume-ing. That would cut the time it takes to save_resume though. I opted not to do that here, but we could consider it in the future if the save_resume becomes too slow.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -459,7 +459,7 @@ class Download(TaskManager):
         if alert.state == lt.torrent_status.downloading and self.tdef.torrent_info is None:
             self.post_alert("metadata_received_alert")
 
-    def on_save_resume_data_alert(self, alert: lt.save_resume_data_alert) -> None:
+    def on_save_resume_data_alert(self, alert: lt.save_resume_data_alert) -> None:  # noqa: PLR0915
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

afbeelding

PR #8870, opened 2026-02-03
Various `DownloadConfig` things

### LINKED ISSUE #4980: Add seeding ratio control for individual torrents
At the moment, we provide only global seeding policy controls. Instead, thhe GUI should provide a way to control seeding ratio for individual torrents.

### LINKED ISSUE #8629: Have Download "post-handle" operations persist between restarts
_[Result of discussion in #8627]_ 

We have several operations that require a `Download` to have a valid handle and would, ideally, persist even after a restart.

https://github.com/Tribler/tribler/blob/36b8365afe884e13f6cf20df016dfac659814bcb/src/tribler/core/libtorrent/restapi/downloads_endpoint.py#L454-L460

Currently, a download that never receives a handle before Tribler is shut down also doesn't have any default trackers added or `.torrent` created. Furthermore, the above code keeps the REST API connection open.

It would be nice to have some generic method to achieve this.

### LINKED ISSUE #8867: Don't use a file for the download config spec
We are still using a download config `.spec` file. This is in line with what `ConfigObj` has in mind, but wholly unnecessary for the way we use the `DownloadConfig` class.

If we use `StringIO`, we can avoid reading the file. We already do this in the unit tests.

https://github.com/Tribler/tribler/blob/c1f822e45edd11a0763499e8e4534ff3d86f1bed/src/tribler/test_unit/core/libtorrent/download_manager/test_download.py#L52

qstokkink:
Fixes #4980
Fixes #8629
Fixes #8867

This PR:

 - Adds individual torrent seeding ratio control. This is the only change in this PR with GUI edits.
 - Updates the "default tracker" and "backup torrent" features to be persistent between restarts.
 - Removes the `configobj` dependency and replaces it with the built-in `configparser` module.

### DIFF
--- a/src/tribler/core/libtorrent/download_manager/download.py
+++ b/src/tribler/core/libtorrent/download_manager/download.py
@@ -16,6 +16,7 @@ from binascii import hexlify
 from collections import defaultdict
 from collections.abc import Callable
 from contextlib import suppress
+from functools import lru_cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypedDict, cast
 
@@ -23,7 +24,7 @@ import libtorrent as lt
 from ipv8.taskmanager import TaskManager, task
 from ipv8.util import succeed
 
-from tribler.core.libtorrent.download_manager.download_config import DownloadConfig
+from tribler.core.libtorrent.download_manager.download_config import DownloadConfig, PostHandleOp
 from tribler.core.libtorrent.download_manager.download_state import DownloadState, DownloadStatus
 from tribler.core.libtorrent.download_manager.stream import Stream
 from tribler.core.libtorrent.torrents import check_handle, get_info_from_handle, require_handle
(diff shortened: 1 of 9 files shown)


egbertbouman (approved):
Looks good to me, although I'm sure about mixing add/substract with bitwise OR/AND. Personally, I think it's safer/clearer to always use bitwise OR/AND for setting/unsettings post-handle operation flags. It's not a deal breaker though.

qstokkink:
I do agree with that general sentiment. However, the reason I opted for this approach is because Python is a little annoying when dealing with clearing bits due to the variable length of `int`'s (and I didn't want to deal with it 🙂). Example, flipping the 0'th bit that contains the value `1`:

```python
>>> ~1  # Intention: flip a 1-bit length '1' value to '0'
-2  # :-( extra data left of the flipped bit
```

I know there are bitwise operator packages and I know you can specify a mask with the flag length, or even do some fancy bit shifting. However, it seemed "easier" to just use `- (minus)`. Of course, it is always possible that I just missed a better solution.

If you can make this implementation sleeker and/or safer, I will absolutely approve a PR on this. For now, let's stick with this until something better arrives.

egbertbouman:
I think something like this would also have worked:

```python
remaining &= ~PostHandleOp.ADD_DEFAULT_TRACKERS
```

Anyway, I'm nitpicking here.

PR #8871, opened 2026-02-04
Added download filters for 'not downloading' and 'inactive, unqueued'

### LINKED ISSUE #8497: FR: Multi-select and/or Button for clearing completed downloads
It would be extremely useful if we had the ability to select multiple torrents and then remove them all in one click instead of clicking on each one, one by one. It becomes very time consuming if you have many files in the UI. 

Another option would be to have a button to remove all completed torrents (stopped or seeding) in order to quickly clean up your list.

Thanks!

### LINKED ISSUE #8793: Remove "Queued" from "Inactive" sub-menu list
Tribler 8.2.3 Build 179, to locate all, "Stopped", torrents I generally go to the, "Inactive", list. When auto-management is being used this can be heavily populated with, "Queued", items making the search for, "Stopped", torrents a bit of an annoyance.

I think it would be better if, "Queued", is filtered out from the, "Inactive", list or, alternatively, there is a, "Stopped" and/ or "Queued", sub-menu under, "Downloads".

qstokkink:
Fixes #8497
Fixes #8793

This PR:

 - Adds two new requested filters: "anything else than actively downloading incomplete torrents" and "not doing anything, but that's not because it's queued".
 - Removes an accidental `console.log()` that I left in #8864.

### DIFF
--- a/src/tribler/ui/src/components/layouts/Footer.tsx
+++ b/src/tribler/ui/src/components/layouts/Footer.tsx
@@ -42,7 +42,6 @@ export function Footer() {
 
         let anonResponse = await ipv8Service.getTunnelPeers();
         if (anonResponse && !isErrorDict(anonResponse)) {
-            console.log(anonResponse);
             if (anonResponse.length == 0) {
                 setAnonymityStatus(AnonStatus.NoPeers);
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

PR #8878, opened 2026-02-11
Prefetch trackers

egbertbouman:
This PR:
* Fixes a "the database session is over" error in `ContentDiscoveryCommunity.send_search_request`
* Add the tracker from `TorrentMetadataPayload` to the database (this was removed in https://github.com/Tribler/tribler/pull/8875)


egbertbouman:
validate

PR #8895, opened 2026-02-19
Added corrupt database cleanup

### LINKED ISSUE #8815: Build 179, Failed to start download! DatabaseError: database disk image is malformed
Tribler Build 179 (Ubuntu), I am receiving error message, "Failed to start download! DatabaseError: database disk image is malformed", whilst adding new torrents.

Image

After the error message, the download continues to start and finish okay.

qstokkink:
Fixes #8815

This PR:

 - Adds a corrupt database removal step in `MetadataStore`. Because the `self.db.bind` happens in the `__init__`, this is also in the `__init__`, which is not necessarily where we want to do slow I/O—but I conformed to the existing structure for now.
 
I tested this with a real corrupt database: `test_fast_integrity_check_malformed` uses the output of a real integrity check.

### DIFF
--- a/src/tribler/core/database/store.py
+++ b/src/tribler/core/database/store.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import enum
 import logging
+import os.path
 import re
+import sqlite3
 import threading
 from asyncio import get_running_loop
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


qstokkink:
Oops: cleaning up branches on my repo and accidentally trashed this one.

PR #8896, opened 2026-02-19
Added Windows quality of life features

### LINKED ISSUE #5972: Suggestions for Tribler UI

qstokkink:
Fixes #5972

This PR:

 - Adds Tribler being able to register itself as a Startup App in Windows.
 - Adds a config option to launch Tribler minimized (not available in `headless` mode).
 - Fixes Tribler appearing in the Windows task manager and Startup Apps as `Privacy enhanced BitTorrent client with P2P content discovery`.

I built https://github.com/qstokkink/tribler/actions/runs/22185429951 and tested it to see if Tribler correctly appears in the Windows Startup Apps list. In this particular build it will show up as `Privacy enhanced BitTorrent client with P2P content discovery` (hence the fix in this PR for that).


src/run_tribler.py
```diff
@@ -267,9 +267,10 @@ def spawn_tray_icon(session: Session, config: TriblerConfigManager) -> Icon:
     api_port = session.rest_manager.get_api_port()
     url = f"http://{config.get('api/http_host')}:{api_port}/ui/#/downloads/all?key={config.get('api/key')}"
     menu = (pystray.MenuItem("Open", lambda: open_webbrowser_tab(url)),
-            pystray.MenuItem("Quit", lambda: session.shutdown_event.set()))
+            pystray.MenuItem("Quit", session.shutdown_event.set))
```
qstokkink: "random" change, but `ruff` insisted on this.

Issue #8904, opened 2026-02-24
get_uniformed_tracker_url `None` url


qstokkink:
Hi! I was using Tribler git and THIS happened! :cry:
```
Traceback (most recent call last):
  File "C:\Users\goqs\AppData\Local\Programs\Python\Python312\Lib\asyncio\events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "tribler\pyipv8\ipv8\taskmanager.py", line 180, in done_cb
    future.result()
  File "tribler\pyipv8\ipv8\taskmanager.py", line 30, in interval_runner
    await interval_task(*args)
  File "tribler\src\tribler\core\rss\rss.py", line 135, in check
    await self.parse_rss(content)
  File "tribler\src\tribler\core\rss\rss.py", line 153, in parse_rss
    await self.resolve(new_entries)
  File "tribler\src\tribler\core\rss\rss.py", line 89, in resolve
    self.notifier.notify(Notification.torrent_metadata_added, metadata=metadata_dict)
  File "tribler\src\tribler\core\notifier.py", line 84, in notify
    observer(**kwargs)
  File "<string>", line 2, in add_ffa_from_dict
  File "tribler\.venv\Lib\site-packages\pony\orm\core.py", line 519, in new_func
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "tribler\src\tribler\core\database\orm_bindings\torrent_metadata.py", line 313, in add_ffa_from_dict
    return cls.from_dict(dict(metadata, public_key=b"", status=COMMITTED, id_=id_))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "tribler\src\tribler\core\database\orm_bindings\torrent_metadata.py", line 348, in from_dict
    return cls(**dct)
           ^^^^^^^^^^
  File "tribler\src\tribler\core\database\orm_bindings\torrent_metadata.py", line 264, in __init__
    self.add_tracker(tracker_info)
  File "tribler\src\tribler\core\database\orm_bindings\torrent_metadata.py", line 276, in add_tracker
    sanitized_url = get_uniformed_tracker_url(tracker_url)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8908, opened 2026-02-25
Fix for RSS feeds with bad tracker URLs

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8904.


egbertbouman:
validate

Issue #8919, opened 2026-03-06
DatabaseSessionIsOver: Cannot load attribute TrackerState[12].url


qstokkink:
I'm getting this error when running on the latest release code:

```python
Traceback (most recent call last):
  File "tribler/pyipv8/ipv8/taskmanager.py", line 180, in done_cb
    future.result()
  File "tribler/src/tribler/core/content_discovery/community.py", line 414, in on_remote_select_response
    request.processing_callback(request, processing_results)
  File "tribler/src/tribler/core/content_discovery/community.py", line 260, in notify_gui
    r.md_obj.to_simple_dict()
  File "<string>", line 2, in to_simple_dict
  File "python3.12/site-packages/pony/orm/core.py", line 538, in new_func
    reraise(exc_type, exc, tb)
  File "python3.12/site-packages/pony/utils/utils.py", line 88, in reraise
    try: raise exc.with_traceback(tb)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.12/site-packages/pony/orm/core.py", line 519, in new_func
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "tribler/src/tribler/core/database/orm_bindings/torrent_metadata.py", line 336, in to_simple_dict
    "trackers": self.tracker_info_list,
                ^^^^^^^^^^^^^^^^^^^^^^
  File "tribler/src/tribler/core/database/orm_bindings/torrent_metadata.py", line 283, in tracker_info_list
    return [tracker_state.url for tracker_state in (self.health.trackers if self.health else [])]
            ^^^^^^^^^^^^^^^^^
  File "python3.12/site-packages/pony/orm/core.py", line 2278, in __get__
    value = attr.get(obj)
            ^^^^^^^^^^^^^
  File "python3.12/site-packages/pony/orm/core.py", line 2288, in get
    val = vals[attr] if attr in vals else attr.load(obj)
                                          ^^^^^^^^^^^^^^
  File "python3.12/site-packages/pony/orm/core.py", line 2243, in load
    if cache is None or not cache.is_alive: throw_db_session_is_over('load attribute', obj, attr)
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.12/site-packages/pony/orm/core.py", line 685, in throw_db_session_is_over
[...]

PR #8922, opened 2026-03-09
Add metadata dict to ProcessingResult

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/8919.

This PR:
* Refactors `process_payload` such that entities are directly converted to dictionaries in the current `db_session`. This should make the process faster and avoids prefetching or "session is over" issues.
* Adds `ProcessingResult.rowid` in case we need the database `Entity` at a later time.
* Removes database logic from the `notify_gui` callback.


egbertbouman:
validate

PR #8924, opened 2026-03-13
Added advanced bandwidth allocation feature

### LINKED ISSUE #1031: Scheduling bandwidth limits
Having the ability to schedule bandwidth during certain times is greatly needed. Some countries still have on-peak / off-peak times and it would be great to only download / upload when the user can. This would obviously be for the torrent traffic only and the underlying protocol would be untouched

qstokkink:
Fixes #1031

This PR:

 - Adds an "advanced settings" graph to control bandwidth rate limits based on the time of day, for each of the four torrent sessions.

Note:
 - Setting the rate limit to `0` bytes/s means `unrestricted` to libtorrent, so the minimum value is actually `1`, which in practice will have peers not send you anything (I tested this with a 500 byte download).
 - I tried really hard to make `const lineChart = useRef<any>(null);` a narrower type, but I can't make it work.
 - The config settings use stringified JSON per hop mostly to save space. By default, Python would put every value in a list on a new line: this made the `configuration.json` extremely long.
 

https://github.com/user-attachments/assets/ceb866d5-8639-4ce6-be7a-94d17a279a5b


qstokkink:
Hold on. Fixing the types broke the config loading. Fixing 🧑‍🏭

PR #8938, opened 2026-03-18
Allow empty state directory in win installer

### LINKED ISSUE #8936: (Optional) Specify a custom directory to store database and configuration files.
Tribler 8.4.1, during installation there is a (optional) prompt to specify a custom directory.

From the attached screenshot, it is not, "optional", if it needs to be specified. Should the empty directory field be pre-populated?

Image

qstokkink:
Fixes #8936

This PR:

 - Fixes the optional state dir in the Windows installer not being optional.
 
Test build here: https://github.com/qstokkink/tribler/actions/runs/23246692147

### DIFF
--- a/build/win/resources/tribler.nsi
+++ b/build/win/resources/tribler.nsi
@@ -77,6 +77,7 @@ BrandingText "${PRODUCT}"
 !define MUI_PAGE_HEADER_SUBTEXT "(optional)"
 !define MUI_DIRECTORYPAGE_TEXT_TOP "Specify a custom directory to store database and configuration files."
 !define MUI_DIRECTORYPAGE_TEXT_DESTINATION "$TSTATEDIR"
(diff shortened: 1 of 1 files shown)


qstokkink:
> Won't this disable path validation completely? If we use `MUI_DIRECTORYPAGE_VERIFYONLEAVE`, maybe we should also have a `MUI_PAGE_CUSTOMFUNCTION_LEAVE`.

Yes, that is correct. I figured Tribler is able to create directories that don't exist yet and it will pull the default when the field is empty. I guess that could be a problem if users enter really random stuff.

I've seen `MUI_PAGE_CUSTOMFUNCTION_LEAVE`, but I don't quite understand it. So, I left it alone. If you know how to work that macro: I'm open to suggestions/help with that. Documentation is a bit sparse.

PR #8980, opened 2026-05-05
ru_RU.json: Fixed multiple mistakes and inaccuracies in translations

pasabanov:
This PR:
 - Fixes multiple mistakes and inaccuracies in Russian translations

Key considerations:
1. The use of the word `прокси` (proxy) for the term `hop`. As far as I know, there is no conventional Russian translation for the word `hop`. Also, the word `прокси` (proxy) was already used in some strings here, so I decided to stick with it.
2. Does `"AutoManageMax": "Maximum number of auto-managed torrents per hop count...",` mean that the specified numbers are counted once per every hop and therefore multiplied by it? The current translation is based on this interpretation.


qstokkink (approved):
Thanks!

PR #8987, opened 2026-05-11
FEATURE: File-level priority

### LINKED ISSUE #8642: Download priority
I would like to set the priority of specific files/folders to be higher, so they download before others files. Other clients like qBittorrent have this feature.

mohanbharath:
Fixes #8642 

This PR introduces granular file-level priorities for torrent downloads, allowing users to designate files as Low, Normal, or High priority directly from the UI.

Backend (libtorrent/DownloadConfig):
- Leverages existing libtorrent priority mechanics (`download_priority_t`).
- Reframes new `file_priorities` config array as the source of truth, with `selected_files` treated as a view over it
- *Backwards Compatibility*: The legacy `selected_files` (inclusion/exclusion) property still remains to ensure existing torrents and existing API consumers do not break.
- Safely defers priority initialization for Magnet links until `on_metadata_received_alert` fires and the total file count is known.
- Extensive unit tests added to test_download_config.py and test_download.py to verify state synchronization.

Frontend (React/TypeScript):
- Added a new Priority dropdown column to Files.tsx.
- Implemented recursive tree traversal using a Map to guarantee folder-level priority assignments correctly propagate to the flat backend array using index mapping.

API:
- GET `/downloads/{infohash}/files` response now includes a `priority` integer per file. PATCH `/downloads/{infohash}` accepts a new corresponding optional `file_priorities` array parameter.

Testing:
- Unit tests on affected portions
- Manually tested via adding torrents and modifying file priorities to verify working as intended
- Fixed misleading docstring on `set_selected_files`: the old docstring claimed empty list meant 'select all,' but the code actually selected none. The old *test* passed only because it mocked the libtorrent handle to return a fixed value, bypassing the production code path. Updated docstring to match actual behavior and rewrote/renamed the test to exercise the real code path
- NOTE: Change to `test_get_torrent_data_v2` was due to extreme test flakiness at HEAD on my machine from scheduling 2,196 event loop tasks; happy to revert that if desire to keep it as-is

**Out-of-scope**: I refrained from completely removing the `files`/`selected_files` logic, both in the backend and frontend, as that seemed overly ambitious for a first contribution. Happy to file a follow-up issue for full removal (including removal of config code and UI checkbox logic) if/when this is merged.

(*Note for reviewer(s): Please be harsh on the Python and generous with the TypeScript/React -- outside of toy projects, this is quite literally my first experience with frontend!*)

**GenAI Disclosure** -- Antigravity code completion, alongside extensive pair-programming/pair-debugging with LLMs.


mohanbharath:
@qstokkink -- friendly ping on this, as I think it's in a decent state for review. Happy to address any feedback/squash bugfix commits/etc!

qstokkink:
Ok, so I had a look at your code. There is no nice way to say this, so I'll just rip the proverbial bandage off: this is the programming equivalent of making fine art with a sledgehammer. I'm not sure if this PR is salvageable, or whether you should be the one saving it.

Before we get to the crux of the matter. There are two things this code should do (equivalent to two line edits):
1. Store file priorites in a `Download`'s `add_torrent_params`, so it can be reloaded with the new priorities.
2. Set file priorities on the active `Download`, using `set_file_priorities`, so the active download has the new priorities.

Your code does both, which is good. However, what you apparently missed is that the `add_torrent_params` is already stored in the `DownloadConfig` (this is the `engineresumedata`). Apart from violating the responsibilities of this class, no edits to it were even necessary. This means that the `+423 -46` of core line edits in this PR should be more in the order of `+10 -2`. Had you discussed your intentions in the related issue, we could have warned you before you starting creating code.

I have some other minor gripes with disabling a unit test that has run for thousands of times without error, adding note comments in the code, and untranslated strings in the GUI. There is also the project risk factor to consider, in that you are touching the soft underbelly of our core logic.

All in all, looking at what would have to change in the bigger picture, I think it's best to just abandon this PR and let the Tribler dev team implement this core feature. We don't want to lead you on, so I'll close this PR to make sure you don't sink any more time/tokens into something that won't get merged.

mohanbharath:
Fair enough -- I did this mostly because I wanted it this feature in my own local Tribler install, so I jumped in and only really thought about "maybe others would also want it" after the fact. Apologies for the noise, and thanks for your work on Tribler broadly!

PR #9032, opened 2026-06-25
GHA CI: enable WebTorrent support

kingomarwashere:
## Summary

- Adds a new `ci_webtorrent.yml` workflow that builds **libtorrent-rasterbar from source** and passes `-Dwebtorrent=ON` for the `RC_2_1` branch
- Uses a version matrix so `RC_2_0` (which matches the current pinned `libtorrent==2.0.13`) still builds without the flag
- Installs Tribler's dependencies with `--ignore-installed libtorrent` so the source-built bindings take precedence
- Mirrors the approach taken in [qbittorrent/qBittorrent#24564](https://github.com/qbittorrent/qBittorrent/pull/24564)

This paves the way for future WebTorrent related development in Tribler.

> **Note:** Tribler's existing CI uses the pre-built `libtorrent==2.0.13` pip wheel. This new workflow adds a dedicated source-build path so WebTorrent-enabled libtorrent bindings can be tested end-to-end ahead of any future RC_2_1 upgrade.

## Test plan

- [ ] Workflow triggers on push/PR
- [ ] `RC_2_1` matrix entry builds with `-Dwebtorrent=ON` and passes unit tests
- [ ] `RC_2_0` matrix entry builds without flag and passes unit tests

### DIFF
--- a/.github/workflows/ci_webtorrent.yml
+++ b/.github/workflows/ci_webtorrent.yml
@@ -1,0 +1,62 @@
+name: CI: WebTorrent
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  test-linux-webtorrent:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        libtorrent:
+          - version: "RC_2_1"
+            libt_build_flags: "-Dwebtorrent=ON"
+          - version: "RC_2_0"
+            libt_build_flags: ""
+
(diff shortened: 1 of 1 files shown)


qstokkink:
Sorry for not getting back to you earlier: we're in the middle of the exam period and vacations.

As a result of a dev team meeting, we ultimately decided that we do not want to create custom `libtorrent` builds, as proposed by this PR. Instead, our approach will be to wait for an _official_ `libtorrent 2.1.x` build with webtorrent enabled by default.

PR #9036, opened 2026-07-07
Added augmented search

### LINKED ISSUE #9015: Allow spelling errors in search
Currently, Tribler search only uses exact keyword matches to retrieve results. This has some odd effects where searching for `ubuntu` may give no results, but `ubuntu-24` may.

Ideally, I would like for Tribler to just understand what we query. Of course, actual understanding would require something like an LLM (or even heavier). However, we can start a bit simpler by simply searching for _near_ results, based on embeddings (cheap and fast).

My current shortlist for this feature (a nearness index) consists of `sqlite-vec` and `hnswlib`, examples in links below.

~~https://github.com/asg017/sqlite-vec/blob/main/examples/python-recipes/openai-sample.py~~ **[REJECTED]** Reason: this library requires `enable_load_extension` on `sqlite3`, which is not shipped by default. We would have to do a custom compile of Python on all our platforms.

~~https://github.com/nmslib/hnswlib~~ **[REJECTED]** Reason:  this library straight-up goes into `SIGABRT` `free(): invalid next size (fast)` with a toy example.

qstokkink:
Fixes #9015

This PR:

 - Adds unigram learning from torrent titles to autocorrect _local_ searches with UTF-8 support (so, not just English or ASCII).
 
Successful build:
https://github.com/qstokkink/tribler/actions/runs/28862381528

There are quite a few parameters that can be tuned here. For these, I simply used my intuition. So, "Why do we retrain/fine-tune at 50 new torrent titles?" and "Why do we build a vocabulary of 8000 unigrams?" are answered with "Feels right." These may/can also change in the future.

In this PR, I use continuous learning with batches/windows of newly-discovered torrent titles. Not completely online learning per title, because that would degenerate the vocabulary. To kickstart vocabularies, I simply sample up to 10000 random torrent titles from the local database (once!). We can play with this in the future by, instead, always training on startup - depending on when we want the slowdown from training to happen.


egbertbouman (approved):
Looks great!

PR #9038, opened 2026-07-07
Added libtorrent outgoing interface setting

### LINKED ISSUE #9037: Add a config option to pin libtorrent's outgoing interface
### Use case

Running headless Tribler behind a VPN (WireGuard in Docker) on a multi-homed host, where torrent traffic must only egress via the tunnel interface.

`libtorrent/listen_interface` already pins the *listen* socket, but outbound peer/tracker connections still leave via the default route. On a host with more than one usable interface, 0-hop download traffic can therefore bypass the VPN entirely (IP leak). libtorrent has a session setting for exactly this: `outgoing_interfaces` (bind-before-connect / SO_BINDTODEVICE). But Tribler never sets it and there is no config key to reach it.

### Proposed change (minimal)

* New string config key `libtorrent/outgoing_interface`, default `""`.
* In `DownloadManager.create_session()` (the `hops == 0` branch, right after `listen_interfaces` is set): `settings["outgoing_interfaces"] = self.config.get("libtorrent/outgoing_interface")`. The config default `""` is exactly libtorrent's own default for `outgoing_interfaces`, so existing users see zero behavior change.
* `tribler_config.pyi` stub regenerated; unit tests for both the set and default paths.

That's one line of feature code plus config keys and tests. I have this [implemented ](https://github.com/Tribler/tribler/pull/9040) and passing the unit suite.

One design point I'd like feedback on: an alternative is to fall back to the listen interface when the new key is unset (so pinning the listen side also pins the outgoing side). That's arguably what a VPN user *wants*, but it changes behavior for anyone who currently sets `listen_interface` on a multi-homed host, so I'd default to the strictly opt-in variant unless you prefer the fallback.

wesleylima:
Fixes #9037

This PR:

 - Adds a `libtorrent/outgoing_interface` config option that pins libtorrent's `outgoing_interfaces` on the hops==0 session, so outbound peer/tracker connections can be bound to a specific interface (e.g. a VPN tunnel) instead of the default route.

The default `""` is exactly libtorrent's own default for `outgoing_interfaces`, so existing users see no behavior change (the unset-path unit test asserts this). One line of feature code; `tribler_config.pyi` regenerated; unit tests included for the set and default paths.


qstokkink:
Just a heads up: setting up a connection test bench to see the effects of these changes is taking quite a while and may take even more time still. With vacations coming up, this PR may still _stay open for several months_ until we've properly evaluated it and made a decision to accept or reject it. If you don't want to deal with that, you can close this PR and we'll pick the code up again later.

wesleylima:
@qstokkink no worries! I can wait. Thanks for the update. Let me know if there's anything I can do to help.

PR #9040, opened 2026-07-07
Made exit node participation configurable with per-protocol flags

### LINKED ISSUE #9039: Allow enabling exit-node mode from config, with per-protocol exit flags
### Use case

Operating a headless Tribler node (Docker, behind a VPN) that contributes relay/exit capacity to the network.

Two obstacles today:

1. **`exitnode_enabled` is not reachable from configuration.** `TriblerTunnelSettings.exitnode_enabled` exists, but `TunnelComponent.get_kwargs()` never reads it from config and there is no `tunnel_community/exitnode_enabled` key — the only assignments in the tree are in the integration tests. So a headless deployment cannot opt in to being an exit node via `configuration.json` at all.
2. **Exit is all-or-nothing.** When `exitnode_enabled` is set (by an embedder), the community adds `PEER_FLAG_EXIT_BT | PEER_FLAG_EXIT_IPV8 | PEER_FLAG_EXIT_HTTP` as one block — so agreeing to be a BitTorrent/IPv8 exit also makes the node an open HTTP proxy. Operators may be willing to carry BT/IPv8 exit traffic but not to run an open proxy.

### Proposed change (minimal)

- New `tunnel_community` config keys: `exitnode_enabled` (default `false` — current behavior), `exit_bt`, `exit_ipv8` and `exit_http` (all default `true` — parity with the current constant).
- `TunnelComponent.get_kwargs()` passes them into `TriblerTunnelSettings`; `TriblerTunnelCommunity.__init__` composes the exit peer flags per-flag instead of the constant triple.
- `tribler_config.pyi` regenerated; unit tests cover flag composition and config defaults.

~15 lines of feature code plus tests. Implemented and passing the unit suite; PR ready if the approach is welcome.

The PR keeps exact parity by default: `exit_bt`, `exit_ipv8` and `exit_http` all default `true`, so `exitnode_enabled=true` yields the same three peer flags as today (verified against the unit and integration suites — nothing changes for any existing user or embedder).

Separate question for you: would you *want* `exit_http` to default to `false`, so that enabling exit-node mode doesn't run an open HTTP proxy unless explicitly requested? Since `exitnode_enabled` currently can't be turned on from config, only embedders that set it programmatically would notice. Happy to make that change in the PR or as a follow-up — your call.

wesleylima:
Fixes #9039

This PR:

 - Adds `tunnel_community/exitnode_enabled` (default `false`) so exit-node participation can be enabled from configuration — the `TriblerTunnelSettings` field existed but was never read from config, so headless deployments could not opt in.
 - Adds `tunnel_community/exit_bt`, `exit_ipv8` and `exit_http` (all default `true` — exact parity with the previous all-or-nothing constant) so the exit peer flags can be composed per protocol, e.g. a BT+IPv8 exit that is not an open HTTP proxy.

No behavior change for any existing user or embedder: with the defaults, `exitnode_enabled=true` yields the same three flags as before (verified against the unit and integration suites). Whether `exit_http` *should* eventually default to `false` is raised as an open question in #9039. `tribler_config.pyi` regenerated; unit tests cover flag composition and config defaults.


qstokkink:
Closed for reasons specified in https://github.com/Tribler/tribler/issues/9039#issuecomment-4912617894

PR #9041, opened 2026-07-08
Fixed Dockerfile for Ubuntu latest/26

### LINKED ISSUE #9024: Build 195, Going from 0 hops to 2 hops, stuck in "Loading 0%" status
In build 195 I have noticed when going from zero hops (no anonymity) to, for example, two hops Tribler is stuck in, "loading", status. The only way to recover is to exit Tribler and restart.

Edit 1: Note, storing torrent info in RAM is enabled.

Edit 2: Actually, this seems to be happening when changing between different anonymity levels (hops) e.g. to two to one or one to three. It is even more of a (noticeable) problem when it is a group of torrents being selected and changed at the same time.

qstokkink:
Fixes https://github.com/Tribler/tribler/issues/9024#issuecomment-4746882841

This PR:

 - Fixes the Docker build failure due to `latest` pointing to Ubuntu 26.

### DIFF
--- a/build/docker/build.Dockerfile
+++ b/build/docker/build.Dockerfile
@@ -9,7 +9,7 @@
 FROM ubuntu:latest
 
 RUN apt update -y \
-  && apt -y install curl dbus-x11 libgirepository-1.0-1 libcanberra-gtk-module libcanberra-gtk3-module libglib2.0-bin
+  && apt -y install curl dbus-x11 libgirepository-1.0-1 libcanberra-gtk3-module libglib2.0-bin
 
 RUN tag=`basename $(curl -Ls -o /dev/null -w %{url_effective} https://github.com/Tribler/tribler/releases/latest)` \
   && vlesstag=$(echo $tag | cut -c2-) \


egbertbouman (approved):

qstokkink:
Regarding the MyPy failure:

```python
pyipv8/ipv8/messaging/anonymization/exit_socket.py:209: error: Argument 1 to "UDPv6Address" has incompatible type "str | int"; expected "str"  [arg-type]
pyipv8/ipv8/messaging/anonymization/exit_socket.py:210: error: Argument 1 to "UDPv4Address" has incompatible type "str | int"; expected "str"  [arg-type]
```

There has been a change in `getaddrinfo` typing:

https://github.com/python/typeshed/blob/a7bcce8b6df32dd85dc31691cef30af61e7b1371/stdlib/_socket.pyi#L927-L929

```python
def getaddrinfo(
    host: bytes | str | None, port: bytes | str | int | None, family: int = ..., type: int = 0, proto: int = 0, flags: int = 0
) -> list[tuple[int, int, int, str, tuple[str, int] | tuple[str, int, int, int] | tuple[int, bytes]]]: ...
```

In the past, the address was only `tuple[str, int] | tuple[str, int, int, int]`. However, now `tuple[int, bytes]` has been added, which - according to https://github.com/python/typeshed/issues/13344 - has to do with Python versions that are compiled with `--disable-ipv6`.

qstokkink:
I offloaded this failure as an issue on IPv8 (https://github.com/Tribler/py-ipv8/issues/1368) and I'll merge this as-is to finally have a Docker build for the most recent Tribler version out.

qstokkink:
🤔 Well, shoot. It seems our Docker builder for a particular version actually builds that particular version now. So, these changes are not backported to fix the "old" 8.4.3 version.

PR #9055, opened 2026-07-14
Various fixes

### LINKED ISSUE #8642: Download priority

### LINKED ISSUE #9051: Build 195, Hashchecking after moving (twice)
Build 195, I moved a couple of torrents (currently in, "Downloading", status) to a different folder. Later on I filtered by name, in the list of torrents, and selected about 10 which, included the two that were already moved. Thinking nothing of it, because they were already in the moved folder, included them in the list. All the torrents moved to the new destination folder whilst the two, previously moved torrents, went to, "Hashchecking", status.

If the torrent is already in the destination folder it should be ignored, and not need to complete a hash check.

qstokkink:
Fixes #9051
Fixes #8642 

Also contains build issues found while making tests builds for the "move progress" issue.

This PR:

 - Adds a context menu action to change file download priority.
 - Updates download moving to not move/checkpoint if the destination and completed dirs have not changed (not filtered by GUI when using multi-selection).
 - Updates the logging to log `error_middleware` unhandled errors (the GUI shows big red errors that don't appear in the logs).
 - Updates the code for `sentencepiece >= 0.2.2`.


qstokkink:
Marking this PR as ready as it now has > 100 line changes. Fat enough 🍔

egbertbouman (approved):

PR #9073, opened 2026-09-08
Added SaveAs dialog queue

### LINKED ISSUE #8993: Build 190, Multiple Processes Created After Importing Torrents
Build 190, whilst, "Always ask download settings", is enabled I am experiencing Tribler creating multiple processes, and failing to queue importing of multiple torrents (similar to https://github.com/Tribler/tribler/issues/8960).

I am using Mullvad browser to import .torrent files from websites.

Firefox browser is set as the default browser application, not Mullvad.

Tribler is not set as the default application for opening .torrent files, qBittorrent is.

The Tribler tab is not open in Firefox when the torrents are being open/ imported via Mullvad, and Tribler is minimised to the tray icon.

If I start importing torrents in Mullvad they are not queuing in the same Tribler instance. Firefox opens a Tribler tab with the issue reported here: https://github.com/Tribler/tribler/issues/8988. After importing the first torrent the second, third, fourth, etc. torrents are not queued and the download settings prompt is not opened.

Additionally, on occasions, multiple Tribler processes are being created. This isn't always consistent but it can be replicated.

Image

qstokkink:
Fixes #8993

This PR:

 - Adds a queue for SaveAs dialogs.

This may seem like just another mundane PR, but I assure you that this covers some next-level arcane React functionality. This is the most complicated thing I've done in TypeScript so far. You're looking at about an hour per line of code here.

But, it does work! Here's a demo that shows me first loading 2 torrents, and then - before I finish adding those - I spawn two separate background Tribler processes that both add a torrent. As you can see, all 4 torrents are queued correctly:

https://github.com/user-attachments/assets/81d71701-5e35-4e69-8dbc-3e8e5315d6a8


egbertbouman (approved):

balupton:
Good work. So this is a sequential pool (concurrency of 1).

If you another approach with a customisable concurrency, you can use my native-promise-pool package:
https://github.com/bevry/native-promise-pool

PR #9081, opened 2026-09-16
Added core CLI error queue

### LINKED ISSUE #9059: Build 195, Tribler is missing a user, "Execution Log", page
Build 195, I occasionally come across issues during the import of a torrent, and I have no place to go to that logs this, an example being the one shown in the image below, of a corrupt .torrent...

Image

You do not see this error message if importing directly from the default browser, because it never shows up (I'll create a separate issue). The only way I can get to see this error message is to download the .torrent file to local disk and manually, "Add torrent", and import in to Tribler.

It would be good if there is an, "Execution Log", page similar to that used in qBittorrent. If there is an error and I don't see the popup/ prompt I can refer to the log page.

With an, "Execution Log", page any disk read/ write issues such as no storage space available could be shown there too.

qstokkink:
Fixes #9059

This PR:

 - Adds a queue to the `DownloadsEndpoint` for errors that result from downloads added in "cli" mode.

There are a million ways to implement this. At the end of the day, this "felt right" to me. Another option would be to use events, but I have personal experience that React doesn't like this. I also tried to avoid active polling, and just piggyback on the `get_downloads`.


qstokkink:
validate

egbertbouman (approved):

